PR #51, opened 2016-12-10
bug fix

krille90:


DooblyNoobly:
opps that was my mistake, was doing some lobby work.

PR #117, opened 2017-01-29
Additional mapping

Epicworlds:


Epicworlds (commented):
Mapped EVA, Security, Partyroom, Vault, Arrvals, Port Hall, Garden, sec checkpoint, north primary hall

PR #179, opened 2017-03-04
Lighting experiment

TheBimmer231:
Added lights and shadows generatos


DooblyNoobly (changes requested):
PR to relevant branch

PR #233, opened 2017-04-17
Mob/Living/SimpleAnimal types

mixxit:
Port a large amount of the Mob/Living/SimpleAnimal code from tgstation, switch some actions to gib/harvest/attack etc


mixxit:
removed kill script
tried to follow the inheritence structure of tgstation to make porting other code into it easier
updated Pete to use the new SimpleAnimal type which is descended from Living which is descended from Mob
need to do more work now on the player which i think is Human descended from Carbon so you'll notice started putting the Carbon code in there for body parts etc
didnt end up giving Pete body parts as thats not how tgstation did it, instead they have a harvest list thats assigned on Living
so he has his meat prefab in there instead
i set the damage of attack to 50 for the knife check you have but really we need to look more at how item damage is calculated

PR #236, opened 2017-04-22
Player health

mixxit:


mixxit:
Not really finished on this but doobly would like to have not so many merge conflicts

PR #266, opened 2017-05-08
little Weapon_Ballistic refactor

deRemo:
-Firing Rate is now working
-It's easier to create new guns(maybe i'll do a wiki page about the format)
-Added L6SAW a sub-machinegun
-Better in-hand reference for the c20rSMG

Issue: need Networking fix


DooblyNoobly:
fixed with #268

PR #319, opened 2017-06-15
Basic EditMode tests for clothing, Minor fix for ItemAttributes

fomalsd:


fomalsd:
Oops, looks like ItemAttributes changes didn't make it here

PR #328, opened 2017-06-24
Resolves compiler warnings for unused variables

MrLeebo:
- Resolves several (but not all) compiler warnings
- Clean up unit test objects
- Removes some noisy debug statements


DooblyNoobly:
looks good

PR #337, opened 2017-06-27
Jobs

mixxit:
Add jobs to game


Assets/scripts/Equipment/Equipment.cs
```diff
@@ -10,160 +10,201 @@
...
+                    syncEquipSprites.Add(-1);
+                }
+                else
+                {
+                    clothingSlots[i].Reference = syncEquipSprites[i];
+                }
+            }
+            isInit = true;
+            //Player sprite offset:
+            clothingSlots[10].Reference = 33;
```
MrLeebo, line 72: What does 10 and 33 mean? Can you define these as consts to make it more clear?

mixxit: You may want to speak to Doobly as this is his change

Assets/scripts/Managers/GameManager.cs
```diff
@@ -6,10 +6,14 @@
...
 using UnityEngine.SceneManagement;
 using PlayGroup;
+using System;
+using System.Linq;
+
 public class GameManager : MonoBehaviour {
 
 	public static GameManager Instance;
 
+    public List<GameObject> Occuptions = new List<GameObject>();
```
MrLeebo: Should be `Occupations`

Assets/scripts/Managers/GameManager.cs
```diff
@@ -32,13 +36,26 @@ public class GameManager : MonoBehaviour {
...
 
+    void ValidateGameConfiguration()
+    {
+        if (Occuptions.Where(o => o.GetComponent<OccupationRoster>().Type == JobType.ASSISTANT).ToList<GameObject>().Count == 0)
+        {
+            Debug.LogError("Refusing to start server, ASSISTANT role must be defined in the GameManager Occupations to start");
+            #if UNITY_EDITOR
+                        UnityEditor.EditorApplication.isPlaying = false;
+            #endif
+            Application.Quit();
```
MrLeebo: This check seems like it should be in an [OnValidate()](https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnValidate.html) handler and call `Debug.LogError` instead of killing the app.

Assets/scripts/Managers/GameManager.cs
```diff
@@ -116,7 +133,53 @@ void OnDisable()
 		}
 	}
 
-	void RestartRound(){
+    int GetOccuptionsCount(JobType jobType)
```
MrLeebo: 🅰️

Assets/scripts/Managers/GameManager.cs
```diff
@@ -116,7 +133,53 @@ void OnDisable()
...
+                }
+            }
+        }
+
+        return count;
+    }
+
+    public List<string> GetOccupationEquipment(JobType jobType)
+    {
+        return Occuptions.Where(o => o.GetComponent<OccupationRoster>().Type == jobType).First().GetComponent<OccupationRoster>().startingItemHiers;
```
MrLeebo: Probably doesn't matter for these methods if they only run on start, but in general I would avoid using Linq extension methods at runtime because they generate a lot of garbage object allocations which will impact performance if we let them get out of control.

Assets/scripts/Equipment/Equipment.cs
```diff
@@ -10,160 +10,199 @@
...
+        public NetworkIdentity networkIdentity { get; set; }
 
-		void Start()
-		{
-			networkIdentity = GetComponent<NetworkIdentity>();
+        private bool isInit = false;
+
+        void Start()
+        {
+            networkIdentity = GetComponent<NetworkIdentity>();
```
MrLeebo, line 25: Is this reference used by anything?

Assets/scripts/Equipment/Equipment.cs
MrLeebo: @DooblyNoobly

Assets/scripts/Managers/GameManager.cs
mixxit: Oh nice i didn't know about this handler, thanks!

PR #366, opened 2017-07-25
Develop

krille90:


MrLeebo:
So what was the purpose of this besides ... ? 
image

DooblyNoobly:
Have nfi how or why but that is why I'm reverting it. Testing the PR now

PR #403, opened 2017-08-30
added new tool to generate map images

krille90:
* can be found under Tools -> Make Map PNG
* PNG will be named like the scene and be saved to top level folder
* ignores sprites on default layer (e.g. ambient sprites)


Epicworlds:
yes plz

PR #589, opened 2017-12-09
splitted structures layer into windows and walls layers

krille90:
### Purpose
Bullets could go through windows, since they couldnt have a collider, because players wouldnt be able to see through windows.

### Approach
Splitted Walls and Windows into two layers, such that both can be handled differently.


krille90:
Do you mean integrating the FOV into the tms?

krille90:
yes, that should already be working

PR #605, opened 2017-12-10
Fix the fix for pushing lockers

levyadams:
### Purpose
fix the fix because the last code had foreach statements which require GC

### Approach
us for loops instead apparently

### Notes:
_Please enter any further information here_

_When you made a function or a tool, please enter a TL:DR manual here_


DooblyNoobly:
Make sure you remove any necessary white space before submitting

PR #606, opened 2017-12-10
Space Death

levyadams:
### Purpose
For people who get spaced to die painfully instead of floating.

### Approach
addDamage in healthbehaviors.cs

### Notes:
Sorry I haven't figured out the formatting issues here. I'll get my VS figured out shortly

run into spess


Em3rgencyLT (commented):
It's worth noting that this will only damage if you're floating away. We should also add damage if you're in space but something is nearby preventing you from floating.

We can't solve this for tiles that are walkable but in space without atmos. I suggest editing the map and removing such areas as a temporary workaround for TDM.

DooblyNoobly:
Yep, although this is sufficient enough for the TDM I would also suggest turning this into a coroutine and only applying damage every second as the 1 hitpoint every frame quickly kills the player

PR #658, opened 2017-12-15
UI health display fixes

DooblyNoobly:
- Now updates the UI correctly by ensuring the data of the PlayerHealth monitor is accurate
- Also adjusted the threshold for a healthy UI display
- Also added more blood to the players blood level when eating for TDM

Network fixes:
- Checks if player is a ghost before doing any calculations and network updates. This stops messages being sent to a players body that is no longer controlled

### Purpose
- Before UI changes weren't happening as expected
- When you respawned your old body was still trying to send UI updates from the carcass

### Approach
- Added blood to the blood levels
- Move the threshold for a healthy UI from 100% to 90%

- Added a check to HealthReporting to see if the player has turned into a ghost to stop updating UI

### Notes:
Its gud

- Stab yourself, eat meat and see that crit overlay goes back to normal (note the health doll doesn't update fully to green, thats another issue)

- Die and respawn, no netmsg errors! yay

### DIFF
--- a/UnityProject/Assets/Scripts/Health/Living/PlayerHealthReporting.cs
+++ b/UnityProject/Assets/Scripts/Health/Living/PlayerHealthReporting.cs
@@ -14,6 +14,7 @@ namespace PlayGroup
     {
         private int bloodLevelCache;
         private float BloodPercentage = 100f;
+        private PlayerMove playerMove;
 
         //server only caches
         private int healthServerCache;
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/Health/Living/PlayerHealthReporting.cs
```diff
@@ -67,7 +69,11 @@ private void ServerMonitorHealth()
         {
             //Add other damage methods here like burning, 
             //suffication, etc
-
+            
+            //If already dead then do not check the status of the body anymore
+            if (playerMove.isGhost)
```
Em3rgencyLT, line 74: No if's without { }! It's easy to miss, prone to produce errors and makes code harder to read!

PR #702, opened 2017-12-21
Rework of access system and fix of bug #675

JacobSweeten:
### Purpose
Reworks access system to reduce unnecessary code redundancy.
Fixes bug #675 

### Approach
I followed the code until I found that the client's check for access. At this point I noticed (With the help of Doobly) that it was using the server's ID instead of the client's. In looking for a fix, I found a rework of the whole system.

### Notes:
The new component can be applied to department cabinets, consoles, etc.

All checks for access are now serverside.


DooblyNoobly (approved):
Thanks, looks good!

PR #786, opened 2018-01-02
Add env switch to enable test server

MrLeebo:
### Purpose
Instead of toggling the GameData "testServer" checkbox in the inspector, you can enable it with an environment switch.

### Approach

Windows
```bash
set TEST_SERVER=True && build.exe
```

OSX
```bash
TEST_SERVER=True open -n build.app/
```


DooblyNoobly (approved):
Cool! I like it

PR #791, opened 2018-01-03
fixing some problems when registering later spawned items

### LINKED ISSUE #772: Post-join spawned items can only be pickuped from X+1 y0 and y+1 X0
# Preliminary information

## Category
BUG

Develop

# Report
## Current Behavior
When an item is spawned it post-you joining the game, it can only be picked up from the right side (x+1 Y0) from above (y+1 X0) Otherwise it gives a not-inreach error

You should be able to pickup items all way arround

## Steps to Reproduce
1. spawn an item (microwave or butcher) after joining
2. See it not be able to picked up
3.

## OS used:
OSX

## System specifications
Macbook pro retina 2012 (late)

## Network
Same pc, high ping (also with headless)

krille90:
### Purpose
This fixes the offset (x-1, y-1) for post-join spawned objects.

### Approach
rerouting every spawning of an item via ItemFactory, which also requires a transform parent to be set, that will be synced for the clients. This fixes #772 .


krille90:
fixes #766

PR #796, opened 2018-01-05
[HEAVY DNM] Adding Energy Weapons

lquinn2015:
### Approach
Added 8 prefabs. FusionCells, LaserPistol, LaserGun, Tazer, 2 Sound manager changes/prefabs, LaserShot, TazerShot. 

Edited Weapon.cs  -> added a toggle for spawning bullet casings. Note i set the default for it to true as to not step on ballistic weapons.

Edited WeaponNetworkActions.cs  -> to use the toggled listed above.

Created BulletEnergyShot.cs -> while this class is empty now i am sure it will expand due to different damage types, maybe it will denote bombs etcs. 

Add the sounds to the sound manager in the death match scene and applied them. 

Additonals
- [x]  Removing map changes
- [x]  Sounds Changed
- [x]  Tested.
- [x] Todo in bullet Behavior script left

## The TODO note
So i left a todo in bullet behavior cs because the damage and damage type will likely need to be set on the call of OnShoot.

### Notes:
I have a bad eye for the guns inhands numbers they should look close enough however. Also my tazer sound would likely work for touching raw wires.


Ornias1993:
Tags added

PR #859, opened 2018-01-14
Fixed #829, removed an old Debug.Log, added suicide PNA-hook.

### LINKED ISSUE #829: Dying inside closet doesn't spawn ghost sprite and opens closet on moving
# Preliminary information

## Category
BUG

Develop

# Report
## Current Behavior
When you die while being in a closet, you don't spawn a ghost sprite, but you do open up the closet when moving after dying. tested with High ping, should be there without it too.

You should just ghost, leaving the closet how it is.

## Steps to Reproduce
1. Die in closet
2. Move as after death
3.

## OS used:
OSX

## System specifications
MBP-r 2012

## Network
High ping, headless

JacobSweeten:
### Purpose
Fixed #829, removed an old Debug.Log, added suicide.

### Approach
Made player closet behaviour stop updating the player when ghosted. Made ghost impossible to make invisible. Removed player closet behaviour when respawning.

### Notes:
Suicide is backspace, Must be un-commented in InputController.cs first. Fixed improper naming on a function.

### DIFF
--- a/UnityProject/Assets/Scripts/Closets/ClosetPlayerHandler.cs
+++ b/UnityProject/Assets/Scripts/Closets/ClosetPlayerHandler.cs
@@ -41,15 +41,8 @@ namespace Cupboards
 		{
 			if (PlayerManager.LocalPlayerScript.playerNetworkActions.isGhost)
 			{
-				Camera2DFollow.followControl.damping = 0.0f;
 				return;
-			}
-			else
-			{
-				// In case we re-enter corpse
-				Camera2DFollow.followControl.damping = 0.2f;
-			}
-			
+			}			
 			if (monitor)
 			{
 				if (CheckForDirectionalKeyPress())
--- a/UnityProject/Assets/Scripts/Closets/ClosetPlayerHandler.cs
+++ b/UnityProject/Assets/Scripts/Closets/ClosetPlayerHandler.cs
@@ -21,9 +21,13 @@ namespace Cupboards
 		public void Init(ClosetControl closetCtrl)
 		{
 			closetControl = closetCtrl;
-			
-			Camera2DFollow.followControl.target = closetControl.transform;
-			Camera2DFollow.followControl.damping = 0.2f;
+
+			if (!PlayerManager.LocalPlayerScript.playerNetworkActions.isGhost)
+			{
+				// TODO: Change this stuff to the proper settings once re-entering corpse becomes a feature.
+				Camera2DFollow.followControl.target = closetControl.transform;
+				Camera2DFollow.followControl.damping = 0.2f;
+			}
 
 			if (!closetControl)
 			{
--- a/UnityProject/Assets/Scripts/Closets/ClosetPlayerHandler.cs
+++ b/UnityProject/Assets/Scripts/Closets/ClosetPlayerHandler.cs
@@ -1,4 +1,5 @@
-﻿using Tilemaps.Scripts;
+﻿using PlayGroup;
+using Tilemaps.Scripts;
 using UnityEngine;
 
 namespace Cupboards
@@ -38,6 +39,17 @@ namespace Cupboards
 
 		private void Update()
 		{
+			if (PlayerManager.LocalPlayerScript.playerNetworkActions.isGhost)
+			{
+				Camera2DFollow.followControl.damping = 0.0f;
+				return;
+			}
+			else
+			{
+				// In case we re-enter corpse
+				Camera2DFollow.followControl.damping = 0.2f;
+			}
+			
 			if (monitor)
 			{
 				if (CheckForDirectionalKeyPress())
--- a/UnityProject/Assets/Scripts/Objects/VisibleBehaviour.cs
+++ b/UnityProject/Assets/Scripts/Objects/VisibleBehaviour.cs
@@ -98,12 +98,8 @@ public class VisibleBehaviour : NetworkBehaviour
 			SpriteRenderer sr = renderers[i] as SpriteRenderer;
 			// Cast and check cast.
 			// This is necessary because some renderers fail the cast.
-			if (sr != null)
+			if (sr != null && !CanBeDisabled(sr))
 			{
-				if (CanBeDisabled(sr))
-				{
-					renderers[i].enabled = _aliveState;
-				}
 			}
 			else
 			{
--- a/UnityProject/Assets/Scripts/Objects/VisibleBehaviour.cs
+++ b/UnityProject/Assets/Scripts/Objects/VisibleBehaviour.cs
@@ -36,8 +36,7 @@ public class VisibleBehaviour : NetworkBehaviour
 		closetHandler
 	};
 
-	public GameObject[] ignoreSpriteUpdates;
-	private SpriteRenderer[] ignoredSpriteRenderers;
+	public SpriteRenderer[] ignoredSpriteRenderers;
 
 	public RegisterTile registerTile;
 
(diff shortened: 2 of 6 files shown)


JacobSweeten:
Can do. I'm at work right now though.

UnityProject/Assets/Scripts/Closets/ClosetPlayerHandler.cs
```diff
@@ -38,6 +39,17 @@ public void Init(ClosetControl closetCtrl)
...
+			{
+				Camera2DFollow.followControl.damping = 0.0f;
+				return;
+			}
+			else
+			{
+				// In case we re-enter corpse
+				Camera2DFollow.followControl.damping = 0.2f;
+			}
+
```
DooblyNoobly: What happens if LocalPlayerScript is null (When there is a few frames of spawning new player) and what happens if Camera2DFollow.followControl.damping needs to be set to something different from another object in a different scenario? This condition in all of the closets Update loops will override that. I would like you to come up with a better solution to changing the damping if a player turns into a ghost. You could just always set the damping property once whenever a player turns into a ghost no matter if they are in a closet or not (maybe in the local client of HealthBehaviour) then you can get rid of this condition.

UnityProject/Assets/Scripts/Objects/VisibleBehaviour.cs
```diff
@@ -36,6 +36,9 @@ public class VisibleBehaviour : NetworkBehaviour
 		closetHandler
 	};
 
+	public GameObject[] ignoreSpriteUpdates;
+	private SpriteRenderer[] ignoredSpriteRenderers;
```
DooblyNoobly: Why not just make this public and drag the spriterenderers into the inspector? You just need to drag the whole game object and it will take the sprite renderer from it

UnityProject/Assets/Scripts/Objects/VisibleBehaviour.cs
```diff
@@ -57,6 +60,16 @@ public override void OnStartClient()
...
+
+		ignoredSpriteRenderers = new SpriteRenderer[ignoreSpriteUpdates.Length];
+		for (int i = 0; i < ignoreSpriteUpdates.Length; i++)
+		{
+			SpriteRenderer sr = ignoreSpriteUpdates[i].GetComponent<SpriteRenderer>();
+			if (sr != null)
+			{
+				ignoredSpriteRenderers[i] = sr;
+			}
+		}
```
DooblyNoobly: You wouldn't need this loop if you make ignoredSpriteRenders public and just drag them in

UnityProject/Assets/Scripts/Objects/VisibleBehaviour.cs
```diff
@@ -109,7 +134,12 @@ private void UpdateState(bool _aliveState)
 		}
 	}
 
-	private bool canBeDisabled(MonoBehaviour script)
+	private bool CanBeDisabled(SpriteRenderer script)
+	{
```
DooblyNoobly: I know its a small thing but the parameter here should not be called script. Call it something related to a renderer

UnityProject/Assets/Scripts/Objects/VisibleBehaviour.cs
```diff
@@ -93,7 +106,19 @@ private void UpdateState(bool _aliveState)
 
 		for (int i = 0; i < renderers.Length; i++)
 		{
-			renderers[i].enabled = _aliveState;
+			// Cast and check cast
+			SpriteRenderer sr = renderers[i] as SpriteRenderer;
+			if (sr != null)
+			{
```
DooblyNoobly: Is this needed? renderers should only have SpriteRenderers in them anyway

UnityProject/Assets/Scripts/Objects/VisibleBehaviour.cs
```diff
@@ -93,7 +106,19 @@ private void UpdateState(bool _aliveState)
...
+			if (sr != null)
+			{
+				if (CanBeDisabled(sr))
+				{
+					sr.enabled = _aliveState;
+				}
+			}
+			else
+			{
+				renderers[i].enabled = _aliveState;
```
DooblyNoobly: This is weird because you are saying that if renderers does not contain a renderer then disable the renderer that according to this condition should not contain a renderer. Very confusing. You don't need this at all. Just remove the whole condition and check if:
```
if(CanBeDisabled(renderers[i]){
renderers[i].enabled = _aliveState;
}
```

UnityProject/Assets/Scripts/PlayGroups/Input/InputController.cs
```diff
@@ -97,6 +98,14 @@ private void CheckAltClick()
 			}
 		}
 
+		private void CheckSuicide()
+		{
+			if (UnityEngine.Input.GetKeyDown(KeyCode.Backspace))
+			{
+				PlayerManager.LocalPlayerScript.playerNetworkActions.CmdCommitSuicide();
+			}
+		}
```
DooblyNoobly: Wouldn't this kill the player if they use backspace in chat? I don't think this should be a feature here anyway. if you want to commit suicide make it a chat command i.e "/suicide or .suicide" something like that

UnityProject/Assets/Scripts/PlayGroups/PlayerNetworkActions.cs
```diff
@@ -448,8 +458,7 @@ public void RpcSpawnGhost()
 			Camera.main.cullingMask &= ~(1 << LayerMask.NameToLayer("FieldOfView"));
```
DooblyNoobly, line 459: This is where you should set the Camera2D followcontrol dampening property, not from the closet update loop

DooblyNoobly (changes requested):

UnityProject/Assets/Scripts/Objects/VisibleBehaviour.cs
JacobSweeten: I have to type cast to pass it to CanBeDisabled, and a few fail cast. So i'm keeping this bit unless you have another solution.

JacobSweeten:
Ready for another review.

UnityProject/Assets/Scripts/PlayGroups/Input/InputController.cs
```diff
@@ -54,6 +54,11 @@ private void Update()
 			CheckHandSwitch();
 			CheckClick();
 			CheckAltClick();
+			// Useful for debugging death and ghost related issues
+			//if(Debug.isDebugBuild)
```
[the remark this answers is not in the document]
JacobSweeten: In case someone is debugging the chat. This gets uncommented if you are specifically debugging death related stuff.

UnityProject/Assets/Scripts/Objects/VisibleBehaviour.cs
```diff
@@ -93,7 +95,20 @@ private void UpdateState(bool _aliveState)
...
+			if (sr != null)
+			{
+				if (CanBeDisabled(sr))
+				{
+					renderers[i].enabled = _aliveState;
+				}
+			}
+			else
+			{
+				renderers[i].enabled = _aliveState;
```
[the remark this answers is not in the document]
JacobSweeten: I needed to get just the sprite renderers to test them for eligibility. The other renderers need no filtering.

UnityProject/Assets/Scripts/PlayGroups/Input/InputController.cs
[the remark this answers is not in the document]
JacobSweeten: Should I leave the PNA commands in, and just remove the input check?

UnityProject/Assets/Scripts/Objects/VisibleBehaviour.cs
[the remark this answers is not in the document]
JacobSweeten: Oh ok I see

JacobSweeten:
I added ghost and chaticon to the ignoredspriterenderers array

UnityProject/Assets/Scripts/PlayGroups/Input/InputController.cs
JacobSweeten: I can do that.

JacobSweeten:
How do I go about making a headless server?

JacobSweeten:
Oh that's what you meant ok.

JacobSweeten:
Tested on 200 ping headless with two clients. Had a camera follow issue, but I fixed and tested it. Should be OK for pull.

PR #896, opened 2018-01-20
Improvements on PushPull system

DooblyNoobly:
- Starting to scrub out some of the most annoying high lag bugs when it comes to push pull. Still not perfect yet but it is now 40% more robust
- Made sure that if player is pulling and object and tries to pull another that the previousily pulled object is removed from pulling correctly across network
- Added a check to  CmdTryPush that checks if object is in space or not before applying impulse
- Added more robust checks on predictive pushing (still needs work, I think the rest of the problems are due to registerTile anomalies)

## Notes
Bugs left to squash: 

- You can't push an object that has floated in space and stopped (register tile problems) you just walk straight through it
- Sometimes you can walk into an obj that is being pushed in lag spike situations (register tile problem again)

### DIFF
--- a/UnityProject/Assets/Scripts/PlayGroups/PlayerNetworkActions.PushPull.cs
+++ b/UnityProject/Assets/Scripts/PlayGroups/PlayerNetworkActions.PushPull.cs
@@ -100,7 +100,11 @@ public partial class PlayerNetworkActions : NetworkBehaviour
 		if (pushed != null)
 		{
 			var netTransform = obj.GetComponent<CustomNetTransform>();
-			netTransform.PushTo(targetPos,playerSprites.currentDirection,true, speed, true);
+			if (netTransform.IsInSpace()) {
+				netTransform.PushTo(targetPos, playerSprites.currentDirection, true, speed, true);
+			} else {
+				netTransform.SetPosition(targetPos, true, speed, true);
+			}
 		}
 	}
 }
--- a/UnityProject/Assets/Scripts/PlayGroups/PlayerNetworkActions.PushPull.cs
+++ b/UnityProject/Assets/Scripts/PlayGroups/PlayerNetworkActions.PushPull.cs
@@ -100,11 +100,7 @@ public partial class PlayerNetworkActions : NetworkBehaviour
 		if (pushed != null)
 		{
 			var netTransform = obj.GetComponent<CustomNetTransform>();
-			if (netTransform.IsInSpace()) {
-				netTransform.PushTo(targetPos, playerSprites.currentDirection, true, speed, true);
-			} else {
-				netTransform.SetPosition(targetPos, true, speed, true);
-			}
+			netTransform.PushTo(targetPos, playerSprites.currentDirection, true, speed, true);
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterTile.cs
```diff
@@ -85,7 +85,7 @@ public void OnDestroy()
 
 		public void UpdatePosition()
 		{
-			Position = Vector3Int.RoundToInt(transform.localPosition);
+			Position = Vector3Int.FloorToInt(transform.localPosition);
```
fomalsd: Are you sure it's better that way? Did it show signs of regression with RoundToInt?
Because FloorToInt works the intended way* **only** if you increment your axis values, but it does **not** if you decrement them (you should use CeilToInt instead)! 
(*By intended way I mean the situation where you want to avoid premature registering)
That's why RoundWithContext() method exists in CNT.
If it's unknown if the axis value has increased or decreased, the safest option would be to just round.

UnityProject/Assets/Scripts/Transform/CustomNetTransform.cs
```diff
@@ -396,17 +397,29 @@ private void CheckSpaceDrift()
 			Vector3Int intGoal = RoundWithContext(newGoal, serverTransformState.Impulse);
 			if (CanDriftTo(intGoal))
 			{
+				if (registerTile.Position != Vector3Int.FloorToInt(transform.localPosition)){
```
fomalsd: same FloorToInt concern, plus this thing is already done in Synchronize() (lines 354 and on)

UnityProject/Assets/Scripts/PlayGroups/PlayerNetworkActions.PushPull.cs
```diff
@@ -100,7 +100,11 @@ public void CmdTryPush(GameObject obj, Vector3 startLocalPos, Vector3 targetPos,
 		if (pushed != null)
 		{
 			var netTransform = obj.GetComponent<CustomNetTransform>();
-			netTransform.PushTo(targetPos,playerSprites.currentDirection,true, speed, true);
+			if (netTransform.IsInSpace()) {
```
fomalsd: These kind of checks should be inside CNT if you insist on them!

UnityProject/Assets/Scripts/Transform/CustomNetTransform.cs
```diff
@@ -419,6 +432,10 @@ private static Vector3Int RoundWithContext(Vector3 roundable, Vector2 impulseCon
 			0);
 	}
 
+	public bool IsInSpace(){
+		return matrix.IsSpaceAt(Vector3Int.FloorToInt(transform.localPosition));
```
fomalsd: FloorToInt again

fomalsd (commented):
Got concerns about FloorToInt usage. 
Also these checks in CmdTryPush should be inside CNT

DooblyNoobly:
Ok made all the requested changes, please review again 👍

PR #922, opened 2018-01-25
Atmospherics all ported to C# but not working yet..

Bod9001:
### Purpose
so before I starte ripping it to pieces to try and get it to work am PRing so you have something to look at

###  TODOs of Porting

- [x] Ported Atmospherics to C#
- [x]  lay out so it should run in an c# environment
- [ ]  Making it work
- [ ]  Testing it
- [ ]  Optimising
- [ ]  Cleaning up the appearance of the code

### Notes:
Currently I'm having problems with a dictionary forgetting keys..


MrLeebo:
@Bod9001 seems like there is something wrong with the encoding your IDE or OS is using; git considers your codefile to be binary. 

1) Types in C# are broken into two categories, reference types and value types. Classes are reference types, values and structs are value types. When you compare two instances of a class `inst1 == inst2`, they will only match if they are both references to the exact same instance; creating 2 instances will never match, even if you give them the same values. In this way, your `Dictionary<List<int>, Dictionary<string, float>>` monstrosity is not going to match the way you expect.

```cs
var list1 = new List<int>() {Number1, Number2};
var list2 = new List<int>() {Number1, Number2};
Assert.That(list1 != list2) // as reference types, multiple instances don't match even if they have the same values
```

2) Instead of using dictionaries for everything, you will probably see a significant performance improvement simply by defining these as fields on objects. I don't fully understand what's going on with this system but this seems like a simpler equivalent to the air tile dictionary:

```cs
public struct AirTile {
  public static const float GasConstant = 8.3144598f;
  public float Temperature;
  public float Moles;

  public AirTile(float temperature, float moles) {
    this.Temperature = temperature;
    this.Moles = moles;
  }

  public float Pressure { 
    get {
      return Moles * GasConstant * Temperature / 2 / 1000;
    }
  }

  public static readonly AirTile Obstructed = new AirTile(0, 0);
  public static readonly AirTile Space = new AirTile(2.7f, 0.000000316f);
}
```

3) Similarly for the gasses dictionaries:
```cs
public class Gas {
  public readonly string Name;
  public readonly float HeatCapacity;
  public readonly float MolarMass;

  Gas(string name, float heatCapacity, float molarMass) {
    this.Name = name;
    this.HeatCapacity = heatCapacity;
    this.MolarMass = molarMass;
  }

  public static Gas Oxygen = new Gas("Oxygen", 0.659f, 31.9988f);
  public static Gas Nitrogen = new Gas("Nitrogen", 0.743f, 28.0134f);
  public static Gas Plasma = new Gas("Plasma", 0.8f, 40f);
  public static Gas CarbonDioxide = new Gas("Carbon Dioxide", 0.655f, 44.01f);  
}
```

```cs
var gas = Gas.Oxygen;
if (gas.HeatCapacity > x) {
  // ...
}
```

That gives you more flexibility, because now you aren't limited to all of the values in your dictionary being the same type.

Bod9001:
@MrLeebo Would these Alterations also apply to the "AirMixes" Dictionaries?

MrLeebo:
Yeah. In general if you are thinking about indexing into several dictionaries for related values, that is a good candidate for a new type.


-------- Original message --------

Bod9001:
@MrLeebo  can you Iterate through the gas mixture of the Tile, so if there is Oxygen and carbon dioxide It will only iterate through those 2, Instead of every gas in the game?

PR #923, opened 2018-01-25
Database

Harcourt1:
### Purpose
Staring approach to a database for Unity Station bans and logs
The C# file is currently just for ref

- [ ]  This PR has been tested in editor


Harcourt1:
Something something something might of forgot to tab everything

Reference Material/harcourt_db/ipbans.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/python
```
PJB3005, line 1: Use `#!/usr/bin/env python3`. This is clearly Python 3 *and* you should use `env` always.

Reference Material/harcourt_db/ipbans.py
```diff
@@ -0,0 +1,69 @@
...
+	
+capture = pyshark.LiveCapture(interface='eth0')
+capture.sniff(timeout=50)
+#print len(capture)
+lst1=[]
+#IP bans
+lst2=[]
+
+def blockip(ip):
+	cmd="/sbin/iptables -A INPUT -s "+ip+" -j DROP"
```
PJB3005, line 43: You really shouldn't be formatting these commands early. At LEAST use arrays.

Reference Material/harcourt_db/ipbans.py
```diff
@@ -0,0 +1,69 @@
...
+	print cmd
```
PJB3005, line 44: This code literally doesn't work because you're mixing prints with and without parentheses.

Reference Material/harcourt_db/ipbans.py
```diff
@@ -0,0 +1,69 @@
...
+	subprocess.call(cmd,shell=True)
```
PJB3005, line 45: You shouldn't be using `shell=True` here.

Reference Material/harcourt_db/ipbans.py
```diff
@@ -0,0 +1,69 @@
...
+ 
+def blockIP1(ip):
+	cmd="/sbin/iptables -A INPUT -s "+ip+" -j DROP"
+	subprocess.call(cmd, shell=True)
+	subprocess.call("kill -9 $(/usr/bin/pgrep dumpcap)", shell=True)
```
PJB3005, line 50: Use [`os.kill`](https://docs.python.org/3/library/os.html#os.kill) instead. Also, SIGKILLing these things is just a horrible approach. This blows up if somebody is using the same executable for something else, and it's 100% an anti pattern and a problem elsewhere that you have to do this at all.

Reference Material/harcourt_db/ipbans.py
```diff
@@ -0,0 +1,69 @@
...
+	subprocess.call("/usr/bin/kill -9 $(/usr/bin/pgrep tshark)", shell=True)
+	for i in range(len(capture)):
+		pack=capture[i]
```
PJB3005, line 53: `capture` implements `__iter__` so you shouldn't use index iteration.

PJB3005 (commented):
You really shouldn't be putting binary dll dependencies in the repo. Use NuGet.

PR #924, opened 2018-01-25
Removed snoop code

Harcourt1:
Removed snooping code


Harcourt1:
Interesting that the discord webhook does not updated my pfp

PR #925, opened 2018-01-25
Create banmanager.py

Harcourt1:
Quick IP ban manager for testing and removal.

### DIFF
--- a/Reference Material/harcourt_db/banmanager.py
+++ b/Reference Material/harcourt_db/banmanager.py
@@ -1,0 +1,50 @@
+
+import sys
+import getopt
+import subprocess
+
+_name    = "banman"
+_version = "0.1"
+
+def main(argv):
+	try:
+		opts, args = getopt.getopt(argv, "ha:d:lv", ["help", "add=", "delete=", "list", "version"])
+		
+		if len(opts) == 0:
+			# Adding without -a or --add: iptables 1.2.3.4 2.3.4.5
+			if len(args) > 0:
+				opts = []
+				for arg in args:
+					opts.append(("-a", arg))
+			# Display help if no args exist
+			else:
+				opts = [("-h","")]
+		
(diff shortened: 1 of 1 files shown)


Reference Material/harcourt_db/banmanager.py
```diff
@@ -0,0 +1,50 @@
...
+	except:
+		print "blockip error (unknown)."
+		sys.exit(2)
+
+def version():
+	global _name, _version
+	return _name + " " + _version
+
+def help():
+	print version()
```
PJB3005, line 46: You should be using Python 3 already man.

Reference Material/harcourt_db/banmanager.py
```diff
@@ -0,0 +1,50 @@
...
+				subprocess.call(["iptables", "-L", "-n"])
+	except getopt.GetoptError, err:
+		print "blockip error: " + str(err)
+		sys.exit(2)
+	except:
+		print "blockip error (unknown)."
+		sys.exit(2)
+
+def version():
+	global _name, _version
```
PJB3005, line 42: `global` is only necessary when you're modifying global variables.

Reference Material/harcourt_db/banmanager.py
```diff
@@ -0,0 +1,50 @@
...
+import sys
+import getopt
+import subprocess
+
+_name    = "banman"
+_version = "0.1"
+
+def main(argv):
+	try:
+		opts, args = getopt.getopt(argv, "ha:d:lv", ["help", "add=", "delete=", "list", "version"])
```
PJB3005, line 11: Any reason you're not using argparse here?

PJB3005 (commented):
Outside of iptables being an awful way to manage IP bans, this doesn't include a way to save the iptables config so bans get cleared upon rebooting your system.

PR #927, opened 2018-01-26
Atmospherics in c# (not optimised)

Bod9001:
yay It works now

- [x] Ported Atmospherics to C#
- [x]  lay out so it should run in an c# environment
- [x]  Making it work
- [x]  Testing it (Tested a tiny bit)
- [ ]  Optimising
- [ ]  Cleaning up the appearance of the code

### Notes:
Runs for me About  0.3 Seconds per Update so Definitely need some optimising


Bod9001:
@Ornias1993 it is Fixed, It's comparing a binary file that was Originally here to a non-binary file that is being PRed so it has nothing to compare with ues "c# atmospherics is working!" if You want to see it

PR #928, opened 2018-01-26
Dragon Key

Harcourt1:
image

### Purpose
DRAGON KEY is a way to generate a unique key for users that should be un-spoof-able.

There are three kinds dragon keys

HARD DRAGON 
SOFT DRAGON
NET DRAGON

Hard Dragon is based on the current computer's hardware. 

Soft Dragon is based on software currently installed and data around that software.

Net Dragon is based on your network.

### Hard Dragon
HD (HARD DRAGON) checks your win32 keys and through them generates a unique multi-section id.
an example key will look like this. C# code for HD is in hdgen.cs

1F9E-CE97-7005-5FE0-6E9B-7587-F3XZ

The things it checks should not / cannot change unless the user goes about changing every part on their PC.

HD generates a key every time Unity Station is launched and checks it against a list of banned DKs.

### Soft Dragon

SD (Soft Dragon) generates a key based on installed software on the users computer. It does not grab software names or other sensitive info just their EstimatedSize or something else.

Soft Dragon keys are expected to change so it is more used to guess the chance that a machine is the same as a banned machine.  To be used in conjuction with HD and ND.

### Net Dragon
Generates a key based on network info like DNS, IP, PROVIDER, etc
I am not fond of caring so much about IP when trying to ban a machine, because personally my IP changes on a daily basis.

### TODO

- [ ] Add a SQLite3 database for banned DKs
- [ ] Quick C# or Python for actually denying access based on DK ban list
- [ ] Master Key implementation to validate keys 

### Notes:
Possible missing library I need to run a bit more tests.


PJB3005:
Hang on.

This is ran on the client?

So... I can just build a modified version of this, copy paste it into my game files, and avoid bans?

Doing *any* sort of finger printing on the client is a losing battle *especially* on an open source project.

Ornias1993:
Well no.
You would not be able to do that.

PJB3005:
And, why wouldn't I be able to do that (provided it's being ran on the client?)

Ornias1993:
This is an outdated PR already, but in regards to the current version (as this is still: Reference material, not ingame content):

If I need to explain you why it wouldn't work, I think you are not into this codebase enough to question its security and I am certainly not going to tell every scriptkiddo that cant find that out themselves what those other security measures are.

PJB3005:
> This is an outdated PR already

Literally the only thing that happened to this code after this PR was slight restructuring, at best.

> (as this is still: Reference material, not ingame content):

I can see that, but it still begs the question of why it's even in the main repo? This argument just seems like "I'm hitting merge on everything that seems useful".

> If I need to explain you why it wouldn't work, I think you are not into this codebase enough to question its security and I am certainly not going to tell every scriptkiddo that cant find that out themselves what those other security measures are.

That's literally security through obscurity. Since this is an AGPL project there's *nothing* you can hide. If you really have a trick up your sleeve it won't last two seconds and you should discuss it instead of going "private hands only!". "Scriptkiddos" will just download a CID scrambler patch from the internet and call it a day, which they're *already doing on SS13*

Ornias1993:
> Literally the only thing that happened to this code after this PR was slight restructuring, at best.

Look again

> I can see that, but it still begs the question of why it's even in the main repo? This argument just seems like "I'm hitting merge on everything that seems useful".

Because this or something very close is going to be implemented after some carefull vetting and restructuring.

> That's literally security through obscurity.

yes it is, in the same way as i am not going to disclose all known loopholes.

> Since this is an AGPL project there's nothing you can hide. 

Its not hidden, i am just not going to tell you what feature it is. Open source does not means: you are free to find out by yourself.

> If you really have a trick up your sleeve it won't last two seconds

I have no secret "trick" up my sleeve, I just wont tell you every part of the security stack if you are not able to find out the basics yourself

> you should discuss it instead of going "private hands only!"

I never said that.

> "Scriptkiddos" will just download a CID scrambler patch from the internet and call it a day, which they're already doing on SS13

Hope they do, that means my final design will pay off bigtime 👍

outbools:
I don't think any security measure will even be 10% effective in free open source game. That's cool you did some obfuscation in a public repo, but once someone find the way to read it will became public knowledge anyway rendering any attempts at forcing this HARD DRAGON security worthless. There is a little you can do security-wise in AGPL software and sadly all these things can be figured out _quickly_. Look at closed source BYOND - it failed to guarantee 100% bans. In fact, ban-evading in free games is a big problem which doesn't have 100% soultion! 
The only thing to stop that is a paywall, which is not your case.

IMO IP ban is a best solution with best worth/effort ratio in this situation.

Ornias1993:
This feature is not (yet) implemented in any shape or form.
I'm leaving it at that.

PJB3005:
@Ornias1993

> Look again

1. The only PR related to these files after this was #930, and that did nothing to the files except change how they output their data (to be broken without further code mind you)
2. It's kinda difficult when these are your merge standards. https://i.imgur.com/oyP1Xai.png

> Because this or something very close is going to be implemented after some carefull vetting and restructuring.

Then it can still totally be in a separate repo. It has no organizational reason to be in here. Also no offense but I'm having doubts about your "careful" vetting.

> yes it is, in the same way as i am not going to disclose all known loopholes.

Putting your head in the sand about this stuff is a terrible approach, because I can punch a hole in this crap in 15 minutes. Acknowledge the issues publicly. You're not fooling anybody except yourself.

> Its not hidden, i am just not going to tell you what feature it is.

So if you're not telling me, you're *hiding* it from me? Damn.

>  Open source does only means: you are free to find out by yourself.

Sure I'll do that when I'm making a ban evasion tool. Except I'm not going to do that. Somebody else will.

> I have no secret "trick" up my sleeve, I just wont tell you every part of the security stack if you are not able to find out the basics yourself

I totally am able to figure that out *if I wanted to*, but I don't want to. Either way client side finger printing is worthless.

> I never said that.

Yes, not literally, but practically you did. If you didn't try to hide this behind private hands you could've just *had a discussion* instead of letting this devolve into security through obscurity discussions.

> Hope they do, that means my final design will pay off bigtime 👍

🤔

PR #931, opened 2018-01-26
DK Cross-Platform

Harcourt1:
Now cross platform for HD. SD and ND are next on my list.

### DIFF
--- a/Reference Material/dragonkey/hdgen.cs
+++ b/Reference Material/dragonkey/hdgen.cs
@@ -7,14 +7,14 @@ using System.IO;
 
 //Dragon Key
 //Hard Dragon
-OperatingSystem os = Environment.OSVersion;
-PlatformID     pid = os.Platform;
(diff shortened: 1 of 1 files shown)


Reference Material/dragonkey/hdgen.cs
```diff
@@ -7,14 +7,26 @@
...
     {
         public string HD()
         {
+		OperatingSystem os = Environment.OSVersion;
+		PlatformID     pid = os.Platform;
+		case PlatformID.WinCE:
 			return Value();
+			break;
+		case PlatformID.Unix:
+			break;
```
fomalsd: you'll never `return` stuff if you `break` first
and you don't need to `break` after you `return`

Harcourt1: noted

Harcourt1:
Ready 2 Go

PR #936, opened 2018-01-27
Database Logging

Harcourt1:
Note: Change file paths and include an actual DB file. The DB file I was using is now on a computer that is dead & I will upload it later. 
Delete ph files when you can.


PJB3005:
Literally all this PR seems to do is add binary dependencies that should've been using nuget.

PR #944, opened 2018-01-31
Fixed #905

### LINKED ISSUE #905: Changing input mode to AZERTY and then pressing A or D causes ambient volume slider to change
# Preliminary information

## Category
BUG
0.2.2 beta 5 steam
# Report
## Current Behavior
After changing input mode to AZERTY if you press the A or D key it will cause the ambient volume slider to change.
Ambient volume slider not to change when pressing A or D after being toggled on.

## Steps to Reproduce
1. Click AZERTY input mode toggle
2. Press A or D key once or more
3. Ambient volume slider will either increase when pressing D or decrease when pressing A

## OS used:
Windows
## System specifications
i7-4790K, gtx980, 16gb, HDD
## Network
Beta server 90-110ms

mikeslet:
### Purpose
Fixes #905, which is caused by Unity's InputManager affecting windows. The bug affected more than just AZERTY mode, it also affected QWERTY and if used with TAB could cause movement to other Upper Right Panel tabs. 

### Approach
Removed InputManager WASD and Arrow key inputs, which only affect Unity windows (because we have our own inputs through scripts). 

### Notes:
Would appreciate verification that it works fine in multiplayer, but I see no reason it would not.


DooblyNoobly:
Squashing this one because of all the merge commits in your PR

PR #946, opened 2018-01-31
Introduction of a Meta Data Layer

krille90:
### Purpose
To implement systems like Atmos, we need a way to save meta data. This is the first approach to accomplish this.

### Approach
Adds a MetaDataLayer and a RoomControl, which initializes rooms, or more precisely Inside vs Outside. Just standing outside the station will now kill you.

### Notes:
solves #739


DooblyNoobly (approved):
looks good

PR #960, opened 2018-02-11
Localchat in range only

DooblyNoobly:
### Purpose
Makes sure that localchat messages are only received by players nearby and in the same room

### Approach
First check the distance and then raycast to the player to see if there are any walls in the way

### Notes:
Also fixed a bug with the closets where you could not walk onto them when they were opened. This was caused by some tms changes


UnityProject/Assets/Scripts/Managers/ChatRelay.cs
```diff
@@ -71,6 +72,28 @@ public void AddToChatLogClient(string message, ChatChannel channels)
...
+			var speaker = PlayerList.Instance.Get(chatEvent.speaker);
+			RaycastHit2D hit;
+			LayerMask layerMask = 1 << 9; //Walls layer
+			for (int i = 0; i < players.Count(); i++){
+				if(Vector2.Distance(speaker.GameObject.transform.position,
+				                    players[i].GameObject.transform.position) > 14f){
+					//Player in the list is too far away for local chat, remove them:
+					players.Remove(players[i]);
+				} else {
+					//within range, but check if they are in another room or hiding behind a wall
```
fomalsd, line 87: could probably use Pulsey's room detection stuff here and remove people whose location's room ids are different to the speaker's

krille90: but that would mean that people standing on both sides of an open door couldn't talk via local chat. The current room detection only happens at the beginning and doesn't update with doors opening / closing.

fomalsd: Oh, good point!
I think we should only make them recalculate on wall placement/deconstruction – when we implement building

PR #986, opened 2018-03-05
Modified AdjustedDirection to recognize replay moves

### LINKED ISSUE #961: Fix the Prediction Queue when ServerState lags behind PredictedState B: 80
# Preliminary information
In connections with high ping sometimes the localPlayer will receive a ServerState update that is a few moves ahead of the predicted state. This makes the player lerp back a few move positions before lerping forward to the position it should be at

## Category
TODO: Figure out why predicted state tries to lerp towards the serverstate when it is a behind predicted state in MoveNumber order. Predicted state should be cycling through the pendingActions to overcome the lag but it seems to fall short.

If the ServerState is lagging behind the PredictedState should be able to use the trimmed pendingActions to cycle through the last actions and then add the next action to the prediction (this is not happening)

## Steps to Reproduce
1. Build with net simulator on at 200ms and play as client
2. Pull a closet and run in random directions really fast
3. Start pushing the closet and sometimes you will lerp in to the closet closet

DooblyNoobly:
- When replaying moves due to the serverState being behind the predictedState all of the matrix checks are done again while the player is no longer at the positions it is checking from. Separated AdjustedDirection to do certain checks when it is a replay action or not. The lerp back problems of high lag players mentioned in #961 has been resolved but it exposes a lot of problems with Matrix Detection and Object pulling

### Purpose
Fixes #961

### Notes:
Exposes underlying problems with PushPull and Matrix detection

### DIFF
--- a/UnityProject/Assets/Scripts/PlayGroups/PlayerMove.cs
+++ b/UnityProject/Assets/Scripts/PlayGroups/PlayerMove.cs
@@ -78,10 +78,10 @@ namespace PlayGroup
 			return new PlayerAction { keyCodes = actionKeys.ToArray() };
 		}
 
-		public Vector3Int GetNextPosition(Vector3Int currentPosition, PlayerAction action)
+		public Vector3Int GetNextPosition(Vector3Int currentPosition, PlayerAction action, bool isReplay)
 		{
 			Vector3Int direction = GetDirection(action);
-			Vector3Int adjustedDirection = AdjustDirection(currentPosition, direction);
+			Vector3Int adjustedDirection = AdjustDirection(currentPosition, direction, isReplay);
 
 			if (adjustedDirection == Vector3.zero) {
 				Interact(currentPosition, direction);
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/PlayGroups/PlayerMove.cs
```diff
@@ -193,69 +193,79 @@ private Vector3Int GetMoveDirection(KeyCode action)
...
+			if (!isReplay) {
+				
+				Vector3 newRayPos = transform.position + direction;
+				rayHit = Physics2D.RaycastAll(newRayPos, (Vector3)direction, 0.2f, hitCheckLayers);
+			//	Debug.DrawLine(newRayPos, newRayPos + ((Vector3)direction * 0.2f), Color.red, 1f);
+
+				//Detect new matrices
+				for (int i = 0; i < rayHit.Length; i++) {
+					//checks to see if the matrix has changed
+					if (rayHit[i].collider.gameObject.layer == 24
```
fomalsd, line 215: We should really use some enum instead of these magic numbers

UnityProject/Assets/Scripts/PlayGroups/PlayerMove.cs
```diff
@@ -193,69 +193,79 @@ private Vector3Int GetMoveDirection(KeyCode action)
...
+						curMatrixCol = rayHit[i].collider;
+						ChangeMatricies(rayHit[i].collider.gameObject.transform.parent);
+						Debug.Log("Change Matricies");
+					}
 
-				//Door closed layer
-				if (rayHit[i].collider.gameObject.layer == 17) {
-					DoorController doorController = rayHit[i].collider.gameObject.GetComponent<DoorController>();
+					//Detected windows or walls across matrices or from space:
+					if (rayHit[i].collider.gameObject.layer == 9
```
fomalsd, line 223: Maybe we should detect "anything solid" instead?

UnityProject/Assets/Scripts/PlayGroups/PlayerMove.cs
```diff
@@ -193,69 +193,79 @@ private Vector3Int GetMoveDirection(KeyCode action)
...
+					}
 
-					// Attempt to open door that could be on another layer
-					if (doorController != null && allowInput) {
-						pna.CmdCheckDoorPermissions(doorController.gameObject, gameObject);
-						allowInput = false;
-						StartCoroutine(DoorInputCoolDown());
+					//Door closed layer (matrix independent)
+					if (rayHit[i].collider.gameObject.layer == 17) {
+						DoorController doorController = rayHit[i].collider.gameObject.GetComponent<DoorController>();
```
fomalsd, line 230: Are we going to list all possible interactions here? (like, what if there is a pushable on the matrix border?) Sounds like a potential maintenance hell

fomalsd (approved):
No questions regarding replay, but matrix switching code needs careful consideration (though it's not a part of the task)

UnityProject/Assets/Scripts/PlayGroups/PlayerMove.cs
DooblyNoobly: The reason I haven't yet is because the number and its layer name can be found so easy by just hitting the layer drop down on any object. But we can have a public enum to make it easier in the future

UnityProject/Assets/Scripts/PlayGroups/PlayerMove.cs
DooblyNoobly: It was a rough draft, trying to detect things on different layers. We def need a better solution and soon

PR #991, opened 2018-03-08
The Nuclear Option

rmhenn:
### Purpose
Added a functioning nuclear weapon usable only by the syndicate, and restricted by a passcode sent to the syndies via their radios
### Approach
1. Created a new prefab, the nuke
2. Created a script to deal with the interaction between the player and the nuke, and the subsequent functions of a nuclear weapon
3. Created a script to control the UI and pass the entered code back to the nuke to validate it

### Notes:
Paypal: noodledanger9@gmail.com


DooblyNoobly (changes requested):
Nevermind, found the nuke prefab

UnityProject/Assets/Scripts/Objects/NukeInteract.cs
```diff
@@ -0,0 +1,116 @@
...
+    public string deniedMessage;
+    [SyncVar]
+    private int nukeCode;
+    private GameObject Player;
+    void Start()
+    {
+        //calling the code generator and setting up a 10 second timer to post the code in syndicate chat
+        codeGenerator();
+        StartCoroutine(WaitForShit());
+    }
```
DooblyNoobly, line 22: You want this to be:
``
public override void OnStartServer(){
base.OnStartServer();
}
``
So that this start method only runs on the server

UnityProject/Assets/Scripts/Objects/NukeInteract.cs
```diff
@@ -0,0 +1,116 @@
...
+using UnityEngine.Networking;
+using UnityEngine.Video;
+
+public class NukeInteract : InputTrigger
+{
+    public float cooldownTimer = 2f;
+    public string interactionMessage;
+    public string deniedMessage;
+    [SyncVar]
+    private int nukeCode;
```
DooblyNoobly, line 15: Probably don't want to be sharing this with every client. Just keep nukeCode on the server and have the nuke_gui send code attempts to the server to be checked against it

UnityProject/Assets/Scripts/Objects/NukeInteract.cs
```diff
@@ -0,0 +1,116 @@
...
+        else
+        {
+            //if no, say bad message
+            UIManager.Chat.AddChatEvent(new ChatEvent(deniedMessage, ChatChannel.Examine));
+        }
+        //posts nuke code for testing
+        Debug.Log(nukeCode);
+    }
+    [Server]
+    //The serverside of the interaction code
```
DooblyNoobly, line 51: Putting a comment after an Attribute voids the Attribute I'm pretty sure

UnityProject/Assets/Scripts/Objects/NukeInteract.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        return true;
+    }
+    IEnumerator WaitForDeath()
+    {
+        yield return new WaitForSeconds(5f);
+        GibMessage.Send();
+    }
+    [Server]
+    //Server validating the code sent back by the GUI
```
DooblyNoobly, line 68: put comment before attribute

UnityProject/Assets/Scripts/Objects/NukeInteract.cs
```diff
@@ -0,0 +1,116 @@
...
+            return true;
+        }
+        else
+        {
+            //if no, tell the GUI that it was an incorrect code
+            return false;
+        }
+    }
+    [ClientRpc]
+    //Server telling the nukes to explode
```
DooblyNoobly, line 85: same here

UnityProject/Assets/Scripts/Objects/NukeInteract.cs
```diff
@@ -0,0 +1,116 @@
...
+        VideoPlayer Video = UIManager.Display.video;
+        Video.Play();
+        //Playing the sound
+        AudioSource audio = GetComponent<AudioSource>();
+        audio.Play();
+        //KILLING EVERYONE!!!!!1!
+        StartCoroutine(WaitForDeath());
+    }
+    [Server]
+    //Generating the code for the nuke
```
DooblyNoobly, line 108: and here

UnityProject/Assets/Scripts/PlayGroups/PlayerMove.cs
```diff
@@ -78,10 +78,10 @@ public PlayerAction SendAction()
 			return new PlayerAction { keyCodes = actionKeys.ToArray() };
```
DooblyNoobly, line 78: All these commits shouldn't be in the PR. I suggest cut and pasting your changes for Nuke PR and starting again on a new branch with an updated fork. Remember to always create a branch for your feature and only rebase it if you want the latest changes so that the history of other commits are not added

UnityProject/Assets/Scripts/UI/Nuke/GUI_Nuke.cs
```diff
@@ -0,0 +1,37 @@
...
+    //what needs to happen when the button is clicked
+    public void BtnOk()
+    {
+        SoundManager.Play("Click01");
+        UIManager.Display.nukeWindow.SetActive(false);
+        GameObject nuke = GameObject.Find("nuke");
+        NukeInteract ni = (NukeInteract)nuke.GetComponent("NukeInteract");
+        string codestring = code.text;
+        //send the code to NukeInteract
+        ni.validate(codestring);
```
DooblyNoobly, line 22: A NetworkMessage should be created (Derive from ClientMessage, see other messages for examples) and then used here to take the inputted code and send it to the server to be validated. This will stop people from hacking as in this PR the code is also synced on clients.

DooblyNoobly (changes requested):
Not a bad attempt. A few things need to be fixed before I can merge this. 

The history of this PR is bad. You should take the changes you've done and start again on a new branch with an updated fork. If you need to pull changes from origin, rebase your branch instead, or alternatively squash your commits before submitting a PR.

The Managers.prefab is missing, so the actual window does not show up and the video does not play.

The code is being shared with all clients so this can easily be hacked. Keep it server side and instead send the code input attempts to the server for validation. 

Now to communicate the code to the syndies you could put a check in Equipment.cs at around line 117 after the job outfit has been determined. This part of the code is only run on the server. So here you can do a scan of the scene for a NukeInteract component and if you find one then get the Code from it (it is fine to make the nukeCode public on NukeInteract now that it is only sitting on the server). Then use a NetworkMessage to communicate that to the player (either a already built chat one or build your own and use the reporttochannel on the client side).

I'll also increase the bounty for this by an extra $10 as it has been a bit of work. Any problems ping me on discord.

UnityProject/Assets/Scripts/Objects/NukeInteract.cs
fomalsd: Yeah, SyncVar means "this info is sent to everyone", definitely not suitable here

PR #1024, opened 2018-04-23
Space death fix

krille90:
### Purpose
Death in space not working. Need to get rid of the "Vacuum of space conspiracy" before it's too late.

### Approach
MetaDataLayer created a new MetaNode, which is by default not Space. Fixed this by not creating that node. If no node exists, it must be space.


fomalsd:
Seems to work

Issue #1018, opened 2018-04-20
Start Menu fixes


krille90:
## Current Behavior
Currently the Start Menu, isn't really comfortable, e.g. no "back" or "cancel" buttons. 

## Expected/Wanted/Requested Behavior
Make the start menu more comfortable.  These are just some small changes, have a look at the menu and add some buttons, so you can't get stuck in the menu.

PR #1034, opened 2018-04-30
[DNM] Near Fix of #1018 | Adds a back button to the job selection menu

Bonfire:
Nearly fixes the issue referenced in #1018, adding a back button to the job selection menu. Issue when re-hosting a server.

### Purpose
Adds a back button so that a user can re-name themself or connect to a different server.

Screenshot:
unity_2018-04-30_18-47-41

### Approach
Created a new button that is later added to the job menu. When the button is clicked, we hide the job menu and attempt to change the scene back to the lobby. If the user is hosting a server, that server is stopped.

### Notes:
**This PR introduces a new bug, so please do not merge**. To reproduce this bug you need to host a server, click back, and attempt to host a server again. Unity says that the Tilemap you're trying to access has been destroyed, when it really shouldn't be. Asking for assistance in fixing this.

### DIFF
--- a/UnityProject/Assets/Scripts/Items/ItemAttributes.cs
+++ b/UnityProject/Assets/Scripts/Items/ItemAttributes.cs
@@ -521,7 +521,15 @@ public class ItemAttributes : NetworkBehaviour
 
 	public void OnMouseEnter()
 	{
-		UIManager.SetToolTip = itemName + " (" + itemDescription + ")";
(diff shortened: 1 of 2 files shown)


DooblyNoobly:
It looks like the server hasn't actually shut down properly. When you reconnect you'll notice that your previous player is still there (look at the connected player list). Strange

Bonfire:
I think this fix is a little too sophisticated with the current server system. Closing this.

PR #1036, opened 2018-05-05
Fix for #836, forced window aspect.

### LINKED ISSUE #836: Cannot maximize window completely
# Preliminary information
If I drag the window to the top of my screen to "snap" it to full screen or press the maximize button, it will, for a second, maximize, then it goes back to being slightly smaller than my actual screen size.

## Category
BUG

Beta V2, but I'm pretty sure it's been happening for a while on various versions.

# Report
## Current Behavior
See Preliminary Information

Window should snap to full screen

## Steps to Reproduce
1. Attempt to maximize the screen
2. ???
3. Not profit.

## OS used:
Windows

## System specifications
i3-3220, GTX 1060 6GB, 8GB RAM, 1920x1080 24" AOC monitor.

## Network
Not relevant

NinjaDanz3r:
ForceGameWindowAspect() now only resizes the game window, not the entire window.

### Purpose
Fixes #836.

### Approach
I first tried to isolate the cause of the issue, then removed the code that was causing the entire window to resize, which turned out to be a call to `Screen.SetResolution(...)`.
Removing this line and the additional lines providing the arguments to the function seems to fix the problem reported in issue #836.

- [x]  This PR has been tested in standalone build

### Notes:
As I am new to the project there might be unintended consequences to it that I'm not seeing, please notify me if there is anything that seems off with this fix.

### DIFF
--- a/UnityProject/Assets/Scripts/UI/ResponsiveUI.cs
+++ b/UnityProject/Assets/Scripts/UI/ResponsiveUI.cs
@@ -109,22 +109,6 @@ namespace UI
 			yield return new WaitForSeconds(0.2f);
 			if (!Screen.fullScreen)
 			{
-				float screenWidth = Screen.height * targetAspect;
-
-				//The following conditions check if the screen width or height
-				//is an odd number. If it is, then it adjusted to be an even number
-				//This fixes the sprite bleeding between tiles:
-				if ((int) screenWidth % 2 != 0)
-				{
-					screenWidth += 1f;
-				}
-				int screenHeight = Screen.height;
-				if (screenHeight % 2 != 0)
-				{
-					screenHeight++;
-				}
-
-				Screen.SetResolution((int) screenWidth, screenHeight, false);
 				if (camResizer != null) {
 					camResizer.AdjustCam();
 				}


UnityProject/Assets/Scripts/UI/ResponsiveUI.cs
```diff
@@ -109,22 +109,6 @@ private IEnumerator ForceGameWindowAspect()
...
-				//This fixes the sprite bleeding between tiles:
-				if ((int) screenWidth % 2 != 0)
-				{
-					screenWidth += 1f;
-				}
-				int screenHeight = Screen.height;
-				if (screenHeight % 2 != 0)
-				{
-					screenHeight++;
-				}
```
DooblyNoobly, line 125: These lines that check the screenHeight and screenWidth are making sure that the Height and Width are even numbers because if the they end up being odd numbers then you see gaps in between the tiles in the game. Removing this will reintroduce that bug.

DooblyNoobly (changes requested):

UnityProject/Assets/Scripts/UI/ResponsiveUI.cs
NinjaDanz3r: Ah, alright, I did a small check and I can confirm that bleedthrough happens when the Screen size isn't divisible by two.
However, using `Screen.SetResolution(...)` isn't good either as it prevents the window from fully resizing.
A possible solution would be to use layout groups and padding in order to avoid bleedthrough and still enable fullscreen window, but I will have to read up on those topics.

PR #1047, opened 2018-05-10
Updated the names of lots of assets

Bonfire:
### Purpose
Renamed most machines from "vending_#" to an actual name. Names were pulled from other projects. Some vending machines still don't have names, and it is possible I mis-named one or two. Unity doesn't allow multiple assets to have the same name, so some assets have custom names to differentiate them (Instead of ClothesMate, we have HatMate, ShirtMate, etc.). On top of that, many, many assets were renamed to their proper names within the editor.

### Approach
Simply renamed the assets in the project, using previous projects as reference

### Notes:
Some item tooltips are not visible in game (don't know whether or not this is just an issue on my part). As a result, _some_ asset names may have some spacing issues that I wasn't able to catch. For the most part though, this shouldn't be the case.

### DIFF
--- a/UnityProject/Assets/Prefabs/Items/Accessories/Resources/SpaceCigarettes.prefab
+++ b/UnityProject/Assets/Prefabs/Items/Accessories/Resources/SpaceCigarettes.prefab
@@ -1,0 +1,260 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!1001 &100100000
+Prefab:
+  m_ObjectHideFlags: 1
+  serializedVersion: 2
+  m_Modification:
+    m_TransformParent: {fileID: 0}
+    m_Modifications: []
+    m_RemovedComponents: []
+  m_ParentPrefab: {fileID: 0}
+  m_RootGameObject: {fileID: 1645856876941636}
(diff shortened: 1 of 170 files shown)


DooblyNoobly:
Nice! I'll have to check this one out before approving

Bonfire:
Added a lot more item names and attempted to follow the naming conventions for files properly. The way that file names are translated into in-game is very abstract. Let me know if there are any issues with it or if I need to push over some missing files. I've also added lots of items to their appropriate layers within the editor.

DooblyNoobly:
looks good!

fomalsd:
you know what's wrong with this one? 
it doesn't rename old ones, just the new ones were created and old prefabs are still in place!
default

DooblyNoobly:
god dammit, a commit with them removed should work?

Bonfire:
Hahaha, I knew it was missing something. Unity's renaming likes to create new assets it seems. I'll correct this.

fomalsd:
Just make sure that scene would reference the correct, renamed prefabs

Bonfire:
I think rolling back the commit (at least for the time being) would be best.

DooblyNoobly:
will revert this one

PR #1077, opened 2018-05-21
Added two new music scores to the title screen

Bonfire:
### Purpose
This pull will add two new music scores to the title screen. The score list would then be comprised of Title01, Title02, Title03, and ClownTitle.

### Approach
The music files were already included in the project, so I simply added them to the title screen. By editing the SoundManager prefab, I was able to add the two new songs and increase the set number of tracks.

### Notes:
No code is able to be commented, but I will leave it unchecked. Also, due to the nature of the feature, I don't think that it requires multiplayer testing, but I will leave that unchecked just in case as well.


DooblyNoobly:
thanks, will check it in a few hours

Issue #970, opened 2018-02-18
Server authorative matrix movement B: 50


DooblyNoobly:
# TODO:
Now that multiple matrices are working it is time to implement the matrix movement system and to sync them over network with the clients. As we don't need to worry about prediction all that is required is a deterministic way to move the matrices on the server and then to update the clients 

Collisions of matrices will be handled in a later TODO and so will the shuttle controls

## Criteria
1. Create a matrix move component and put it on the parents of the separate matrices
2. Develop a movement system that is server authoritative making sure the movement happens on the server first and then it syncs with the clients. You could create a Vector3.MoveTowards that is always running and sync the targets + speed.
3. Make sure that when a new player joins they get an updated stat of the matrix positions and its target + speed.
4. Test the system on a headless server with 2 clients (Creating a test input command to trigger the test)

Bounty of 50usd will be paid to the contributor on a successful merge

PR #1082, opened 2018-05-26
Shuttle UI

### LINKED ISSUE #971: Build new shuttle control console B: 150
# TODO

Build the console computer that will sit in the bridge of each shuttle. When the console is clicked on it should show a UI window with the shuttle controls on it. At the moment we need a control system that takes a waypoint that the shuttle can automatically guide its way too. In the future we will add the feature for manually controlled shuttles.

A radar window should be shown that shows near by objects (just shows the other matrices positions) and the radar window should be clickable to set waypoints for the shuttle. A fly, no fly button should be placed under the radar to toggle the ships flight progress.

## Criteria:
1. Build the Shuttle control console object, using sprites from the codebase, create a ShuttleControl component and derive it from InputTrigger so it can be clicked on.
2. Create a UI window and place it in the Windows parent underneath UIManager in the Manager prefab.
3. Have the Window show when someone clicks on the control console and have the window disappear if you walk away from the console
4. Create all of the UI elements to control the shuttle with. Window should know the current shuttle it is controlling and it should send movement requests to the server (To the serverside of the MatrixMove component) to be validated.
5. Ensure there are necessary checks on the server to prevent hacking
6. Create all of the control elements and graphics and have them operational
7. Connect into the MatrixMove system which is being implemented in #970 
8. Test on headless server with 2 clients

Bounty of 150usd will be paid to the contributor on a successful merge

NinjaDanz3r:
### Purpose
This PR is mainly a request for comments and is not complete in its current state.
Adds a functional shuttle UI and a shuttle console tile that can be interacted with in order to bring up the UI.
[Obligatory GIF](https://i.imgur.com/lr09Hmt.gifv)
Resolves #971.

### Approach
Trial and error, looking through the wiki and reading through the code.
@fomalsd recommended that I do this as a tab instead of a window, and thus the layout had to change in order to accommodate the smaller screen space.

- [ ]  This PR does not include files without specific need to do so.

### Criteria
- [x] Build the Shuttle control console object, using sprites from the codebase, create a ShuttleControl component and derive it from InputTrigger so it can be clicked on.
- [ ] Create a UI window and place it in the Windows parent underneath UIManager in the Manager prefab.
- [x] Have the Window show when someone clicks on the control console and have the window disappear if you walk away from the console
- [ ] Create all of the UI elements to control the shuttle with. Window should know the current shuttle it is controlling and it should send movement requests to the server (To the serverside of the MatrixMove component) to be validated. _(Might be fixed, not entirely sure. I am using a reference to a MatrixMove component but that reference might be only client-side. I will look into what's needed to make this serverside.)_
- [ ] Ensure there are necessary checks on the server to prevent hacking
- [x] Create all of the control elements and graphics and have them operational
- [x] Connect into the MatrixMove system which is being implemented in #970
- [ ] Test on headless server with 2 clients

### Notes:


UnityProject/Assets/Scripts/UI/ControlTabs.cs
```diff
@@ -67,6 +74,11 @@ private void SelectWindow(WindowSelect winSelect)
 					moreTab.image.color = selectedColor;
 					panelMore.SetActive(true);
 					break;
+				case WindowSelect.shuttleControl:
+					UnselectAll();
+					moreTab.image.color = selectedColor;
```
fomalsd, line 79: ShuttleControlTab.image.color

UnityProject/Assets/Tilemaps/Palettes/Static Objects.prefab
```diff
@@ -149,7 +149,7 @@ Tilemap:
       m_TileColorIndex: 0
       m_ObjectToInstantiate: {fileID: 0}
       m_TileFlags: 0
-      m_ColliderType: 0
+      m_ColliderType: 2
```
fomalsd, line 152: hm, any idea what that means?

UnityProject/Assets/Scripts/Shuttles/ShuttleInteract.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    public string interactionMessage;
+    public MatrixMove ShuttleMatrixMove;
+    private GameObject player;
+    
+    // Method that brings up shuttle controls.
+    public override void Interact(GameObject originator, Vector3 position, string hand)
+    {
+        player = originator;
+        ControlTabs.ShowShuttleTab(ShuttleMatrixMove);
```
fomalsd, line 15: In this case tab opening should be granted by server.
In fact, server should know when it's opened on clients and when it's not, it would be required for pushing radar updates etc.
So [Command] should be introduced here, or InteractMessage should be sent.
(At this point it's not very intuitive, I know)
So, InteractMessage way would look like this:
```cs
	public override void Interact(GameObject originator, Vector3 position, string hand)
		{
                      var playerScript = originator.GetComponent<PlayerScript>();
                      if (playerScript.canNotInteract() || !playerScript.IsInReach( gameObject.transform.position ))
                      { //check for both client and server
                          return;
                      }
			if (!isServer)
			{ 
			        //Client wants this code to be run on server
				InteractMessage.Send(gameObject, hand);
			}
			else
			{
				//Server actions
				//you'll have to write this yourself. it would call Open() for client when Process()'ed 
                               TabActionMessage.Send( originator, console, TabAction.Open ); 
                               //... some code to mark this one as opened by someone  
			}
		}
```

And Command way would be like that:
```cs
  // Method that brings up shuttle controls.
    public override void Interact(GameObject originator, Vector3 position, string hand)
    {
        player = originator;
        CmdShowShuttleTab( gameObject, originator );
    }

    /// Request to open shuttle tab
    [Command]
    public void CmdShowShuttleTab( GameObject console, GameObject originator ) {
        var playerScript = originator.GetComponent<PlayerScript>();
        if (playerScript.canNotInteract() || !playerScript.IsInReach( gameObject.transform.position ))
        { //check for server
            return;
        }
        //you'll have to write this yourself. it would call Open() for client when Process()'ed 
        TabActionMessage.Send( originator, console, TabAction.Open ); 
        //... some code to mark this one as opened by someone  
    }

    public void Open() {
        ControlTabs.ShowShuttleTab(ShuttleMatrixMove);
    }

    public void Close() {
        ControlTabs.HideShuttleTab();
    }
....
public interface ITab
{
    void Open();
    void Close();
}
```

UnityProject/Assets/Scripts/Shuttles/ShuttleUIControl.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    public MatrixMove matrixMove;
+    private float maxSpeed = 0;
+    private bool active = false;
+
+    /// <summary>
+    /// Starts or stops the shuttle.
+    /// </summary>
+    /// <param name="off">Toggle parameter</param>
+    public void TurnOnOff(bool off)
```
fomalsd, line 15: Currently these only work for serverplayer, as server methods are called directly. They won't do anything on clients.
You should convert these to Commands so that they would run on server, for example:
```
[Command]
public void CmdTurnOnOff(bool off)
```
You should also add some validation to each of these (until we have our own networking layer all checks should be explicit): return if player `ps.canNotInteract() || !ps.IsInReach(position)`

UnityProject/Assets/Scripts/Shuttles/ShuttleInteract.cs
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    /// <summary>
+    /// Checks if the player has moved away from the shuttle. If the player has moved away it hides the tab.
+    /// </summary>
+    public void Update()
+    {
+        if ( player != null && Vector3.Distance(player.transform.position, transform.position) > 1.5f )
```
fomalsd, line 23: 1) this check is clientside-only in your case, it should be serverside as well (more on that in other comments)
2) It's unnecessary to check for that every frame, I think checking only when player moves would be better
3) use PlayerScript.IsInReach() for range checks instead

fomalsd (commented):
At the moment it only works for server, will need some additional work to make it have server authority and support for things like radar updates from server

UnityProject/Assets/Tilemaps/Palettes/Static Objects.prefab
NinjaDanz3r: Nope! Got a load of updated binaries when I added ran tile2prefab in order to add my shuttle console prefab to the tile palette.

DooblyNoobly: btw did the prefab2tile tool work? I thought there was problems with it.

NinjaDanz3r: I did not notice any immediate issues, the Console prefab was converted to a tile and I could subsequently use it.
What issues could occur as a consequence of using prefab2tile?

DooblyNoobly: Oh good. I don't think there will be any issues, just the tool itself has not been used for a long time and I think it does not create the preview sprite to put on the asset (for when you are hovering over the scene with the tile selected as a brush). But we will fix that stuff later.

krille90:
Is this still being pursued? Otherwise this PR should be closed.

krille90:
Closing because of inactivity

PR #1088, opened 2018-06-03
first atmos stuff

krille90:
### Purpose
This is the first try of implementing Bod's Atmos (see Reference Material). The code still needs clean up and improvements, but the base concept, atmospherics, exists. Atmos runs in a separate thread.
Currently, there is no interaction between atmos and other things, although no networking is happening atm. Furthermore, Atmos between Matrices isn't handled yet.


krille90:
wrong branch

PR #1095, opened 2018-06-13
Fixes players going through each other - #992

### LINKED ISSUE #992: Players aren't solid anymore (can walk through each other) B: 30
# Preliminary information

## Category
BUG

Develop

# Report
## Current Behavior
In title.
Looks like it's a regression caused by matrix switch functionality and/or changes to colliders

Players should not be able to pass freely through each other (until we have intents in, at least)

## Steps to Reproduce
1. Get two players on the server
2. Try walking through each other

## Network
All types

badbon:
### Purpose
Fixes #992

### Approach
I went through a bit of hell because I am still new to codebase, but I believe I've found a fix. 👍 Basically the past code wasn't even checking for players, it was just checking for tiles (I guess?) I made a specific checks for a registered player "tile", but then issue came up with network replays which I also had to fix by checking if that player tile is same player that called the function for player check, because player was stopping moving thinking it hit a player, but it was actually itself, just being called from network replay. (I hope this makes sense lol)

### Notes:
Going to PM my PayPal to Doobly in case this gets merged.

### DIFF
--- a/UnityProject/Assets/Scripts/PlayGroups/PlayerMove.cs
+++ b/UnityProject/Assets/Scripts/PlayGroups/PlayerMove.cs
@@ -229,7 +229,7 @@ namespace PlayGroup
 				}
 			}
 
-			if (curMatrix.IsPassableAt(currentPosition, newPos) || curMatrix.ContainsAt(newPos, gameObject)) {
+			if (!curMatrix.ContainsAt(newPos, gameObject) && curMatrix.IsPassableAt(currentPosition, newPos)) {
 				return direction;
 			}
 
--- a/UnityProject/Assets/Scripts/PlayGroups/PlayerMove.cs
+++ b/UnityProject/Assets/Scripts/PlayGroups/PlayerMove.cs
@@ -73,7 +73,7 @@ namespace PlayGroup
 				}
 			}
 
-			return new PlayerAction { keyCodes = actionKeys.ToArray() };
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/PlayGroups/PlayerMove.cs
```diff
@@ -200,49 +200,62 @@ private Vector3Int GetMoveDirection(KeyCode action)
...
-			if (playerSync.PullingObject != null) {
-				if (curMatrix.ContainsAt(newPos, playerSync.PullingObject)) {
-					return direction;
-				}
-			}
-
-			//could not pass
-			return Vector3Int.zero;
-		}
+                if (isGhost)
```
fomalsd: please use tabs for indentation instead of spaces!

badbon:
Not sure if I messed it up worse or fixed it, try now :c

DooblyNoobly:
@NicholasM10 I'm having trouble seeing where the Player check is in PlayerMove. Can you direct me to the line # of the changes (hard to tell with the indent changes)

DooblyNoobly:
Oh right you added it to the ContainsAt on Matrix

DooblyNoobly:
I tested in high lag mp. Works well. We may need a new issue for Machines and Objects

DooblyNoobly:
Also I see issues in the future when intent comes into play. But we will cross that bridge when we get there

Issue #1001, opened 2018-03-19
Pulling containers doesn't work B: 35


fomalsd:
# Preliminary information

## Category
BUG

## Bug: Release or Development?
Develop after a80a0a1357fe44b61597f657a79072e1eb3a069c (regression in that commit) of PR #968 

# Report
## Current Behaviour
Ctrl+clicking containers like closets just doesn't do anything.

## Steps to Reproduce
Please enter the steps to reproduce the bug or behaviour:
1.Come by a closet
2.Ctrl+click it while standing next to it
3.Walk away, it won't follow

## OS used:
Windows

## Network
Editor self-host

DooblyNoobly:
Criteria: 

 - Fix the bug above

A bounty of 35 usd will be paid to the contributor on a successful merge

StrawberryMoses:
I'm working on this. Push / Pull seems kind of broken in general right now as well.

StrawberryMoses:
Confirmed working in f394325, broken by #968 . Looks like OnMouseDown isn't being called anymore in PushPull for those objects

fomalsd:
Push code has been commented out in #1084 PR as it's not adapted to work across matrices and is hacky in general

dapocalypse:
If I cannot recreate can I provide proof and claim the bounty?

dapocalypse:
Nvmd

krille90:
FYI: If an issue can't be reproduced and doesn't occur anymore, the issue will be closed. In this case, no bounty will be paid.

DooblyNoobly:
yeah closing this. The problem is because push pull is dead. We need to fix it now that player move has all be sorted out. Will make another TODO

PR #1101, opened 2018-06-16
[ DNM ] Push / Pull #1001

StrawberryMoses:
### Purpose
Beginning of work on #1001. 

Fixing interactions with pushing and pulling objects, most notably containers.

### Approach

- Removed OnMouseDown from PushPull and ObjectBehaviour
  - In PushPull, OnMouseDown has changed to be called from InputController
    - InputController has been slightly refactored and cleaned to handle this
  - Previously, this would be incorrectly called when the tile the object is on was clicked
  - Cancelling the pull previously done in ObjectBehaviour when picking up is now handled in PickUpTrigger
- Added functionality to stop pulling an object when hitting forward delete key
- Added variable to toggle whether a container can be walked through when open
  - For all closets, this is now true, staying false for boxes

### Open Issues
- Cancel Pull Issues
  - Previously, when cancelling the pull, the object would shift one tile down and to the left
  - This is changed to position instead of localposition
  - Object appears to blink out of existence when this happens
    - Best guess is that matrix position or something like that updates, and when it does, the object is moved from the position it was picked up to the new location
- When pulling, objects will blink behind the player in certain rotations 
- Pushing doesn't work at all

- [ ]  I read the contribution guidelines and the wiki.
- [ ]  This fix is tested on the same branch it is PR'ed to.
- [ ]  This PR does not include files without specific need to do so.
- [ ]  This PR does not bring up any new compile errors
- [ ]  This PR has been tested in editor

### Notes
Previously, there was some code in playermove that would push objects that would not be allowed to be pushed (if block at ~243, "not to be checked when doing a replay"). Rebasing to develop seems to have fixed this.

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Objects/Closets/Unlockable/Closet.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Objects/Closets/Unlockable/Closet.prefab
@@ -215,9 +215,8 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   ObjectType: 1
   AtmosPassable: 1
-  Offset: {x: 0, y: 0, z: 0}
   Passable: 0
-  IsClosed: 1
+  PassableWhileOpen: 1
 --- !u!114 &114485734239509306
 MonoBehaviour:
   m_ObjectHideFlags: 1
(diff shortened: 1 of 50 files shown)


DooblyNoobly:
Converted this PR into a branch for further development. Checkout Push/Pull/Fixes branch to continue working on a solution

krille90:
Closing since @StrawberryMoses is busy atm, the PR will be closed.

If anyone wants to pick this up, the current changes are in the branch `Push/Pull/Fixes`

PR #1107, opened 2018-06-26
Simple chemistry system

Bod9001:
### Approach

Requires Newtonsoft.Json

Uses a "Reactions.json" To store all the reactions
Once its initialise it makes a dictionary of what chemicals are in what reactions
Then you can just call it with a Dictionary<String,Float> Area (For the chemicals)  and Float Temperature
And it should return Dictionary<String,Float> back With reactions applied to it
### Supports

- [x]  Catalysts
- [x]  Temperature
- [x]  Reactions
- [ ]  Stuff like explosions
- [ ]  foam

### Notes:
It's a consul application because I'm not quite sure how to make sure its initialise when being called from somewhere else, It roughly can do what TG's system does but doesn't support stuff like explosions, basically anything that produces something else then chemicals

### DIFF
--- a/Reference Material/Bod9001_c#_chemistry_reaction/Program.cs
+++ b/Reference Material/Bod9001_c#_chemistry_reaction/Program.cs
@@ -1,0 +1,166 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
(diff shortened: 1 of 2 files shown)


fomalsd:
very nice!

PR #1115, opened 2018-06-30
Set default zoom level based on the Screen width

DooblyNoobly:
### Purpose
As per the title. Tested on multiple screens.


fomalsd:
Why did you remove "auto" slider position tho? It worked fine, and now zoom defaults to 1 on 720p

fomalsd:
Also it doesn't save my 2x scale preference (in editor) and keeps starting up in 1x

DooblyNoobly:
Auto was broken for me that is why I tried to make it work with player prefs on start up instead. I'll fix it shortly. Maybe 2 should just be the default for everything

PR #1119, opened 2018-07-02
Fix for #900 "Running along walls and then pressing to run into the wall as well, causes you to stop "

### LINKED ISSUE #900: Running along walls and then pressing to run into the wall as well, causes you to stop
# Preliminary information
Running along walls and then pressing to run into the wall as well, causes you to stop 

## Category
 this is a  BUG,

it was  on the Latest steam version Beta 5 

# Report
## Current Behavior
Running along walls and then pressing to run into the wall as well, causes you to stop 

like in TG code you keep on running along the wall instead of stopping

## Steps to Reproduce
1. Find long stretch of wall
2.Run along 
3.Press key to go into the wall while still pressing the same key to go along the wall

## OS used:
Windows

## Network
server That is default on steam

Bod9001:
### Purpose
To fix #900 

### Approach
Instead of having generate directions from keys then that being passed to the function thta checks if it's ok to do that

it Generates a direction from the keys and keeps the buffer of what directions it is made up of,  Then tests the direction   If it is not suitable then tests the directions in the buffer

- [I think so?]  I correctly commented my code 

### Notes:
If there's any changes that need to be done,  I'll do them


DooblyNoobly:
Not bad. But it introduces a bug where you can randomly move through walls when trying to run diagonally into a wall. Tested on 1 headless server and 1 client with 400ms lag

UnityProject/Assets/Scripts/PlayGroups/PlayerMove.cs
```diff
@@ -250,32 +229,81 @@ private Vector3Int AdjustDirection(Vector3Int currentPosition, Vector3Int direct
...
                 {
-                    return direction;
+                    return true;
                 }
             }
 
             if (isReplay)
             {
-                return direction;
+                bool is_not_replay = false;
```
fomalsd, line 252: What's the point of introducing this variable here? And you're passing that isNotReplay to the method that accepts isReplay, that makes no sense and brings a lot of confusion

Bod9001:
@DooblyNoobly 
I was able to recreate the same bug In the non modified code
(Where it would Teleport you through walls bringing you to an alternative path you could have gone to)
Using 2 Compiled clients 
one with 400ms Network delay

Image
https://i.imgur.com/Yvv3HCa.png
(To the left of bridge/command area)
If you run through those 3 doors (with the 400ms Client) and Just keep on doing it even if you get pinged back. it becomes pretty uncontrollable An this is in the non modified version and my version

fomalsd:
could you record a gif perhaps? (GifCam for Windows set to 25FPS or more should do fine)

UnityProject/Assets/Scripts/PlayGroups/PlayerMove.cs
Bod9001: Image 
If it doesn't check it can move their, it ends up giving false locations to move to  Since if "isReplay" Returns always true

Bod9001:
Example
https://i.imgur.com/Z4zCVwo.gifv

Bod9001:
Apparently this is the issue for my PR Is this correct?   
Image

Bod9001:
I'll closed due to the prediction system needing reworking before this becomes viable

Issue #1019, opened 2018-04-21
WinDoor doesn't close when tile is occupied


krille90:
## Current Behavior
When the tile of the door is occupied (not passable), it won't close. However, this is not the wished behaviour for WinDoors / Sliding glass doors.

## Expected/Wanted/Requested Behavior
WinDoors should always close, as it can not be obstructed.
Make sure that the original behaviour stays the same for the other doors.

## Picture
![image](https://user-images.githubusercontent.com/7312905/39081572-025c6a64-4544-11e8-8117-12d96a60d2c5.png)

this door should eventually close, even if the player keeps standing there.

PR #1131, opened 2018-07-08
Updated Diagonal Player Movement

FilPag:
### Purpose
Updated diagonal movement for players when they walk into walls to make it more like the original /tg/. Players will now slide alongside walls when walking into them diagonally instead of stopping completely.

### Approach
The "AdjustDirection" function now checks if it's possible to walk alongside the obstacle instead of completeley stopping the player. Also changed the function to repeat the last move when "Replaying" client moves.

### Notes:

### DIFF
--- a/UnityProject/Assets/Scripts/PlayGroups/PlayerMove.cs
+++ b/UnityProject/Assets/Scripts/PlayGroups/PlayerMove.cs
@@ -47,8 +47,6 @@ namespace PlayGroup
 		private RegisterTile registerTile;
 		private Matrix matrix => registerTile.Matrix;
 
-		private Vector3Int lastMove = Vector3Int.zero;
-
 		/// temp solution for use with the UI network prediction
 		public bool isMoving { get; } = false;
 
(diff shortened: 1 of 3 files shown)


DooblyNoobly:
won't have a chance to check this out until tomorrow. Unless @krille90 or @fomalsd can get to it first

fomalsd:
Looks like this aims to fix #900, just like the other opened PR #1019 by Bod
Needs to be tested throughoutly under high ping, will check this out later today

fomalsd:
1) Minor: player can face the wall while sliding, IMO it's wrong
gif2

2) Players make a weird wiggling move after they finish a diagonal move and start walking straight again
gif
It gets worse with high ping (green is prediction, white is last confirmed move from server):
gif3
Things can get really messy in general:
gif4
I think problem is that several extra predictive moves are made after change of direction. 
See the gif below.

3) Also you can walk through players again while you aren't supposed to:
gif5

yeah, movement prediction is a delicate thing

PR #1139, opened 2018-07-16
New debug system

### LINKED ISSUE #1111: Log / Console clean up
# Preliminary information

## Category
TODO

# Report
## Current Behavior
Currently there are too many things logged, especially on system start.

Either clean up the logging such that only necessary messages are shown or introduce a tool to better log message. See #943 

On startup there are already a lot of messages:
image

Bod9001:
### Purpose
To help manage with debug, by allowing the user to set what categories are shown in the debug
Fixes #1111

### Notes:
How to use
Example of how to call Logger.Log("Message",Categories.#Something#);
You can set what you want to see in FilterDictionary 
-1 will not show anything
 0 Only will show errors 
 1 will show errors and warnings
 2 will show errors, warnings and logs

### DIFF
--- a/UnityProject/Assets/Light2D/Scripts/LightingSystem.cs
+++ b/UnityProject/Assets/Light2D/Scripts/LightingSystem.cs
@@ -136,7 +136,7 @@ namespace Light2D
 		{
 			if (GameData.Instance.testServer || GameData.IsHeadlessServer)
 			{
-				TADB_Debug.Log("Turn off lightsystem as this is a server",TADB_Debug.Category.Light2D.ToString());
+				Logger.Log("Turn off lightsystem as this is a server",Categories.Light2D);
 				enabled = false;
 			}
 #if UNITY_EDITOR
@@ -149,19 +149,19 @@ namespace Light2D
 
 			if (LightCamera == null)
 			{
-				TADB_Debug.LogError("Lighting Camera in LightingSystem is null. Please, select Lighting Camera camera for lighting to work.",TADB_Debug.Category.Light2D.ToString());
+				Logger.LogError("Lighting Camera in LightingSystem is null. Please, select Lighting Camera camera for lighting to work.",Categories.Light2D);
 				enabled = false;
 				return;
 			}
 			if (LightOverlayMaterial == null)
 			{
-				TADB_Debug.LogError("LightOverlayMaterial in LightingSystem is null. Please, select LightOverlayMaterial camera for lighting to work.",TADB_Debug.Category.Light2D.ToString());
+				Logger.LogError("LightOverlayMaterial in LightingSystem is null. Please, select LightOverlayMaterial camera for lighting to work.",Categories.Light2D);
 				enabled = false;
 				return;
 			}
 			if (AffectOnlyThisCamera && _camera.targetTexture != null)
 			{
-				TADB_Debug.LogError("\"Affect Only This Camera\" will not work if camera.targetTexture is set.",TADB_Debug.Category.Light2D.ToString());
+				Logger.LogError("\"Affect Only This Camera\" will not work if camera.targetTexture is set.",Categories.Light2D);
 				AffectOnlyThisCamera = false;
 			}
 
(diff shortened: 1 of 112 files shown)


UnityProject/Assets/Scripts/Debug/TADB.cs
```diff
@@ -0,0 +1,127 @@
...
+		ShutterController,
+		ShipNavigation,
+		Lighting,
+		Electrical, 
+	};
+	//-1 will not show anything
+	// 0 Only will show errors 
+	// 1 will show errors and warnings
+	// 2 will show errors, warnings and logs
+	private static Dictionary<string, int> NameDictionary = new Dictionary<string, int>{
```
fomalsd: Suggest introducing default log level to config as well.
People absolutely shouldn't be forced to add categories here, it could rather list exceptions from default log level

UnityProject/Assets/Scripts/Debug/TADB.cs
```diff
@@ -0,0 +1,127 @@
...
+		["PlayerJobManager"] = 0,
+		["PushPull"] = 0,
+		["NetSpriteImage"] = 0,
+		["ShutterController"] = 0,
+		["ShipNavigation"] = 0,
+		["Lighting"] = 0,
+		["Electrical"] = 0,
+	};
+	//Example of how to call TADB_Debug.Log("Message",TADB_Debug.Category.#Something#.ToString());
+	private static Dictionary<int, string> LevelMessage = new Dictionary<int, string>{
```
fomalsd: why not use an enum for these? would be much more efficient:
public enum LogLevel {
Error = 0,
Warning = 1,
Log = 2,
Trace = 3
};
and then getting ordinal by casting to int

fomalsd (changes requested):
Props for manually setting all the categories throughout the codebase!
But log strings are way too long at the moment:
1) TADB_Debug is a way too long name, just call it `Logger` (if it's not ambiguous) or something short
2) Category enum should be outside of that logger class, having it inside just makes you write unnecessary logger name _each time_
3) Pass enums as categories! There is absolutely no reason you should pass string representations instead. Some names could also be shorter.

So I'd imagine typical log to look like
`Logger.LogError("Encryption keys cannot be None type!",Category.Telecoms);`
and not
`TADB_Debug.LogError("Encryption keys cannot be None type!",TADB_Debug.Category.Telecommunications.ToString());`

UnityProject/Assets/Scripts/Debug/TADB.cs
Bod9001: If it's not in the list of categories it just logs it anyway
"if (NameDictionary.ContainsKey(Category)) " 
"else
Debug_Level_process((Category+" "+LevelMessage[PriorityLevel]+Message+"\n"+ memberName), PriorityLevel);"

fomalsd: In case it's unlisted it should check against the default logging level setting, I think that's reasonable

UnityProject/Assets/Light2D/Scripts/LightingSystem.cs
```diff
@@ -136,7 +136,7 @@ private void Start()
 		{
 			if (GameData.Instance.testServer || GameData.IsHeadlessServer)
 			{
-				Debug.Log("Turn off lightsystem as this is a server");
+				TADB_Debug.Log("Turn off lightsystem as this is a server",TADB_Debug.Category.Light2D.ToString());
```
PJB3005: Having to do a `.ToString()` call to use logging categories is absolutely too much boiler plate.

UnityProject/Assets/Scripts/Debug/TADB.cs
```diff
@@ -0,0 +1,127 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using System.Runtime.CompilerServices;
+
+
+public class TADB_Debug
+	{
```
PJB3005: What is this, the GNU coding standards? These tabs man.

UnityProject/Assets/Scripts/Debug/TADB.cs
```diff
@@ -3,108 +3,120 @@
...
+	private enum Levels{
+		Error = 0,
+		Warning = 1,
+		Log = 2,
+		Trace = 3
 		};
-	public static void Log(string Message, string Category = "Unknown ",[CallerMemberName]string memberName = "")
+
+	private static List<string> PriorityMessage = new List<string>{
+		"Error ",
```
fomalsd: Hehe, what's this one for? You've got a Levels enum for that already

fomalsd (approved):

PR #1152, opened 2018-07-23
Again fixing Chemistry lib

Bod9001:
### Purpose
Fixes all the stuff I missed...
added code so Initialization can only be run once
Fixed an Edge case with two reactions That could happen


DooblyNoobly:
I'll setup a CraftingManager at some point to deal with all of the crafting libs and then we can get rid of those globals.

PR #1185, opened 2018-09-01
[DNM] New Lighting System Preview

LeonardKr:
Implementation of a new Light System (WIP). 

This version still has old ray-cast fov, so you may still experience non optimal performance.
Generally its deep in the progress, am actively experimenting and shuffling solutions so there will be some bugs and quirks.


DooblyNoobly:
converted to lighting/system branch for changes before merging

PR #1199, opened 2018-09-11
Adds the ability to remove or add tiles at runtime

DooblyNoobly:
### Purpose
- This adds the TileChangesManager which handles tilemap changes over the network at runtime. This opens up the ability to start working on construction and deconstruction. 
- Added an example of deconstruction with the crowbar. Use it on a floor tile to see it in action. 
- Also fixed up some Rcon problems while testing headless mode

### Notes:


UnityProject/Assets/Scripts/Messages/Server/TileChangesNewClientSync.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Networking;
+
+//long name I know. This is for syncing new clients when they join to all of the tile changes
+public class TileChangesNewClientSync : ServerMessage
+{
+	public static short MessageType = (short) MessageTypes.TileChangesNewClientSync;
+	public string data;
+	public NetworkInstanceId ManagerSubject;//fixme: Recipient is redundant! Can be safely removed
```
fomalsd: hehe, why use ManagerSubject if it's redundant?

DooblyNoobly: I struggled with that name for ages. Its actually the TileChangeManager in question

DooblyNoobly: As you can eventually have more then 1 station which can have construction / deconstruction on it

fomalsd: Ah, I guess that fixme is not valid here then

DooblyNoobly: oh right I didn't even see that. I cut and pasted one of your messages and altered it

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/TileChangeManager.cs
```diff
@@ -99,10 +100,151 @@ void LoadAllTileAssets ()
...
+		changeEntry.cellPosition.y, 0), newTile);
+
+		if(isServer){
+			RpcActionChange(changeEntry.cellPosition, changeEntry.tileKey, changeEntry.layerToChange);
+		}
+		//TODO ADD TO THE CHANGE REGISTER
+	}
+
+	[ClientRpc]
+	private void RpcActionChange(Vector2 cellPos, string tileKey, TileChangeLayer layer){
```
fomalsd: Using ClientRpc will lead to inconsistencies on client side (some tiles will not be updated), keep that in mind

DooblyNoobly: it's on the default reliable channel so it should work. Will keep an eye on it

PR #1262, opened 2018-09-28
Right click radial menu! [Work in progress]

Bod9001:
### Purpose
Right click radial menu 
For all your clicking needs

### Notes:
Still work in progress
Jibby/alex180500 Said they would be doing some graphics improvement to it and The current icons are meant to be temporary


DooblyNoobly:
--------------
You will need to remove the Right click canvas object from the managers tree. 
image
Pull it out and work on it separately like that. 

You will also need to think about how it will be loaded and where. You can either put it in the hierarchy (outside of managers tree) in each game scene and save the scene. Or you can have a manager instantiate it at the start of any game scene by checking GameData if the newly loaded scene is an ingame scene or lobby scene. 

So I'd recommend instantiating it on the Managers.SetScreenForGame() method:
https://github.com/unitystation/unitystation/blob/bae12986d682021a19d46a9a1834810023cec1be/UnityProject/Assets/Scripts/Managers/Managers.cs#L42

Just after the Await call. (Don't put it in the isServer condition)

That way you don't have to submit a scene file in your PR and you will know whenever someone creates a new map that they don't have to worry about dragging in RightClickMenu prefab to their scene and saving it

---------------------

I'm still reviewing the rest. Will update comments as I go over the next day or so

UnityProject/Assets/Scripts/UI/Right click/Rightclick.cs
```diff
@@ -0,0 +1,155 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using System;
+using System.Reflection;
+using UnityEngine;
```
DooblyNoobly, line 5: two of the same using statements. Remove one

UnityProject/Assets/Scripts/UI/Right click/Rightclick.cs
```diff
@@ -0,0 +1,155 @@
...
+		public string title;
+		public GameObject Item;
+		public List<Menu> SubMenus = new List<Menu>();
+		public MethodInfo Method;
+		public MonoBehaviour Mono;
+	}
+		
+
+	void Awake(){
+		ins = this;
```
DooblyNoobly: Everytime a scene loads you will have static instances of this class being created. That means after the scene has been destroyed, static instances from the last round still remain. You will have memory stack increases over time as the instances pile up. You need to check in awake if the instance has already been assigned and if it has then to destroy the instance that is doing the checking. Like this:

```cs
void Awake(){
    if(ins == null){
        ins = this;
    } else {
       Destroy(this);
    }
}
```
This will clean up any static instances of an object from the previous scene

UnityProject/Assets/Scripts/UI/Right click/Rightclick.cs
```diff
@@ -0,0 +1,155 @@
...
+		//Make sure to add your sprite on load
+		SpriteDictionary ["hand"] = Resources.Load<Sprite> ("UI/RightClickButtonIcon/" + "hand");
+		SpriteDictionary ["Magnifying_glass"] = Resources.Load<Sprite> ("UI/RightClickButtonIcon/" + "Magnifying_glass");
+		SpriteDictionary ["question_mark"] = Resources.Load<Sprite> ("UI/RightClickButtonIcon/" + "question_mark");
+		SpriteDictionary ["Drag_Hand"] = Resources.Load<Sprite> ("UI/RightClickButtonIcon/" + "Drag_Hand");
```
DooblyNoobly, line 53: A better way would be to make public Sprite fields on this class and drag the sprites into the inspector then save the prefab. What you have done is fine though as it only happens once on the start of the scene. So leave it if you want.

UnityProject/Assets/Scripts/UI/Right click/RadialMenuSpawner.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class RadialMenuSpawner : MonoBehaviour {
+
+	public static RadialMenuSpawner ins;
+	public RadialMenu menuPrefab;
+
+	void Awake(){
```
DooblyNoobly, line 10: See below about the need to check if the instance has already been assigned.

i.e. static instances need to be set up like this:
```cs
void Awake(){
    if(ins == null){
        ins = this;
    } else {
       Destroy(this);
    }
}
```

DooblyNoobly (changes requested):
Some small changes and I think we can merge this. Please see my comments too about instantiating the right click handler on InGame scenes.

UnityProject/Assets/Scripts/UI/Right click/Rightclick.cs
```diff
@@ -0,0 +1,155 @@
...
+				}
+			}
+
+			ins.options.Add (newMenu);
+		}
+	}
+}
+
+[AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
+public class contextMethod : Attribute
```
fomalsd: Please rename `contextMethod` to `ContextMethod`, classes are meant to be PascalCase

UnityProject/Assets/Scripts/UI/Right click/Rightclick.cs
```diff
@@ -0,0 +1,155 @@
...
+
+		}
+		
+	}
+	private void Generate(List<GameObject> objects){
+		options = new List<Menu> ();
+		for (int i = 0; i < objects.Count; i++) {
+			Menu newMenu = new Menu();
+			newMenu.colour = Color.gray;
+			ItemAttributes Nameues = objects[i].GetComponent<ItemAttributes>();
```
fomalsd: Nameues? what's that? :)

fomalsd (commented):
Looks good code wise apart from what Doobly suggested and sometimes weird naming

UnityProject/Assets/Scripts/UI/Right click/Rightclick.cs
Bod9001: yeah, it should be fine Because to the engine it's basically the same thing

Bod9001:
The changes should be made?

Bod9001:
you can Technically ignore all the changes to  the Scenes but I don't know how to do that?

PR #1282, opened 2018-10-04
Colorization for FOV masked Shader

LeonardKr:
Colorization for FOV masked Shader
Unity Fog removed from shaders


DooblyNoobly:
Thanks. Adding changes to Player/Accounts branch

PR #1283, opened 2018-10-04
Fixing StopCoroutine calls

YotaXP:
### Purpose
`StopCoroutine` was being called incorrectly in most places, and therefore not actually stopping any coroutines.  This fixes that.  It also cleans up some sprite swapping code for the APCs to waste less memory.

### Notes:
Noticed that the field generator isn't animating on clients.  Can someone confirm I'm not causing this?


fomalsd:
Thanks! (Didn't know we had NPC_Clown, that code is likely ancient)

fomalsd:
Yeah, just checked out field generator on my dated branch and it's gives NREs on client interact attempts, guess it's not production ready. Also APC's sprite isn't synced when client joins.
So you're not causing this for sure

PR #1285, opened 2018-10-05
Remove Methods from TMS that are only based on the position

### LINKED ISSUE #942: Remove Methods from TMS that are only based on the position B: 15
# Preliminary information

## Category
TODO

dev

# Report
## Current Behavior
Currently, methods like IsPassableAt(position) are being used. In the context of windoors, it doesn't makes sense though, as a direction from which it could be passable is always needed. IsPassable(position) and IsAtmosPassable(position) shouldn't exist in the first place.

Remove IsPassable(position) and IsAtmosPassable(position)  and rewrite those to the IsPassable(origin, position) and IsAtmosPassable(origin, position) 

## Criteria

- [ ] Search the codebase to find all uses of IsPassable(position) and IsAtmosPassable(position) and convert them to IsPassable(origin, position) and IsAtmosPassable(origin, position)
 - [ ] Test the game thoroughly and make sure there are no bugs
 - [ ] Remove the IsPassable(position) and IsAtmosPassable(position) methods from the matrix

Itsnotjames:
### Purpose
Fixes #942 Remove Methods from TMS that are only based on the position.

### Notes:
First PR ever, wahoo.

Turned Matrix.IsPassableAt(position) into a wrapper for Matrix.IsPassableAt(origin, position) which passes the displaced position (position + Vector3Int.up) as the origin.
Because: of the classes: DoorController, RandomMove, ExplodeWhenShot, PushPull, PlayerSync, PlayerSync.Server, and MatrixTest which call Matrix.IsPassableAt(position) for various reasons.

Removed IsPassableAt(position) in Layer and it's override in objectLayer.cs.
Because: They now are unused and unneeded.

Turned MetaTileMap.IsPassableAt(position) and MetaTileMap.IsAtmosPassableAt(position) into a wrapper for their (from, to) counterparts, passing the position for both arguments.
Because: of the calls in RoomControl and TileMapCheckEditor which do not have an origin position in their contexts.

Removed the "if (from == to) return true" from layer.IsPassableAt(from, to).
Because: this if statement returns walls as passable, despite them not being passable ingame. This is shown with the "Passable" gizmo in the TileMapCheckEditor tool (under window/TileMapCheck).
		
As shown with the Tilemap Check tool:
Non-"Passable" tiles currently are now objects, walls, and closed doors.
Non-"AtmosPassable" tiles currently are now walls and closed doors.

There is currently no Passable or AtmosPassable code that compares the orgin coordinate/tile with position coordinate/tile for the context of windoors. The code for this, which would compare the "from" and "to" coordinates/tiles should be written in layer.IsPassableAt(from, to) and layer.IsAtmosPassableAt(from, to).

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Matrix.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Matrix.cs
@@ -35,7 +35,7 @@ public class Matrix : MonoBehaviour
 
     public bool IsPassableAt(Vector3Int position)
     {
-        return IsPassableAt(position + Vector3Int.up, position);
+        return IsPassableAt(position, position);
     }
 
 	public bool IsPassableAt(Vector3Int origin, Vector3Int position)
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Matrix.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Matrix.cs
@@ -33,21 +33,19 @@ public class Matrix : MonoBehaviour
 		initialOffset = Vector3Int.CeilToInt(gameObject.transform.position);
 	}
 
+    public bool IsPassableAt(Vector3Int position)
+    {
+        return IsPassableAt(position + Vector3Int.up, position);
+    }
+
 	public bool IsPassableAt(Vector3Int origin, Vector3Int position)
 	{
 		return metaTileMap.IsPassableAt(origin, position);
 	}
 
-	//TODO:  This should be removed, due to windows mucking things up, and replaced with origin and position
-	public bool IsPassableAt(Vector3Int position)
+	public bool IsAtmosPassableAt(Vector3Int origin, Vector3Int position)
 	{
-		return metaTileMap.IsPassableAt(position);
-	}
-
-	//TODO:  This should also be removed, due to windows mucking things up, and replaced with origin and position
-	public bool IsAtmosPassableAt(Vector3Int position)
-	{
-		return metaTileMap.IsAtmosPassableAt(position);
+		return metaTileMap.IsAtmosPassableAt(origin, position);
 	}
 
 	public bool IsSpaceAt(Vector3Int position)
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Matrix.cs
```diff
@@ -33,21 +33,19 @@ private void Awake()
 		initialOffset = Vector3Int.CeilToInt(gameObject.transform.position);
 	}
 
+    public bool IsPassableAt(Vector3Int position)
+    {
+        return IsPassableAt(position + Vector3Int.up, position);
```
fomalsd: Why Vector3Int.up? it's (0, 1, 0), so it will always check if one's approaching from the north, this will cause all kinds of quirks and inconsistencies

Itsnotjames: Most contexts will have an origin, so IsPassableAt(origin, position) will be called and the wrapper method IsPassableAt(position) will not be called.
IsPassableAt(position) should be only called when the origin is irrelevant, therefore the origin argument does not matter as long as it is adjacent to the position. Vector3.Int.up/down/left/right should all return the same result in the contexts that the wrapper method IsPassableAt(position) is called.
One example of a call to IsPassableAt(position) is in the DoorController class. IsPassableAt(position) is called to see if a normal door (not a sliding door) is passable. Since normal doors can be passed from all directions, the origin argument is irrelevant as long as it is adjacent to the door position. Thus IsPassableAt(position) is called because the origin is not needed.

fomalsd: still its behaviour would depend on whether tile above is solid, as `IsPassableAt(origin, target)` returns false if `origin` is solid and `target` is not.
Let's take ExplodeWhenShot for example, in your case it won't spawn fire tile if there's a wall/solid object on the tile above

fomalsd: I did some tests and I can say that replacing this with
`return IsPassableAt(position, position);` certainly doesn't harm and has consistent behaviour regardless of surroundings

fomalsd:
Awesome, thanks!

Issue #1291, opened 2018-10-09
TODO: Map 4 or 5 Asteriods on their own matricies for OutpostStation scene B: 50


DooblyNoobly:
### Description

Please read the 0.4.0 scope: https://github.com/unitystation/unitystation/wiki/Start-of-Round and understand the requirements of the Asteroids.

Use the hierarchy structures of the station and shuttles matrices to help you create new ones for the Asteroids. 

## Criteria:

- Using Tilemap Palette, map 4 or 5 Asteroids on their own matrices in the OutpostStation scene
- Add a matrix move component to all Asteroids 
- Move the Asteroids to random locations at the start of the round (move on the server through matrix move)

Bounty of 50usd will be paid to the contributor who meets all of the criteria on a successful merge.

PR #1297, opened 2018-10-10
Adds mining asteroids (mapping and matrixmove stuff)

badbon:
### Purpose
Implements #1291, adds 6 asteroids for use in mining (tiny, very small, small, medium, large, very large - each with their tilemaps) and gives them random matrixmove position on round start.

They all have their own matrixmove, as updated on "station" prefab by Dooby. (I think)

Thanks to Foma and Doobly for help with different stuff.

### Notes:
On minimap, asteroids show up as shuttles with their icon because thats how matrixmove is recognised, I will try to fix this up in a bit, feel free to merge this though, if it looks good.

### DIFF
--- a/UnityProject/Assets/Scripts/Map/Asteroid.cs
+++ b/UnityProject/Assets/Scripts/Map/Asteroid.cs
@@ -1,25 +1,30 @@
 ﻿using System.Collections;
 using System.Collections.Generic;
 using UnityEngine;
+using UnityEngine.Networking;
 
-public class Asteroid : MonoBehaviour {
+public class Asteroid : NetworkBehaviour {
 
 	private MatrixMove mm;
 
 	private float asteroidDistances = 550; //How far can asteroids be spawned
 
 
-	private void Awake()
+	private void Start()
 	{
 		mm = GetComponent<MatrixMove>();
 
 		SpawnNearStation();
 	}
 
+	[Server]
 	public void SpawnNearStation()
 	{
-		//Based on EscapeShuttle.cs
-		mm.SetPosition(Random.insideUnitCircle * 500 + new Vector2(500, -500));
+		if (isServer)
+		{
+			//Based on EscapeShuttle.cs
+			mm.SetPosition(Random.insideUnitCircle * 500 + new Vector2(500, -500));
+		}
 	}
 
 }
--- a/UnityProject/Assets/Scripts/Map/Asteroid.cs
+++ b/UnityProject/Assets/Scripts/Map/Asteroid.cs
@@ -1,0 +1,25 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class Asteroid : MonoBehaviour {
+
+	private MatrixMove mm;
+
+	private float asteroidDistances = 550; //How far can asteroids be spawned
+
+
+	private void Awake()
+	{
+		mm = GetComponent<MatrixMove>();
+
+		SpawnNearStation();
+	}
+
+	public void SpawnNearStation()
+	{
+		//Based on EscapeShuttle.cs
+		mm.SetPosition(Random.insideUnitCircle * 500 + new Vector2(500, -500));
+	}
+
+}
--- a/UnityProject/Assets/Scripts/Map/Asteroid.cs
+++ b/UnityProject/Assets/Scripts/Map/Asteroid.cs
@@ -25,8 +25,18 @@ public class Asteroid : NetworkBehaviour {
 	[Server]
 	public void SpawnNearStation()
 	{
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/Map/Asteroid.cs
```diff
@@ -0,0 +1,32 @@
...
+	private float asteroidDistance = 550; //How far can asteroids be spawned
+
+	private float distanceFromStation = 175; //Offset from station so it doesnt spawn into station
+
+
+	private void Start()
+	{
+		mm = GetComponent<MatrixMove>();
+
+		SpawnNearStation();
```
fomalsd: `if (isServer)` goes here, also you should really check it in multiplayer: it might accidentally execute before matrices are initialized

UnityProject/Assets/Scripts/Map/Asteroid.cs
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	[Server]
+	public void SpawnNearStation()
+	{
+		if (isServer)
```
fomalsd: [Server] method won't run on clients anyway, checking after it's already executing is useless

UnityProject/Assets/Scripts/Map/Asteroid.cs
```diff
@@ -0,0 +1,32 @@
...
+		{
+			//Based on EscapeShuttle.cs
+			mm.SetPosition(Random.insideUnitCircle * asteroidDistance + new Vector2(distanceFromStation, -distanceFromStation));
```
fomalsd: This won't work:
That `Vector2(distanceFromStation, -distanceFromStation)` is actually **always** Vector2( x=175; y=-175 ),
which equals to bottom right
So it doesn't guarantee that it won't spawn inside station, as first part can generate -175; 175 and asteroid will land exactly at the station, so please test your calculations before PRing!
What you need to do is exclude the entire range of ( x=-175..175; y=-175..175 ) from final randomised position by per-axis `Mathf.Clamp`ing or something else

fomalsd (changes requested):

UnityProject/Assets/Scripts/Map/Asteroid.cs
fomalsd: Also it would be nice if asteroids had random rotation on start

fomalsd: ...And some could be moving very slowly

badbon:
Okay, I think I've fixed everything that was mentioned here and on Discord, except the issue where asteroid tiles are showing up on screen at spawn for a frame.

I'd give a random guess that it might have something to do with the fact that my new functions are still being called on game startup rather than player spawn, so matrix move might be acting weird because of that?

badbon:
Oh and I also added random orientations for asteroids, good suggestion ^^

UnityProject/Assets/Scripts/Map/Asteroid.cs
```diff
@@ -0,0 +1,75 @@
...
+		
+		if(clampVal.x > 0)
+		{
+			clampVal.x = Mathf.Clamp(clampVal.x, distanceFromStation, asteroidDistance);
+		}
+		else
+		{
+			clampVal.x = Mathf.Clamp(clampVal.x, -distanceFromStation, -asteroidDistance);
+		}
+		mm.SetPosition(clampVal);
```
fomalsd, line 41: You forgot about Y axis

UnityProject/Assets/Scripts/Map/Asteroid.cs
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	[Server] //Asigns random rotation to each asteroid at startup for variety.
+	public void RandomRotation()
+	{
+		int rand = Random.Range(0, 3);
```
fomalsd, line 47: `Note that max is exclusive for integers, so using Random.Range( 0, 10 ) will return values between 0 and 9`
It will never rotate left in your case

UnityProject/Assets/Scripts/Map/Asteroid.cs
```diff
@@ -0,0 +1,75 @@
...
+				mm.RotateTo(Orientation.Right);
+				break;
+			case 3:
+				mm.RotateTo(Orientation.Left);
+				break;
+		}
+	}
+
+	//Delays start functions to avoid issues with matrixmove
+	IEnumerator DelayedStart()
```
fomalsd, line 67: Delaying is a pretty dirty way, it will lead to inconsistent behaviour and bad bugs on slower machines, don't do that!

Issue #1317, opened 2018-10-23
Bug: FoV Flicker on shuttle rotation


DooblyNoobly:
# Description
As per the gif below, when rotating a shuttle the FoV flickers:

https://thumbs.gfycat.com/RawWanGrizzlybear-size_restricted.gif

This is caused by the updated lighting system and its reliance on pixel perfect positions: #1315 

I have created some methods that are triggered for the Local Player when they are on a matrix that is rotating and placed them in the LightingSystem: https://github.com/unitystation/unitystation/blob/74c560991ce269a5ab74ba3afdfda36f314b9140/UnityProject/Assets/Scripts/Lighting%20System/LightingSystem.cs#L392

```cs
public void MatrixMoveStartRotation()
{
	Debug.Log("Matrix has started rotating for local player");
}

public void MatrixMoveStopRotation()
{
	Debug.Log("Matrix has stopped rotating for local player");
}
```

## TODO:

> switch to billinear filter for occlusion sampling to reduce the jitter when the shuttle is rotating

fomalsd:
fixed in #1323

PR #1323, opened 2018-10-29
Lighting system: Matrix rotation mode

LeonardKr:
Lighting system: Matrix rotation mode for rotating matrices.


fomalsd (approved):
Yeah, it works, fixes #1317 
wall lights become visible during rotation, but it's not very important I guess
gif

PR #1351, opened 2018-12-02
maping! for Lights, Emergency Lights, Light Switches and APCs 

Bod9001:
### Purpose
to map out the electrical system and also the lights

the Radiation collector should turn on at game Start so the lights should be powered 
all lights are Powered by The Power System until voltage drops too low then the light switches turn off (**Warning causes a big lag  spike**) (they dim according to the voltage going down)
if there is no Light switch nearby lights don't turn on


DooblyNoobly:
@Bod9001 what are these "LOW" sprites? 
image

Bod9001:
I still need to do some sprite work, I couldn't get them to use a different spite  in the tile palette  (since the medium and low look really similar) So had to change The sprite,

PR #1391, opened 2018-12-20
Fixes #1373 - fix blood splatter effect position

### LINKED ISSUE #1373: Bloodsplats spawn with an offset
# Description
Bloodsplats spawn with an offset from their designated position

Most likely a world<->local coords issue

## Category
BUG

## Steps to Reproduce
1. Host from editor and join from client
2. Start shooting one another
giff

chairbender:
### Purpose
Fixes #1373 - fix blood splatter effect position so it appears under the mob rather than off to the side

### Notes:
The BloodSplatter method was trying to instantiate the BloodSplatter GameObject under a parent. The parent caused the blood splatter to be offset.

There doesn't seem to be a need to have the splatter created underneath the parent (none of the other effects do it when they call PoolNetworkInstantiate) so I think it's fine to leave the parent as null.

### DIFF
--- a/UnityProject/Assets/Scripts/Factories/EffectsFactory.cs
+++ b/UnityProject/Assets/Scripts/Factories/EffectsFactory.cs
@@ -1,15 +1,13 @@
 ﻿using UnityEngine;
 using UnityEngine.Networking;
 
-/// <summary>
-/// Creates various visual effects
-/// </summary>
 public class EffectsFactory : NetworkBehaviour
 {
 	public static EffectsFactory Instance;
 
 	//Parents to make tidy
 	private GameObject shroudParent;
+	private GameObject bloodParent;
 
 	private GameObject fireTile { get; set; }
 	private GameObject scorchMarksTile { get; set; }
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Factories/EffectsFactory.cs
```diff
@@ -71,8 +74,7 @@ public GameObject SpawnShroudTile(Vector3 pos)
 	[Server]
 	public void BloodSplat(Vector3 pos, BloodSplatSize splatSize)
 	{
-		GameObject b = PoolManager.Instance.PoolNetworkInstantiate(bloodTile, pos, Quaternion.identity, 
-			MatrixManager.AtPoint( Vector3Int.RoundToInt( pos ) ).GameObject.transform);
```
DooblyNoobly, line 75: `MatrixManager.AtPoint( Vector3Int.RoundToInt( pos ) ).GameObject.transform` this is the parent that the blood splat needs to be a child of. So removing it would make the blood splat not a child of anything. If the blood was in a shuttle then the shuttle would fly off and leave the blood behind. Maybe try setting the parent on a separate line after it has instantiated and see what happens.

so: `b.parent = MatrixManager.AtPoint( Vector3Int.RoundToInt( pos ) ).GameObject.transform;`

DooblyNoobly (changes requested):

DooblyNoobly:
> There doesn't seem to be a need to have the splatter created underneath the parent (none of the other effects do it when they call PoolNetworkInstantiate) so I think it's fine to leave the parent as null.  

yeah still some bugs I think because of this. It just has been neglected for awhile.

chairbender:
Ah, I assumed since other effects did it that it wasn't actually needed. Sorry for the oversight. I'll figure out a different way that maintains the parenting.

chairbender:
@DooblyNoobly should be fixed now. All the other effects are properly parented.

DooblyNoobly:
looks good. Thank you : )

PR #1393, opened 2018-12-20
Fixes #1376 - you can now shoot yourself

### LINKED ISSUE #1376: Suicide shot does not work
# Description

## Category
BUG

## Detailed Bug Report
Suicide shot does not work, instead you just melee punch yourself

## Steps to Reproduce
1.Grab a handgun
2.Try clicking yourself to suicide
gif

chairbender:
### Purpose
Restore the ability to shoot yourself with a gun. Also marked LaserGun.prefab with ItemType.Gun.

### Notes:
This will probably change a bit once we refactor the weapon interaction logic, but the only reason this stopped working was just that the MeleeTrigger was handling it before P2PInteraction and returning true, which stops processing of further interactions. But this suicide check only happens in P2PInteraction. Checking for this special case and returning false from MeleeTrigger (gun in hand, clicking self should not be a melee action so we exit from MeleeTrigger in that case) solves this issue.

### DIFF
--- a/UnityProject/Assets/Prefabs/Items/Weapons/Energy/Resources/LaserGun.prefab
+++ b/UnityProject/Assets/Prefabs/Items/Weapons/Energy/Resources/LaserGun.prefab
@@ -118,7 +118,7 @@ MonoBehaviour:
   inHandReferenceRight: 32
   itemDescription: 
   itemName: LaserGun
-  itemType: 0
+  itemType: 16
   size: 1
   spriteType: 2
   type: 16
@@ -139,17 +139,10 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: c23c49138f42840a1bc22ade283c1694, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
-  isPlayer: 0
   ignoredSpriteRenderers: []
   registerTile: {fileID: 0}
   visibleState: 1
-  allowedToMove: 1
-  currentPos: {x: 0, y: 0, z: 0}
-  isPushable: 0
-  pulledBy: {fileID: 0}
-  custNetActiveState: 0
-  pushing: 0
-  pushTarget: {x: 0, y: 0, z: 0}
+  isNotPushable: 0
   closetHandlerCache: {fileID: 0}
 --- !u!114 &114449825066554946
 MonoBehaviour:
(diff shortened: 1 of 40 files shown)


chairbender:
One sec, let me fix the bad indentation

chairbender:
That's better. Ready for review.

fomalsd:
thanks for adding extra comments! was unable to make a suicide shot though, will post a video through discord

chairbender:
@fomalsd  Should be fixed.

fomalsd:
ohh, didn't know we had both `type` and `itemType`...ItemAttributes needs a refactor
Noticed that automatic weapons sometimes skip suicide shot and/or continue firing normally while mouse button is held
Also shooting yourself doesn't generate a text message so it's harder to see if shot was successful, would you mind adding `PostToChatMessage.SendItemAttackMessage()` somewhere like here: https://github.com/unitystation/unitystation/blob/218c19d4c8eaeae9706e84b0ee0fc0aadb39b253/UnityProject/Assets/Scripts/Weapons/Projectiles/BulletBehaviour.cs#L57
(and also not randomize body part in case if it's suicide shot)

chairbender:
@fomalsd okay, ready for re-review.

Just to summarize:

1. Modified how InputTrigger works, it now has a virtual method for drag events and all Interact methods return a boolean indicating to InputController if no further triggers should be processed on the current update. Prior to this change, the logic was the same as if all trigger methods were returning true (which caused the gun to not fire sometimes due to the interaction being triggered with a tile before the trigger to fire the gun), so I modified almost all the trigger methods to default to returning true to preserve the existing functionality. This change allows us to have multiple interactions trigger in a single update if needed for other situations. Instead of doing this refactoring, I also could've added code to the weapon shoot method to do raycasting and figure out which object was clicked (to see if the player clicked on themselves), but doing it this way prevents the need for this additional call to raycast - I can just depend on the P2PInteraction to pass this information down to the Weapon (because P2PInteraction already knows the gameobject that was clicked on since it is passed from the InputController after InputController did the raycasting).
1. Added the attack message as requested.
1. Fixed a few other issues related to how guns were working (like being unable to shoot when the mouse gets too close to the body)
2. Added various /// comments to improve readability.
3. Re-formatted the documents that I touched according to the standard code style

fomalsd:
Thanks for adding a lot of documentation!
There is a NRE on this line when you shoot a fuel tank or throw something on it: HealthBehaviour isn't limited to players
https://github.com/unitystation/unitystation/blob/215b8efe35ce322e294c6e13c0e8eb050d50cdc0/UnityProject/Assets/Scripts/Health/HealthBehaviour.cs#L86
The rest looks good, fine job!

chairbender:
@fomalsd ready again.

fomalsd (approved):
Awesome!

Issue #1377, opened 2018-12-14
Able to generate/choose empty name in character editor. Messes up server.


fomalsd:
# Description

## Category
BUG

## Detailed Bug Report
~~1) Character name generator is wonky and can generate empty firstname and/or lastname~~
![gif](https://user-images.githubusercontent.com/10403536/50007075-9fc66200-ffa7-11e8-878a-9392ef44770c.gif)
2) Joining with empty name messes up server, it doesn't start
![gif](https://user-images.githubusercontent.com/10403536/50007175-02b7f900-ffa8-11e8-85f7-c2bbac0b7f66.gif)

## Steps to Reproduce
Please enter the steps to reproduce the bug or behaviour:
1. Start game and open character editor
2. ~~Keep randomizing names until you generate an empty one~~ Enter space instead of name
3. Apply changes and try hosting a server

fomalsd:
Seems like name generator is fine on latest develop. But if someone finds a way to join with an empty name it will still mess up server!

fomalsd:
Joining with empty name still messes up server

donovanhubbard:
I can do this one.

While I'm validating the character name, are there any other rules we want to enforce? Like first name and lastname?

fomalsd:
I think almost any non-empty name will do for now (like 2-3+ non-digit chars, maybe)
But I find it more important to handle the situation with empty name on server side so that it wouldn't get messed up (last time I checked, joining existing server with empty name generated some game-breaking exceptions on server side and server went unresponsive)

donovanhubbard:
I plan on tackling it on that side as well but I havent looked at it yet. What would be the desired behavior if a null or empty name is passed to the server? Should it assign a name? Prevent the user from connecting?

fomalsd:
yeah, either refuse connection or assign default name (Cuban Pete)

PR #1407, opened 2018-12-29
Empty name

donovanhubbard:
### Purpose
Fixes issue #1377.

Players can no longer enter an empty name at character select. It will give them an error instead.

If somehow a player does manage to connect with a empty, whitespace-only, or null name, it will assign the ConnectedPlayer.ANONYMOUS_NAME value  "Anonymous Spessman"

- [X ]  This fix is tested on the same branch it is PR'ed to.
- [ X]  I correctly commented my code
- [X ]  My code is indented with tabs and not spaces
- [ X]  This PR does not include any unnecessary .meta, .prefab or .unity (scene) changes
- [ X]  This PR does not bring up any new compile errors
- [ X]  This PR has been tested in editor
- [ X]  This PR has been tested in multiplayer

### Notes:
Players can no longer enter an empty name at character select. It will give them an error instead.
errormessage

If somehow a player does manage to connect with a empty, whitespace-only, or null name, it will assign the ConnectedPlayer.ANONYMOUS_NAME value  "Anonymous Spessman"

anonymous-log
anonymous-welcome

### DIFF
--- a/UnityProject/Assets/Scenes/Lobby.unity
+++ b/UnityProject/Assets/Scenes/Lobby.unity
@@ -50,7 +50,6 @@ LightmapSettings:
     m_BounceScale: 1
     m_IndirectOutputScale: 1
     m_AlbedoBoost: 1
-    m_TemporalCoherenceThreshold: 1
     m_EnvironmentLightingMode: 0
     m_EnableBakedLightmaps: 0
     m_EnableRealtimeLightmaps: 0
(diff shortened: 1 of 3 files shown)


donovanhubbard:
After looking over this again, there are a ton of changes to UnityProject/Assets/Scenes/Lobby.unity even though I only added a new text field. I assume some of these are just from opening the scene, but it seems like an awful lot.

DooblyNoobly:
Yeah the diffs on scene files are mystery. This looks good btw. Thank you

PR #1409, opened 2019-01-02
Fixes #389 - Occlude wallmounts and cleanup shader code

### LINKED ISSUE #389: Wallmount visibility
# Preliminary information
## Category
BUG

# Report
## Current Behavior
Wallmounts are visible on the wrong side of a wall

They should only be visible on the side the player is actually on.

## Requiremetns

* [ ] Hide wallmount when player is not within the front 180 degrees of the wallmount

chairbender:
### Purpose
Fixes #389  . Wallmounts are now occluded when out of view. Also includes various refactorings / comments to make it easier to understand what is going on with all of our various shader logic. I renamed some shaders / textures to more accurately correspond to what they are doing.

As part of this, I needed to go through and all a new behavior to all wallmount prefabs, update their sprites to use the one which is based on the FOV, and edit a bunch of wallmounts in OutpostStation so their orientation matches up with the side of the wall they are on.

### Notes
I'm unsure why commits from an already merged PR (1376) are showing up in the commit history here. I think it's because I branched from that branch rather than branching from develop when creating my new branch. I don't think this should be a problem.

As for how this was implemented...the previous shader logic generated an "FOV Mask" texture which masks out the areas of the floor that aren't visible, but it ignored the walls. This texture has been renamed to floorOcclusionMask. The new logic adds an additional texture (wallFloorOcclusionMask) which calculates if wall tiles should be visible / hidden based on the floorOcclusionMask. It essentially just checks up, down, left, and right from the current pixel at a distance of at most one wall tile width away and sees if green touches red (indicating that the wall tile we are on is touching a non-occluded floor). If green touches red in any direction, it keeps the wall visible, otherwise it blanks it out.

The amount of precision used when checking in each direction (which uses a lerp) can be adjusted using the LERP_ITERATIONS constant if performance of the shader is becoming an issue. If it is set too low, the occlusion in the wallFloorOcclusionMask becomes very jagged and can make wallmounts which are partially obscured look weird.


DooblyNoobly:
> I'm unsure why commits from an already merged PR (1376) are showing up in the commit history here. I think it's because I branched from that branch rather than branching from develop when creating my new branch. I don't think this should be a problem.

It's ok I'll squash this one when merging. And yeah it would of happened because you didn't branch from the updated develop

DooblyNoobly:
nice solution! thanks

PR #1432, opened 2019-01-12
#1110 - fix issue where client can get stuck in door

### LINKED ISSUE #1110: Doors close sometimes with player on them
# Preliminary information

## Category
BUG

development

# Report

Door detects reliably if there is a person on it.

## Steps to Reproduce
1. Go to some doors
2. Walk into them and out
3. eventually, the door will close with the player on it

image

chairbender:
### Purpose
Fixes #1110 - an issue where the client can get stuck in the door in the .15 s window between when the door close begins and the door is registered as impassable.

### Notes:
See my long comment in DoorController. In order to reproduce this, you have to have very precise timing on the client. Basically you need to enter the door right when the door close animation starts but before the server registers it as impassable. Otherwise you will get the normal lerpback behavior where it kicks you back out of the door. You know you've reproduced it if the character stays where they are in the door and the door reopens once it's done closing.

I originally implemented this check the entire time the door is closing (so it reopens as soon as it detects a player in it rather than once the door is finished closing), but I didn't like what I saw in the profiler with that approach (since it runs in UpdateManager.Update and checks the RegisterTiles every Update), and considering this is a very rare event I think it's fine to just have a simple behavior of reopening the door only once it's done closing.

### DIFF
--- a/UnityProject/Assets/Scripts/Doors/DoorController.cs
+++ b/UnityProject/Assets/Scripts/Doors/DoorController.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections;
+using System.Linq;
 using UnityEngine;
 using UnityEngine.Networking;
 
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Doors/DoorController.cs
```diff
@@ -1,222 +1,259 @@
...
+			openLayer = LayerMask.NameToLayer("Door Open");
+
+
+			registerTile = gameObject.GetComponent<RegisterDoor>();
+		}
+
+		/// <summary>
+		/// Invoked by doorAnimator once a door animation finishes
+		/// </summary>
+		public void OnAnimationFinished()
```
chairbender, line 78: This method is the actual only change to this file

UnityProject/Assets/Scripts/Doors/DoorController.cs
```diff
@@ -1,222 +1,259 @@
...
+			// Case 2 won't be handled by the rollback code because the client enters the passable tile while the
+			//	server still thinks its passable. So, for the rare situation that case 2 occurs, we will apply
+			// the below logic and reopen the door if the client got stuck in the door in the .15 s gap.
+
+			//only do this check when door is closing, and only for doors that block all directions (like airlocks)
+			if (!IsOpened && !registerTile.OneDirectionRestricted)
+			{
+
+				//Check if there's anything occupying this space that's impassable (other than this door)?
+				var registeredTiles = MatrixManager.GetAt<RegisterTile>(transform.position.RoundToInt());
```
fomalsd: any reason why you didn't use 
`MatrixManager.IsPassableAt(registerTile.WorldPosition, registerTile.WorldPosition, true, this.gameObject)` here?

chairbender: Yes. The first time I tried fixing it, I DID use that method, but it doesn't work. 

The door itself is already marked as impassable, so this method always returns false. Due to the way passability is determined, it was not feasible to add the ability to exclude the door from the passability check. If you dig into it, passability is checked using TileUtils.IsPassable, which checks against BasicTile - BasicTile does not check passability for the individual objects on it - it's an all or nothing thing.

chairbender:
(marked the wrong fix # in the original PR, it's actually supposed to be #1110)

UnityProject/Assets/Scripts/Doors/DoorController.cs
fomalsd: `Layer.IsPassableAt()` overridden in ObjectLayer, and airlock is an Object, check it out:
https://github.com/unitystation/unitystation/blob/0fc17b6faa8a885c292b2d468c40e2fd4144643d/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/ObjectLayer.cs#L68

chairbender: nice find, thanks for the tip. I'll try that

chairbender:
fyi @fomalsd there was a bug in ObjectLayer.IsPassableAt which was responsible for MatrixManager.IsPassableAt not working for my case. I fixed the bug and switched to use MM.IsPassableAt. Ready for re-review

fomalsd:
Sweet, thanks!

Jibbajabbafic:
Does this mean windoors have been fixed too (with the IsPassableAt fix)?

PR #1468, opened 2019-01-27
Shutters should be already closed from start #1437

### LINKED ISSUE #1437: Shutters should be already closed from start
# Description
Currently shutters always start opened and will close after a few seconds (it plays the normal closing animation). This affects things like Atmos, e.g. if the shutter is next to a space tile, the gases from the room will be drained at round start.

## Category
BUG

## Detailed Bug Report
Desired behaviour is that the shutters already start closed (like doors) and don't play the closing animation. Beware that shutters could also start opened, when the "IsClosed"-Flag the "RegisterDoor"-script is not selected.

## Steps to Reproduce
1. Look in scene editor at a shutters
2. press Play
3. See how the shutters start opened and will close after a few seconds

christmoore:
### Purpose
_Fixes #1437_<br>
_Allows the opening/closing of shutters at game start up by ticking is Closed flag of Shutter's Register Door script_

Example of running in edit mode, the left two shutters have IsClosed to *false*, the right two to *true*
  
  
The below example has two switches. One is set to open the two shutters on the left (set to isClosed = true). The set at the bottom left is a switch set to close a shutter (set to isClosed = false)

* [x]   This fix is tested on the same branch it is PR'ed to.
* [x]  I correctly commented my code
* [x]  My code is indented with tabs and not spaces
* [x]  This PR does not include any unnecessary .meta, .prefab or .unity (scene) changes
* [x]  This PR does not bring up any new compile errors
* [x]  This PR has been tested in editor
* [x]  This PR has been tested in multiplayer

### Notes:
_The only odd change needed to be done was having to move animator.SetBool("close", isClosed). There was only a single Shutter that would not close, and for some reason would not reach the animator.SetBool("close", isClosed) line when it was below the if{} else{} block within SetState() method_

### DIFF
--- a/UnityProject/Assets/Animations/Doors/Shutters/04_shuttersClosed.anim
+++ b/UnityProject/Assets/Animations/Doors/Shutters/04_shuttersClosed.anim
@@ -6,7 +6,7 @@ AnimationClip:
   m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInstance: {fileID: 0}
   m_PrefabAsset: {fileID: 0}
-  m_Name: 04_shuttersClose
+  m_Name: 04_shuttersClosed
   serializedVersion: 6
   m_Legacy: 0
   m_Compressed: 0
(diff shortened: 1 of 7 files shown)


christmoore:
# Changes to ShutterController Animator
## FROM
image

## TO
image


The new 04_shutterClosed and 04_shutterOpen are determined at game start by the IsClosed empty state, which checks the value of _close_ parameter and sets the initial state of the door (open or closed) corresponding to the "IsClosed" flag from "RegisterDoor".

This is made possible by triggering the shutter SetState() (through Trigger method) at the object's creation.

image

DooblyNoobly:
thanks! There is a problem with Matrix.Get as Matrix.Objects is not initialized yet but I'll fix that separately.

christmoore:
Ah sorry - that was missed!

PR #1481, opened 2019-01-28
Make debug log more usable

christmoore:
### Purpose
_Fixes issue #1472_<br>
_Implements a new console logger and removes the previous one. This new logger allows us to filter through informational messages, warnings, and errors, as well as provide input to the console to run console commands during runtime._

* [x]  This fix is tested on the same branch it is PR'ed to.
* [x]  I correctly commented my code
* [x]  My code is indented with tabs and not spaces
* [x]  This PR does not include any unnecessary .meta, .prefab or .unity (scene) changes
* [x]  This PR does not bring up any new compile errors
* [x]  This PR has been tested in editor

Not tested in multiplayer due to no direct multiplayer impact.

.meta changes include necessary .meta changes from asset import
Manager.prefab changed to remove previous logger within UIManager object.
Lobby.unity scene changed to include IngameDebug

### Notes:

F5 to open/close the logger console window


DooblyNoobly:
Did you submit this using unity 2018.3? All the prefabs are missing in the lobby scene:
image

DooblyNoobly:
I also noticed in your GIF that you are using F5 to open the actual console. F5 should be used to toggle the whole system not for switching between the small notification pop out and the console itself. We don't want the notification bar on by default as this system will be in the steam builds.

christmoore:
I should be sticking with 2018.2? The wiki states `Note that you should use Unity 2018.2 and above as of 01.08.2018.`, so I assumed that 2018.3 was fine. I will close and rework using the earlier unity version

fomalsd:
Actually we're moving to 2018.3 the day after tomorrow, so you can just keep this open for a few more days and we'll re-review it after update

christmoore:
Alright! I will fix the console F5 implementation and we can hold off until then.

christmoore:
F5 implementation now shows/hides the popup window during runtime, and will close the log window if it is opened as well.

christmoore:
This should be ready now to be merged once 2018.3 comes along. Let me know if there's any issue.

DooblyNoobly:
cool thanks

DooblyNoobly:
Looks good! I might play around with the GUI and fonts after I merge this.

PR #1515, opened 2019-02-05
#1467 improve the appearance of rotated stuff, fix unrelated bugs

### LINKED ISSUE #1467: Improve the appearance of rotated things
# Description
In ss13, sprites for objects always appear upright regardless of how the matrix / object is rotated. But in unitystation, currently we rotate the sprite based on how the matrix is rotated. This looks a bit weird and makes the game less legible. 

You can see this by spawning as syndicate, going up to the ship controls, and using the arrows to rotate the ship. You can also see this by looking at some of the wallmounts throughout the station, as many of them are rotated based on which side of the wall they are on.

For example, look at how the table / vending machine looks when we rotate the ship. The legs stick out to the left, when it would be better if they pointed down:
rotation

For this issue, improve the logic of how rotated objects appear so that they all look nice. This will involve some judgement and experimentation, as some objects might actually be better off rotating with the matrix rather than always remaining upright.

Here's the general rules that should be used:
1. If the object has a multi-directional sprite, use the appropriate sprite based on its facing (this is how Chair, clothing, and player already works).
2. If the object has only one sprite, it should always be upright (the top of the sprite points to the top of the screen.
3. When matrix rotation occurs, sprites should only update their facing at the end of rotation.
4. Things which have some sort of directionality (like windoors and wires) can rotate with the matrix and are exempt from the above rules.

This should be configurable to facilitate experimentation. For example, we are not sure if airlocks and wallmounts should always stay upright. There are others that might be questionable as well. It should be possible to have some which rotate with the matrix and some that always remain upright.

If you're uncertain about how a particular thing should rotate, please get in touch on Discord so we can get some eyes on it and decide together.

chairbender:
### Purpose
Fixes #1467 

Makes everything remain upright unless it is explicitly set
not to (using the checkbox on RegisterTile
called rotateWithMatrix). Currently, everything snaps to
the new orientation at the end of a matrix rotation, but this
can be easily changed by modifying the ROTATE_AT_END constant.

Also fixes a bug where matrix.GetOffset
was returning incorrect values due to caching logic, which
may have caused oxygen to disappear when turning a matrix and
caused player to teleport into space when changing matrices. There
may have been other latent bugs caused by this that are now fixed.

As usual, added lots of comments as well.

### Notes:
Also did some perf testing locally, seems fine. Here's my rough notes from perf testing:
develop branch
FPS:
63 to 65 on station, 65-80 on syndi

non-spike perf:
PlayerLoop - 10 KB, 11.42 ms
Updatemanager.Update - 8.8 KB, 3.25 MS
playersync.update 404 B gc, .15 ms

Large spikes:
LightingSystem
UpdateManager.Update
PlayerSync

my branch:
on station
FPS 65-70
non-spike perf:
PlayerLoop 9.4 kb, 10 ms
Updatemanager.Update 7.8 kb 1.9 ms
playersync.update 404 B, .1 ms

Large spikes:
LightingSystem
UpdateManager.Update
PlayerSync

So the takeaway is that there shouldn't be any perf problems introduced with this change. I tested on syndi ship while rotated and outpost station for both of these.

### DIFF
--- a/UnityProject/Assets/Prefabs/Items/Other/emergency_lights.prefab
+++ b/UnityProject/Assets/Prefabs/Items/Other/emergency_lights.prefab
@@ -1,21 +1,11 @@
 %YAML 1.1
 %TAG !u! tag:unity3d.com,2011:
---- !u!1001 &100100000
-Prefab:
-  m_ObjectHideFlags: 1
-  serializedVersion: 2
-  m_Modification:
-    m_TransformParent: {fileID: 0}
-    m_Modifications: []
-    m_RemovedComponents: []
-  m_SourcePrefab: {fileID: 0}
-  m_RootGameObject: {fileID: 1275749971873228}
-  m_IsPrefabAsset: 1
 --- !u!1 &1128005765242482
 GameObject:
   m_ObjectHideFlags: 0
   m_CorrespondingSourceObject: {fileID: 0}
-  m_PrefabInternal: {fileID: 100100000}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
   serializedVersion: 6
   m_Component:
   - component: {fileID: 4185858802756340}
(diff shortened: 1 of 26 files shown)


DooblyNoobly:
Done a fair bit of testing, this is actually a good solution. I like it. For some reason the chair sprites on Cargo shuttle changed back to the default ones so we just need to change them in the next mapping job.

PR #1523, opened 2019-02-08
fix rotation of bullet casings, shards, and blood

chairbender:
Fix them so they randomly rotate like before and stay rotated with the matrix.


chairbender:
merging as this is a small change to a few prefabs

PR #1528, opened 2019-02-08
#1506 Player speaking while in critical state

### LINKED ISSUE #1506: Player can chat in crit
# Description
Trying to talk while you're in crit shows the dialogue bubble

## Category
BUG

## Detailed Bug Report
OS: Windows 10 Pro x64

Under no circumstances should someone be able to talk while they are in critical condition. If this actually does transmit the player's words it's worse, but if this is just a client-side visual quirk it should still be fixed to avoid ambiguity in what the player expects vs. what happens.

## Steps to Reproduce
1. Start unitystation
2. Get into crit
3. Attempt to talk

Fishfish458:
### Purpose
When a user was in critical state, they were still able to communicate
Fixes #1506

### Notes:
I chose the method of adding it as a modifier over a simple fix ignore talk because I believe this would make any future changes to critical state easier (death whisper or such). I added a check just after modifiers that if the message is empty after modifiers are applied that there shouldn't be any message. If this is too much I can just go with the simple block text chat in ControlChat.cs

### DIFF
--- a/UnityProject/Assets/Scripts/Chat/ChatEvent.cs
+++ b/UnityProject/Assets/Scripts/Chat/ChatEvent.cs
@@ -33,6 +33,7 @@ public enum ChatModifier
 	None = 0,
 	Drunk = 1,
 	Stutter = 2,
+	Crit = 3,
 	Hiss = 4,
 	Clown = 8
 }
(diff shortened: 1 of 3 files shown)


chairbender:
Very nice, I like how you did it. Thank you!

PR #1529, opened 2019-02-09
#1521 Fixes various issues related to rotating matrices / rotations

### LINKED ISSUE #1521: Various MatrixMove Issues
# Description
There are a few various MatrixMove issues that need fixing.

1. When FPS is low enough, chairs don't complete rotation and end up at a non-90 degree angle. Apparently this can be reproduced in editor by going into the cargo shuttle and turning on atmospheric overlay which displays the O2 levels.
2. Drag a lockerful of items onto cargo shuttle and open it. Objects fly away when you start moving or move around when you rotate. Probably the items are not being re-parented to the new matrix.
3. Turn on the atmsopheric overlay to see the O2 levels on the cargo shuttle. Rotate the ship. Notice how some tiles drop to 0. They should remain at their previous values. If you go into these tiles you'll be unable to breathe. It looks like the atmospherics on each tiles remain fixed where they are in world space and do not rotate with the matrix (just look at the numbers).
4. When trying to melee the glass on the cargo shuttle while moving, the character suddenly gets frozen in world space (rather than local space) during the melee animation, then lerps back to the correct position. Instead they should remain where they are in local space.
5. When trying to melee the glass on the syndie shuttle near the thrusters, we get an exception in the logs (Missing Key Exception or something like that).
6. Anything else that you notice related to moving matrix. In particular, make sure to test what happens when things cross matrices and the matrices rotate.

Here's a gif from foma showing point 2 and 3:
gif

chairbender:
### Purpose
Fixes #1521 

Fix issue where rotated objects are not quite precisely at
90 degree increments.

When object is hidden, preserve everything else in its
transformstate but set only its position to HiddenPos. Fixes
issue where closets dragged to a different matrix were parented
to outpost station and got stuck in world space.

Also sets uninitialized transform state matrix IDs to -1 so uninitialized
transforms no longer parent to outpost station by default and we can
detect the error earlier (fail fast)

Fix missing key exception when trying to show damaged window tile

Fix issue where player gets stuck in world position
when lerping during the melee attack animation

Fix issue where doors were not rotating and clients were
not getting rotated matrices when joining

Fix issue where locker and other stuff rotates when
on a rotated matrix and pushed / pulled

Fix NRE on load related to closets.

### Notes:
Also note that I changed TransformState so that IsLocalRotation is no longer there and clarified the purpose of the Rotation float by renaming it to SpinRotation, and that it is always used for localRotation only as it is only used for spinning objects.

### DIFF
--- a/UnityProject/Assets/Scripts/Camera/Camera2DFollow.cs
+++ b/UnityProject/Assets/Scripts/Camera/Camera2DFollow.cs
@@ -10,7 +10,7 @@ public class Camera2DFollow : MonoBehaviour
 	private readonly float lookAheadMoveThreshold = 0.1f;
 	private readonly float lookAheadReturnSpeed = 0.5f;
 
-	private readonly float yOffSet = -0.5f; 
+	private readonly float yOffSet = -0.5f;
 
 	private Vector3 cachePos;
 	private Vector3 currentVelocity;
@@ -82,7 +82,7 @@ public class Camera2DFollow : MonoBehaviour
 		if(!PlayerManager.LocalPlayerScript.weaponNetworkActions){
 			return;
 		}
-		if (target != null && !isShaking && !PlayerManager.LocalPlayerScript.weaponNetworkActions.lerping)
+		if (target != null && !isShaking)
 		{
 			// only update lookahead pos if accelerating or changed direction
 			float xMoveDelta = (target.position - lastTargetPosition).x;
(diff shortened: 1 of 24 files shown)


chairbender:
I think it's fine to go ahead and merge this as its a bunch of fixes which should help other people's tickets. Would still be good to get some testing but if there's anything I didn't notice people can bring it up in discord.

PR #1543, opened 2019-02-10
Makes drinks containers now

Bod9001:
Very small  Change/tweak

### Purpose
All drinks are containers now 
So you can mess around with the chemistry system now without having to spawn in stuff

### Notes:
I want to get player chemical reaction working before I add more interactions

### DIFF
--- a/UnityProject/Assets/Scripts/Chemistry/ReagentContainer.cs
+++ b/UnityProject/Assets/Scripts/Chemistry/ReagentContainer.cs
@@ -4,11 +4,10 @@ using UnityEngine;
 
 public class ReagentContainer : Container {
 	public float CurrentCapacity;
-	public List<string> Chemicals;
-	public List<float> Amounts;
+	public List<string> Chemicals; //Specify chemical
+	public List<float> Amounts;  //And how much
 
-	// Use this for initialization
-	void Start () {
+	void Start() {//Initialise the contents if there is any
 		for (int i = 0; i< Chemicals.Count; i++)
 		{
 			Contents[Chemicals[i]] = Amounts[i];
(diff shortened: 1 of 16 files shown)


UnityProject/Assets/Scripts/Chemistry/ReagentContainer.cs
```diff
@@ -68,7 +71,7 @@ public class ReagentContainer : MonoBehaviour {
 		double DivideAmount = Using / CurrentCapacity;
 
 		Dictionary<string, float> Transfering = new Dictionary<string, float> ();
-		Dictionary<string, float> BrokenCS = new Dictionary<string, float> (Contents); 
+		Dictionary<string, float> BrokenCS = new Dictionary<string, float> (Contents);  //Because apparently am changing the order just by modifying the values That are tied with keys
```
Jibbajabbafic: If order is important you could use an OrderedDictionary instead: https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.ordereddictionary?redirectedfrom=MSDN&view=netcore-2.2

The order in normal dictionaries isn't guaranteed to stay the same.

Bod9001: It doesn't matter too much About the order,  It's mainly because stuff like "Contents [Chemical.Key] = 0;" Apparently messes up the for loop so That's why it's there, I could change the Contents to an OrderedDictionary then Change the loop to use that or Does that not allow me to change Contents [Chemical.Key] = 0; to  Contents.Remove(Chemical.Key)?

Jibbajabbafic: If you need to remove entries from a dictionary you're better off checking if values are 0 and using the Remove method on the dictionary. Adding things with names like "BrokenCS" to try and work around C# is going to make the codebase more confusing, and is probably unnecessary since it can be achieved in a different way.

Jibbajabbafic (changes requested):
See other  comment

UnityProject/Assets/Scripts/Chemistry/ReagentContainer.cs
Bod9001: yeah, That's how it set up currently But setting the value to zero still causes the loop to break

Bod9001: That's why it's called BrokenCS  Because it makes no sense Why should cause the dictionary to change length/break a for loop

Jibbajabbafic (approved):

PR #1544, opened 2019-02-10
Added 'none' damage state for windows

### LINKED ISSUE #1541: Window cracks remaining after window is broken B: 25
# Description
In title

## Steps to Reproduce
1. Throw a weapon / heavy object on window till it breaks

scr 2019-02-10 2 57 17 2019-02-10 02_58_58

hecksadecimal:
### Purpose
Fixes #1541
There was no state change taking effect when windows got broken, leaving behind the cracks. A new 'none' WindowDamage state addresses this.

### Notes:
Adding the 'none' WindowDamage state caused unity to automatically change a few other files in the process:
POst-PPRT Transform Material.mat
Post-Final Blit Material.mat
Post-FloorFovGenerator Material.mat
Post-FovGenerator.mat
Managers.prefab
UIManager.prefab
none.asset.meta

The files I personally added or changed were:
TilemapDamage.cs
none.asset

Additionally as there was only one line of code added, commenting it seemed redundant.

### DIFF
--- a/UnityProject/Assets/Materials/PostProcess/POst-PPRT Transform Material.mat
+++ b/UnityProject/Assets/Materials/PostProcess/POst-PPRT Transform Material.mat
@@ -5,8 +5,7 @@ Material:
   serializedVersion: 6
   m_ObjectHideFlags: 0
   m_CorrespondingSourceObject: {fileID: 0}
-  m_PrefabInstance: {fileID: 0}
-  m_PrefabAsset: {fileID: 0}
+  m_PrefabInternal: {fileID: 0}
   m_Name: POst-PPRT Transform Material
   m_Shader: {fileID: 4800000, guid: c812736daeccca7439b94c819cadb594, type: 3}
   m_ShaderKeywords: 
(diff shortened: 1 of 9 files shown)


Jibbajabbafic:
For futue reference, please avoid commiting all changes and only include changes that you made where possible. If the only change you made was to the script below, then that should be the only file you include. https://github.com/unitystation/unitystation/blob/f5cd85c1c548d369a33f9151b0de3546bad4382a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/TilemapDamage.cs#L146

If you change some prefabs or scripts and it changes the meta files associated with them then it's fine to include those too. Since this change adds 'none.asset' it's fine to include 'none.asset.meta', but the other prefabs shouldn't be necessary since this fix has nothing to do with the Managers.prefab or UIManager.prefab. 

As for the .mat files I think it might because they were made with an older version of unity and they're getting updated. I'll have a look into it.

Jibbajabbafic (changes requested):
See other comment

fomalsd:
Please ignore all .mat files starting with "Post" from commits. These materials are related to FoV system and are modified locally every time you play game.
Also prefabs and scenes shouldn't be included unless you explicitly created/modified them.
But if you created a new prefab (like that `none` one), be sure to include its .meta file, too

hecksadecimal:
Alright, I set those files back to their initial states before I sent the PR.

Jibbajabbafic (approved):

Jibbajabbafic:
Just doing a quick test

Jibbajabbafic:
Don't see any cracks leftover!

PR #1546, opened 2019-02-11
Refactor ShowTab to take into account NetTab.isPopOut

### LINKED ISSUE #1533: Refactor how popup tabs are checked B: 25
# Description
New popout tabs have to be placed into a hardcoded check to make them popout.

https://github.com/unitystation/unitystation/blob/f9c5597aac500808ad4d67615d89a69f6caab47c/UnityProject/Assets/Scripts/UI/ControlTabs.cs#L426

This isn't ideal because the tabs already have a `IsPopOut` variable to determine this,

image

The code should be refactored so this hardcoded check isn't necessary anymore.

## Category
TODO

hecksadecimal:
### Purpose
Fixes #1533 

ShowTab used to rely on a static list of tab types, it's been refactored to instead rely on the tab's isPopout variable.

### Notes:
If there's a better way to do this please let me know and I'll work on changing it. I'm not too familiar with Unity just yet.

### DIFF
--- a/UnityProject/Assets/Scripts/UI/ControlTabs.cs
+++ b/UnityProject/Assets/Scripts/UI/ControlTabs.cs
@@ -420,13 +420,10 @@ public class ControlTabs : MonoBehaviour
 			return;
 		}
 
-		bool isPopOut = false;
+		NetTab thisTab = NetworkTabManager.Instance.Get(tabProvider, type);
 
-		//Add the popout types here:
-		if ((type == NetTabType.Paper) || (type == NetTabType.ChemistryDispenser))
-		{
-			isPopOut = true;
-		}
+		//Make use of NetTab's fancy isPopOut bool instead of depending on the NetTabType.
+		bool isPopOut = thisTab.isPopOut;
 
 		if (!Instance.rolledOut && !isPopOut)
 		{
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/UI/ControlTabs.cs
```diff
@@ -420,13 +420,10 @@ public static void ShowTab(NetTabType type, GameObject tabProvider, ElementValue
 			return;
 		}
 
-		bool isPopOut = false;
+		NetTab thisTab = NetworkTabManager.Instance.Get(tabProvider, type);
```
fomalsd: Won't work for client – NetworkTabManager is serverside-only:
https://github.com/unitystation/unitystation/blob/c29f55cf197f2d2a2e8b56ac77765c6c0327cc37/UnityProject/Assets/Scripts/Managers/NetworkManagement/NetworkTabManager.cs#L7

fomalsd (changes requested):
You're trying to access a tab list that's only present on server.
Try getting isPopOut property after you actually spawned it locally

UnityProject/Assets/Scripts/UI/ControlTabs.cs
hecksadecimal: > 
> 
> You're trying to access a tab list that's only present on server.
> Try getting isPopOut property after you actually spawned it locally

How are these latest changes?

hecksadecimal:
Made changes based on the input given.

fomalsd (approved):

fomalsd:
Good job!

PR #1556, opened 2019-02-15
Linux workflow fixes

hecksadecimal:
### Purpose
Attempting to work on the project in the Linux Unity Editor has been a troubling task for a variety of different reasons. 

1. SunVox, for some strange reason, causes the editor to crash when SunVox.sv_deinit() is called in a linux environment.

2. Additionally, when importing the project in unity for the first time it makes sweeping changes to a lot of .meta files.

This PR addresses both of these problems via the following changes: 

1. SunVox.sv_deinit() is altered with a preprocessor to avoid the crash in the Linux Unity Editor while leaving the call unaltered in full builds. 

2. A script has been added that when used will go through all the changed .meta files and perform a `git update-index --assume-unchanged` on them. This prevents git commit pollution with unintentional merging of altered .meta files

### Notes:

Additionally, a `using static` directive was added to Synth.cs to fix VSCode intellisense and codelense not being able to locate the reference.


hecksadecimal:
I would appreciate if someone could test this PR in a Windows or macOS environment to ensure my changes don't break anything in those systems.

Jibbajabbafic:
What changes are shown on the meta files when you do a git diff? It could just be permissions getting changed. If that's the case I'd recommend setting the filemode setting to false instead since meta file changes can be important

hecksadecimal:
It isn't just permissions unfortunately. The editor makes actual changes to the contents of the file. Here's an example diff of `UnityProject/Assets/Textures/Turf/walls.meta` Which represents a full deletion of the .meta file that the editor wanted to perform.

```
@@ -1,9 +0,0 @@
-fileFormatVersion: 2
-guid: ed3ef2eb89e34c46aee258b42550d970
-folderAsset: yes
-timeCreated: 1477562706
-licenseType: Pro
-DefaultImporter:
-  userData: 
-  assetBundleName: 
-  assetBundleVariant: 
```
And here's one which is modified rather than deleted.
`UnityProject/Assets/Textures/Turf.meta`

```
@@ -1,8 +1,6 @@
 fileFormatVersion: 2
-guid: 042cc493bdd7b414c88e586a1234b4ff
+guid: c29d2d460db5b73d39101afd46d6e7dc
 folderAsset: yes
-timeCreated: 1509379128
-licenseType: Free
 DefaultImporter:
   externalObjects: {}
   userData:
```

As you can see, the Linux Unity Editor has a habit of changing meta GUIDs among other things on import.

UnityProject/Assets/Scripts/Sound/SunVox.cs
```diff
@@ -47,6 +47,8 @@ public struct sunvox_note {
 
 #if UNITY_EDITOR_OSX || UNITY_STANDALONE_OSX
   private const string LIBRARY_NAME = "sunvox";
+#elif UNITY_EDITOR_LINUX || UNITY_STANDALONE_LINUX
+  private const string LIBRARY_NAME = "sunvox";
 #elif UNITY_EDITOR_WINDOWS || UNITY_STANDALONE_WINDOWS || UNITY_STANDALONE_WIN
   private const string LIBRARY_NAME = "sunvox";
 #elif UNITY_STANDALONE_LINUX
```
Jibbajabbafic: This bit won't be needed anymore (line 54) since it's covered by the elif you added

assume_unchanged.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
```
Jibbajabbafic: This file would be better off in a dedicated scripts directory so it doesn't clutter the top level

krille90:
```
#if UNITY_EDITOR_OSX || UNITY_STANDALONE_OSX
  private const string LIBRARY_NAME = "sunvox";
#elif UNITY_EDITOR_LINUX || UNITY_STANDALONE_LINUX
  private const string LIBRARY_NAME = "sunvox";
#elif UNITY_EDITOR_WINDOWS || UNITY_STANDALONE_WINDOWS || UNITY_STANDALONE_WIN
  private const string LIBRARY_NAME = "sunvox";
#elif UNITY_STANDALONE_LINUX
  private const string LIBRARY_NAME = "sunvox";
```

Couldn't this be shortened? (It always just the same name)

hecksadecimal:
It probably could be shortened, but as it was a list like that before I touched it to add the Linux entries changing it to shorten may be beyond the intended scope of this PR.

My only intention was to add fixes to make some crashes go away in the Linux editor and help prevent automatic .meta changes from poisoning commits.

Jibbajabbafic (changes requested):
Only minor nitpicks, but I won't be able to test it today unfortunately.

Jibbajabbafic:
> ```
> #if UNITY_EDITOR_OSX || UNITY_STANDALONE_OSX
>   private const string LIBRARY_NAME = "sunvox";
> #elif UNITY_EDITOR_LINUX || UNITY_STANDALONE_LINUX
>   private const string LIBRARY_NAME = "sunvox";
> #elif UNITY_EDITOR_WINDOWS || UNITY_STANDALONE_WINDOWS || UNITY_STANDALONE_WIN
>   private const string LIBRARY_NAME = "sunvox";
> #elif UNITY_STANDALONE_LINUX
>   private const string LIBRARY_NAME = "sunvox";
> ```
> Couldn't this be shortened? (It always just the same name)

Agreed, most of this seems unnecessary

fomalsd:
That `assume_unchanged` could backfire, I suggest using separate changelists and just keeping unnecessary files there. You might want to commit some of them eventually.

default

hecksadecimal:
@Jibbajabbafic I made the changes you requested, but should I also refactor the whole #if #elif block to cut down on the entries in this PR?

hecksadecimal:
@fomalsd Changelists? I'm using VSCode, isn't that an IntelliJ Idea feature?

fomalsd:
I think GUI git clients should have this feature? but yeah, that screenshot is from Rider

hecksadecimal:
Never used a git GUI, I've always worked with CLI. In the event I do want to apply a commit with an asset .meta I intentionally changed, I can just manually `git update-index --no-assume-unchanged` on it to be able to commit it.

fomalsd:
Ok, it's your call, feel free to keep that script in

fomalsd (approved):
No side effects on win10

krille90:
> @Jibbajabbafic I made the changes you requested, but should I also refactor the whole #if #elif block to cut down on the entries in this PR?

Yea, I would prefer that.

hecksadecimal:
@krille90 How's that? Works just as well on my system and it really cuts down on how many entries there is.

Issue #1512, opened 2019-02-05
pressing E and clicking on a cloth slot dupes the icon B: 25


DiscountVee:
# Description
If you press E and click fast on a clothing slot, the icon of the clothing in question duplicates
## Category
BUG 

## Detailed Bug Report
If you mouse over and click on an equipped piece of clothing, then start mashing E and clicking, the item duplicates, one remains in your hand and the other on your character
## Steps to Reproduce
Please enter the steps to reproduce the bug or behaviour:
1.Click on an equipped piece of clothing
2.Press E and click together fast while pointing at the clothing slot
3.Duplicate the item

Fishfish458:
I looked into this a little bit and this is happening when slots are being changed quickly no matter the method. It seems that items ARE still being moved but sometimes the sprite doesn't properly update(get removed) afterwards.

DooblyNoobly:
Bounty of 25 usd will be awarded to the contributor on a successful merge of a solution to this bug

Fishfish458:
Going to attempt to look into this one

PR #1565, opened 2019-02-19
Drop items when unconcious, fix UI not clearing while incapped

Fishfish458:
### Purpose
This isn't listed as an issue but I came across it while fixing another one
If you die with something in your hand, the UI icon will [still be there on death](https://imgur.com/wAmEy9T)
This removes the interaction check when clearing a UI slot and adds dropitem upon unconcious state

### Notes:

### DIFF
--- a/UnityProject/Assets/Scripts/UI/UI Bottom/UI_ItemSlot.cs
+++ b/UnityProject/Assets/Scripts/UI/UI Bottom/UI_ItemSlot.cs
@@ -145,7 +145,7 @@ public class UI_ItemSlot : MonoBehaviour, IDragHandler, IEndDragHandler
 	public GameObject Clear()
 	{
 		PlayerScript lps = PlayerManager.LocalPlayerScript;
-		if (!lps || lps.canNotInteract())
+		if (!lps)
 		{
 			return null;
 		}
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/UI/UI Bottom/UI_ItemSlot.cs
```diff
@@ -145,7 +145,7 @@ public void SetSecondaryImage(Sprite sprite)
 	public GameObject Clear()
 	{
 		PlayerScript lps = PlayerManager.LocalPlayerScript;
-		if (!lps || lps.canNotInteract())
+		if (!lps)
```
fomalsd, line 148: does it fix things like #1512?

Fishfish458: Unfortunately it doesn't seem relevant to #1512 since it still occurs. clear should be clearing the UI space when called and it was causing some issues when it's checking if the player can interact or not.

Sorry if this pr is slightly unconventional, the issue I'm trying to fix is
1. Grab an item
2. Go into crit with item, notice character still holds onto item
3. Die with item, notice UI still has item

DooblyNoobly:
thanks Fishfish!

PR #1568, opened 2019-02-19
Chat hotkeys, UI tweaks and channel tooltips

### LINKED ISSUE #1510: Add ability to open radio chat and department chat with keybindings
# Description
When attempting to local chat, no text is displayed. Radio and department chat buttons do not even bring up the text entry box.

## Category
TODO
## Detailed Bug Report
OS: Windows 10 Pro x64

This one seems pretty critical. Unsure at this point if it effects connecting to a server that isn't localhost, but even if it only does effect localhost client-servers it is a glaring issue.

## Steps to Reproduce
1. Run unitystation
2. Join round
3. Attempt to talk using any of the bound chat keys

Jibbajabbafic:
### Purpose
- Adds hotkeys for OOC and radio channel (fixes #1510 )
- Stopped player from selecting None channel
- Made chat UI neater
- Added tooltips for all the channels which don't go off screen

### DIFF
--- a/UnityProject/Assets/Prefabs/Chat/ChatSystem.prefab
+++ b/UnityProject/Assets/Prefabs/Chat/ChatSystem.prefab
@@ -1,21 +1,11 @@
 %YAML 1.1
 %TAG !u! tag:unity3d.com,2011:
---- !u!1001 &100100000
-Prefab:
-  m_ObjectHideFlags: 1
-  serializedVersion: 2
-  m_Modification:
-    m_TransformParent: {fileID: 0}
-    m_Modifications: []
-    m_RemovedComponents: []
-  m_SourcePrefab: {fileID: 0}
-  m_RootGameObject: {fileID: 1897758349740974}
-  m_IsPrefabAsset: 1
 --- !u!1 &1026355951731664
 GameObject:
-  m_ObjectHideFlags: 1
+  m_ObjectHideFlags: 0
   m_CorrespondingSourceObject: {fileID: 0}
-  m_PrefabInternal: {fileID: 100100000}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
   serializedVersion: 6
   m_Component:
   - component: {fileID: 224820951895999306}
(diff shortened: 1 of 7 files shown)


fomalsd:
Very neat! Tooltips are really helpful! 
A few minor things:
- Channel selection bar doesn't refresh when available channel list gets changed (until you reopen it manually) 
- Ghost channel still appears like None (probably text doesn't refresh in time)

Consider using events to update bar and text list the moment available channels field gets modified!
gif
- Text you just posted appears cropped until you close/reopen chat
gif

Jibbajabbafic:
Thanks for having a look!

Are we ok to merge for now? I'm worried about merge conflicts happening as people work on other things.

Also, any recommendations on where to look in the code to see how events work? And are they unity events or custom event manager ones?

fomalsd:
yeah, it's good to merge.
I use UnityEvents, you can check out PlayerSync for usage examples:
https://github.com/unitystation/unitystation/blob/3588b1088dce2038b3ee7da992064eeae1c1b9f2/UnityProject/Assets/Scripts/Player/PlayerSync.Server.cs#L18-L19
scr 2019-02-20 21 35 34

donovanhubbard:
There are multiple event types. Unity Events are one type that are implemented. They are often good because you can put them in the editor if you want.

We also have a EventManager which you can use to send events with no arguments. These all go through a central system.  For examples on how to use this one, you can check out https://github.com/unitystation/unitystation/blob/develop/UnityProject/Assets/Scripts/UI/OxygenButton.cs I'll probably right a wiki up on it in the next week or so. Ultimately, I'd like to add something like the System.Reflection.EventInfo class to this one so we can make it more robust. 

There is also the EventController which looks a lot like the EventManager, but more complicated. It also doesn't look like it's actually used anywhere. I'd be tempted to remove it from the code base entirely. It looks like it hasn't had any significant modifications since 2017.

This does bring up a good point though. I love events, and I think they are criminally underused in the project. However, we have three different event systems. We should probably standardize and pick one or two and give clear instructions and guidance on how and when to use them.

Issue #1361, opened 2018-12-04
TODO: Player light emission B: 65


DooblyNoobly:
## Description

Whenever a player picks up a light, the light2d sprite that was on the Light item is disabled. A player light handler should then control the light emission that comes from the player when they are holding a light that is toggled on in any of the hand or belt slots. It should also work with helmets that have lights on them (coming in the future)

### Criteria:

 - Create a GameObject on the PlayerPrefab that will handle the control of the Light Emission from the player
 - Create a component/script to handle the behaviour of the player light
 - Add the new script to the root object of the player prefab and derive it from NetworkBehaviour
 - Drag the Light2D prefab onto the new object and add it to a field in the PlayerLightHandler script so that it can be enabled/disabled and the properties of the SpriteRenderer can be changed
 - Create a method that accepts a SpriteRenderer as a parameter (the other light2d sprite from the item)
 - Create a way to detect when a light has been picked up or dropped
 - Create a way to detect when a light has left any of the slots or entered any of the slots where the light emission will be visible
 - Switch the light2D sprite on the player on and off based on the required conditions, also replicate the attributes of the light2D that is being mimicked
 - Sync the behavior over the network (Ask me for help if you get stuck with this one)

A bounty of 65usd will be awarded to the contributor on the successful merge of these features.

dapocalypse:
claim

chairbender:
fyi to anyone, this is up for grabs to be worked on again

PR #1584, opened 2019-02-24
player Light Emission!

Bod9001:
### Purpose
to add player Light emission
Due to the way it set up only one light can be present at one time (as in seen)The mechanism that determines which one is which one has the highest intensity (brightness basically)

- [Tell me to add more if there's not enough]  I correctly commented my code

### Notes:
Any changes need I can do


DooblyNoobly (changes requested):
Can you look at disabling the lights when player is hidden. Look at the VisibleBehaviour script and remove the ignore meshes part (I think that was for the old FOV that is no longer being used).

Also can you test in MP if other players can see that default visible areas when they are hiding. And also if it's possible can you make the radius a bit smaller for it (it is a bit bigger then what is on TG)

DooblyNoobly:
btw this job fixes #1361

Issue #1504, opened 2019-02-05
Opening and closing lockers on your character make you lose 02 B: 35


DiscountVee:
# Description
Opening and closing lockers while you're inside them makes you get o2 signal for suffocation
## Category
BUG

## Detailed Bug Report
If you stand in a locker and then click on the locker fast to open and close you will get an o2 signal and start suffocating.
## Steps to Reproduce
Please enter the steps to reproduce the bug or behaviour:
1.Open a locker and stand in it
2.Start clicking fast on the locker to open and close it
3.Start suffocating

DooblyNoobly:
Bounty of 35 usd will be awarded to the contributor on a successful merge of a solution to this bug

Fishfish458:
Sounds hilarious and I'll look into it

coolcat1711:
Between a mix of fishfish's and my changes, I'm fairly sure that this issue is resolved.

DooblyNoobly:
@coolcat1711 I just rebuilt the steam clients and pushed them to steamworks. Log in and test it out. If it's fixed report back here and I'll split the bounty.

coolcat1711:
![unitystation issue 1504](https://user-images.githubusercontent.com/3880412/54009834-5e72c800-413a-11e9-920d-a575b25709bb.gif)

DooblyNoobly:
cool, looks like that flashing bug has been fixed too (where you would appear, disappear and reappear again after opening closet)

PR #1594, opened 2019-03-02
breathing/damage in lockers

Fishfish458:
### Purpose
This fixes the player not breathing in a locker as well as take damage while in a locker
This issue doesn't actually fix any current open issue, though it has made #1504  far less common.

### Notes:
This PR adds a new location grabbing recursive method to ObjectBehaviour. This method was put in objectbehaviour since it is intended to be on any object that can be stored inside something else.
If this should be located somewhere else, let me know and I can move it or make changes to it.
 I'll continue to expand on container layers if this is the proper way to go about it.


DooblyNoobly:
Thanks Fishfish, the approach looks good

PR #1602, opened 2019-03-03
Fix downed player sprite / rotation issues and melee on MatrixMove

### LINKED ISSUE #1240: Player corpses appearing upright when joining round in progress
# Description

## Category
BUG

## Steps to Reproduce
1. Host a server and die once as serverplayer
2. Join with a 200ms ping client
3. Take a look at serverplayer's corpse as client

### LINKED ISSUE #1567: Dying without crit state causes character to still be standing around
# Description
if the death state is triggered before the player goes into crit, the player character will still be standing around

## Category
TODO

## Detailed Bug Report
The crit state uses methods to cause the character to drop that the death state does not,
death should be able to work independent of if the character goes into a crit state or not.

## Steps to Reproduce
1. In the editor, use Networking > Gib All(server)
2. Notice how the character is still standing around

chairbender:
### Purpose
Fixes #1240 
Fixes #1567 

Basically, fixes network sync issues w.r.t. downed players (crit, soft crit, or dead) where they did not appear in sync between all players or looked wrong when matrices were rotated.

Also fixes melee interactions on rotated / moving matrices. The player would become stuck in world space while the attack animation occurred and the animation would be in the wrong direction.

### Notes:
Removes ForceRotation, instead handling rotation logic in RegisterPlayer.


UnityProject/Assets/Scripts/Player/PlayerSprites.cs
```diff
@@ -110,52 +142,119 @@ private IEnumerator WaitForLoad()
...
 	{
-		if(playerSync.isBumping)
+		if (registerPlayer.IsDown || playerSync.isBumping)
 		{
 			//Don't face while bumping is occuring on this frame
+			//or when player is down
 			return;
 		}
-		SetDir(direction);
+		if (playerMove.isGhost)
```
DooblyNoobly, line 220: If the player is down this condition is never reached therefore disabling ghost directional sprites from being updated.

DooblyNoobly (changes requested):
Ghost directional sprites are not updated

UnityProject/Assets/Scripts/Player/PlayerSprites.cs
DooblyNoobly: going to push through a fix for this after merging so we can get the PR through

PR #1611, opened 2019-03-06
Fixing player light emission, Fixing wrongly rotated wall mounts

Bod9001:
### Purpose
Fixing player light emission, Fixing wrongly rotated wall mounts
, Added gizmo for direction of wallmounts, Added self powered mode, and Fixed APC detection

### Notes:
if (RelatedAPC != null ) //|| SelfPowered
Technically shouldn't work but it does so Just take it out of the comment if it starts being Strange

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/WallmountBehavior.cs
+++ b/UnityProject/Assets/Scripts/Objects/WallmountBehavior.cs
@@ -44,8 +44,7 @@ public class WallmountBehavior : MonoBehaviour
 	void OnDrawGizmos ()
 	{
 		Gizmos.color = Color.green;
-        //Gizmos.DrawSphere(transform.position , 3);
-		Gizmos.DrawLine(transform.position,(transform.position-transform.up ));
+		GizmoUtils.DrawArrow(transform.position,(new Vector3() - transform.up) );
 	}
 
 	/// <summary>
--- a/UnityProject/Assets/Scripts/Objects/WallmountBehavior.cs
+++ b/UnityProject/Assets/Scripts/Objects/WallmountBehavior.cs
@@ -44,7 +44,7 @@ public class WallmountBehavior : MonoBehaviour
 	void OnDrawGizmos ()
 	{
 		Gizmos.color = Color.green;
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Objects/WallmountBehavior.cs
```diff
@@ -39,6 +40,14 @@ public bool IsFacingPosition(Vector3 worldPosition)
 		return difference >= 91 || difference <= -91;
 	}
 
+	//[DrawGizmo(GizmoType.Selected | GizmoType.NonSelected)]
+	void OnDrawGizmos ()
+	{
+		Gizmos.color = Color.green;
+        //Gizmos.DrawSphere(transform.position , 3);
+		Gizmos.DrawLine(transform.position,(transform.position-transform.up ));
```
fomalsd: FYI, there's `GizmoUtils.DrawArrow` method that draws an actual arrow

Bod9001: yeah, I was looking for that but everywhere said it wasn't inbuilt, I can add it now

fomalsd: yep, check an example from PushPull:
```cs
GizmoUtils.DrawArrow( transform.position, PulledBy.transform.position - transform.position, 0.1f );
```
last parameter is arrowHeadLength, you can keep it at 0.1

Bod9001: the Default seems to be good, 
Since that will be the exact tile that the game uses for calculations

fomalsd: try applying this
```suggestion
	GizmoUtils.DrawArrow( transform.position, (transform.position-transform.up) ); 
```

fomalsd:
should syndie shuttle finally have lights with these changes?

Bod9001:
nope, 
"if (IsWithinReach(startPos, localObjectPos, distance))"  Light switch trigger.cs
It fails on that, so  I think it's something to do with minuses  Since should syndie  is Minus from 0

PR #1613, opened 2019-03-06
Update WindowDrag.cs

### LINKED ISSUE #1609: Moving the login box outside the window will make it unobtainable
# Description
If you click and drag the login box outside of the window while in windowed mode, you can't find it anymore even in fullscreen
## Category
BUG

## Detailed Bug Report
welp
threwitaway

## Steps to Reproduce
1. start the game in windowed
2. click and drag the login box outside the window
3. try to find it

Itsnotjames:
Updates BeginDrag() to make it more efficient.
Updates OnDrag() to make it more efficient, and to clamp the dragged window within the screen.
Adds summaries to BeginDrag() and OnDrag().

### Purpose
Prevents the window from going out of the screen bounds.
Fixes #1609 
47bsh3a

### Notes:
When the resolution changes, the windows do not move with the new resolution and may end up out of bounds: 
i9lhx3n
Although, when an out of bound window gets dragged again the position is fixed.

All window's bounds are correct except for the "Alliance" select window's bounds: 
fxdgj5s
Not sure why.

### DIFF
--- a/UnityProject/Assets/Prefabs/SceneConstruction/NestedManagers/UIManager.prefab
+++ b/UnityProject/Assets/Prefabs/SceneConstruction/NestedManagers/UIManager.prefab
@@ -16079,7 +16079,7 @@ RectTransform:
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
   m_AnchorMin: {x: 0.5, y: 1}
   m_AnchorMax: {x: 0.5, y: 1}
-  m_AnchoredPosition: {x: -184, y: -122.700195}
+  m_AnchoredPosition: {x: -184, y: -260}
   m_SizeDelta: {x: 429.3, y: 53.6}
   m_Pivot: {x: 0.5, y: 0.5}
 --- !u!222 &8479790989508148053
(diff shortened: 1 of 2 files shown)


Bod9001:
no spaces Either side  Of the x For the checklist format
- [X]  This fix is tested on
" - [X]  This fix is tested on  " like so

Itsnotjames:
Fixed, thanks mate

Bod9001:
No problem

fomalsd:
While this fixes #1609, it's convenient to be able to partially hide a window.
Do you think it's possible to allow partial hiding (up to 90% of window size)?

As for resolution change, I think a safeguard of some kind should be in place. 
Possible solution: 
- An event that's invoked when resolution change is detected. Notifies all `WindowDrag`s 
- Special `CheckPosition` method inside `WindowDrag` that checks if window is out of bounds and resets their position if necessary. Method is called on resolution change event

Jibbajabbafic:
You might be able to reset the window to its startPosition when the resolution changes, although the startPosition could still be off screen depending on the anchoring

Jibbajabbafic:
If the startPosition is wrong when the resolution changes then that will need to be recalculated too

Edit: I tested it and the start position is indeed wrong

Itsnotjames:
I made some changes that:

- let the window be hidden by 90%,
u1400gZ
- put the window in the correct startPosition when the script is disabled,
ezgif-1-12665d355832
- fixed the positions of these two text elements:
XHhxybW
- and I added a rough way to keep the windows inside the screen bounds when the resolution is changed.

As there is no event for resolution changes, I used the OnRectTransformDimensionsChange() event as a workaround.
This event is called whenever the window's RectTransform's dimensions change, like when the window size is scaled to a new resolution.
Unfortunately the window's RectTransform's dimensions only scale based on the resolution's width, so OnRectTransformDimensionsChange() is not invoked when only the resolution height changes.

Here is the resolution being changed horizontally:
ezgif-1-9041f28b48cc
Here is the resolution being changed **only** vertically:
ezgif-1-6be2908aaaf0

Jibbajabbafic:
This is awesome, nice work

PR #1615, opened 2019-03-06
[DNM] Grenades

Fishfish458:
### Purpose
This is a first pass at adding grenades and a generally more "refined" explosive method. I hope to generalize it more. This originally started by using explodeOnDamage.cs as a base so it uses some of the same methods though modified a bit.
https://gfycat.com/halftensekookaburra

### Notes:
Issues that become noticeable with grenades:

self-removing items in backpacks will not clear upon  UpdateInvSlot(true, "", gameObject, invSlot.UUID);
self-removing items in closets will still exist on using disappearfromworldserver
ObjectBehaviour.parentContainer needs a server authorative setter.
location
some odd issues checking positions of explosions being slightly offset.

current todo:
generalize explosions into it's own file and have anything that explodes use methods from it.
explosions should have it's own tile reaction which should make CalcDamage obsolete and setFire should instead create those tile reactions directly.
extend explosion shake to clients
another explosion shape as rounded
create a better system for checking what objects will block an explosion


Fishfish458:
Going to close this one since I needed to do some rebasing and just changed it to a different local branch

PR #1616, opened 2019-03-09
#1606 Change swap to check for item interaction first

### LINKED ISSUE #1606: Hands always swap even if one item is interacting on another
# Description
If you have two items in either hand and use them on each other, swapItem will still run, causing the item to change places if one disappears.
## Category
BUG

## Detailed Bug Report
I noticed this when changing how guns work. UI_InteractOtherSlot runs when an object or empty hand is used on an item. Regardless of what action happens, swapitem will still run.

It could be beneficial if this was made into an optional parameter on UI_InteractOtherSlot

## Steps to Reproduce
1. get a gun and eject the magazine
2. Load the magazine into the gun and notice that the gun has swapped hands
3. This interaction occurs anytime two items are clicked together, however since both slots are full the hands do not swap.

Fishfish458:
### Purpose
This changes itemswap to always use tryItemInteract first.
UI_intereactOther is now a bool which determines if swapItem will be run.
(true means SwapItem)
Fixes #1606 

### Notes:
I am using "this" in itemSlot for swapping item. if that's an issue an issue I can correct it. I just figured I'd post everything for feedback.
The change to the UI_Manager.prefab is the removal of the button on hands which directly runs tryItemInteract since swapItem is still run on slots regardless
It seems to be working without issue in multiplayer.

### DIFF
--- a/UnityProject/Assets/Prefabs/SceneConstruction/NestedManagers/UIManager.prefab
+++ b/UnityProject/Assets/Prefabs/SceneConstruction/NestedManagers/UIManager.prefab
@@ -2072,7 +2072,6 @@ GameObject:
   - component: {fileID: 8479207382714481023}
   - component: {fileID: 8587403945929970393}
   - component: {fileID: 8587432192721568523}
-  - component: {fileID: 8587769863163809915}
   - component: {fileID: 8587712180830150811}
   m_Layer: 5
   m_Name: itemSlot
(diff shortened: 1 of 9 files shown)


DooblyNoobly:
Oh snap we have a conflict with the UIManager prefab. Any chance you can revert UIManager, rebase your branch and then make the prefab changes again?

Fishfish458:
Yes, sorry about that, I'll do that now

PR #1649, opened 2019-03-19
Prevent lateral shifting without creating undue PPRT artifacts

### LINKED ISSUE #1386: MatrixMove – ships can move with offsets B: 50
# Description
Some time ago, due to change of smooth MatrixMove lerping to the custom blocky one, ships can sometimes stop on non-integer positions. This was quickly patched by snapping to grid when ships stop moving.
Now, while shuttles always snap to grid when stopped, they restore the offset when movement is started and keep moving like that.
## Category
BUG

## Steps to Reproduce
1. Get on shuttle and start moving on lowest speed, then try stopping when ship isn't perfectly aligned to grid. (should introduce that offset)
2. Turn left or right
3. Start/stop moving to see if you see wiggling like on the gif

chairbender:
### Purpose
Fixes #1386 

### Notes:
There MIGHT be a bug in PixelPerfectRTParameter.GetFilteredRendererPosition which only shows up when movement is starting, inducing lateral motion, but I wasn't sure. So rather than fix that method I added logic in MatrixMove.clampedPosition to prevent the lateral motion, and it seems to work well without creating more artifacts, but a second set of eyes will be helpful.

### DIFF
--- a/UnityProject/Assets/Scripts/Shuttles/MatrixMove.cs
+++ b/UnityProject/Assets/Scripts/Shuttles/MatrixMove.cs
@@ -70,8 +70,6 @@ public struct MatrixState
 /// </summary>
 public class MatrixMove : ManagedNetworkBehaviour
 {
-	public bool IsMoving => isMovingClient; //?
-
 	//server-only values
 	public MatrixState State => serverState;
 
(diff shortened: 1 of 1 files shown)


DooblyNoobly:
Very nice, its much better! The stopping feels really solid now too. Thanks

PR #1650, opened 2019-03-19
Fixed wallmouts visiblility on 1x zoom

LeonardKr:
Fixed wallmouts visible when they should not be on 1x zoom.

Problem: I'm cheating a little in the lighting system to improve performance. When zoom is set to 1x I'm reducing lighting texture by half from 14 pixels per unit to 6. It helps a lot since jump form 2x to 1x requires a doubling of lighting texture size which leads to a bad performance at 1x zoom. New wallmount shader had tile to pixel ratio hardcoded, so it would not be able to catch the change. I've changed lighting system to set global property _LightingTilePixelsPerUnit to be available to all the shaders if needed, and replaced hardcoded pixel ration with it.


DooblyNoobly:
Thanks @MrMeatpie !

PR #1673, opened 2019-03-24
Players with help intent can swap places

### LINKED ISSUE #1596: Two players with help intention should trade places on push B: 50
# Description
When two players have help intent, they should be able to pass each other by swapping places when either one or both are moving toward each other.

## Detailed Bug Report
[example](https://imgur.com/0mBmOTz)
Only when both players have help intent enabled should they be allowed to pass each other, any other intent on both parties should result in pushing. moving diagonally will not trade places unless both players are moving diagonally at the same time.
## Steps to Reproduce
1.  have one player run at another player with help 
2. players should swap places with each other if they are both help and continue their movement

chairbender:
### Purpose
Fixes #1596 

### Notes:
Had to make a few design decisions about how things worked, so these are up for debate (not like you all need any prompting for that 😄 ). Here's the logic that swapping has:

- If a player with help intent begins to move into a tile occupied by another stopped player (with help intent), they will swap. This is the basic functionality described in the ticket.
- Players with help intent can pass through each other when they are both moving. This is intended to make movement smooth when 2 players move through each other even if the swap logic doesn't catch this case and explicitly initiate a swap. Also I think this just makes movement less of a hassle when players are trying to move around in the same area.
- If 2 players with help intent end up on the same tile, the swap will be performed with the player who last reached the tile initiating it (this is a failsafe for the above logic to prevent 2 players from being able to stop on the same tile and remain there)
- When a player is pulling something, they cannot swap or be swapped with. They revert back to the normal behavior of being pushable. I just don't see a way where it would make sense. If they were pulling an impassable object, for example, where would the swapee go? Or what if they were against a wall and already pulling someone behind them? 
- The exception to the above is that if a player is pulling another player and both have help intent, they can swap places and the pull will not be broken. I think this is more convenient than having to do a circle when trying to turn around and still maintain a pull.
- A player who is BEING pulled can be swapped with by another player, thus breaking the pull.

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/MatrixManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/MatrixManager.cs
@@ -122,10 +122,19 @@ public class MatrixManager : MonoBehaviour
 	/// <param name="worldOrigin">current position in world coordinates</param>
 	/// <param name="dir">direction of movement</param>
 	/// <param name="bumper">GameObject trying to bump / move, used so we don't bump against ourselves</param>
+	/// <param name="isHelpIntent">if the bumper is a player with help intent, hence help intent bump is possible</param>
 	/// <returns>the bump type which occurs at the specified point (BumpInteraction.None if it's open space)</returns>
-	public static BumpType GetBumpTypeAt(Vector3Int worldOrigin, Vector2Int dir, GameObject bumper)
+	public static BumpType GetBumpTypeAt(Vector3Int worldOrigin, Vector2Int dir, GameObject bumper, bool isHelpIntent)
 	{
 		Vector3Int targetPos = worldOrigin + dir.To3Int();
+		if (isHelpIntent)
+		{
+			//check for other players with help intent
+			if (GetHelpIntentAt(targetPos, bumper) != null)
+			{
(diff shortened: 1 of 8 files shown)


fomalsd:
getting NREs on join:
```
UIManager.set_CurrentIntent (Intent value) (at Assets/Scripts/UI/UIManager.cs:110)
ControlIntent.SetIntent (Intent intent) (at Assets/Scripts/UI/UI Bottom/ControlIntent.cs:61)
ControlIntent.Start () (at Assets/Scripts/UI/UI Bottom/ControlIntent.cs:20)
```

fomalsd (changes requested):
Really great feature! Makes pulling players much more comfortable.
Here's a few things:
 - You can't trade places with another player if you're not grounded: 
Scr 2019-03-24 в 23 35 12
 - Something's not right after you try to trade places with player who just left a locker

no_trade_after_decloset

chairbender:
- also have to make sure it works properly when swapping between matrices

chairbender:
Alright @fomalsd , ready again.

fomalsd (approved):
Amazing job, works wonderfully in all situations!

Issue #1664, opened 2019-03-24
When round restarts, UI is permanently empty B: 50


chairbender:
# Description
After the round restarts and you select a job and spawn, the UI is shown as empty and never appears to update even when picking things up and even though your sprite is shown as holding something / having clothes.

It also seems that is becomes impossible to shoot, throw, or melee. Basically becomes impossible to use any item in your hands.

Expected behavior is that the UI shows up just like it does when you first join a game and that you can use items and the UI is updated when you pickup / drop stuff.

(Currently only discovered on headless)

DooblyNoobly:
Bounty of 50 usd will be awarded to the contributor on a successful merge of a solution to this bug

DooblyNoobly:
looks like #1677 has fixed this, I just did a test on steam server (Spawned as nuke ops, blew up station and when round restarted everything was fine with the UI slots)

fomalsd:
are you sure it wasn't just because of messed up server back then?

DooblyNoobly:
well Inventory manager errors probably contributed to it on the server. As was pointed out in the note fixes `lambda expressions and null propagation operators dont get along` which seems to hold true when looking for explanations why, I found this on stack overflow: `tree lambdas (unlike delegate lambdas) are interpreted by already existing LINQ providers which don't yet support null propagating.` which makes sense. Learn something new every day.

Also I just done another test to make sure and its working, this time I called restart-round from admin portal

PR #1677, opened 2019-03-25
Fixes job slots lock after player dc

### LINKED ISSUE #1663: Job slots still occupied after round restart or player disconnecting B: 30
# Description
When users disconnect, the job slot they chose is still marked as occupied and the job list is not updated, preventing anyone else from taking on the job. The player count is also not reduced on the job screen.

Additionally, when the round is restarted, the job list doesn't clear out and all the slots that were occupied the previous round still show as occupied. It should show as completely empty at the start of a new round.

(Currently only discovered on headless)

Aranclanos:
### Purpose
Fixes the logout item dropping process runtiming, locking jobs for new logins
Fixes #1663 

- [X]  This PR has been tested in multiplayer (with 2 clients, if applicable)

### Notes:
lambda expressions and null propagation operators dont get along

### DIFF
--- a/UnityProject/Assets/Scripts/Inventory/InventoryManager.cs
+++ b/UnityProject/Assets/Scripts/Inventory/InventoryManager.cs
@@ -333,11 +333,11 @@ public class InventoryManager : MonoBehaviour
 	{
 		//drop everything
 		AllServerInventorySlots
-			.FindAll(x => x.Owner?.gameObject == owner && x.Item)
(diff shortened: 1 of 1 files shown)


Aranclanos:
Looks like it also fixes #1664

fomalsd:
nice, thanks!
don't think it fixes #1664 though, it's only been reported on headless server

Aranclanos:
on the form I was able to reproduce the bug, it definetely fixes it

DooblyNoobly:
ok we should be able to test this now

PR #1681, opened 2019-03-27
Adds canUse() in InputTrigger, for Interact() calls

### LINKED ISSUE #1652: Crowbars, Vendors and MessageOnInteract work outside player's reach B: 30
# Description
Triggering things out of reach has recently been allowed in MouseInputController (that does clientside validations). This uncovered some bugs and security flaws that were unnoticed:
- Crowbar action requests aren't distance checked on server
- Vending machines share some code that's executed on both C and S. When they become out of sync (triggered on client but not on server) they get stuck (clientside cooldowns interfere with requesting items on server)
- MessageOnInteract is a very basic component but it shouldn't work if player is out of reach
- Probably something else?

## Steps to Reproduce
Crowbar:
1. Grab a crowbar
2. Start clicking on the floor tile in your room out of reach

Vendor:
1. Find a vend machine
2. Step away several tiles from it
3. Click it and get a clientside interact message, but notice nothing was ejected
4. Come close to the same vendor and try clicking it again
5. You'll get nothing until clientside cooldown timer runs off

Aranclanos:
### Purpose
Adds canUse() to InputTrigger, to handle range and mob status checks of Interact() calls, for both client and then server. Vendors and tiles interactions will now use it, fixing their issues with infinite range
Fixes #1652

- [X]  This PR has been tested in multiplayer (with 2 clients, if applicable)

### Notes:
Eventually all items should be using canUse() for their Interact()
The logic of canUse() is similar to canUseTopic() in the byond version of ss13 to handle href links


UnityProject/Assets/Scripts/Input System/Triggers/InputTrigger.cs
```diff
@@ -106,5 +106,32 @@ public bool Trigger(Vector3 position)
...
 
+    /// <Summary>
+    /// This is called by the Interact() of objects twice, one by the client and then by the server if the first call returns true
+    /// </Summary>
+    /// <param name="originator">game object that is performing the interaction upon this gameobject</param>
+    /// <param name="hand">hand of the originator which is being used to perform the interaction</param>
+    /// <param name="position">position of the interaction</param>
+	/// <param name="allowSoftCrit">if true it allows mobs in soft crit perform the action</param>
+    /// <returns>true if the server confirms the interaction is valid. The Interact() following this will be carried on serverside</returns>
+    public bool canUse(GameObject originator, string hand, Vector3 position, bool allowSoftCrit){
```
fomalsd: I'd set a default value for softcrit, also please rename method to PascalCase, it's not java
```suggestion
    public bool CanUse(GameObject originator, string hand, Vector3 position, bool allowSoftCrit = false){
```

UnityProject/Assets/Scripts/Input System/Triggers/InputTrigger.cs
```diff
@@ -106,5 +106,32 @@ public bool Trigger(Vector3 position)
...
+        var playerScript = originator.GetComponent<PlayerScript>();
 
+        if ((playerScript.canNotInteract() && (allowSoftCrit && !playerScript.playerHealth.IsSoftCrit)))
```
fomalsd: did you test this clause in different scenarios? feel like something is wrong here, perhaps you meant something like this:
```suggestion
        if ( playerScript.canNotInteract() && (!playerScript.playerHealth.IsSoftCrit || !allowSoftCrit) )
```

UnityProject/Assets/Scripts/Objects/VendorTrigger.cs
```diff
@@ -16,41 +16,33 @@ public class VendorTrigger : InputTrigger
...
 			}
-
-			ServerVendorInteraction(originator, position, hand);
+			ServerVendorInteraction(position);
 			StartCoroutine(VendorInputCoolDown());
 		}
-
-		return true;
+	
+        return true;
```
fomalsd: please adjust your IDE to use tabs and not spaces!

fomalsd (changes requested):
I like this idea! just needs a couple of tweaks

Aranclanos:
Done

fomalsd (approved):

fomalsd:
wallmountsinteract 2019-03-30 13_40_41 simple wallmounts with MessageOnInteract (+ extinguisher cabinet) still work out of range

fomalsd (changes requested):
please fix range check for wallmounts

Aranclanos:
There are around 50 different Interact() that should be reviewed honestly

Aranclanos:
```		
if(!CanUse(originator, hand, position, false)){
	return false;
}
if(!isServer){
	return true;
}
```
At the end the check ended up like this, I'm not convinced but I needed it to avoid doubleclicking and not blocking the current Update because you clicked on something out of reach. For now I'd like to finish it here and in the future take a look again, maybe adding the check in InputTrigger instead.
Oh and adding a serverside clickspam check.

Aranclanos:
And the idea would be making Interact() a void and making the 'interacted' check a variable that would be accessed in CanUse() or something like that

fomalsd (approved):

fomalsd:
Works, very nice!

PR #1682, opened 2019-03-28
Fix player sprite facing sync issue

### LINKED ISSUE #1659: Clients see other clients permanently stuck facing forward B: 20
# Description
Clients always appear facing forward to other clients, even though they appear with the correct facing between client / server.

(Currently only discovered on headless)

## Steps to Reproduce
1. Host.
2. Join with 2 clients.
3. Change direction and observe that it always appears to one client that the other is facing forward.

Probably involves just tweaking the logic in UserControlledSprite

chairbender:
### Purpose
Fixes #1659 

Set up Android build and tested it with IL2CPP configured before my change. I could reproduce the IL2CPP compilation issue. After making the change in this commit the issue went away, so I'm assuming this won't reintroduce the IL2CPP error that happened when trying to build for iOS. I think IL2CPP just didn't like having a syncvar hook being an abstract method, but having the hook be a method in the base class and then invoking the abstract method in that hook is fine.

### DIFF
--- a/UnityProject/Assets/Scripts/Player/UserControlledSprites.cs
+++ b/UnityProject/Assets/Scripts/Player/UserControlledSprites.cs
@@ -12,7 +12,7 @@ using UnityEngine.Networking;
 /// </summary>
 public abstract class UserControlledSprites : NetworkBehaviour
 {
-	[SyncVar(hook = nameof(FaceDirectionSync))]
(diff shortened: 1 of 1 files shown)


fomalsd:
you just added 2 empty lines?

chairbender:
Oops, didn't stage the right file.

chairbender:
@fomalsd fixed

fomalsd:
ok, thanks

PR #1688, opened 2019-03-31
Shuttle Rotations #1502

### LINKED ISSUE #1502: Shuttle rotations not collision checked B: 120
# Description
When a shuttle is in a position where rotating it would merge it with a wall, it is able to rotate anyway

## Category
BUG

## Detailed Bug Report
OS: Windows 10 Pro x64

nice

A check should be added that detects when a rotation will cause a collision with static objects like walls, floors, grates, etc, and prevents the shuttle from rotating in that scenario. Causing the same check for small dynamic items may be overkill, but neglecting to protect those smaller objects from rotations may cause a situation where someone may exploit this behavior by moving a shuttle and rotating it over an item to cause the item to appear inside of the shuttle. A separate check may be necessary for that case. Maybe make items bounce off of the shuttle and get sent off into space?

## Steps to Reproduce
1. Start unitystation
2. Go to shuttle
3. Rotate shuttle nearby a wall

kaffo:
Not complete as shuttles still do not collide with floors
I could have hacked in a workaround specifically for shuttles and floors, but I think this should be an opportunity to extend the BasicTile/RegisterTile classes to support multiple types of Passable objects

* Fixes Syndie, command and cargo escape shuttle pivots
* Adds Rotation Sensors to probe a rotation's collisions before commiting
* Edits IsPassable to optionally exclude Matrixes, in this case Shuttles colliding with themselves

### Purpose
Fixes #1502 and also makes the shuttles a little more sensible to fly by setting their pivots just behind their centre, allowing them to spin away from "crashing" into a wall

### Notes:
I'm not confident I changed the pivots on the shuttles correctly due to the way the TileMaps are setup. They seem to work totally correctly, but I basically had to set arbitary offsets to get them to centre correctly.


kaffo:
This has become a bit of a monster, I don't really want to do more on it as it now fixes the really core issue of shuttles being able to rotate into stuff. There are caveats though and I'd like some input:
* I didn't enable any collisions with Base tiles, I felt it was too expensive for what it's worth, but it'd be pretty simple to enable
* I had to add the Floors layer back into the SolidLayersValues list in MetaTileMap.cs but I added an ignore for players
  * It might be laggy for shuttles, it'll need some proper testing and I understand if it gets bounced back to optimise it
* All tiles in the game currently got converted over to the new Passable dict, seems to work fine
* Objects could also get the same treatment and be pretty simple to support different collision types too
* The Syndie shuttle is freaking huge. I added 6 rotation sensors but I'm not sure it's enough. I was still able to get it stuck without trying too hard. Since they are only used for rotating, I'd debate putting in 8 or 10, but it'd need testing
* Pivots for the shuttles got "fixed" but I'm not especially sure it's done correctly. Might be worth looking into setting this properly via a script or gizmo rather than offsetting a child object

fomalsd (commented):
took a quick look, code looks fine at a glance (except maybe for IsPassableAt's gigantic signature)
needs ingame testing

kaffo:
> 
> 
> took a quick look, code looks fine at a glance (except maybe for IsPassableAt's gigantic signature)
> needs ingame testing

Yeah, I agree the massive signature made me scream internally, but I also don't think it's correctly to split out a new method. Maybe includePlayer and context can be consolidated somehow or depreicated at somepoint

kaffo:
I'm happier with this now. It seems to run alright with a client + server and network simulation 200ms 
I'll welcome any other reviews/comments

DooblyNoobly (approved):
Looks good! Thanks

PR #1705, opened 2019-04-06
Implement searchable dev spawner, cloner, and destroyer

### LINKED ISSUE #1488: Refactor ItemFactory into hassle-free universal item/prefab spawn system B: 100
## Category
TODO

## Details

Refactor ItemFactory into either universal static factory and/or implement custom constructors for scripts, so that all spawned items would just work (have all parameters correct on both server and clientside)
That will help create a convenient in-game item spawner later on (and maybe even allow drawing items with tilemap tools in runtime).

As @chairbender mentioned a while ago:
> I'm noticing an issue with constructing stuff from prefabs. Because we are using an object pool pattern to instantiate things from prefabs, sometimes we are not instantiating our objects in a "fully initialized" state. There's a risk that a developer creating a new object from a prefab will not fully initialize all the required state. For example if creating a bullet, they might fail to set some property of it like the behavior's damage amount. A static factory would avoid this issue (and still allow us to use an object pool) by ensuring that any needed state is provided when the object is grabbed from the pool. Currently we seem to generally be relying on so called "procedural instantiation", which has some downsides compared to a constructor / static factory.

It's a relatively big and delicate task that will require a lot of thorough multiplayer testing.

chairbender:
### Purpose
Fixes #1488
Implements a new dev tab containing options for searching for objects to spawn, cloning existing objects, and destroying objects.

Adds functionality / cleans up PoolManager and ClothFactory to make them simpler to use and support the functionality needed by the new dev tools.

Adds Lucene (an in-memory search index) to support the spawning functionality. That's what most of the "files changed" are.

### Notes:
Cloning and spawning of some objects is going to be buggy. There is no general-purpose way to ensure that cloned / spawned items behavior properly outside of adding logic to each of them to handle their own individual needs. However, I've added hooks for allowing components to know when they are being spawned or cloned (including the object they are being cloned from), and I've already implemented cloning functionality for Weapons and uni-items (ItemAttributes component).

Some game objects, such as machines, may not initialize themselves in a "just works" sort of way. If someone has a particular need to make these work, they should simply add logic to the components to support that use case by implementing the hooks I mentioned above.

I think it would be a waste of time to go through and try to implement cloning logic for every object in our game - it is not a "free" sort of thing. If someone has a need to support cloning of a specific object, such as if they want to use the new NetworkClone method on poolmanager to implement some sort of game logic, I think they should implement the logic at that time for the objects they want to support cloning.

In the meantime, cloned objects which have not implemented cloning behavior will simply behave as if you've spawned a new instance of them from their prefab, which can be somewhat convenient as opposed to searching for them in the spawner menu. For many objects, this is really all that's needed.

As for how these operations are supported, refer to my comments at the top of PoolManager, reproduced here:
When an object is spawned in game, each component is broadcast OnSpawnedServer on the server and then OnSpawnedClient on the client. This allows the component to set itself up after being spawned in a good "default" state that "just works". What that entails is up to the object / component.

When an object is cloned in game, each component is broadcast OnClonedServer on the server and then OnClonedClient on the client. This method also includes the game object that was cloned. This allows the component to inspect the cloned object and change its own state to match the object. It's up to each component / object to decide how to do this.

Neither of these will be called when spawning only a client-side object (via PoolClientInstantiate).


DooblyNoobly:
Very, very, very nice! Now all we need to do is lock it down too admins only which I'll do at a later time

PR #1708, opened 2019-04-07
Assign order in layer to all items and respect it in mouse hit detection

### LINKED ISSUE #1690: From lockers, you'll always pick up breathmasks first
# Description
When opening a locker, its possible to find guns on them. Even when mouseovering the gun and clicking on it, the object you'll find yourself picking up is the breathmask from the locker.

## Category
BUG

## Steps to Reproduce
1.Open an oxygen locker with a gun inside
2.Click on top of the gun to pick it up
3.Find breathmask in your hand

chairbender:
### Purpose
Fixes #1690 and a whole host of related issues.

### Notes:
Mouse interactions were only going by sorting layer and were not checking order within the sorting layer. On top of that, all items had the same order in layer, so I had to set a value for each of them so they were different. If they are the same, Unity does not guarantee their rendering order.

For uni-items I have it ordering them based on the cloth offset, which seems good enough for now.

Added a sorting layer specifically for bullets as well.

### DIFF
--- a/UnityProject/Assets/Prefabs/Items/Accessories/Resources/Pepper Shaker.prefab
+++ b/UnityProject/Assets/Prefabs/Items/Accessories/Resources/Pepper Shaker.prefab
@@ -1,21 +1,11 @@
 %YAML 1.1
 %TAG !u! tag:unity3d.com,2011:
---- !u!1001 &100100000
-Prefab:
-  m_ObjectHideFlags: 1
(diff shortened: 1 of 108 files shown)


chairbender:
going to merge as it's just a very small change to a lot of prefabs and the actual code change is small as well, and don't want to get lots of prefab merge issues

PR #1713, opened 2019-04-08
Remove SpriteRotate and PreviewObject

### LINKED ISSUE #1405: Clients having problems with lights, APCs, and wires rotations
# Description
Was bored so did tiny bit of investigation. I was told some other people haven't seen it and couldn't reproduce so this might have to be closed but oh well, still weird.

Basically, when you join as a client (when other game instance hosts) lights, wall mounts, wires, etc are randomly rotated around. Here's screenshots:

https://i.gyazo.com/6359c1924e30296a0c6d2a9cf1b52d1b.png
https://i.gyazo.com/f9178eb9b7032744661686b78c67fc64.png

## Category
BUG

## Detailed Bug Report
See above.
## Steps to Reproduce
1. Build a game, host on one of them (Doobly said it only happens on editor, but it didn't matter for me, both editor and game build seem to be broken on my side, whichever is client at least)
2. Join from other, pick assistants as both to compare.
3. You can see wires in maintance are off, and lights may or may not be off because they are rotated around.

chairbender:
### Purpose
In my ongoing effort to fix #1405 I wanted to eliminate one possibility that was messing with rotations, so I looked into SpriteRotate. 

It's on a lot of objects and looks like it possibly messes with their sprite. It also looks like it might not actually be used anymore and seems to be tied to PreviewObject, which appears to be part of some old mapping tool.

I removed SpriteRotate and PreviewObject so they would at the very least not clutter up our objects, and also to prevent confusion (anyone attempting to actually use these components would probably cause rotation logic to not work, and these don't appear to be used anymore anyway). There still might be some outdated mapping stuff in the codebase but I didn't want to mess with anything further and I haven't seen that stuff showing up on the prefabs we are actually using.

I tested making various changes using the tile palette and assigning rotations to wallmounts / lights / chairs and it all seems to be working even with this stuff removed.


chairbender:
still want someone to at least take a look at this to get a second opinion on if it was okay to remove

DooblyNoobly:
Yep you are just removing left overs from the old map editor. I just tested this PR with TilePalette and tiles still rotate there so it does not seem that TMS is using the SpriteRotate.cs stuff. Merging..

PR #1716, opened 2019-04-08
Lockers can be clicked when open

### LINKED ISSUE #1715: Can no longer close lockers by clicking them
# Description
You can open lockers, but can't close them back by clicking them – only with rightclick option.

## Category
BUG

## Steps to Reproduce
1. Find a closet
2. Open it
3. Try closing it by clicking its door or base

Not sure if chaining of ThenByDescending actually works here:
https://github.com/unitystation/unitystation/blob/5ccccb85400f9e3fd58de40001254855cb5a93e3/UnityProject/Assets/Scripts/Input%20System/MouseInputController.cs#L317

chairbender:
### Purpose
Fixes #1715 while still ensuring bullets and people pass through open lockers.

Previously this was accomplished by turning off colliders. That doesn't work because colliders are used for clicking. Now, it moves the lockers down into the items layer (since it basically functions as an item when it is open and the Items layer does not collide with projectiles).

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterCloset.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterCloset.cs
@@ -14,8 +14,6 @@ public class RegisterCloset : RegisterObject
 	/// </summary>
 	private ClosetControl closetControl;
 	private bool isClosed = true;
-	// cached colliders so they can be disabled
-	private Collider2D[] colliders;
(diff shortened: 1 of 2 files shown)


chairbender:
Even though this is small, I already messed up once with the logic related to this issue so I want to allow someone to review it before merging. Not sure if moving the lockers down into the items layer when they are open is the right approach, but it does seem to work correctly.

fomalsd:
seems to work fine now, thanks!

PR #1717, opened 2019-04-08
Misc file fixes

Jibbajabbafic:
### Purpose
Committed some files to clear up the git changelist.

It might be worth adding these files to the .gitignore:
```
UnityProject/Logs/Packages-Update.log
UnityProject/Packages/manifest.json
```

Although it would be good to have everyone using the same package version, so perhaps not.


fomalsd:
doesn't cause any trouble on 2018.3.0f2 so it's safe to merge

PR #1737, opened 2019-04-16
Client can mop, refactor FinishProgressAction to delegate to component

### LINKED ISSUE #1668: Mopping floors doesn't work for clients
## Steps to Reproduce
1. Join steam server
2. Grab a mop
3. Try mopping blood

chairbender:
### Purpose
Fixes #1668 

Refactors FinishProgressAction so that it isn't doing all of the actual logic of what happens at the end of every different kind of progress action. Instead, it invokes a callback so that the component that originally started the action can decide what to do - better separation of responsibility that way IMO.

### Notes:
I noticed an issue with tool->floor interactions where TileTrigger was "eating" interactions on the client side without doing anything - returning true from Interact even in cases where an interaction would've have actually occurred, and thus false should've been returned. I made a change to avoid that issue. I tested to see if interaction is still working fine but I'm not sure why that code was there to begin with and it seemed like it might've been a hack for something I'm unaware of. The code I changed seems to pertain to #1681 if you happen to know what the purpose of it was.

### DIFF
--- a/UnityProject/Assets/Prefabs/Items/Tools/Resources/Mop.prefab
+++ b/UnityProject/Assets/Prefabs/Items/Tools/Resources/Mop.prefab
@@ -202,15 +202,15 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   clothingReference: -1
   hierarchy: 
-  inHandReferenceLeft: 984
-  inHandReferenceRight: 996
+  inHandReferenceLeft: 312
+  inHandReferenceRight: 324
   itemDescription: A Mop
   itemName: mop
   itemType: 0
   size: 2
   spriteType: 0
   type: 0
-  ConnectedToTank: 0
+  CanConnectToTank: 0
   throwDamage: 7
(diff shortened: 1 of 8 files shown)


fomalsd:
yep, that's pretty good, thanks!

Aranclanos:
CanUse() broke and now vending machines and others that use it wont respond to clients

chairbender:
ty @Aranclanos , I'll reopen the ticket and take a look

chairbender:
@Aranclanos

I wouldn't call it broken. The only change I made was that it no longer has the side effect of sending an interaction message to the server when it is invoked on the client side. I removed this behavior because it doesn't make sense to have a method called "CanUse" which has a quite significant and undocumented side effect of performing network communication - the method was doing too much by having this side effect. CanUse should only be used to check if something is usable.

Please refactor your code to send this message in a different way.

In the meantime I will work on the other users of this method to ensure they do the same thing.

PR #1790, opened 2019-04-28
Transfering bodies while being logged out

Aranclanos:
WIP transfering bodies while being logged out

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/NetworkManagement/SpawnHandler.cs
+++ b/UnityProject/Assets/Scripts/Managers/NetworkManagement/SpawnHandler.cs
@@ -12,16 +12,11 @@ public static class SpawnHandler
 		GameObject joinedViewer = Object.Instantiate(networkManager.playerPrefab);
 		NetworkServer.AddPlayerForConnection(conn, joinedViewer, 0);
 	}
+
 	public static void SpawnDummyPlayer(JobType jobType = JobType.NULL)
 	{
-		var conn = new NetworkConnection();
 		GameObject dummyPlayer = CreatePlayer(jobType);
-		var connectedPlayer = PlayerList.Instance.Get(conn);
-		PlayerList.Instance.UpdatePlayer(conn, dummyPlayer);
-		NetworkServer.AddPlayerForConnection(conn, dummyPlayer, 0);
-		if (connectedPlayer.Script.PlayerSync != null) {
-			connectedPlayer.Script.PlayerSync.NotifyPlayers(true);
-		}
+		TransferPlayer(null, 0, dummyPlayer, null, EVENT.PlayerSpawned);
 	}
 
 	public static void RespawnPlayer(NetworkConnection conn, short playerControllerId, JobType jobType, GameObject oldBody)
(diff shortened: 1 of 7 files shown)


UnityProject/Assets/Scripts/Managers/NetworkManagement/SpawnHandler.cs
```diff
@@ -12,60 +12,69 @@ public static void SpawnViewer(NetworkConnection conn, short playerControllerId,
...
-		NetworkServer.ReplacePlayerForConnection(conn, ghost, playerControllerId);
-		TriggerEventMessage.Send(ghost, EVENT.GhostSpawned);
-		if (connectedPlayer.Script.PlayerSync != null) {
-			connectedPlayer.Script.PlayerSync.NotifyPlayers(true);
+		if (connectedPlayer == ConnectedPlayer.Invalid)
+		{
+			//conn = new NetworkConnection();
+			ConnectedPlayer loggedOffPlayer = PlayerList.Instance.UpdateLoggedOffPlayer(newBody, oldBody);
+			conn = loggedOffPlayer.Connection;
+			NetworkServer.ReplacePlayerForConnection(conn, newBody, 0); // <- bajillion warnings
```
Aranclanos: These three lines are the part where I can't wrap my head around, the purpose is to leave a clientless body. Both dummy creation and body transfer while logged out end up here. The new body was already created but if I dont call for `NetworkServer.ReplacePlayerForConnection()`, the new mob just starts runtiming to hell and back. If I DO call for this function, unity gets 1k warnings that even lag the editor for half a second.
Need help!

DooblyNoobly: What are the warnings?

Aranclanos: image
its a spam of these bad boys, all the same

DooblyNoobly: So its all coming from here: https://github.com/unitystation/unitystation/blob/2f53e62956dbfa191cb6b3be42b0bff2c2084962/UnityProject/Assets/Scripts/Messages/Server/ServerMessage.cs#L24

Interesting....

DooblyNoobly: See if anything in this thread helps: https://forum.unity.com/threads/reconnect-client-to-server.436033/

Aranclanos:
It's finished, the white bodies character preferences is another issue that it'll fix in another PR

fomalsd:
what cases does this pr address? 
like, someone getting cloned while being offline? any other scenarios?

Aranclanos:
as for now dying and respawning while being logged off, also makes the F7 husk spawning not pop 1k warnings

fomalsd (commented):
Code generally looks ok, except maybe for a few missed opportunities of reducing GetComponent<> calls (as PlayerScript already has references for RegisterPlayer, ObjectBehaviour etc)
But I found out that spawning multiple husks doesn't add more players to PlayerList (only the first one is added), that doesn't seem like a good sign

fomalsd:
ok, let's test it on steam server

PR #1792, opened 2019-04-29
Refactor ItemAttributes.cs

RedlineTriad:
### Purpose
Refactors ItemAttributes to get rid of unnecessary fields, assignments, and comment the code better<br>

### DIFF
--- a/UnityProject/Assets/Scripts/Items/ItemAttributes.cs
+++ b/UnityProject/Assets/Scripts/Items/ItemAttributes.cs
@@ -469,13 +469,11 @@ public class ItemAttributes : NetworkBehaviour
 	public void OnHoverStart()
 	{
 		// Show the parenthesis for an item's description only if the item has a description
-		string tooltip = itemName;
-
-		if (!String.IsNullOrEmpty(itemDescription)) {
-			itemName += $" ({itemDescription})";
-		}
-
-		UIManager.SetToolTip = tooltip;
+		UIManager.SetToolTip =
+			itemName +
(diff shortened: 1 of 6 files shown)


fomalsd:
Thanks! The only thing I'm worried about is whether any of our prefabs used custom values for removed fields

RedlineTriad:
Almost all of the removed fields were private

RedlineTriad:
The only public field that was removed was `type` which was set to `itemType` on construction so the only way this could be a problem was if after being created, they changed type.
If this is the case, I am pretty sure we want it to burn down as quickly as possible so the mistake is found.
Though I haven't found any errors in the editor.

fomalsd:
yes, thankfully itemType is correct in all prefabs I've found, so it's all good!

UnityProject/Assets/Scripts/Items/ItemAttributes.cs
```diff
@@ -415,72 +423,64 @@ private static ItemSize getItemSize(string s)
...
-		{
-			UIManager.SetToolTip = itemName + " (" + itemDescription + ")";
-		}
-		else
-		{
-			UIManager.SetToolTip = itemName;
+		string tooltip = itemName;
+
+		if (!String.IsNullOrEmpty(itemDescription)) {
+			itemName += $" ({itemDescription})";
```
fomalsd: Don't append item's description to its name:
image

fomalsd (changes requested):
Found a bug with `itemName`:

fomalsd (approved):

PR #1811, opened 2019-05-01
Increases backpack size to "Huge" so that it cannot be stored in the pocket slots or other backpacks

EZZE-0:
### Purpose
Set the UniBackpack prefab size to "Huge" so it doesen't fit into pocket slots or into oher backpacks<br>
Fixes #1770 

### Open Questions and Pre-Merge TODOs

- [X]  This fix is tested on the same branch it is PR'ed to.
- [X]  I correctly commented my code
- [X]  My code is indented with tabs and not spaces
- [X]  This PR does not include any unnecessary .meta, .prefab or .unity (scene) changes
- [X]  This PR does not bring up any new compile errors
- [X]  This PR has been tested in editor
- [ ]  This PR has been tested in multiplayer (with 2 clients and late joiners, if applicable)

PR #1813, opened 2019-05-02
Revert "Increases backpack size to "Huge" so that it cannot be stored in the pocket slots or other backpacks"

DooblyNoobly:
Reverts unitystation/unitystation#1811

Looks as if this might of broken storage


EZZE-0:
What exactly broke? As it seems to work fine on my end.

DooblyNoobly:
Worked for me too locally but not working on Steam. Btw are you on discord? There was a discussion about this in our engineering channel. PM me if you are

PR #1815, opened 2019-05-03
fixed vending

### LINKED ISSUE #1573: Vending machine can eject things on weird positions
# Description

## Category
BUG

## Steps to Reproduce
It's not always reproducible, results are inconsistent.
Try clicking coffee machine in the main hall (host in editor)

gif

Likely has something to do with parenting.
When fixing please also make sure that it still works on moving shuttles as well (syndie shuttle cigarette dispenser)

EZZE-0:
### Purpose
After an object from a vending machine was consumed and put into the object pool and then vended again it would pull from the pool but the parent would be set to null(default) as no parent is passed, making it so that the vended object is parented to root. This causes strange behaviour as the object would be pulled to a different tile by presumably the NetTransform as they mismatched. 

Made it so that it takes the parent of the vending machine as the parent of the vended object

Commit 2: Made it so the vending machine dispenses on top of itself instead of ejecting objects, added a boolean to allow for turning on of this functionality in cases like the machine being hacked. Also updated all the vending machine prefabs to eject downwards<br>
Fixes #1573 

### Notes:
This may possibly need to be done in either the PoolManager.cs or RegisterTile.cs scripts, but as i don't know how those scripts are intended to work this will fix the issue for now.

### DIFF
--- a/UnityProject/Assets/Prefabs/Items/Weapons/Explosive/Resources.meta
+++ b/UnityProject/Assets/Prefabs/Items/Weapons/Explosive/Resources.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: 22be88e59a8eb264c8ab3c2bffa2ef92
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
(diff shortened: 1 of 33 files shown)


fomalsd:
Thanks! works fine!

PR #1827, opened 2019-05-06
Tweaked atmos to make gas freeze in closed airlocks/shutters

### LINKED ISSUE #1689: With lag, an opening door can damage you
# Description
Walking over a turf with a door opening can damage your mob

## Category
BUG

## Steps to Reproduce
1.have a bit of lag
2.bump a door causing it to open
3.keep walking in the direction of the door, you'll be damaged

easier way to trigger it is is enabling gizmos on the editor, causing the game to lag. While this way is unrelated to ping but the issue persists

### LINKED ISSUE #1826: Airlocks squeeze out all the air when closing, causing pressure damage and strong winds
# Description
Suggested approach is to:
Initialize pressure on airlock tiles with an average of neighbor atmos passable tiles;
Freeze current gas mixture on airlock tile when it gets closed (like on /tg/)

fomalsd:
No longer get pressure damage under normal conditions
Fixes #1826 
Fixes #1689

### Purpose
Closed airlock/shutters are now initialized with air.
Pressure inside airlocks no longer drops to zero, gas mixes get frozen instead.

### DIFF
--- a/UnityProject/Assets/Scripts/Editor/Matrix/Views/MetaDataView.cs
+++ b/UnityProject/Assets/Scripts/Editor/Matrix/Views/MetaDataView.cs
@@ -26,6 +26,8 @@ public class MetaDataView : BasicView
 		localChecks.Add(new OxygenCheck());
 		localChecks.Add(new CarbonDioxideCheck());
 		localChecks.Add(new RoomNumberCheck());
+		localChecks.Add(new AirlockCheck());
+		localChecks.Add(new SlipperyCheck());
 	}
 
 	public override void DrawContent()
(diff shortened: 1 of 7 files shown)


Jibbajabbafic:
Tested it and it works pretty well.

My only concern is when the door has a vastly different pressure to the connected rooms, like so:
ezgif-2-5b69a41b0c00

One way this could be avoided is if the door inherits the highest/lowest pressure from any adjacent nodes when it's opened, although this could potentially result in excess gas creation/removal

fomalsd:
I'm inclined to think of it as a feature, it behaves the same way in /tg/
airlock could also have some burning plasma stuck inside it, I think it's fun

Jibbajabbafic:
We can always tweak it later, let's merge!

PR #1834, opened 2019-05-07
Removing VS Code plugin because it's redundant

Jibbajabbafic:
### Purpose
Unity now supports VS code natively and so this plugin is no longer necessary.

It hasn't been updated in a long time and the debugger still works without it so the only feature it really adds is auto opening VS code.

We can leave it in if people really want to keep it around though.


fomalsd:
I don't mind, any VSCode users have a different opinion?

DooblyNoobly:
alrighty! As long as vscode works as usual I'm happy

Aranclanos:
should the `UnityProject/Assets/Plugins/Editor.meta` change be there?

Aranclanos:
anyways tested locally and vscode seems to be working fine

fomalsd:
aight, merging

PR #1842, opened 2019-05-09
Chemistry refactor / fix

### LINKED ISSUE #1807: Some chemistry reactions broken
# Description
Example is this entry when you add 5u of hydrogen and 5u nitrogen you get 5u of ammonia:
```
    {
        "Name": "ammonia_from_hydrogen_nitrogen_at_0K",
        "Results": {
            "ammonia": 3
        },
        "Reagents_and_ratio": {
            "hydrogen": 3,
            "nitrogen": 1
        }
    }
```

## Category
BUG

## Steps to Reproduce
1. Get empty glass (spawn an icy soda or grab one from the bar and empty it)
2. Put in chem dispenser
3. Set to 5u and add hydrogen and nitrogen
4. Notice it creates 5u ammonia instead of 3u ammonia, 2u hydrogen and 4u nitrogen like the reaction entry would suggest

EZZE-0:
### Purpose
Started trying to fix the chemistry bug and ended up refactoring the chemistry lib 
<br>
- Simplified chemistry lib as it was pulling all the possible reactions and then recursing, which would get all the possible reactions again, which seemed kinda pointless. Simplified it to pull the next possible reaction in a loop until all the reactions are finished. Changed the current way the chemistry works by using "CompatibleChem" into simply getting a reaction that works with present chems and then deciding on a leading chem, which is going to be the chemical that is going to be used up first, and then adjusting the output and removal of chems out of the container relative to that chem.
- Adjusted inconsistent naming of chemicals and reagents to use reagents everywhere as that seems to be the dominant name
- Updated all the food prefabs as they lost their contents when the Chemicals was renamed to Reagents
- Updated the recipe for grappa that was causing a warning when doing the chem test
- Added rounding to chemistry as full length floats just look ugly in the chemistry menu and seem unnecessary/insignificant. Currently rounding to 3rd decimal point, let me know if you think this should be higher/lower.
<br>
Fixes #1807 

### Notes:
Found the problem with chemistry, forgot what it was along the way, this should work more reliably now.
<br>
Also, let me know if there are any other prefabs that use ReagentContainer script so i can update them

### DIFF
--- a/UnityProject/Assets/Scripts/Chemistry/ReactionLib.cs
+++ b/UnityProject/Assets/Scripts/Chemistry/ReactionLib.cs
@@ -186,10 +186,14 @@ public static class Calculations
 	{
 		foreach (string requiredReagent in reaction.ReagentsAndRatio.Keys)
 		{
-			if (!reagents.ContainsKey(requiredReagent) || reagents[requiredReagent] <= 0)
-			{
-				return false;
-			}
+				if (!reagents.ContainsKey(requiredReagent) || reagents[requiredReagent] <= 0)
+				{
+					return false;
+				}
+				else if (reaction.Catalysts.ContainsKey(requiredReagent) && reagents[requiredReagent] < reaction.ReagentsAndRatio[requiredReagent]) // If Catalyst, there should be atleast the ammount of catalyst specified in the recipe
+				{
+					return false;
+				}
 		}
 		return true;
 	}
--- a/UnityProject/Assets/Scripts/Chemistry/ReactionLib.cs
+++ b/UnityProject/Assets/Scripts/Chemistry/ReactionLib.cs
@@ -102,7 +102,7 @@ public static class Calculations
 			DoReaction(modifiedReagents, reaction);
 		}
 
-		RemoveEmptyReagents(modifiedReagents);
+		modifiedReagents = RemoveEmptyReagents(modifiedReagents);
 
 		return modifiedReagents;
 	}
--- a/UnityProject/Assets/Scripts/Chemistry/ReactionLib.cs
+++ b/UnityProject/Assets/Scripts/Chemistry/ReactionLib.cs
@@ -86,35 +86,22 @@ public static class Initialization
 
 public static class Calculations
 {
-	/// <summary>
-	/// Decimal point all reagent will be rounded to
-	/// </summary>
-	private const int REAGENT_SIGNIFICANT_DECIMAL_POINTS = 3;
+	private const int REAGENT_SIGNIFICANT_DECIMAL_POINTS = 3; // Decimal point all reagents will be rounded to
 
 	/// <summary>
 	/// Ok, so you're wondering how to call it <see cref="Reaction"/>
 	/// <paramref name="reagents"/> would look something like this {"Reagent":5,"Another reagent":2}
 	/// It will return The modified <paramref name="reagents"/>
 	/// </summary>
-	public static Dictionary<string, float> Reactions(Dictionary<string, float> reagents, float Temperature)
+	public static Dictionary<string, float> Reactions(Dictionary<string, float> reagents, float temperature)
 	{
-		bool reactionsFinished = false;
-
-		do
+		Dictionary<string, float> modifiedReagents = new Dictionary<string, float>(reagents); // Creates a copy that will be returned after all the reactions are done so that the original is not modified
+		Reaction reaction;
+		while ((reaction = GetValidReaction(modifiedReagents, temperature)) != null)
 		{
-			var nextReaction = GetValidReaction(reagents, Temperature); // Get next possible reaction
-			if (nextReaction != null)
-			{
-				DoReaction(reagents, nextReaction);
-			}
-			else
-			{
-				reactionsFinished = true;
-			}
-
-		} while (!reactionsFinished); // Do reactions until all have finished
-
-		return RoundReagents(reagents); // Round reagents to a significant decimal point
+			DoReaction(modifiedReagents, reaction);
+		}
+		return modifiedReagents;
 	}
 
 	/// <summary>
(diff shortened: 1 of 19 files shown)


RedlineTriad:
The rounding doesn't fix anything as shown in this [video](https://youtu.be/EftNgzYc-bY)
I think I have a way to fix it that won't require rounding.
And if the reason is that it looks ugly in the UI then I think the rounding should be there.

EZZE-0:
The rounding was just to make it look visually better, as in some cases the numbers got really long when chemicals were over-spilling.
The original bug was that you could make 5 ammonia out of 5 hydrogen and 5 nitrogen although the ration is 3:1. It basically didn't care if there was enough hydrogen for the reaction.
Also, i know this kinda contradicts what the other guy said is the expected behaviour in the bug report but isn't the intended behaviour when you mix 2 chemicals at a 3:1 ration (notice how ratio is used in the reaction definitions and not amount) it should use up the chemical that there is less of after adjusting for ratios. e.g. 5 Hydrogen + 5 Nitrogen at 3:1 ration should produce 5 Ammonia and leave 3.333 Nitrogen leftover in the reaction, as its more abundant. 
I'm not 100% sure that's how it works in the original, but i'll test it later when i get home to make sure)

RedlineTriad:
> 5 Hydrogen + 5 Nitrogen at 3:1 ration

What do you mean? That is a one to one ratio.

EZZE-0:
That it takes 3 Hydrogen and 1 Nitrogen to make 3 ammonia
EDIT: Basically that it takes 3 times more hydrogen than nitrogen in the reaction
Sorry, just noticed my mistake

EZZE-0:
Yeah, just noticed what's wrong, I'll see if i can fit this when i have some time

RedlineTriad:
The solution I think I will implement is to save reagents as a custom `struct Ratio` that uses a numerator and denominator that are saved as `int`s and then manually implement all mathematical operations.
That way we have the determinism of ints and the freedom of `floats`.

EZZE-0:
Wait, nevermind, that is the expected behaviour as the recipe is 3 Hydrogen + 1 Nitrogen = 3 Ammonia, so that is correct.
So, 5 Hydrogen + 1.666... Nitrogen = 5 Ammonia

RedlineTriad:
Yes, the problem is that the rounding sometimes magnifies inaccuracies.

EZZE-0:
The rounding is done at the very end, after all the reactions are finished. So it should't affect stuff much.

RedlineTriad:
Did you watch the video? it shows how due to the rounding 0.002 is left over, the more aggressive the rounding the bigger the leftover will be, there are two ways, either only round in the UI and live with it or stop using floats.

EZZE-0:
I see, yeah, after a bunch of reactions it accumulates a bit. The problem with rounding in UI is that there will be 0.000 ammounts of chems in the beaker.
This could potentially be fixed by simply removing the chems under 0.001 in the remove empty chems function

RedlineTriad:
Can we continue this on discord?

EZZE-0:
If merged will introduce the UI rounding improvement later, probably sunday.

Jibbajabbafic:
Please make a ticket so we don't forget about it if something happens

Bod9001:
ok, I've had a look through it, And I suggest 
Instead of using leadingQuantity 
you could calculate a multiplier of the reaction

you'll have to find the best fitting one first 

reagents[leadingReagent.Key] / leadingReagent.Value 
so how much there is in the container divided by the value in ReagentsAndRatio  
and that will give you a multiplier

so lets so 3H,1N = 3A
and you got  2.5H and 6N

You get the best fitting chemical
2.5H /  3H =  5/6

so - 5/6 * 3H = 2.5
\- 5/6 * 1N = 1.6666666
\+ 5/6 * 3A = 2.5
in code =
reagents[reagent.Key] -= multiplier * reagent .Value;
 (reagent .Value is from ReagentsAndRatio )

that would help simplify it a bit more 
Tell me what you think?

EZZE-0:
Gonna work on this a bit more to make it better

EZZE-0:
Added improvements suggested by others and fixed a few small issues found. Can now be merged.

fomalsd:
@Bod9001 can you look at this again please?

Bod9001:
ok, question about catalysts, should you have a minimum amount of catalysts To do the reaction?
or should it just be 0.01 would be enough just have to do the reaction a lot of times?
 
potential exploits could be Having the lowest amount of catalysts possible
causing 
while ((reaction = GetValidReaction(modifiedReagents, temperature)) != null)
to loop for excessively long

UnityProject/Assets/Scripts/Chemistry/ReactionLib.cs
```diff
@@ -70,167 +70,156 @@ private static void JsonImportInitialization()
...
-				ChemistryGlobals.reactionsStoreDictionary[chemical].Add(reaction);
+				ChemistryGlobals.reactionsStoreDictionary[reagent].Add(reaction);
 			}
 		}
 	}
 }
 
 public static class Calculations
 {
+	private const int REAGENT_SIGNIFICANT_DECIMAL_POINTS = 3; // Decimal point all reagents will be rounded to
```
Bod9001, line 89: this is not used anywhere?

I assume it's supposed to go with this "if (reagent.Value <= 0.001)"

UnityProject/Assets/Scripts/Chemistry/ReactionLib.cs
```diff
@@ -70,167 +70,156 @@ private static void JsonImportInitialization()
...
-			yield return chemical;
-		}
-	}
 
 	/// <summary>
-	/// Removes all chemicals from <paramref name="area"/> where the value is not more than zero
+	/// Gets a reaction that is possible with <paramref name="reagents"/>
 	/// </summary>
-	private static void RemoveEmptyChemicals(Dictionary<string, float> area)
+	private static Reaction GetValidReaction(Dictionary<string, float> reagents, float temperature)
```
Bod9001, line 164: I had a talk in the discord and due to how this is programmed, If you have a tiny amount of catalysts  it could cause a lot of looping in the 
"while ((reaction = GetValidReaction(modifiedReagents, temperature)) != null)"
so a good suggestion could be to Have a minimum amount of catalysts before the reaction triggers?

Bod9001 (changes requested):
looks good, Just need to finish up and  look over the catalyst issues

UnityProject/Assets/Scripts/Chemistry/ReactionLib.cs
EZZE-0: Should i maybe make it so that the catalyst cannot lead the reaction and it will just check if its present and get the leading reagent out of the non-catalyst ones? As it's not going to be used up, so there is no reason for it to lead. What do you think?

EZZE-0:
Disregard the merge thing, that happened after i fixed the typo online and forgot to get that commit into the original

UnityProject/Assets/Scripts/Chemistry/ReactionLib.cs
EZZE-0: I'll set it to the minimum in the recipe for now and we can change it later on

EZZE-0:
Requested changes done, should be ready to merge now.

Bod9001 (approved):

Bod9001:
tested seems to be all good. just waiting on Redline  to see if they have anything to say about it

RedlineTriad:
The thing I don't understand is that every call to `Reactions` is followed by a ` RemoveEmptyReagents` call so why not call it directly from `Reactions`?

EZZE-0:
It is called separately because the reagent transferring method in the ReagentContainer uses it to remove empty reagents after reagents are transferred into another container;

RedlineTriad:
I created a pull request on your repository with my suggested changes.

EZZE-0:
> I created a pull request on your repository with my suggested changes.

There was a little issue with your changes. I left a comment with the issue on your pull request.

EZZE-0:
Never mind, i just merged the suggestions and fixed the bug myself.

Bod9001:
good job, sorry took so long to merge ^^

PR #1847, opened 2019-05-13
Fixes conscious state not updating once a client logged in

Aranclanos:
### Purpose
Fixes conscious state not updating once a client logged in
Changes its behaviour from a ServerMessage sent to all to a SyncVar


fomalsd (changes requested):
I'd rather use messages than syncvars and just send conscious state message when syncing client after join:
https://github.com/unitystation/unitystation/blob/2709e6a68a7f965ba5dc63b9b979b5893176918f/UnityProject/Assets/Scripts/Managers/NetworkManagement/CustomNetworkManager.cs#L313

Aranclanos:
I disagree, this way achieves and does the same thing while having way less bloat

fomalsd:
with syncvars every player will always know every other player’s conscious state, and you can’t limit that using your own culling algorithms

Aranclanos:
I don't see that being a thing anytime soon

fomalsd:
Anyway conscious state should only be sent to subject player

fomalsd:
I still think it's a good idea to scrap that PR and make conscious state be broadcasted to player in question only – unless there are side effects due to other clients not knowing your conscious state?
(Upright state is synced elsewhere, in PlayerUprightMessage)

Aranclanos:
PlayerUprightMessage SendToAll or a synced var ;v)

fomalsd:
Using SendToAll for upright message is fine for now, it doesn’t elaborate on player’s health state except for whether he’s down or not

fomalsd:
Closing this one

Issue #1771, opened 2019-04-23
Crit bloodscreen stays after logging out


Aranclanos:
# Description
![image](https://user-images.githubusercontent.com/701959/56591072-7dc89600-65be-11e9-8fb1-45b736a4e416.png)
If you logout in crit, the bloodscreen/redscreen will stay on the lobby screen

## Category
BUG

## Steps to Reproduce
Please enter the steps to reproduce the bug or behaviour:
1.Get to crit state and get the bloodscreen
2.Logout
3.Notice the bloodscreen not going away

PR #1874, opened 2019-05-17
Ensure crit overlay disappears when returning to lobby, fix health monitor state after logging out.

VitrescentTortoise:
### Purpose
Fixes crit overlay not going away when logging out. Also fixes health monitor not updating from previous state after logging out. (Was broken when starting a second game). #1771 

### DIFF
--- a/UnityProject/Assets/Scripts/UI/OverlayCrits.cs
+++ b/UnityProject/Assets/Scripts/UI/OverlayCrits.cs
@@ -64,14 +64,23 @@ using UnityEngine.UI;
 			await Task.Delay(TimeSpan.FromSeconds(0.1f));
 			if (!pref.shroudActive)
 			{
-				shroudImg.enabled = false;
+				if (shroudImg != null)
+				{
+					shroudImg.enabled = false;
+				}
+
 				return;
 			}
+
 			DoAdjust();
-			shroudImg.enabled = true;
 			holeMat.SetColor("_Color", pref.shroudColor);
 			holeMat.SetFloat("_Radius", pref.holeRadius);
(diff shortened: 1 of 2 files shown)


DooblyNoobly:
This should work but for some reason there is an NRE with shroudImg on CritOverlay: 
image

This is strange because the reference is there in the inspector:

image

I've even tried reimporting the whole project to make sure it wasn't a serialization issue on my side. Can you check this PR again and see if you get the same results?

DooblyNoobly:
I've confirmed that this issue doesn't exist on latest develop branch, so its something to do with this PR

VitrescentTortoise:
Yes, I get that NRE. Very sorry for missing that, don't know what I was doing.

So it appears upon scene change that shroudImg reference goes null briefly. No code modifies the reference, it is deserialized when the OverlayCrit prefab is loaded, and remains untouched afterwards. I do not know what the mechanisms are for scene change in Unity, and have no idea why it may appear null briefly. Observing the scene hierarchy, I find no indication of the image's instance leaving the scene, or the reference being cleared.

I have applied a crappy bandaid fix checking for null upon access, in the future if this reference is accessed during scene change we will get the same error. Shall I create an issue so this doesn't slip through the cracks? I suspect it's gonna be hard to fix, but maybe could be done with some research/experiments.

DooblyNoobly:
Very strange, will look into it some more

DooblyNoobly:
Figured it out, it was the async method causing the grief. Changed it a coroutine. Closing this PR as I made a branch to add the fix and will merge that. Thanks @VitrescentTortoise

Issue #1799, opened 2019-04-30
Problems when clicking walls to weld


Aranclanos:
# Description
![image](https://user-images.githubusercontent.com/701959/56994434-76ccf500-6b75-11e9-8c2b-eb1534a27035.png)
Here I roughly painted with green the area that you CAN click and with red the area that you CAN'T click.
If you click on the green area, you'll start welding the wall. If you click on the red area, nothing will happen.

## Category
BUG

VitrescentTortoise:
Gonna look into this.

DooblyNoobly:
roger that, set to in progress

VitrescentTortoise:
Okay, so I found the source of the bug, I have a fix, but I do not know exactly what the desired behavior is so I have not made a PR.

![Capture](https://user-images.githubusercontent.com/1689336/57911180-0b617380-7855-11e9-91b4-0455806a7af5.PNG)
It is comparing player position which is (73, 65, 0) and world cell position which is actually the mouse position: (72.9, 66.3, -1) clicked on tile above. We also have the cell position that was clicked on (72, 65, 0) (rounded from mouse pos somehow using baseTilemap.WorldToCell).

I find taking the Euclidean distance between the player and the mouse position strange, even more so in 3D with that -1 z coord, but I haven't played OG ss13 in a while and I don't remember how it works. I assume clicking anywhere on an adjacent tile is the desired behavior, I do not know if diagonals should be included.

Feel free to fix, or let me know what it should do and I will make it work. I might be over complicating this, this stuff hurts my head and it's 4 am.

fomalsd:
proper distance check method should be used, in your case it would be
player.Player.Script.IsInReach()

VitrescentTortoise:
Ah, thanks. Hopefully I will get familiar with the idiomatic methods for these types of things soon. Will fix and make a PR.

fomalsd:
Fixed in #1875

PR #1875, opened 2019-05-17
Fix Deconstruct Request using incorrect reach distance test.

VitrescentTortoise:
### Purpose
Deconstruct request used an incorrect distance check to verify if tile could be deconstructed. Fixed to use PlayerScript.IsInReach.  #1799

### DIFF
--- a/UnityProject/Assets/Scripts/Construction/Managers/Deconstruction.cs
+++ b/UnityProject/Assets/Scripts/Construction/Managers/Deconstruction.cs
@@ -24,11 +24,12 @@ public class Deconstruction : MonoBehaviour
 	public void ProcessDeconstructRequest(GameObject player, GameObject matrixRoot, TileType tileType,
 		Vector3 cellPos, Vector3 worldCellPos)
 	{
-		if (Vector3.Distance(player.transform.position, worldCellPos) > 1.5f)
+		if (PlayerScript.IsInReach(player.transform.position, worldCellPos) == false)
 		{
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Construction/Managers/Deconstruction.cs
```diff
@@ -24,11 +24,12 @@ public void TryTileDeconstruct(TileChangeManager tileChangeManager, TileType til
 	public void ProcessDeconstructRequest(GameObject player, GameObject matrixRoot, TileType tileType,
 		Vector3 cellPos, Vector3 worldCellPos)
 	{
-		if (Vector3.Distance(player.transform.position, worldCellPos) > 1.5f)
+		if (PlayerScript.IsInReach(player.transform.position, worldCellPos) == false)
```
fomalsd: transform.position/localPosition aren’t good references for player or items/objects. they can represent out of date positions or not work properly across different matrices.
```suggestion
		if (player.Player().Script.IsInReach(worldCellPos, true) == false)
```
This one uses RegisterTile under the hood, so it’s fine to use

fomalsd (changes requested):
proposed a safer approach

VitrescentTortoise:
> 
> 
> proposed a safer approach

I see, did not know that. Looks like I used the wrong IsInReach instead of the one suggested. Thank you, I committed it.

fomalsd (approved):

PR #1879, opened 2019-05-19
Replaced Color collection with NativeArray in TextureDataRequest.cs

DooblyNoobly:
### Purpose
Reduces the high CPU usage of converting the NativeArray to a List. Had to import Jobs and Entities to use the updated NativeArray collections.

### Notes:
You will get errors about not disposing of the native array if you make changes to the class and recompile it. This is fine. The array is disposed of properly at runtime.

### DIFF
--- a/UnityProject/Assets/Scripts/Lighting System/TextureDataRequest.cs
+++ b/UnityProject/Assets/Scripts/Lighting System/TextureDataRequest.cs
@@ -1,22 +1,29 @@
-﻿using Unity.Collections;
+﻿using System.Collections.Generic;
+using Unity.Collections;
+using Unity.Jobs;
 using UnityEngine;
 using UnityEngine.Rendering;
-using System.Collections.Generic;
 
 public class TextureDataRequest
 {
 	private bool iRequestLoaded;
-	private List<Color32> textureBuffer = new List<Color32>();
 	private int height;
 	private int width;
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Lighting System/TextureDataRequest.cs
```diff
@@ -1,22 +1,29 @@
...
 
 public class TextureDataRequest
 {
 	private bool iRequestLoaded;
-	private List<Color32> textureBuffer = new List<Color32>();
 	private int height;
 	private int width;
+	NativeArray<Color32> colors;
+
+	public void DeallocateOnClose()
```
fomalsd, line 14: Would editor errors go away if you do it like this:
```suggestion
	public void TryDispose()
	{
	if (colors?.Length > 0)
			{
				colors.Dispose();
			}
	}
```
And also use it inside StoreData?

DooblyNoobly: Nay, this is a memory leak error so the NativeArray has been allocated.

DooblyNoobly: And looks like it only happens if you play in the editor. Stop and then recompile the script. Weird

PR #1880, opened 2019-05-19
Implement a toolchain for custom Static Code Analyzers

MrLeebo:
Capture

Purpose is to establish static code analysis integrated into developers' IDEs so that they get guidance and best-practices notes while coding. Goal is to reduce the effort it takes to perform code reviews and maintain a high code quality.

Some code analyzers will be imported from open source projects such as https://github.com/meng-hui/UnityEngineAnalyzer and those rules will be supplemented with custom rules that are specific to unitystation. Code analyzers will be implemented using the roslyn API.

- For custom analyzers to work, you must add the assembly containing your analyzers to the Project References > Analyzers list in VS 2017, or install the omnisharp extension in VS Code or other IDEs (vim, atom, sublime, etc)
- ~~Seems to require net framework 4.7.1 but maybe a specific patch or something? I still get assembly failure errors about this even after installing the SDK of 4.7.1.~~ I originally installed the .Net Compiler SDK using the Extension Store in Visual Studio 2017, but the plugin is only compatible with VS 2015. You must uninstall it and use the Manage Features portion of the Visual Studio installer instead to get a VS 2017-compatible SDK.
- Since the unity csproj files are generated, we will need some automated way to add the Analyzer package reference to the csproj after it's generated for VS 2017 users, or just encourage people to switch to VS Code or an IDE that offers an Omnisharp plugin.
- Even when the analyzers are working correctly, it can take a long time for them to initialize; you may not see any feedback for a couple of minutes.
- A standard code formatting template has been applied, if you enable the "Format on Save" setting in your IDE after installing Omnisharp, it will automatically fix your indents, brace positions, whitespace issues, and more.

### Analyzers Implemented

- [x]  YieldHelper
- [x]  TransformPosition
- [x]  SyncVar prohibited in favor of NetMessages? (UnityEngineAnalyzer gives a deprecation warning for SyncVar already)
- [ ]  `[SerializeField]` suggestion


MrLeebo:
I have the first analyzer implemented. This PR could do with a peer review and suggestions for any additional analyzers that would be useful.

RedlineTriad:
I tried it out and managed to get it to work with vs code.
It is definitely useful, even just the UnityAnalyzer is worth it IMO.
But I think I have an idea on how to change `YieldHelper` so even refactor functionality should be possible.
Also, a `[SerializeField]` attribute refactor/suggestion for public fields that aren't referenced would be nice as well.

MrLeebo:
I added formatting rules to the omnisharp.json config, which you can apply in your IDE by installing the omnisharp extension and then enabling the "Format on Save" option so the rules are applied as soon as you save the file. I used the default configuration options plus any guidelines I found in the wiki (e.g. _Tabs_ 🤮).

Regarding the other analyzer suggestions:

1) should `transform.position` always be replaced with `registerTile.WorldPosition`? Are there circumstances where `transform.position` is appropriate? Or some other coordinate?

2) `[SyncVar]` is already marked as obsolete with UnityEngineAnalyzers so that's probably good enough

3) For `[SerializeField]` you mean this? https://docs.unity3d.com/ScriptReference/SerializeField.html According to the docs public fields are already serialized. Do you mean a suggestion to refactor a public field into a private `[SerializeField]` if it is a unity behavior and the public field is never referenced outside that class?

MrLeebo:
Capture

Added a first stab at a TransformPositionAnalyzer, it's kind of naive, looking for the pattern `transform.position` or `something.transform.position` specifically, so I don't know if it should not allow `any.position` either in case the identifier is not called `transform` or if it should allow `transform.position` in certain contexts?

fomalsd:
> should transform.position always be replaced with registerTile.WorldPosition?

Not _always_, it's still used in some non-important places, like for playing sounds (`SoundManager.PlayNetworkedAtPos`) because using WorldPos is more expensive. It's also being set during lerping. But keeping a warning is fine

RedlineTriad:
> 3\. For `[SerializeField]` you mean this? https://docs.unity3d.com/ScriptReference/SerializeField.html According to the docs public fields are already serialized. Do you mean a suggestion to refactor a public field into a private `[SerializeField]` if it is a unity behavior and the public field is never referenced outside that class?

Yes, the reason it would be useful is that it would make inspector fields more explicit. I had multiple cases where i made a mistake because i didn't know the field was set by the inspector instead of by another class. Also using `public` to serialize a field instead of an attribute is something i, personally, would consider accessor misuse/abuse.

MrLeebo:
@RedlineTriad Do you suppose we should split the `SerializeFieldAnalyzer` into another PR and let this one through? I'm struggling to get that one working because, it turns out counting the references of a field is not as straightforward as it sounds. The `SymbolFinder` helper just doesn't give me any references back even with simple code files with obvious references.

Even without the custom analyzer, you can click on the "X references" label above the field declaration to see a preview window of the field's references, so you can quickly determine if you should refactor a public field by checking to see if there is more than one *.cs file shown in the preview window.

Screen Shot 2019-06-02 at 11 41 07 PM

RedlineTriad:
Yes, i think that it would be worth it's own PR.The other great analyzers should not get delayed because of this one.

fomalsd:
Could've gone one step further and added support for Rider. 
It's used by quite a few people around here and our org has keys for open source development provided by @JetBrains team.
img
Just add this script
https://gist.github.com/42dbc1d8d824b45918e47c05e95d3b24
For more details see gist description

MrLeebo:
@fomalsd Where do you save the script? In the UnityProject/ folder?

fomalsd:
Guy from JetBrains advised putting it in Scripts/Editor folder

fomalsd (commented):
Yeah, it works now
The only thing is that `YieldHelper` is no more since #1894 – replaced by `WaitFor`

MrLeebo:
I renamed YieldHelper to WaitFor and added some markdown files to document the analyzers if in the future we ever want to generate an online styleguide/documentation site.

fomalsd (approved):

PR #1881, opened 2019-05-19
Fix flying objects like glass shards getting stuck on player

### LINKED ISSUE #1865: Shards ejected from window get stuck on you and kill you
# Description
When breaking a window from space the shards continuously pummel you until you drop dead like so:
blqpgULCbO

They should hit you once and continue flying past you.

## Category
BUG

## Steps to Reproduce
1. Go into space
2. Bash window until it smashes
3. Die

VitrescentTortoise:
### Purpose
Floating objects over a speed threshold will now damage what is in their path, and continue floating. They are no longer stopped by the player, and do not spam damage until player is dead. This
fixes #1865.

There is room for more complexity in this system (perhaps changing momentum of flying objects), or 
handling heavy objects differently (knock down player?). This is not related to the bug.

### Notes:
Changes that may have incurred side effects:
I removed the prerequisite for OnHit requiring object to have been thrown to do damage. Fast flying objects that were moved with space wind or other pushing forces should not need to be thrown to do damage. This should allow for emergent damage behavior instead of having to manually fake the throw data. This is only usage of OnHit, so shouldn't affect anything else at the moment.

I removed that !IsPlayerNearby check because I do not believe it makes sense. It seems this code intended to allow an object to drift into a solid object if no players are around. No idea what this is about.

### DIFF
--- a/UnityProject/Assets/Scripts/Transform/CustomNetTransform.Move.cs
+++ b/UnityProject/Assets/Scripts/Transform/CustomNetTransform.Move.cs
@@ -424,24 +424,16 @@ public partial class CustomNetTransform {
 		Vector3Int intGoal = Vector3Int.RoundToInt( goal );
 		var info = serverState.ActiveThrow;
 		List<LivingHealthBehaviour> hitDamageables;
-		if ( CanDriftTo( intOrigin, intGoal, isServer: true ) & !HittingSomething( intGoal, info.ThrownBy, out hitDamageables ) )
+
+        if(serverState.Speed > SpeedHitThreshold && HittingSomething( intGoal, info.ThrownBy, out hitDamageables ))
 		{
-			//if object is solid, check if player is nearby to make it stop
-			return registerTile && registerTile.IsPassable(true) || !IsPlayerNearby(serverState);
+			OnHit( intGoal, info, hitDamageables, MatrixManager.GetDamagetableTilemapsAt( intGoal ) );
 		}
 
-		if ( serverState.Speed > SpeedHitThreshold ) {
-			serverState.ActiveThrow = new ThrowInfo {
-				Aim = BodyPartType.Chest.Randomize(0),
-				OriginPos = origin,
-				TargetPos = goal,
-				SpinMode = SpinMode.None
-			};
-			info = serverState.ActiveThrow;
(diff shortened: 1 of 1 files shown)


fomalsd:
Loving it! 
GIF
Yeah, I can't remember why `IsPlayerNearby` was there. 
Very smooth experience so far, we'll tweak it later if something comes up

fomalsd:
Oh, looks like I found something:
GIF
So items actually no longer stop when flying through player, at all. Need to figure out a way to differentiate stop/no stop situations

fomalsd (changes requested):
Flying items still need to stop when hitting player in some scenarios (like when thrown by other player), perhaps when thrownBy isn't null? 
Or maybe adding a special passThrough flag to ThrowInfo that would be true for push-induced collisions...

fomalsd:
any updates?
IMO two most viable options so far are:
1) single algo – if thrown from afar: stop on tile adjacent to player and hurt him
When thrown from adjacent tile: hurt and go through
(downside is double damage)
2) adding a special `passThrough` flag to ThrowInfo that would be true for wind push-induced collisions, and process things differently depending on that flag

VitrescentTortoise:
Moving into a new apartment atm, so might be a few days before I have time for next iteration.

fomalsd:
It would be really nice if you figured it out soon, as items actively punching player sound very annoying

VitrescentTortoise:
I apologize for starting this and not finishing this, a lot of my free time has recently evaporated. Anyone please feel free to pull those changes and pick this up. I'd tell you I'll wrap it up soon but I can't honestly promise that.

DooblyNoobly:
@fomalsd where does the wind push stuff actually happen where we could set a passThrough flag on ThrowInfo? Right now I can't see where wind push uses any of the throwinfo stuff.

fomalsd:
Wind push is done here:
https://github.com/Aranclanos/unitystation/blob/442b440e0238cad6f399e51fe431aba9eb4fc000/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/ReactionManager.cs#L53-L73
initial idea was that sliding (not flying) objects don't harm player, but they get in the air again if speed is high enough

PR #1883, opened 2019-05-20
Make ignore more restrictive

lfricken:
### Purpose
Prevent more binary files from getting committed and stop blowing up GitHub after building.

### Notes:
None

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,24 @@ ExportedObj/
 *.booproj
 *.svd
 
+# Binary windows
+*.dll
+*.lib
+*.exe
+*.bin
+*.zip
+
+# Binary linux
+*.so
+*.o
+*.a
+*.dylib
+
+# Generated Files and Folders
+[Bb]in/
+[Dd]ebug/
+[Rr]elease/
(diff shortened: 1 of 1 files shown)


fomalsd (commented):
hm, not sure about zip... we might need that eventually
also I've added binary stuff before, namely SunVox for all supported platforms, and who knows if we want to add something else

lfricken:
A binary type can always be permitted later just as easily. But IMO not until you need it.

.gitignore
```diff
@@ -43,6 +43,24 @@ ExportedObj/
 *.booproj
 *.svd
 
+# Binary windows
+*.dll
```
DooblyNoobly, line 47: Not really keen on gitignoring dll libs as there is a possibility of using appropriately licensed libs of that type in the assets. Everything else is fine

DooblyNoobly (changes requested):

fomalsd:
any updates?

fomalsd:
Closed due to inactivity, create a new PR if you want to try again

PR #1888, opened 2019-05-23
Canister behaviour with pipes

Aranclanos:
### Purpose
Adds a sprite overlay for securing a canister to a connector
Blocks canisters from being pushed or dragged when connected.
Pipes will now change its sorting layer when secured or unsecured to the floor.
image

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Pipenets/Pipe.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Pipenets/Pipe.cs
@@ -33,6 +33,7 @@ public class Pipe : MonoBehaviour
 		if (anchored)
 		{
 			anchored = false;
+			spriteRenderer.sortingLayerID = SortingLayer.NameToID("Items");
 			Detach();
 		}
 		else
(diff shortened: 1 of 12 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Pipenets/Pipe.cs
```diff
@@ -33,6 +33,7 @@ public void WrenchAct()
 		if (anchored)
 		{
 			anchored = false;
+			spriteRenderer.sortingLayerID = SortingLayer.NameToID("Items");
```
fomalsd, line 36: you could cache these and make static, but it’s a minor optimisation

fomalsd (approved):
Nice!

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Pipenets/Pipe.cs
Aranclanos: yeah I kinda want to copy all layers from tg too

Aranclanos: actually just an idea because I dunno how they are defined

PR #1894, opened 2019-05-24
Use cached YieldInstructions

RedlineTriad:
### Purpose
Make it easier to use cached `YieldInstruction` by only having hardcoded values where they are needed and elsewhere create them as necessary at runtime.
This should also allow for refactoring support to be added to #1880

Also replaced all places where `new YieldInstruction`  was used with their cached equivalent.


UnityProject/Assets/Scripts/WaitFor.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System.Collections.Generic;
+using UnityEngine;
+
+public static class WaitFor
+{
+	public static readonly WaitForEndOfFrame EndOfFrame = new WaitForEndOfFrame();
+	public static readonly WaitForFixedUpdate FixedUpdate = new WaitForFixedUpdate();
+	
+	private static Dictionary<float, WaitForSeconds> cachedWaitForSeconds = new Dictionary<float, WaitForSeconds>();
```
MrLeebo, line 9: Do you think these dictionary needs to implement a equality comparer to reliably use a float for the key? Due to floating point (im)precision, it may not memoize on computed values, e.g.:

```cs
(0.1f + 0.2f) == 0.3f // false
```

Could also consider having an `int` key and multiplying the float by 1000 (or whatever fraction of a second is the shortest reasonable wait).

RedlineTriad: I don't think there should be a reason floating point accuracy should matter, it should return exactly the same `YieldInstruction` as if the constructor was used.

If they fail equality comparison, even if just do to float wonkiness then they should return a different `YieldInstruction`. If the floats fail equality, so too, should the `YieldInstruction` returned.
This is just my opinion of course. 

But I think rounding (which your suggestion essentially is) would just introduce unnecessary complexity with no benefit that I can currently see.

Of course, if I am misunderstanding something then please correct me.

MrLeebo: I think you understand it. I was concerned if there end up being a lot of waits that are calculated, you might see a "memory leak" where the cache is holding onto references to a bunch of yield instructions that are more-or-less identical to each other and rarely/never reused.

fomalsd:
ok, I don't mind this being in (direct field access is still faster than dictionary lookup though)
please pull from latest develop and resolve conflicts (also watch out as new code might still contain YieldHelper), then it will be good to merge

PR #1924, opened 2019-06-05
Canister wrenching sound

Aranclanos:
### Purpose
Canisters will play the wrenching sound when connecting and disconnecting them to connectors

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Canister.cs
+++ b/UnityProject/Assets/Scripts/Objects/Canister.cs
@@ -36,7 +36,7 @@ public class Canister : InputTrigger
 		{
 			if(isConnected)
 			{
-				SoundManager.PlayAtPosition("Wrench", registerTile.WorldPositionServer);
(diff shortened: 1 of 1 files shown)


chairbender:
Tested in multi. Client never hears the sound, even when performing the action themselves, since the logic which plays the sound only happens server side. The old interaction system strikes again!

Is the sound intended to only be heard locally, or should it be heard by all players? If it should be heard by others, then using PlayNetworkedAtPos should solve this issue easily.

chairbender (changes requested):
Tested in multi. Client never hears the sound, even when performing the action themselves, since the logic which plays the sound only happens server side. The old interaction system strikes again!

Is the sound intended to only be heard locally, or should it be heard by all players? If it should be heard by others, then using PlayNetworkedAtPos should solve this issue easily.

Aranclanos:
good catch

Aranclanos:
Fixed

PR #1926, opened 2019-06-07
Wind will now check if the object is not pushable

Aranclanos:
### Purpose
Wind will now check if the object is not pushable, no more anchored pipes flying off

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/ReactionManager.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/ReactionManager.cs
@@ -50,7 +50,7 @@ public class ReactionManager : MonoBehaviour
 						foreach ( var pushable in matrix.Get<PushPull>( windyNode.Position, true ) )
 						{
 							float correctedForce = windyNode.WindForce / ( int ) pushable.Pushable.Size;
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/ReactionManager.cs
```diff
@@ -50,7 +50,7 @@ private void Update()
 						foreach ( var pushable in matrix.Get<PushPull>( windyNode.Position, true ) )
 						{
 							float correctedForce = windyNode.WindForce / ( int ) pushable.Pushable.Size;
-							if ( correctedForce >= AtmosConstants.MinPushForce )
+							if (!pushable.isNotPushable && correctedForce >= AtmosConstants.MinPushForce )
```
fomalsd, line 53: do you propose to add this check everywhere where push can happen? bad practice
already told you that this check is integrated in tile-snapped Push, but only missing in Nudge that's used for sub-tile pushing

so all you have to do is add this check to CNT.Nudge():
```cs
    public void Nudge( NudgeInfo info ) {
        if ( PushPull.isNotPushable )
        {
            return;
        }
```

fomalsd (changes requested):
see comment

DooblyNoobly:
All fixed on develop

PR #1943, opened 2019-06-13
Vendor gui

vlandemart:
### Purpose
GUI for all vending machines. 

### Notes:
GUI frame color, items and their amount is set in VendorTrigger.
Also, includes VendingRestock.prefab which can restock items in machines.
[Picture of GUI](
https://cdn.discordapp.com/attachments/295186861377323009/588760102673317895/unknown.png)


fomalsd (changes requested):
- VendingRestock isn't in Resources folder therefore can't be spawned in runtime, also will have problems in multiplayer
- Serverplayer doesn't get any feedback messages from vendors (wrong kind of "send to chat" message perhaps?)
- Empty glass, empty plate and UniCloth are displayed as white squares for some reason (but it's probably a NetPrefabImage issue)

other than that it's really cool!
is this intended btw?
img

vlandemart:
I removed feedback messages for now, as I was having troubles with getting addressee of message - I was storing originator on `Interact` with vending machine, but it led to problems when two players interacted with one vendor.

Meat dispensers are intended (for now at least), these had VendorTrigger before.

fomalsd:
Pretty sure you can broadcast vending messages to all current observers. That will be quite sufficient for now since no credits are involved yet.
Still, there should be a way to get click originator without any caching, but I'll look into that later

vlandemart:
Addressed these issues.

fomalsd:
Won't be able to test until Monday, maybe someone else could do it until then. But code looks fine!

fomalsd (approved):

PR #1956, opened 2019-06-18
Fixed #1922

### LINKED ISSUE #1922: Igniting plasma makes pressure grow in a geometrical progression until it is NaN
# Description
You don't want to open a room that has a pressure of numbers that are wider than tile
GIF

## Steps to Reproduce
1. Start as Miner (for pressure protection), go to Atmospherics
2. Open plasma tank for a second and close it
3. Throw a lit welder (find one in yellow closet) and quickly leave the room
4. Turn on Matrix Gizmos. Observe plasma moles reappearing (while they should burn out) and pressure kpa eventually reaching MaxValue
GIF

krille90:
Gas wasn't removed correct by taking out a volume from a tile.

### Purpose
In #1922, it was shown that plasma wasn't removed from a correctly. Leading it to always exist, i.e. a fire wouldn't never go out and lead to huge temperatures.


Aranclanos:
what was the cause of bug?

PR #1959, opened 2019-06-21
Player sprites syncing rework and jobtype var swap to minds

Aranclanos:
### Purpose
Swaps jobtype from playerscript to minds.
Player job loadouts will now only be set on respawning a player and will not using ClonePlayer().
Removes the unused ItemChanged class.
Players will no longer appear naked for half a second before their job equipment spawns.
Removes a dummy case in PlayerHealth OnStartClient().
Removes the SyncListInt snycEquipSprites from Equipment.
Renames the enum Epos to EquipSlot.
ClothingItem will now handle the syncing of their sprites, the variables color and reference will now be syncvars.
CharacterSettings will no longer be synced.

### Notes:
the syncvars hooks arent being called on clients, DNM


UnityProject/Assets/Scripts/Items/ClothingItem.cs
```diff
@@ -8,13 +9,14 @@ public enum SpriteHandType
...
-public class ClothingItem : MonoBehaviour
+public class ClothingItem : NetworkBehaviour
 {
 	/// <summary>
 	/// Absolute orientation
 	/// </summary>
 	private Orientation currentDirection = Orientation.Down;
-	public int reference = -1;
+	[SyncVar(hook = nameof(SyncSprite))] public int reference = -1;
+	[SyncVar(hook = nameof(SyncColor))] public Color color;
```
fomalsd: try
```suggestion
	[SyncVar(hook = nameof(SyncColor))] private string hexColor;
	public Color color;
```
and then use `DebugTools.ColorToHex( color )` and `DebugTools.HexToColor( hexColor )`

UnityProject/Assets/Scripts/Player/PlayerSprites.cs
```diff
@@ -32,9 +29,6 @@ protected void Awake()
 		{
 			clothes[c.name] = c;
 		}
-		playerScript = GetComponent<PlayerScript>();
-		playerScript.OnCharacterSettingsChange.AddListener(OnCharacterSettingsChange);
-		OnCharacterSettingsChange(playerScript.CharacterSettings);
```
chairbender, line 37: I don't like this removal of the UnityEvent. This event hook was here for inversion of control, allowing any component to listen for character setting changes. Now this is forcing the SpawnHandler to know about and call each component that might want to know about settings changes rather than letting each component subscribe to the event. You are also exposing OnCharacterSettingsChange as a public method now, allowing anything to invoke it. This change removes the inversion of control and more tightly couples PlayerSprites with SpawnHandler.

chairbender: Though on second thought maybe it needed to be this way in order to get around issues with the order in which the Awake methods are called and ensure it happens without delay?

chairbender: Yeah nvm, untiy events makes it needlessly complicated.

Aranclanos:
Scrapped syncvars, there's a problem with network identity on childs inside of prefabs. Swapped to server messages.
After adding the login syncing for clients it should be finished

Aranclanos:
Ready to merge

fomalsd:
oopsie, inventory's gone for some reason
oops

Aranclanos:
seems to happen only if you login through the lobby scene

fomalsd:
well yeah, that's the intended way of running game

fomalsd (changes requested):
Empty inventory slots issue should be addressed first

Aranclanos:
Fixed

fomalsd:
some items don't initialize in time and show wrong item sprites.
reproducible when hosting in build and joining with editor
oops2

Aranclanos:
I can't reproduce or try to find sense on that, the base sprite of the unicloth prefab is null and the UpdateSlotMessage waits until the client's object that wants to be equipped is ready

Aranclanos:
like, why red hats?

fomalsd:
I think I remember why, as it happened before – long time ago:
if hierarchy string is empty, a random cloth is generated
https://github.com/unitystation/unitystation/blob/ac4c98768ffd4bfe7ea47781cae812c1f75c8c9d/UnityProject/Assets/Scripts/Items/ItemAttributes.cs#L378-L384
and because these two items did it at the same time, the random number was identical so they ended up having the same random item – red hat in this case.
Someone tried disabling this `RandomizeClothHierIfEmpty` behaviour before, but it didn't do any good because otherwise UniClothes kept that default holy robe icon

fomalsd:
Merging this despite it re-introducing side effect of sometimes wrong icons for UniClothes given to player on spawn – on slower machines

PR #1965, opened 2019-06-24
Updates screen_alert.png and deletes its two duplicates

Aranclanos:
### Purpose
Updates screen_alert.png and deletes its two duplicates

### Notes:
Updated it because of the very important fix of https://github.com/tgstation/tgstation/pull/44348


fomalsd:
did you make sure that all references for old sheet are still correct?
I think @chairbender added a new v2 sheet to avoid messing with references as there can be quite a lot of them
Not sure how reliable "find references in scene" option is, but it shows a lot of objects found in Lobby
img

Aranclanos:
Yeah all should be gucci

the 'find references in scene' works, but if you get a lot of objects that don't have a direct reference. Unpacking the prefabs of the scene will filter out those, they just have a child with a reference

fomalsd (approved):
ok, looks fine, didn't spot any missing references

PR #1974, opened 2019-06-27
Fix chat window not closing

Farseer2:
### Purpose
The chat window would never close when I pressed enter or submitted a message, because the event that opens it because enter was pressed occurred later. I didn't find an issue about it, though.
Also, this change makes pressing enter without a message close the chat.

### DIFF
--- a/UnityProject/Assets/Scripts/UI/ControlChat.cs
+++ b/UnityProject/Assets/Scripts/UI/ControlChat.cs
@@ -129,14 +129,11 @@ public class ControlChat : MonoBehaviour
 
 		if (UIManager.IsInputFocus && KeyboardInputManager.IsEnterPressed())
 		{
-			if (KeyboardInputManager.IsEnterPressed())
+			if (!string.IsNullOrEmpty(InputFieldChat.text.Trim()))
 			{
-				if (!string.IsNullOrEmpty(InputFieldChat.text.Trim()))
-				{
-					PlayerSendChat();
-				}
-				CloseChatWindow();
+				PlayerSendChat();
 			}
+			CloseChatWindow();
 		}
 
 		if (!chatInputWindow.activeInHierarchy) return;
--- a/UnityProject/Assets/Scripts/UI/ControlChat.cs
+++ b/UnityProject/Assets/Scripts/UI/ControlChat.cs
@@ -127,7 +127,7 @@ public class ControlChat : MonoBehaviour
 			RefreshChannelPanel();
 		}
 
-		if (UIManager.IsInputFocus)
+		if (UIManager.IsInputFocus && KeyboardInputManager.IsEnterPressed())
 		{
 			if (KeyboardInputManager.IsEnterPressed())
 			{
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/UI/ControlChat.cs
```diff
@@ -73,6 +73,7 @@ public class ControlChat : MonoBehaviour
 	/// Are the channel toggles on show?
 	/// </summary>
 	private bool showChannels = false;
+	private bool debounceCloseChat = false;
```
Jibbajabbafic: This debounce logic is unnecessary. Behaviour seemed identical with and without it.

Jibbajabbafic: This debounce var and the logic is unnecessary, the behaviour seemed the same with and without it

UnityProject/Assets/Scripts/UI/ControlChat.cs
```diff
@@ -126,10 +129,12 @@ private void Update()
 
 		if (UIManager.IsInputFocus)
```
Jibbajabbafic: You can reduce the if statement nesting a little by combining these
```suggestion
		if (UIManager.IsInputFocus && KeyboardInputManager.IsEnterPressed())
```

Jibbajabbafic (changes requested):
Some minor things you could improve in the code.

Can confirm it fixes the issue where pressing enter didn't close the chat if it wasn't in focus 👍

UnityProject/Assets/Scripts/UI/ControlChat.cs
Farseer2: Are you sure? On my machine the chat never closes when sending a message without it... Maybe its the script load order? Or rebound keys?

Jibbajabbafic: What version of unity are you using, and what OS? Mine closes just fine without it. 
@fomalsd did you experience something like this when you tested before?

Farseer2: Windows 10, 2018.3.14f - The README states 2018.3

fomalsd: current behaviour on develop is like that:
pressing enter with empty message never closes chat (regardless of focused state), 
pressing esc always closes chat and shows esc menu
pressing enter with non-empty message always sends it (regardless of focused state)

Jibbajabbafic:
@Farseer2 try it now, I removed the debounce logic from the PR

Jibbajabbafic:
Current behaviour with this PR:
- Press Enter with chat window open and no message: 
  - Close chat window
- Press Enter with chat window open and message typed out:
  - Send message
- Press Esc with chat window open:
  - Close chat window
  - Keeps message in input box if one is typed out

Farseer2:
All of these work - but the chat doesn't close after sending a message (Which was the problem).
If it shouldn't, line 138 should be removed

Jibbajabbafic (approved):

Jibbajabbafic:
I'm happy to merge this PR as is, but I can't reproduce the issue you're having:
NnQ0BjtVTK

If you're still experiencing a problem you should make an issue for it and we'll look into it more

Jibbajabbafic:
What's your name on discord btw?

Farseer2:
Maybe I'm using a different unity version? Something that changes script execution order?

PR #1978, opened 2019-06-29
Adds RequireComponents to ItemAttributes.cs

Jibbajabbafic:
### Purpose
Added these:
```cs
[RequireComponent(typeof(NetworkIdentity))]
[RequireComponent(typeof(Pickupable))]
[RequireComponent(typeof(ObjectBehaviour))]
[RequireComponent(typeof(RegisterItem))]
[RequireComponent(typeof(CustomNetTransform))]
[RequireComponent(typeof(SpriteMatrixRotation))]
```

Updated some prefabs that were missing RegisterItem and made sure to remove RegisterObject from them too:

### DIFF
--- a/UnityProject/Assets/Prefabs/Items/Other/Resources/Emergency Oxygen Tank.prefab
+++ b/UnityProject/Assets/Prefabs/Items/Other/Resources/Emergency Oxygen Tank.prefab
@@ -90,13 +90,13 @@ GameObject:
   - component: {fileID: 4370604793044084}
   - component: {fileID: 61625481682341612}
   - component: {fileID: 114277704357941144}
-  - component: {fileID: 114651414057651298}
   - component: {fileID: 114342755618585978}
   - component: {fileID: 114288380597099474}
   - component: {fileID: 114267664904661214}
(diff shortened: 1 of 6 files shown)


fomalsd:
Why should SpriteMatrixRotation be required?

fomalsd:
Also I think Network Identity is implicitly required, because all NetworkBehaviours do

Jibbajabbafic:
SpriteMatrixRotation is definitely needed on items or else this happens:
d8o1ownJoe

You're right about NetworkIdentify though so I've removed that

Aranclanos:
the networkidentity is autoadded, but it can be removed by mistake, maybe? I think?

Jibbajabbafic:
I tested this and it can't be removed since a lot of components depend on it

fomalsd:
I feel like SpriteMatrixRotation should still be optional (thinking about pipes and bloodsplats)

Jibbajabbafic:
Will those be items though? If pipes aren't anchored we'll want them to rotate still

Aranclanos:
I like it

PR #1990, opened 2019-07-02
Added tooltip component

Farseer2:
### Purpose
This PR adds a tooltip component, which you can add to GUI elements so hovering over them for a second will display information.
Added this since as a new player, I have no idea what half the GUI buttons do :)
image (print screen removes the mouse, but its there)

First PR that actually involves code, so sorry if there are any mistakes or bad conventions

- [ ]  This PR has been tested in multiplayer (with 2 clients and late joiners, if applicable

### Notes
This PR doesn't add the tooltip to any elements, since I don't really know what anything does, its just the minimal component and prefab.


UnityProject/Assets/Scripts/UI/Tooltip.cs
```diff
@@ -0,0 +1,56 @@
...
+    const float TOOLTIP_INTERVAL = 1.0f;
+    private float enterTime = 0;
+    private GameObject tooltipObject = null;
+    [SerializeField]
+    private string tooltipText = "";
+
+    void Start()
+    {
+        if (tooltipResource == null) {
+            tooltipResource = Resources.Load<GameObject>("UI/GUI/Tooltip");
```
Aranclanos: dont load with strings, just use the references in the editor

UnityProject/Assets/Scripts/UI/Tooltip.cs
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    void Update()
+    {
+        if (tooltipObject) {
+            // Move tooltip to mouse
+            tooltipObject.transform.position = Input.mousePosition - new Vector3(0, 20, 0);
+        } else if (enterTime != 0 && Time.realtimeSinceStartup - enterTime > TOOLTIP_INTERVAL) { 
+            // Create tooltip if needed
+            tooltipObject = Instantiate(tooltipResource, Vector3.zero, Quaternion.identity);
```
Aranclanos: you dont have to constantly create and destroy this object, just reuse it

Aranclanos:
I really like them, ask me in discord how to do it if you're confused

Jibbajabbafic:
See the channel toggle script if you want some more [inspiration](https://github.com/unitystation/unitystation/blob/develop/UnityProject/Assets/Scripts/UI/UIToggleChannel.cs), along with the ChannelToggleTemplate prefab in the editor

UnityProject/Assets/Scripts/UI/Tooltip.cs
Farseer2: But its always the same one, should you always drag the same component there?

DooblyNoobly: Yep make toolTipResource public or make it private and put the [Serializable] tag above it then you can drag and drop the Tooltip prefab into the Tooltip component on the inspector. Then Apply changes to the UI prefab or whatever prefab it is nested in and commit that.

UnityProject/Assets/Scripts/UI/Tooltip.cs
DooblyNoobly: yep when you are done with it just set the toolTipObject.SetActive(false) and turn it back on when you need it

Farseer2:
Sorry it took a while, can't work on it most days. Anyway, I think it looks worse this way, so maybe I misunderstood something...

DooblyNoobly:
Yep works as intended. I think next you will need to add the tooltip to all of the GUI elements you said you had trouble with and apply the specific prefabs for those. So for the bottom hud you just add it to all of the slots and apply and commit the Panel_Hud_Bottom prefab as shown in the screenshot:
image

One other thing you could do to improve this in the future is to move the update loop to UpdateManager as after awhile there could be many many tooltip components in the scene which will load the cpu as MonoBehaviour.Update does a lot of checks each frame before actually calling Update() where as our UpdateManager does not.

PR #2004, opened 2019-07-05
DNA scanner drag and drop

Aranclanos:
### Purpose
Adds the layer Items to the MouseDraggable layer of the player prefab.
Sets the DNA scanner IsClosed on by default.
Adds drag and drop interaction for the DNA scanner, now players can place other mobs inside it this way.
Adds all the validations that the DNAscanner should check for its drag and drop.
Adds an IsAdjacent() to Validations.cs


chairbender:
Looks good, thanks!

PR #2006, opened 2019-07-07
Fire Damage and Burning

### LINKED ISSUE #1453: Plasma fire should apply damage to objects and tiles B: 50
Currently, plasma fire is only for show and doesn't affect objects, player or tiles.

## Criteria

- [ ] Implement a way to apply damage to objects, players or tiles (use already existing ways if possible)
- [ ] Add Health to all things
- [ ] Destroy things when health is 0, e.g. explode plasma containers

chairbender:
### Purpose
Fixes #1453 

In action:
https://cdn.discordapp.com/attachments/315652489544335361/597557191620427806/burning.webm
https://cdn.discordapp.com/attachments/315652489544335361/597564070455017537/station_fire.webm

Fire now damages players and objects. Any component can implement IFireExposable to allow it to receive fire exposure events when in contact with fire.

Hotspots damage adjacent atmos-impassable tiles, like doors / windows / grills. This can allow fires to spread and be more destructive.

Integrity now has a minimum temperature resistance, which is set pretty high for doors so doors currently don't burn up since you can't create fires hotter than 700.

Integrity now has a Resistances configuration, allowing objects to indicate if they are flammable. 

Players and objects can catch on fire and burn (see Paper or any of the wooden looking crates). Small objects use the small burning sprite, large objects use some burning sprites I found that look a little different from the normal hotspots. Burned up objects turn into ash. Raw meat cooks when burnt up.

Players accumulate "fire stacks" just like in tg, and player sprite shows the lightly burning vs. fully engulfed sprite depending on how many stacks they have. There isn't currently a way to eliminate the fire stacks other than staying out of fire and letting the stacks slowly decrease or go into a cell with no oxygen.

Unlike tgstation, fire damage on players is currently applied to Chest and does not burn any equipment or check if player is wearing inflammable stuff. This should be changed to more closely match tg at some point, probably after we migrate more off of unicloths.

Clean up Decal prefabs and components, there were uneeded extra components and inheritance (this will be overhauled when we implement a new decal system but this should help somewhat in the meantime). All decals use the same decal component which allows random variations of sprite appearance.

Scorch marks appear on floor when there's a hotspot, but this overwrites the appearance of the tile and should be refactored to use a new and improved decal system we have yet to create.

Canisters refactored to use prefab variants. I like this a LOT and the more things we can convert to prefab variants the more time we will save when making changes that effect lots of objects of the same "type". Other things which could be converted would be crates, lockers, guns, etc...

Note that canister refactor required me to update tile palette and remap the cansiters in OutpostStation.

Added additional ways to expose hotspots - player burning, shooting a gun, a new badmin right click option "Hotspot". You can use badmin options even as ghost, which can make testing easier.


UnityProject/Assets/Scripts/Effects/Fire/BurningDirectionalOverlay.cs
```diff
@@ -0,0 +1,86 @@
...
+	/// stop displaying the burning animation
+	/// </summary>
+	public void StopBurning()
+	{
+		spriteRenderer.sprite = null;
+		spriteRenderer.enabled = false;
+		burn = false;
+	}
+
+	private void Update()
```
DooblyNoobly, line 76: I would put a Profile Sample (https://docs.unity3d.com/ScriptReference/Profiling.Profiler.BeginSample.html) around this update loop when you profile large fires for performance. If this is the cause of the low FPS then try using the UpdateManager instead

UnityProject/Assets/Scripts/Effects/Fire/BurningOverlay.cs
```diff
@@ -0,0 +1,59 @@
...
+	/// stop the burning animation
+	/// </summary>
+	public void StopBurning()
+	{
+		spriteRenderer.sprite = null;
+		spriteRenderer.enabled = false;
+		burn = false;
+	}
+
+	private void Update()
```
DooblyNoobly, line 49: Same here (RE: profiling)

UnityProject/Assets/Scripts/Health/IFireExposable.cs
```diff
@@ -0,0 +1,13 @@
+/// <summary>
```
DooblyNoobly, line 1: Not sure if this should be in the Health folder. Maybe a dedicated folder for fire?

DooblyNoobly: Same with all of the other fire scripts

DooblyNoobly (approved):

PR #2008, opened 2019-07-08
Huge Sprites Refactor

Bod9001:
That's a lot of changes

### Purpose
To make it easier to find sprites and also to update them

### Notes:
Security Console needs to be worked on,  I would need a bit of help from  fomalsd  to Use the net image stuff
dev cloth Spawn needs to be implemented
I can't work out how to set sprites for the shutters


Aranclanos:
panic

DooblyNoobly:
image

DooblyNoobly:
I would like to talk to the heads first before we merge. Putting on DNM.

Aranclanos:
Is it possible that 'Delays' have an extra pair of brackets? I reverted them on the temp jsons of SpriteHandler

see link: https://github.com/unitystation/unitystation/pull/2060/commits/242d28ac87067626c4fe1a59cd6c2f100cdac78d

Bod9001:
it's so With direction variance you can have different delays for each frame, Annoyingly  you can't have different amount of Frames for each direction, Since that would mess up the formula

fomalsd (dismissed):
Took me ~11 minutes to import all the extra stuff when switching from another branch on i7-3770, not as bad as I expected.

But! Game startup time in editor is now helluva 90 seconds (and 3 minutes when regenerating atlases) – vs 15 seconds before! in my opinion this is **unacceptable**. Long game startups really piss people off and it was the reason I had to upgrade all my machines before. Returning to more-than-a-minute-startup hell is something I can't tolerate.

Also getting thousands of warnings like this: 
`Sprite xmashat_0 matches more than one built-in atlases. Default to use the first available atlas.`
it's known that iconstate names aren't unique across different spritesheets, I suspect it would mean that atlases ignore such duplicates and only take the first one found. this should be addressed.

I sincerely hope that such long startup time is solely because of thousands of sprite warnings that had to be printed, otherwise comfy sprite lookup isn't worth the 6x startup time difference.

Updated UI is a nice touch though!

Bod9001:
Wasn't the main reason for atlases for compression?
Will have to test what it's like without atlases

Bod9001:
yeah, It was strange when I was Testing out the UI To make sure I set it up Correctly it didn't take too long to load so Must be some  janky setting somewhere

DooblyNoobly:
if you are testing without spritepacker you want to look at the Draw calls and fps to get a clear idea of the impact. Make sure to test at different spots around the station

Bod9001:
with no Sprite Packer, the draw calls is about 180 top 200
this is with the Almost all of the objects moved over to the new sprites

Bod9001:
I think the packing also breaks mouse interaction stuff, it doesn't seem to work anymore With the Atlases?

fomalsd:
just mentioning that startup time is no longer an issue here

fomalsd (approved):
There are still some things left to work on, but overall it's done and works nicely

PR #2018, opened 2019-07-13
Fixes pipes sprites on shuttle rotations

Aranclanos:
### Purpose
Fixes pipes sprites on shuttle rotations
Pipes will now use their registerTile position to set their sprite.
Removes the SpriteMatrixRotation from pipes and the requirement of them in ItemAttributes.

### Notes:
commit says server player but its both client/server, fixed it and forgot to change that
I'm thinking on renaming the bitflag later on so it isn't also called direction, but for another PR

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Objects/Atmos/AirVent.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Objects/Atmos/AirVent.prefab
@@ -17,10 +17,10 @@ GameObject:
   - component: {fileID: 8354343867630940287}
   - component: {fileID: 966013011358701611}
   - component: {fileID: 6479037130962053368}
-  - component: {fileID: 114247328083268004}
   - component: {fileID: 114207068481858528}
   - component: {fileID: 3953549719403137607}
   - component: {fileID: 3771872541723634421}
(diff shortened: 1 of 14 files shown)


Jibbajabbafic:
I like what you've done with atmos pipes so far but I don't think removing SpriteMatrixRotation from the required list for items is a good solution. 

When pipes aren't anchored they don't rotate like normal objects should:
eCJgJTc98i

You could argue that they should keep their rotation even when anchored, but 99% of items will still need that behaviour so you're removing a big safeguard for new people creating items.

You could easily change it to enable/disable the SMR component when anchoring so I suggest doing it that way:
eJCNwci2jv

Jibbajabbafic (changes requested):
See other comment

Aranclanos:
I like them not rotating when not anchored I think it looks better, plus it wouldnt drive you crazy if you're building and someone turns the ship

Aranclanos:
I'll readd the requirement and just disable it for pipes

Jibbajabbafic:
Sounds good, ty

Aranclanos:
At the end I can't do that easily so i'd rather get this merged as it is

PR #2026, opened 2019-07-16
Fixes ID cards in security records console not updating properly

Aranclanos:
### Purpose
Fixes ID cards in security records console not updating properly

### Notes:
Basically moves the ID gameobject handling to the console rather than the GUI


Jibbajabbafic:
Unsure if this fixes anything, but having the id removal and insertion on the console makes more sense, so yeah.

PR #2042, opened 2019-07-20
Fix some issues with the Right-click menu

### LINKED ISSUE #2040: Cannot open Right-click menu for items in matrix the player is not in
# Description
Any item that is not in the same matrix as the player cannot be interacted with using the right-click menu.

## Category
BUG

## Detailed Bug Report
As the code only retrieves items that are in the same matrix as the player, attempting to open the right-click menu of any item not in the same matrix will not call `GenerateRightClickOptions()` on that item.

## Steps to Reproduce
1. Go to any shuttle.
2. Drop an item into the shuttle
3. Step out of the shuttle and attempt to open the right-click menu on the item

### LINKED ISSUE #2041: Right-click shows items that are obscured by a floor tile
# Description
Right clicking any tile containing a hidden cable will reveal the cable in the right-click menu.

## Category
BUG

## Detailed Bug Report
A cable is hidden if the tile it is on is does not have `LayerType.Base` as LayerType, which there currently is a correct check for. However, cables occur on layer 22, which the check does not correctly do, currently(instead, it checks if `obj.layer < 1`).

## Steps to Reproduce
1. Right-click on any tile that is covering a cable

0xJoey:
### Purpose
Fixes #2040 
Fixes #2041 

### Notes:
Also fixes an `if (<cond>) { return true; } return false;` anti-pattern.

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Right click/RightclickManager.cs
+++ b/UnityProject/Assets/Scripts/UI/Right click/RightclickManager.cs
@@ -114,6 +114,7 @@ public class RightclickManager : MonoBehaviour
 	{
 		LayerTile tile = UITileList.GetTileAtPosition(obj.WorldPosClient());
 
+		// Layer 22 is the 'Floor' layer
 		return (tile != null && tile.LayerType != LayerType.Base && obj.layer == 22);
 	}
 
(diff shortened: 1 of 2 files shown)


Jibbajabbafic:
Can I suggest using a constant for the layer instead of a magic number

0xJoey:
Good point. Is there already an Enum or something similar that describes cable layers or should I just define my own?

Jibbajabbafic:
I'm unsure, if you don't find one feel free to make it yourself in an appropriate place

DooblyNoobly:
I don't mind people just putting a comment next to it explaining what name the layer is. I don't think its necessary to create an enum for it as unity already has that in a way, just unroll the layers list on any gameobject to see.

Jibbajabbafic (approved):
Looks good, and it works in-game.

Nice work!

Issue #2044, opened 2019-07-20
Account login 'Enter' keyboard button doesn't work


Aranclanos:
# Description
The only way to login is to use your mouse and click on the 'Login' button. Pressing enter does not work, it should!

## Category
TODO

vladb99:
Could I try work on it? This would be my first attempt to contribute to an open source project.

Aranclanos:
there's a first time for everything

PR #2046, opened 2019-07-21
Update GUI_LobbyDialogue.cs

vladb99:
### Purpose
Adds #2044 

### Notes:
_

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Lobby/GUI_LobbyDialogue.cs
+++ b/UnityProject/Assets/Scripts/UI/Lobby/GUI_LobbyDialogue.cs
@@ -370,5 +370,14 @@ namespace Lobby
 			loggingInPanel.SetActive(false);
 			connectionPanel.SetActive(false);
 		}
+		
+		void Update()
+		{
+			// Log in when the enter key gets pressed
+			if (Input.GetKeyDown(KeyCode.Return) || Input.GetKeyDown(KeyCode.KeypadEnter))
+			{
+				OnLogin();
+			}			
+		}
 	}
 }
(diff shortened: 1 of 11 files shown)


UnityProject/Assets/Scripts/UI/Lobby/GUI_LobbyDialogue.cs
```diff
@@ -370,5 +370,14 @@ void HideAllPanels()
 			loggingInPanel.SetActive(false);
 			connectionPanel.SetActive(false);
 		}
+		
+		void Update()
+		{
+			// Log in when the enter key gets pressed
+			if (Input.GetKeyDown(KeyCode.Return) || Input.GetKeyDown(KeyCode.KeypadEnter))
```
DooblyNoobly: Only problem with this is that if you press enter on any other screen in the lobby windows, it'll call that Login method. You need to determine when the Account login screen is active. Also it might be a better idea to separate the "enter to continue" behaviour into its own component like the tab next component found here: https://github.com/unitystation/unitystation/tree/develop/UnityProject/Assets/Scripts/UI/GUI/Components 

You will need to make the component know when it is visible (or just put it on the last field of Account login as it will be disabled on the next screens anyway, preventing the behaviour for working) and then you'll need a public UnityEvent to invoke when the user does hit enter so that you can add the OnLogin method to it via the inspector. Don't forget to save it to the lobby scene and commit the scene also.

DooblyNoobly (changes requested):

UnityProject/Assets/Scripts/UI/Lobby/GUI_LobbyDialogue.cs
vladb99: Alright, I will change that. Thanks for the help!

PR #2050, opened 2019-07-23
Nettab sprite loading change

Aranclanos:
### Purpose
NetSpriteAndColor and NetSpriteImage will now load their sprites from a resource load rather than getting them from the PlayerSprites manager.

### Notes:
Now vendor machines can use this instead of spawning an object and taking the snapshot for the sprite

### DIFF
--- a/UnityProject/Assets/Scripts/UI/SecurityRecords/GUI_SecurityRecordsEntryPage.cs
+++ b/UnityProject/Assets/Scripts/UI/SecurityRecords/GUI_SecurityRecordsEntryPage.cs
@@ -96,25 +96,25 @@ public class GUI_SecurityRecordsEntryPage : NetPage
 
 		if(characterSettings != null)
 		{
-			torso.SetComplicatedValue("human_parts_greyscale", characterSettings.torsoSpriteIndex, characterSettings.skinTone);
-			head.SetComplicatedValue("human_parts_greyscale", characterSettings.headSpriteIndex, characterSettings.skinTone);
+			torso.SetComplicatedValue("icons/mob/human_parts_greyscale", characterSettings.torsoSpriteIndex, characterSettings.skinTone);
+			head.SetComplicatedValue("icons/mob/human_parts_greyscale", characterSettings.headSpriteIndex, characterSettings.skinTone);
 			rightLeg.SetValue = characterSettings.skinTone;
 			leftLeg.SetValue = characterSettings.skinTone;
 			rightArm.SetValue = characterSettings.skinTone;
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/UI/SecurityRecords/GUI_SecurityRecordsEntryPage.cs
```diff
@@ -96,25 +96,25 @@ public void UpdateEntry()
...
+			head.SetComplicatedValue("icons/mob/human_parts_greyscale", characterSettings.headSpriteIndex, characterSettings.skinTone);
 			rightLeg.SetValue = characterSettings.skinTone;
 			leftLeg.SetValue = characterSettings.skinTone;
 			rightArm.SetValue = characterSettings.skinTone;
 			leftArm.SetValue = characterSettings.skinTone;
 			eyes.SetValue = characterSettings.eyeColor;
-			beard.SetComplicatedValue("human_face", characterSettings.facialHairOffset, characterSettings.facialHairColor);
-			hair.SetComplicatedValue("human_face", characterSettings.hairStyleOffset, characterSettings.hairColor);
+			beard.SetComplicatedValue("icons/mob/human_face", characterSettings.facialHairOffset, characterSettings.facialHairColor);
+			hair.SetComplicatedValue("icons/mob/human_face", characterSettings.hairStyleOffset, characterSettings.hairColor);
```
Bod9001, line 107: Is this to make the loading a little bit better?
Since my only worry with this as it will make it a bit harder to move sprites around

DooblyNoobly: @Aranclanos can you clarify this?

Aranclanos: I was using SpriteManager.PlayerSprites[string] but since SpriteManager doesnt contain all sprites I'm swapping to resource load, which needs the exact path

DooblyNoobly: Problem is, when we swap over to the new sprite system this will probably get overlooked and break. It would be best to add what you need to SpriteManager so there is only one place we need to fix instead of searching through code for these Resource references

Aranclanos: in that case this can be closed and left as it is

PR #2060, opened 2019-07-26
Adds the class SpriteHandler, it syncs sprites and provides animations.

Aranclanos:
### Purpose
Adds the class SpriteHandler, it syncs sprites and provides animations.
Adds said class to the DNA scanner.
hZKLUByCdE

### Notes:
I took the sprites from bod's PR and placed them on a different folder so it wouldnt add conflicts
The animations are at a fixed rate, dont load from the json
The bad thing about this system is having to load the prefab SpriteHandler with this:
image
and make sure when you call ChangeSprite(index number) you have to make sure that the argument is the one you want, it has a "magic number" feeling to it

Also not all objects in the game should have this, only the ones that change its sprite or have an animation
The good thing is that it does the syncing and animation for you, you only have to call ChangeSprite() and forget about everything else


Bod9001:
so For the animations couldn't you just drag in the Sprite sheet And let the The code handle the Loading of the Animations, Instead of having to set each animation frame individually, that would help to save time?

Bod9001:
Since you get the string  location  for the Animation data, Doesn't that mean you could technically synchronise what sprites are shown without having to specify them on the client?

Bod9001:
ok, One question how does it get the delays from the .json file for the animations?

UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
```diff
@@ -0,0 +1,94 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Networking;
+
+public class SpriteHandler : NetworkBehaviour
```
DooblyNoobly: Please add some Summary comments here

DooblyNoobly (changes requested):

UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
```diff
@@ -0,0 +1,94 @@
...
+		waitTime = animationStills.waitTime;
+		spriteRenderer.sprite = animationStills.sprite;
+	}
+
+	void SetSpriteList(List<Sprite> newSprites)
+	{
+		infoList = new List<SpriteInfo>();
+		for (int i = 0; i < newSprites.Count; i++)
+		{
+			var newSprite = new SpriteInfo(newSprites[i], 0.1f);
```
DooblyNoobly: The delay data needs to be set from the json data

DooblyNoobly (changes requested):

Aranclanos:
Done

UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
```diff
@@ -0,0 +1,120 @@
...
+		timeElapsed = 0;
+		waitTime = animationStills.waitTime;
+		spriteRenderer.sprite = animationStills.sprite;
+	}
+
+	void SetSpriteList(List<Sprite> newSprites)
+	{
+		if (newSprites.Count > 1)
+		{
+			LoadJson(AssetDatabase.GetAssetPath(newSprites[0]));
```
Bod9001: Since this is part of the UnityEditor Namespace
https://docs.unity3d.com/ScriptReference/UnityEditor.html
https://docs.unity3d.com/ScriptReference/AssetDatabase.html

It doesn't work in builds, So you have to keep a note of that if you want to do it in runtime, You'll only be able to run this in the Editor

DooblyNoobly: I would just convert this to Resources Load if its meant to run in a build. Or you can make a CustomEditor for this component that can do all that at edit time and cache the reference.

UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
```diff
@@ -0,0 +1,120 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Networking;
+using UnityEditor;
```
DooblyNoobly: I did not see this in my initial review. You can include UnityEditor libs in runtime script if you put them in compile time definitions. So you can add:
```
#if UNITY_EDITOR
using UnityEditor;
#endif
```
And it will remove this line at compile time. Any other code relying on this lib will also need to be placed in these definitions. But the code won't work in a build obviously as the code is removed at compile time.

DooblyNoobly:
Is this one good for another review?

Aranclanos:
>63k files changed

w-what happened

Bod9001:
sorry for messing up your Fork, don't trust the web merge conflict thing it's betrayed me too many times

Aranclanos:
gonna redo it in another pr

PR #2065, opened 2019-07-27
Fix for closet issues

### LINKED ISSUE #1970: Player sprite blinks when going out from a closet
# Description
When leaving a closet, the player sprite appears, disappears and re-appears in a quick succession 

## Category
BUG

## Detailed Bug Report
https://cdn.discordapp.com/attachments/312454684021620736/593202325229797377/uC477BeUXn.mp4
.mp4 files are too high tech for github, a small video of it happening

vladb99:
### Purpose
Fixes #1970 for lockers and dna scanners
Fixes occasional bug, where player gets brute damage due to low pressure when exiting a locker

- [x]  This PR has been tested in multiplayer (with 2 clients)

### Notes:
The removed if condition from ClosetPlayerHandler was causing the sprite flickering. It also has no purpose, I tested without it in multiplayer and players inside pulled lockers could still interact with them.

### DIFF
--- a/UnityProject/Assets/Scripts/Closets/ClosetPlayerHandler.cs
+++ b/UnityProject/Assets/Scripts/Closets/ClosetPlayerHandler.cs
@@ -69,11 +69,6 @@ public class ClosetPlayerHandler : MonoBehaviour
 					closetControl.Interact(HandApply.ByLocalPlayer(closetControl.gameObject));
 				}
 			}
-			//take the player with the closet so they can interact with it
-			if (closetControl.transform.position != transform.position)
-			{
-				transform.position = closetControl.transform.position;
-			}
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


Bod9001:
yeah, is good, the only issue that is That the player pops in after opening the locker but that would probably require some unet wizardry, so is outside of the issue

vladb99:
Yes I noticed the small delay when player gets out of the locker. I also noticed that on multiplayer this delay doesn't exist for others, so if you get out of a closet you appear immediately on their screen. I thought it might be a fundamental problem with the networking system plus I have not much experience with networking so I left it untouched.

fomalsd:
FYI this change introduces other bugs, because player's assumed position never changes right after spawning (before getting into any lockers).
For example, wallmounts act weird because of this

Issue #1454, opened 2019-01-22
TODO: NetTab GUI for GasContainer B: 150


krille90:
### Description
With the introduction of Atmos, the gas containers are also now functional in the way that you can open and close the valve by clicking on the container. This needs to be extended to a GUI, in which information of the gas are shown, where the release pressure can be changed and the valve can be opened/closed.

### Criteria

- [ ] Clicking on a gas container should open a GUI window
- [ ] Show information of the current content of the gas container, e.g. pressure, temperature, etc.
- [ ] Release pressure can be changed in the GUI
- [ ] Introduce a button to open / close the containers valve
- [ ] Any changes should be synchronized to other players

### Tips
Look at the class `GasContainer` to get a better idea:
https://github.com/unitystation/unitystation/blob/7728a5750651e761f5d495d38a8394abe98e6920/UnityProject/Assets/Scripts/Objects/GasContainer.cs#L10

DooblyNoobly:
Bounty of 150 usd will be awarded to the contributor on a successful merge of all completed criteria above

Carbonhell:
Gonna try to do it when possible

DooblyNoobly:
All yours, changing to in progress

NinjaDanz3r:
Talked with Carbonhell and I'm shanghaiing this task.
Quick mockup of the Valve:
https://imgur.com/Zp9bN7d

NinjaDanz3r:
![gastankui mockup](https://user-images.githubusercontent.com/10200316/52434538-d6d07400-2b0f-11e9-8714-8c7759310730.png)
A quick mockup of the whole interface.

Jibbajabbafic:
Remember, all this will need to be synced across clients using NetObjects (NetSlider, NetToggle... etc) so you might need to make a new one for the valve depending on how it's being implemented

NinjaDanz3r:
I might as well do that, I think rotary controls might show up in a few other cases.
Also, new preliminary UI design according to input from Foma:
![gasui](https://user-images.githubusercontent.com/10200316/52522180-fe474e00-2c81-11e9-9a00-9ab172ca6af4.png)

chairbender:
@NinjaDanz3r any update on this?

fomalsd:
Need this rather sooner than later

JacobSweeten:
Haven't contributed in a while, but I can give it a shot this weekend. I'll update you on whether or not it's feasible for me to complete when I get to look at it.

[...]

PR #2067, opened 2019-07-29
Basic Canister UI

chairbender:
### Purpose
Adds a UI for canister, part of #1454 .

Remaining work to be done:
* Sounds
* Flashing red status LED when super low (currently just shows solid red when low)
* Ability to drag window (disabled drag because it was preventing from using the wheel)
* Polish the drag interaction of the wheel (currently the rim changes color slightly when you click which looks weird, and it would be nice to add more of a signifier such as changing mouse icon when over the wheel or something)
* Potentially add client interpolation between updates of the internal pressure (currently it only updates each 0.5 seconds, client could interpolate these values so the numbers change more smoothly)
* Color variations and text changes for each canister (want to make it so I can also avoid having to create a separate asset for each canister background if possible, can probably do this using tinting)
* Connected / disconnected status indicator should update
* Click the release pressure to allow typing in the value
* Clean up remaining TODO comments


DooblyNoobly:
Damn! The canister UI progress is looking really good. Nice work

PR #2083, opened 2019-08-06
Spray mechanic and effects

Aranclanos:
### Purpose
The fire extinguisher will now contain reagents, water by default.
Adds a security toggle mechanic for the fire extinguisher, it'll change its sprite.
Fire extinguishers will now remove hotspots if used in the general direction of the fire.

Adds a space cleaner sprayer.
Adds space_cleaner to the chemical reagents list, it'll clean floors.
Adds a particle effect for the extinguisher and space cleaner spray.

Adds NBAimApplyHandActivateInteractable for both of the extinguisher interactions.

### Notes:
Particle effects are locked in 0,0 due to a bug where registeritem is not saving the position when picked up

### DIFF
--- a/UnityProject/Assets/Prefabs/GUI/Resources/GUI_ChemistryDispenser.cs
+++ b/UnityProject/Assets/Prefabs/GUI/Resources/GUI_ChemistryDispenser.cs
@@ -38,6 +38,7 @@ public class GUI_ChemistryDispenser : NetTab {
 		"sulfur",
 		"water",
 		"welding_fuel",
+		"space_cleaner"
 	};
 
 	private List<string> DispenseAmounts = new List<string>(){ //Some security bizz
--- a/UnityProject/Assets/Prefabs/Items/Tools/Resources/Extinguisher.prefab
+++ b/UnityProject/Assets/Prefabs/Items/Tools/Resources/Extinguisher.prefab
@@ -248,7 +248,7 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   clothingReference: -1
-  hierarchy: /obj/item/weapon/extinguisher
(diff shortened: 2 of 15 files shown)


DooblyNoobly:
Just did a quick play test these are the issues I encountered:

- Item sprites are not being updated from server to client: 
image

- No particle effect can be seen. I guess this is what you mentioned in your notes. Let me know if you need a hand with that

- No soundfx

fomalsd:
Tried it and see no reaction at all (yes, I've toggled the extinguisher), floors aren't getting wet, just nothing seems to happen

fomalsd:
oh, it seems to show fx if you click and drop it rapidly
I'm starting to think about some kind of ShowParticleMessage, because single syncvar won't cover it – it needs both start position (or object to parent to) and angle, also syncvars are unrobust.
Clients don't know Assumed Positions for items/players in containers, and they won't – it's secret info, so you'll have to send particle start positions from server

fomalsd:
https://gist.github.com/fomalsd/fdbfb5c0d06fdff1a423ed091f226cb2
made a dirty fix that will work if you pull my honk branch (#2092)
but it involves two syncvars, so it will be unreliable af
GIF
still it's cool that you made these particles collide with walls
– but where's slippery floor? it's near impossible to check if spray is working without particles on

Aranclanos:
The single syncvar wasnt the best but it was elegant enough, I can't really accept two, I'm willing to set up a particle manager so it works as the soundmanager (without strings) but in another pr
A particle manager could also kill all those particles on cables 

also extinguishers do not wet the floors

Aranclanos:
added the sounds and fixed the sprite at client login

fomalsd:
it's good to merge as long as you address particles in following PRs

PR #2094, opened 2019-08-09
SpriteHandler 2

Aranclanos:
### Purpose
continuation of https://github.com/unitystation/unitystation/pull/2060
Adds the class SpriteHandler, it syncs sprites and provides animations.

### Notes:
It loads from the sprites currently in develop without using the new fractured ones, so the animation has a static value of 0.1 seconds per image, there are no json files to load from

### DIFF
--- a/UnityProject/Assets/Scripts/Medical/DNAscanner.cs
+++ b/UnityProject/Assets/Scripts/Medical/DNAscanner.cs
@@ -8,9 +8,7 @@ public class DNAscanner : ClosetControl, IAPCPowered
 	public LivingHealthBehaviour occupant;
 	public string statusString;
 	[SyncVar(hook = nameof(SyncPowered))] public bool powered;
-	public Sprite closedWithOccupant;
-	public Sprite doorClosedPowerless;
-	public Sprite doorOpenPowerless;
+	public SpriteHandler spriteHandler;
 
 	public override void OnStartServer()
 	{
@@ -69,24 +67,24 @@ public class DNAscanner : ClosetControl, IAPCPowered
 		{
 			if (!powered)
 			{
-				spriteRenderer.sprite = doorOpenPowerless;
+				spriteHandler.ChangeSprite(6);
 			}
 			else
 			{
-				spriteRenderer.sprite = doorOpened;
+				spriteHandler.ChangeSprite(2);
 			}
 		}
 		else if (!powered)
 		{
-			spriteRenderer.sprite = doorClosedPowerless;
+			spriteHandler.ChangeSprite(5);
 		}
 		else if (value == ClosetStatus.Closed)
 		{
-			spriteRenderer.sprite = doorClosed;
+			spriteHandler.ChangeSprite(0);
 		}
 		else if(value == ClosetStatus.ClosedWithOccupant)
 		{
-			spriteRenderer.sprite = closedWithOccupant;
+			spriteHandler.ChangeSprite(1);
 		}
 
 	}
--- a/UnityProject/Assets/Scripts/SpriteHandler/SpriteHandler.cs
+++ b/UnityProject/Assets/Scripts/SpriteHandler/SpriteHandler.cs
@@ -1,0 +1,93 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Networking;
+
+public class SpriteHandler : NetworkBehaviour
+{
+	public SpriteRenderer spriteRenderer;
+
+	public List<SpriteList> spriteList = new List<SpriteList>();
+	[SyncVar(hook = nameof(SyncSprite))] public int spriteIndex;
+
+	private List<SpriteInfo> infoList = new List<SpriteInfo>();
+	private int animationIndex = 0;
+	private float timeElapsed = 0;
+	private float waitTime;
+
+	public override void OnStartClient()
+	{
+		SyncSprite(spriteIndex);
+	}
+
+	public void UpdateMe()
+	{
+		timeElapsed += Time.deltaTime;
+		if (timeElapsed >= waitTime)
+		{
+			animationIndex++;
+			if (animationIndex >= infoList.Count)
+			{
+				animationIndex = 0;
+			}
+			SetSprite(infoList[animationIndex]);
+		}
+	}
+
+	void SetSprite(SpriteInfo animationStills)
+	{
+		timeElapsed = 0;
+		waitTime = animationStills.waitTime;
+		spriteRenderer.sprite = animationStills.sprite;
+	}
+
+	void SetSpriteList(List<Sprite> newSprites)
+	{
+		infoList = new List<SpriteInfo>();
+		for (int i = 0; i < newSprites.Count; i++)
+		{
+			var newSprite = new SpriteInfo(newSprites[i], 0.1f);
+			infoList.Add(newSprite);
+		}
+		if (newSprites.Count > 1)
+		{
+			UpdateManager.Instance.Add(UpdateMe);
+		}
+		else
+		{
+			UpdateManager.Instance.Remove(UpdateMe);
+		}
+		SetSprite(infoList[0]);
+	}
+
+	public void ChangeSprite(int newSprites)
+	{
+		if (spriteIndex != newSprites)
+		{
+			spriteIndex = newSprites;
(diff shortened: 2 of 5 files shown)


UnityProject/Assets/Scripts/SpriteHandler/SpriteHandler.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Networking;
+
+public class SpriteHandler : NetworkBehaviour
+{
+	public SpriteRenderer spriteRenderer;
+
+	public List<SpriteList> spriteList = new List<SpriteList>();
+	[SyncVar(hook = nameof(SyncSprite))] public int spriteIndex;
```
chairbender, line 11: If I understand correctly, users of this class are supposed to change the sprite using the ChangeSprite method. Therefore spriteIndex should be a private field to avoid confusion when someone tries to set it manually.

Also, syncvar field should basically never be public anyway.

UnityProject/Assets/Scripts/SpriteHandler/SpriteHandler.cs
```diff
@@ -0,0 +1,93 @@
...
+			UpdateManager.Instance.Add(UpdateMe);
+		}
+		else
+		{
+			UpdateManager.Instance.Remove(UpdateMe);
+		}
+		SetSprite(infoList[0]);
+	}
+
+	public void ChangeSprite(int newSprites)
```
chairbender, line 63: I don't think this would work well if called from client code. If that's the case, please add [Server] tag, and probably rename to ServerChangeSprite so it's more clear that it should only be called from the server. It could be confusing for a new dev calling this from client code and then the value gets overwritten by the server.

UnityProject/Assets/Scripts/SpriteHandler/SpriteHandler.cs
```diff
@@ -0,0 +1,93 @@
...
+	{
+		if (spriteIndex != newSprites)
+		{
+			spriteIndex = newSprites;
```
chairbender, line 67: This might have bugs in some situations from the server's perspective. From my wiki page on SyncVar (https://github.com/unitystation/unitystation/wiki/SyncVar-Best-Practices-for-Easy-Networking), I recommend only setting the value of spriteIndex inside the SyncVar hook method. I.e., first line of SyncSprite should be spriteIndex = value. Server should call SyncSprite when it wants to change the value.

UnityProject/Assets/Scripts/SpriteHandler/SpriteHandler.cs
```diff
@@ -0,0 +1,93 @@
...
+		}
+	}
+
+	void SyncSprite(int value)
+	{
+		SetSpriteList(spriteList[value].spriteList);
```
chairbender, line 73: You never assign spriteIndex = value. You need to assign the field in the syncvar hook. Thus the client will still change the sprite (due to calling SetSpriteList), but the spriteIndex will always stay at the original value.

This should be refactored to set spriteIndex = value, and server should only change the field via calling SyncSprite.

UnityProject/Assets/Scripts/SpriteHandler/SpriteHandler.cs
chairbender: ...and to still allow assigning this in editor, you can have a separate public field for initialSpriteIndex.

UnityProject/Assets/Scripts/SpriteHandler/SpriteHandler.cs
```diff
@@ -0,0 +1,93 @@
...
+	private List<SpriteInfo> infoList = new List<SpriteInfo>();
+	private int animationIndex = 0;
+	private float timeElapsed = 0;
+	private float waitTime;
+
+	public override void OnStartClient()
+	{
+		SyncSprite(spriteIndex);
+	}
+
```
chairbender, line 22: it's hard to reproduce, but if this object is pooled, its animation when spawned will be initially set to whatever it had when it was put back into the pool. You should define a hook for OnSpawnedServer which resets the sprite to its correct initial value (probably need a new initialSpriteIndex field for assigning this in editor, as mentioned earlier).

UnityProject/Assets/Scripts/Medical/DNAscanner.cs
```diff
@@ -8,9 +8,7 @@ public class DNAscanner : ClosetControl, IAPCPowered
 	public LivingHealthBehaviour occupant;
 	public string statusString;
 	[SyncVar(hook = nameof(SyncPowered))] public bool powered;
-	public Sprite closedWithOccupant;
-	public Sprite doorClosedPowerless;
-	public Sprite doorOpenPowerless;
+	public SpriteHandler spriteHandler;
```
chairbender, line 11: this can be private, and get it using GetComponent<SpriteHandler> during Awake()

UnityProject/Assets/Scripts/SpriteHandler/SpriteHandler.cs
```diff
@@ -0,0 +1,93 @@
...
+	{
+		infoList = new List<SpriteInfo>();
+		for (int i = 0; i < newSprites.Count; i++)
+		{
+			var newSprite = new SpriteInfo(newSprites[i], 0.1f);
+			infoList.Add(newSprite);
+		}
+		if (newSprites.Count > 1)
+		{
+			UpdateManager.Instance.Add(UpdateMe);
```
chairbender, line 54: won't this risk adding UpdateMe to UpdateManager multiple times?

UnityProject/Assets/Scripts/SpriteHandler/SpriteHandler.cs
```diff
@@ -0,0 +1,93 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Networking;
+
+public class SpriteHandler : NetworkBehaviour
+{
+	public SpriteRenderer spriteRenderer;
+
+	public List<SpriteList> spriteList = new List<SpriteList>();
```
chairbender, line 10: structure makes sense to me in editor, but might not to others. If you rename this to animationList I think it will help. You have a list of animations, and each animation has a list of sprites. So calling this a spriteList confuses between what an animation is and what a sprite is.

UnityProject/Assets/Scripts/SpriteHandler/SpriteHandler.cs
chairbender, line 11: I would prefer if spriteIndex was called animationIndex, since you are using it to select which animation is being played, and animationIndex should instead be called spriteIndex, because you are using it select which sprite of the animation to display.

UnityProject/Assets/Scripts/SpriteHandler/SpriteHandler.cs
```diff
@@ -0,0 +1,93 @@
...
+			spriteIndex = newSprites;
+		}
+	}
+
+	void SyncSprite(int value)
+	{
+		SetSpriteList(spriteList[value].spriteList);
+	}
+
+	[System.Serializable]
```
chairbender, line 76: Why is this needed? I think it can be removed. This makes me think you are trying to send this class over the network, which is not possible (Sprite cannot be serialized, pretty sure)

chairbender (changes requested):
This will be a great component to have, animation is currently a mess (and many of our objects don't animate despite having everything we would need to do it) and Unity's way of controlling animations is overkill.

Just have some issues I noticed related to syncvar usage, also things relating to usability for other developers like naming and making things private that don't need to be public.

fomalsd:
any news?

chairbender:
I just removed my assignment so it's clear the PR isn't waiting for me to do something

chairbender:
Status update - according to @Aranclanos , @Bod9001 is going to be making changes to this along with his sprite overhaul, so this can most likely be closed.

UnityProject/Assets/Scripts/SpriteHandler/SpriteHandler.cs
Bod9001: it's so the inspector can view The class properly , If it's not there sometimes it just doesn't show up at all

PR #2092, opened 2019-08-09
Bike horn (and some refactoring)

fomalsd:
### Purpose
Fixes #2069 
Fixes #2073
Fixes #2090 
Gives clown something to do (has a chance to harm eardrums when honked in close proximity)
GIF

Synchronizes player (dis)appearance with closet door
Adds "polyphonic sound" option, so that more than one instance of sound could be played
Removes VisibleBehaviour and ClosetPlayerHandler
Makes ObjectBehaviour kinda redundant (as it's identical to PushPull)
If you need to make something (dis)appear you can just set its VisibleState boolean

### Open Questions and Pre-Merge TODOs

- [x]  This fix is tested on the same branch it is PR'ed to.
- [x]  I correctly commented my code
- [x]  My code is indented with tabs and not spaces
- [x]  This PR does not include any unnecessary .meta, .prefab or .unity (scene) changes
- [x]  This PR does not bring up any new compile errors
- [x]  This PR has been tested in editor
- [x]  This PR has been tested in multiplayer (with 2 clients and late joiners, if applicable)

### Notes:
You'll need to spawn it manually as there's no way to spawn prefabs as part of job outfits yet (see #2091)
This needs to be tested on live server to see if WaitForCameraToReachCloset stuff is still relevant

PR #2096, opened 2019-08-11
Removes the ability to shoot inside a locker

### LINKED ISSUE #2072: You can shoot guns inside lockers
# Description
You can shoot inside lockers.

And it doesn't look too good, you actually just get recoil effect and literally nothing else happens, not even bullets appear.

## Category
BUG

## Detailed Bug Report
TODO: Remove shooting insided locker.

## Steps to Reproduce
Get in the locker with a machine gun and ratatatatatata

zrparsons:
### Purpose
Removes the ability to shoot inside lockers fixes #2072 

### Notes:
I'm still not super familiar with the code base so this might not be the best way to do this but it is my understanding that 'IsHidden' is used exclusively for being inside lockers.

### DIFF
--- a/UnityProject/Assets/Scripts/Input System/InteractionV2/DefaultWillInteract.cs
+++ b/UnityProject/Assets/Scripts/Input System/InteractionV2/DefaultWillInteract.cs
@@ -72,9 +72,13 @@ public static class DefaultWillInteract
 	/// </summary>
 	public static bool AimApply(AimApply interaction, NetworkSide side)
 	{
-		return  Validations.CanInteract(interaction.Performer, side) &&
-					!(interaction.Performer.GetComponent<PlayerScript>() != null &&
-					interaction.Performer.GetComponent<PlayerScript>().IsHidden);
+		if ( !Validations.CanInteract(interaction.Performer, side) )
+		{
+			return false;
+		}
+
+		bool? isNotHidden = !interaction.Performer.Player()?.Script.IsHidden;
+		return isNotHidden.GetValueOrDefault( true );
 	}
 
 	/// <summary>
--- a/UnityProject/Assets/Scripts/Input System/InteractionV2/DefaultWillInteract.cs
+++ b/UnityProject/Assets/Scripts/Input System/InteractionV2/DefaultWillInteract.cs
@@ -72,7 +72,9 @@ public static class DefaultWillInteract
 	/// </summary>
 	public static bool AimApply(AimApply interaction, NetworkSide side)
 	{
-		return Validations.CanInteract(interaction.Performer, side);
+		return  Validations.CanInteract(interaction.Performer, side) &&
+					!(interaction.Performer.GetComponent<PlayerScript>() != null &&
+					interaction.Performer.GetComponent<PlayerScript>().IsHidden);
 	}
 
 	/// <summary>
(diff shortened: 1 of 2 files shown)


zrparsons:
just saw that this one might have been fixed already by #2092 so you can probably close this one

fomalsd:
haven’t checked if it’s the case, but I didn’t do it on purpose

zrparsons:
oh sorry you referenced it in that issue thought you might have fixed it.

fomalsd:
yeah, it was a typo in an issue number but the reference didn’t go away for some reason

UnityProject/Assets/Scripts/Weapons/Gun.cs
```diff
@@ -179,6 +179,10 @@ protected override bool WillInteract(AimApply interaction, NetworkSide side)
 			return false;
 		}
 
+		if (interaction.Performer == PlayerManager.LocalPlayer && PlayerManager.PlayerScript.IsHidden){
+			return false;
+		}
+
```
Aranclanos: grab the playerscript with a getcomponent() call so both the server and client can do the check, this specific function acts like that

Aranclanos (changes requested):

Aranclanos:
though at the same time im not sure about making this check here, since all other items would need it but chairbender would know better

UnityProject/Assets/Scripts/Weapons/Gun.cs
zrparsons: I just pushed another commit anyway with what i think you wanted and yea thats a much better way of doing it, sorry still learning.

fomalsd:
@chairbender could you take a look at this PR? maybe “inside locker” check should be added to default validations for PositionalHandApply or something like that?

chairbender:
I would suggest adding it to the default AimApply validation, since that's the interaction used for shooting.

chairbender (commented):
I would suggest adding it to the default AimApply validation, since that's the interaction used for shooting.

That can be found in DefaultWillInteract.AimApply

zrparsons:
just pushed a new commit moving it to defaultWillInteract tell me if that boolean expression is too gross.

also thanks for all the help guys.

UnityProject/Assets/Scripts/Input System/InteractionV2/DefaultWillInteract.cs
```diff
@@ -72,7 +72,9 @@ public static bool MouseDrop(MouseDrop interaction, NetworkSide side)
 	/// </summary>
 	public static bool AimApply(AimApply interaction, NetworkSide side)
 	{
-		return Validations.CanInteract(interaction.Performer, side);
+		return  Validations.CanInteract(interaction.Performer, side) &&
```
fomalsd: Yeah, the idea seems right. That bool inversion is hard to spot though and it's best to avoid calling GetComponent when it's not required. How about this:
```cs
		if ( !Validations.CanInteract(interaction.Performer, side) )
		{
			return false;
		}

		bool? isNotHidden = !interaction.Performer.Player()?.Script.IsHidden;
		return isNotHidden.GetValueOrDefault( true );
```

zrparsons: yea that does look a lot better just pushed another commit, thanks.

fomalsd:
sweet, cheers

PR #2098, opened 2019-08-11
SoundManager's methods accept wildcards

### LINKED ISSUE #2029: Implement a solid way to do sound variations
# Description
Right now, SoundManagers has 5 different sounds for rustling (Rustle01..Rustle05). To play a random one of these the code that is used is:

`SoundManager.PlayAtPosition("Rustle0" + UnityEngine.Random.Range(1, 6).ToString(), PlayerManager.LocalPlayer.transform.position);` (one example from InteractableStorage.cs)

This *works*, but it's hardly elegant. Disregarding the unnecessarily verbose syntax above, coding like this means that to add another rustle sound you would need to go to every place where a rustling sound is used and update the code there. This actually didn't happen for this example in PushPull.cs, where only the first 4 rustle sounds are actually used:
`SoundManager.PlayNetworkedAtPos( "Rustle0" + Random.Range(1, 4), pullable.transform.position );`

A better way to do this would be a function that takes only the name of the multi-sound (i.e. "Rustle") and then plays a random one (something like `SoundManager.PlayNetworkedAtPos("Rustle", position);`)

## Category
TODO

YotaXP:
### Purpose
Allows the sound names passed to `SoundManager` to contain "#" wildcards which match any number. When multiple matches are found, one is chosen at random. This will make it easier to use sounds that have multiple variations, such as hit sounds.

Fixes #2029

### Notes:
Although perhaps redundant, the wildcard names are resolved ahead of time in the `PlayNetworked...` methods so that the clients will not choose the sound themselves. This is primarily intended to ensure that everyone hears the same sound.

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/SoundManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/SoundManager.cs
@@ -107,7 +107,8 @@ public class SoundManager : MonoBehaviour
 	}
 
 	/// <summary>
-	/// Serverside: Play sound for all clients
+	/// Serverside: Play sound for all clients.
+	/// Accepts "#" wildcards for sound variations. (Example: "Punch#")
 	/// </summary>
 	public static void PlayNetworked( string sndName, float pitch = -1,
 		bool shakeGround = false, byte shakeIntensity = 64, int shakeRange = 30 )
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Managers/SoundManager.cs
```diff
@@ -129,6 +168,7 @@ public static void Play(string name, float volume, float pitch = -1, float time
 	/// </summary>
 	public static void Play(string name)
```
DooblyNoobly, line 175: Could you add an example to the summary comments, showing how to use the wildcard to play a random sound

UnityProject/Assets/Scripts/Managers/SoundManager.cs
```diff
@@ -137,6 +177,7 @@ public static void Play(string name)
 	/// </summary>
 	public static void PlayAtPosition(string name, Vector3 pos, float pitch = -1)
```
DooblyNoobly, line 185: Same here

DooblyNoobly (changes requested):

DooblyNoobly:
Thanks @YotaXP

Issue #1743, opened 2019-04-16
Can't examine equipped items


Aranclanos:
# Description
It seems to be impossible to inspect an item equipped or in-hand, right clicking doesnt seem to do anything and shift-click is also rejected

## Category
TODO

## Steps to Reproduce
Please enter the steps to reproduce the bug or behaviour:
1.Pick up an item and leave it in-hand or equip it
2.Try to inspect it
3.Find no option to do so

chairbender:
This ticket is similar to #1464 , but is more expansive.

Issue #1464, opened 2019-01-26
Tooltip doesn't show up when mousing over UI


chairbender:
# Description
Mousing over UI doesn't show the tooltip for the items being moused over in the player's inventory. It should show the tooltip for whatever item is moused over.

## Steps to Reproduce
1. Launch in editor and host.
2. Mouse over an item in the inventory
3. Note that the tooltip on the bottom left doesn't change to show what is moused over.

![mouseover](https://user-images.githubusercontent.com/1329062/51791454-24380300-2158-11e9-8840-1f6ebf845999.gif)

Fishfish458:
the tooltip feature as a whole should definitely get a second pass into a more generalized form. it should display only the name of the item and possibly seperated from the register object/register item scripts. byond's could give a name for any tile as well which may be a far goal but it would be useful.

![generalized tooltip](https://user-images.githubusercontent.com/47410468/54500053-c7023780-48e6-11e9-9d02-ef9dd6b21b9d.gif)

PR #2113, opened 2019-08-18
Added text placeholder to display ping separately

plnqux:
### Purpose
Added text placeholder to Panel_ToolTip prefab to display the ping separately from the hover tool tip.

### Notes:
I was looking for some good first issues and found #1743 and then #1464.  While trying to figure out where to add code to handle UI hover updating the tool tip I found that the ping display was wiping out any tool tips that were currently hovered.  To avoid this issue, I've setup another placeholder in the prefab anchored in the bottom right with right justified text so it moves to the right when the window expands.  I was having trouble with the text turning black after running the build & run process and found that I had to set the material for both placeholders to the appropriate Font Texture for Font Awesome to fix that issue.

Currently I've left some text saying 'ping:' in the placeholder since it helped with testing but maybe that needs to be blank until the PlayerScript first sets the text.

I've tested this with one host and one client using the network simulation but no late joiners.

### DIFF
--- a/UnityProject/Assets/Prefabs/SceneConstruction/NestedManagers/UIManager.prefab
+++ b/UnityProject/Assets/Prefabs/SceneConstruction/NestedManagers/UIManager.prefab
@@ -611,6 +611,7 @@ MonoBehaviour:
   playerListUIControl: {fileID: 8587271803756561913}
   alertUI: {fileID: 5602923776831303726}
   toolTip: {fileID: 8587246508841551039}
+  pingDisplay: {fileID: 2039997815268504786}
   walkRunControl: {fileID: 8584883978471619497}
   storageHandler: {fileID: 8587659954310627797}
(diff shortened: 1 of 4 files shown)


fomalsd:
Hi! Thanks for contributing!
It works fine, just one subjective observation: wouldn't it be better if ping part would be fixed instead of moving when ping amount changes?
I mean this
GIF
vs this:
GIF1

I don't think ping is going to be more than several characters long, so I guess it could be fixed in place not too far off the edge

plnqux:
I've pushed a new commit to pad the string now and I also removed the string from the placeholder so it's blank on startup.  Did you want six character alignment?  I saw your gif was showing six as a sample and I tried that but the spacing seemed weird when the network simulator was showing numbers with three digits.  Five character alignment seemed a bit better but I guess any way it will look a bit weird since the text is right aligned.

fomalsd:
yeah, five is fine, thanks!

PR #2117, opened 2019-08-20
#2103 sync player up / down status on join.

### LINKED ISSUE #2103: Corpses appear standing up for late joiners
# Description
This was an old issue but was fixed, and I think a recent change has caused it to regress.
Probably

Reproduction steps:
1. Host in editor and kill yourself somehow (in the game). Respawn. Validate that your corpse appears laying down.
2. Join as client. Corpse appears standing up.

Expected behavior is that corpse appears laying down, just as it does for the server player.

chairbender:
### Purpose
Fixes #2103 

### DIFF
--- a/UnityProject/Assets/Scripts/Editor/Tools/Networking.cs
+++ b/UnityProject/Assets/Scripts/Editor/Tools/Networking.cs
@@ -87,7 +87,7 @@ public class Networking : Editor
 	{
 		if (CustomNetworkManager.Instance._isServer)
 		{
-			PlayerManager.LocalPlayerScript.playerHealth.Death();
+			PlayerManager.LocalPlayerScript.playerHealth.ApplyDamage(null, 99999f, AttackType.Internal, DamageType.Brute);
 		}
 	}
 
--- a/UnityProject/Assets/Scripts/Player/PlayerSync.Server.cs
+++ b/UnityProject/Assets/Scripts/Player/PlayerSync.Server.cs
@@ -264,6 +264,7 @@ public partial class PlayerSync
 	{
 		serverState.NoLerp = noLerp;
 		var msg = PlayerMoveMessage.Send(recipient, gameObject, serverState);
+		PlayerUprightMessage.Sync(recipient, gameObject);
 		Logger.LogTraceFormat("Sent {0}", Category.Movement, msg);
 	}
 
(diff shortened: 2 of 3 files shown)


UnityProject/Assets/Scripts/Player/PlayerSync.Server.cs
```diff
@@ -264,6 +264,7 @@ public void NotifyPlayer(GameObject recipient, bool noLerp = false)
 	{
 		serverState.NoLerp = noLerp;
 		var msg = PlayerMoveMessage.Send(recipient, gameObject, serverState);
+		PlayerUprightMessage.Sync(recipient, gameObject);
```
fomalsd, line 267: why not use
https://github.com/unitystation/unitystation/blob/0142d58a829c05cfc825b6dc0162b30c44c7db3c/UnityProject/Assets/Scripts/Managers/NetworkManagement/CustomNetworkManager.cs#L313
method instead?

UnityProject/Assets/Scripts/Editor/Tools/Networking.cs
```diff
@@ -87,7 +87,7 @@ private static void KillLocalPlayer()
 	{
 		if (CustomNetworkManager.Instance._isServer)
 		{
-			PlayerManager.LocalPlayerScript.playerHealth.Death();
+			PlayerManager.LocalPlayerScript.playerHealth.ApplyDamage(null, 99999f, AttackType.Internal, DamageType.Brute);
```
fomalsd, line 90: So is Death() the wrong way of killing creatures?
If so, maybe it should be made private

UnityProject/Assets/Scripts/Player/PlayerSync.Server.cs
chairbender: Not sure what you mean by "use"...maybe be more specific about the change you are suggesting. This method gets called from SyncPlayerData. Because this new logic has to do with an individual player's status it made sense to me to put it in playersync. I can't think of any situation where you'd want to sync PlayerSync without syncing the player's upright status as well so putting it all in this method prevents someone from forgetting to do both.

UnityProject/Assets/Scripts/Editor/Tools/Networking.cs
chairbender: Good point, will do

UnityProject/Assets/Scripts/Player/PlayerSync.Server.cs
fomalsd: ahh, sorry, didn't notice that it's called inside `NotifyPlayer`, not `NotifyPlayers`

PR #2123, opened 2019-08-24
Fix meta file casing

RedlineTriad:
### Purpose
Hopefully fixes some case-sensitive file system issues for Linux-users.

- [ ]  This PR does not include any unnecessary .meta, .prefab or .unity (scene) changes

### Notes:
The issue was that sometimes when hard resetting my git repository, it would give me `Case sensitivity refresh error`s (about 1000 and it did it every time I re-focused the unity editor, the only fix was to copy the repository to a case-sensitive file system open it with the editor and then moving it back to my case-insensitive shared partition.
I hope this fixes it but it is not guaranteed, I learned that case-sensitivity issues are hell hole I never want to enter again.


fomalsd (approved):
Doesn't seem to cause any issues on case-insensitive filesystems, at very least

PR #2121, opened 2019-08-24
Added hover tooltips for UI elements

plnqux:
### Purpose
Fixes #1464 by adding tooltips for most UI elements.

### Open Questions and Pre-Merge TODOs
- [x]  This fix is tested on the same branch it is PR'ed to.
- [x]  I correctly commented my code
- [x]  My code is indented with tabs and not spaces
- [x]  This PR does not include any unnecessary .meta, .prefab or .unity (scene) changes
- [x]  This PR does not bring up any new compile errors
- [x]  This PR has been tested in editor
- [x]  This PR has been tested in multiplayer (with 2 clients and late joiners, if applicable)

### Notes:
- Tooltips exist for bottom UI hud, ghost hud, and player health hud, and zoom/options hud
- Added generic UI_HoverTooltip class for hover tooltips
- Updated UI_ItemSlot to capture slot hover name and display item hovered
- Updated UI_ItemSwap to display hover name for slot
- Changed heart monitor image to be a raycast target since hovering is determined by raycast
- Removed OxygenButton component reference from buckled button

This PR needs to wait #2120 is merged and then I'll rebase my changes on that.  Is there a specific SS13 codebase I should use to test what things should look like?  I used tg for these changes and I believe I verified everything but "High Pressure" and "Freezing" since I wasn't able to get those to pop up.  Some of the Control*.cs classes used to have a namespace and their formatting is a bit weird right now.

PR #2124, opened 2019-08-24
Updated UI_OxygenAlert to use UpdateManager

plnqux:
### Purpose
Updated UI_OxygenAlert to use the UpdateManager to remove an instance of InvokeRepeating.

### Notes:
While working on #2121 I found that the UI_OxygenAlert was using InvokeRepeating and saw a way to make this class use the UpdateManager instead based on the UI_HeartMonitor class.  I'm not sure what all the cons are for using InvokeRepeating but one I thought of was that the CycleImg function could have been called when the O2 alert was not on screen.  I believe using OnEnable and OnDisable to register the UpdateMe function will prevent that.

### DIFF
--- a/UnityProject/Assets/Scripts/UI/UI_OxygenAlert.cs
+++ b/UnityProject/Assets/Scripts/UI/UI_OxygenAlert.cs
@@ -5,8 +5,9 @@ using UnityEngine.UI;
 
 public class UI_OxygenAlert : MonoBehaviour {
 
+	private const float SpriteCycleTime = 1f; // cycle every 1 second
 	public Sprite[] statusImages; //images to cycle between when active
-	private int activeImageIndex = 0;
+	private int nextImageIndex = 1;
 
 	public Image img;
 	private Sprite sprite;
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/UI/UI_OxygenAlert.cs
```diff
@@ -10,12 +10,36 @@ public class UI_OxygenAlert : MonoBehaviour {
...
+		{
+			UpdateManager.Instance.Remove(UpdateMe);
+		}
+	}
+
+	void UpdateMe()
+	{
+		timeWait += Time.deltaTime;
+		// Cycle images every 1 second
+		if (timeWait > 1f)
```
RedlineTriad: The value that `timeWait` is compared to should either be a constant or a serialized value.
It's easier to understand and modify if there are fewer magic numbers.

UnityProject/Assets/Scripts/UI/UI_OxygenAlert.cs
```diff
@@ -10,12 +10,36 @@ public class UI_OxygenAlert : MonoBehaviour {
...
+		{
+			CycleImg();
+			timeWait = 0f;
```
RedlineTriad: This would technically only be accurate with infinite FPS, because when we have, say 400ms per frame it would
```
timeWait = 0;
timeWait = 0.4;
timeWait = 0.8;
timeWait = 1.2 -> 0 // Cycle image called;
```
This means that that the lower the FPS the more inaccurate the cycle time is.
Should be changed to `timeWait -= CycleTime;`

RedlineTriad (changes requested):

UnityProject/Assets/Scripts/UI/UI_OxygenAlert.cs
plnqux: I've created a constant now.

UnityProject/Assets/Scripts/UI/UI_OxygenAlert.cs
plnqux: I've updated the code to subtract the new constant instead of setting timeWait to 0f.

plnqux:
I've pushed a commit that should fix the two issues identified.  I also saw some odd behavior where when the alert first showed up it was showing the first image for two cycles.  I've changed the activeImageIndex to nextImageIndex to reflect how it works now and set it to 1 so it chooses the next image instead of using the same image for the first cycle.  I also added a reset function so the alert is enabled in the same state every time.

RedlineTriad (approved):
I would merge this however I am currently very scared of switching branches due to some Git/Linux/Unity issues, so I can't test the code running in unity.

fomalsd (commented):
it works, but: it can be simpler if you take advantage of our `array.Wrap()` [extension method](https://github.com/unitystation/unitystation/blob/ce52efc6c589888101d389b2c50bc1cde53f317e/UnityProject/Assets/Scripts/Util/SweetExtensions.cs#L65), also making local `sprite` field redundant. 
see for yourself: https://gist.github.com/fomalsd/f662a3995805bb44cf2e1f65327f24f1

plnqux:
That gist looks good to me.  I had one change to that gist to still assign nextImageIndex to 1 when resetting the cycle as it could pop up the alert and display the first sprite twice without it.

fomalsd (approved):

PR #2125, opened 2019-08-24
Fixed hiding tabs to rollup the window

### LINKED ISSUE #1750: Closed tab should minimize tab holder if it was minimized before opening tab
# Description
image
While the window is no longer related to the APC and it turns into the player list, it should close instead of staying open (unless you had it open before interacting with the APC?)

## Category
BUG

## Steps to Reproduce
1.Open an APC window
2.Walk away from it
3.Notice how the screen turns into the player list instead of closing

plnqux:
### Purpose
Fixes #1750, hides the tab window if the tab that was selected is being hidden.

### Notes:

- Fixed Item List tab to rollout tab window if not visible
- Changed to not always select the Stats tab, only select it if the tab selected is being hidden

This fix does not remember the state of the window or the tab that was selected before opening a contextual tab so it could still be improved.  It does handle when a user has a different tab open walking away from a object though.

### DIFF
--- a/UnityProject/Assets/Scripts/UI/ControlTabs.cs
+++ b/UnityProject/Assets/Scripts/UI/ControlTabs.cs
@@ -356,9 +356,18 @@ public class ControlTabs : MonoBehaviour
 	private void HideTab(Tab tab)
 	{
 		tab.Hidden = true;
+		bool wasActive = tab.gameObject.activeSelf;
 		tab.gameObject.SetActive(false);
 		RefreshTabHeaders();
-		SelectTab(ClientTabType.Stats, false);
+		if(wasActive)
+		{
+			SelectTab(ClientTabType.Stats, false);
+			var netTab = tab as NetTab;
+			if (rolledOut && (netTab == null || !netTab.isPopOut))
+			{
+				ToggleTabRollOut();
+			}
+		}
 	}
 
 	private void HideTab(ClientTabType type)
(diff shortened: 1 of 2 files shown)


fomalsd (changes requested):
it can occasionally fail:
GIF

plnqux:
I saw that during my testing, I think it's due to the window being rolled out when it gets a request to rollup.  I'll have to figure out a way to queue it up somehow.

fomalsd:
maybe swap boolean isRolledOut logic with something like RollOutStatus enum {Hidden, RollingOut, Shown, RollingUp}
would be easier to track and process them this way

plnqux:
I've rebased on develop and pushed some changes to handle this now by adding new methods to open/close the window.  These new methods wait for any current animation to finish.  I've changed the location where the rolledOut variable is updated to signal that the window is open or will be open.

One of my reasons for changing the rolledOut variable logic was for thread safety since I didn't realize Coroutines were sequentially executed.  I don't think there will be any side effects due to this but there could be odd behaviors if multiple coroutines are waiting to act.

fomalsd (commented):
AnimTabRoll code looks confusing (hard to understand what's going on), but it seems to work
Still think that it would be cleaner with some kind of TabRollState and UnityEvent listeners to track rollout state and fire desired action when it's done moving
To be fair whole tab rollout thing needs a refactor, currently it's wonky and is almost always a few pixels off because it's not guaranteed to stop at exact position, but it's low priority
Merging this nevertheless

PR #2140, opened 2019-09-07
Object lifecycle hooks, Gun and MagazineBehavior fixes

### LINKED ISSUE #2070: Object Lifecycle - Consistent Spawn and despawn logic / hooks
# Description
Currently many items which are pooled will not reset to their correct initial state when spawned because their state is initialized only in Start / Awake or in their initial settings. 

Due to us using a custom object pooling technique, the standard Unity lifecycle methods like Start / Awake are not sufficient.

We should create a method that is invoked when the object is spawned regardless of if it's via the pool, via Object.Instantiate, or scene loading. Then initialization logic for each component should be moved there. For consistency, we should also have a method that is called when an object is going to be destroyed or despawned. Make sure its possible to have the logic distinguish between client and server. 

As a way to more easily test this, also add a way for a dev to despawn and respawn a given object immediately as if it was taken from the pool, so it can be obvious if it's state isn't being reset. 

Overall we should just have our "lifecycle" methods for our objects which is separate from Unity's lifecycle (due to our custom object pooling), and should document how it can be hooked into.

chairbender:
### Purpose
Fixes #2070 

Added wiki page on this info: https://github.com/unitystation/unitystation/wiki/Creating-Pool-Safe-and-Cloneable-Components

Also added a wiki page to be a checklist for working with components: https://github.com/unitystation/unitystation/wiki/Component-Development-Checklist

Implemented (hopefully) proper lifecycle logic for Integrity, Gun, HealsTheLiving, ElectricalModuleInheritance classes, 

Also fixes some stuff and refactors Gun, I noticed some syncing issues and also figured out how to get rid of the unreliable / bandaid usage of WaitForSeconds there.


fomalsd:
Useful stuff! 
Are all things spawned in runtime via spawner pooled? (thinking of objects shuttle consoles, not items)
For example, I've added some init logic for shuttle consoles recently, so that they would work when spawned via spawner, or built (when we add support). There are still a few things left to consider when spawning/destroying though, so would it be ok to move that logic over to your system even if I'm not sure that consoles are pooled?

chairbender:
Basically everything that is spawned during the game, no matter if it's pooled or not, goes through the same custom lifecycle hooks, because everything is using PoolManager to spawn stuff (even non-pooled stuff) which is where those hooks are kicked off. That includes dev spawner. Includes even things like machines, etc...there's basically nowhere in the game that bypasses PoolManager.

What you are asking would be ok. In fact, I recommend every component should implement spawn hooks (typically just IOnStageServer) even if it is not pooled currently. But it's true that things which aren't pooled technically wouldn't need to implement those hooks as long as they have good init logic in Start / Awake etc, since there is no danger of them being reused. I will add a note about that.

chairbender:
I should add, as far as moving your init logic into this system...you still should have that init logic in the Unity init methods, so that the console will properly initialize when the scene loads. If you want to use my system, you should ALSO add that logic to the OnStageServer hook.

chairbender:
See my update on the wiki to clarify:
**What lifecycle methods must you implement?**
1. All components should put their init logic in the normal Unity methods (`Start`, `Awake`, etc...) regardless of the situation.
1. If your component is going to be on pooled objects, you should implement init logic using `IOnStageServer` / `IOnStageClient` / `IOffStageServer` so it can be safely reused. Non pooled objects can ignore these.

chairbender:
gonna edit the wiki to make it more about object lifecycle in general and clarify some of this stuff

chairbender:
ok sorry for the spam, @fomalsd I've updated the wiki to make it more clear and clarify my own thinking about this: https://github.com/unitystation/unitystation/wiki/Object-Lifecycle-Best-Practices

If there is no chance that consoles will ever be pooled, then you don't need to worry about implementing the new custom lifecycle methods.

EDIT: And yes, all things spawned during runtime use PoolManager, even if they aren't pooled

DooblyNoobly:
Introduces a NetID bug (this is on the client in the editor connecting to a host in a build): 
image

Not sure what is causing it as UNet errors are horrible to debug

DooblyNoobly:
The Unknown Message ID error is not apparent when joining as a client from a build but there is a SyncList error:

image

fomalsd:
are you sure that SyncList error is not one that's many months old?

DooblyNoobly:
Will check on develop

fomalsd:
Also "Unknown message id" are known to happen when client isn't fully loaded yet while already getting some messages from server (1012 is PlayerMoveMessage, just serverplayer running around station in your case), they're not introduced by this PR

DooblyNoobly:
Having major problems trying to get a build working on develop because my filesystem is being stubborn. Will check it out tomorrow

UnityProject/Assets/Scripts/Managers/PoolManager.cs
```diff
@@ -155,132 +260,54 @@ public static GameObject NetworkClone(GameObject toClone, Vector3? position = nu
...
-	/// usually take care of properly parenting themselves when spawned so in many cases you can leave it null.</param>
-	/// <returns>the newly created GameObject</returns>
-	public static GameObject PoolClientInstantiate(GameObject prefab, Vector3? position = null, Transform parent = null, Quaternion? rotation = null)
-	{
-		if (!IsInstanceInit())
+		var objBehavior = target.GetComponent<ObjectBehaviour>();
+		if (objBehavior == null)
 		{
-			return null;
+			Logger.LogErrorFormat("{0} has no ObjectBehavior thus cannot be pooled.", Category.ItemSpawn, target.name);
```
fomalsd:
```suggestion
			Logger.LogErrorFormat("{0} has no ObjectBehavior thus cannot be pooled.", Category.ItemSpawn, target.name);
			return null;
```
Tried to respawn something that isn't poolable and got an NRE that broke right click menu, I think it's best to avoid them if possible
Scr 2019-09-08 в 10 26 42

```
PoolManager.PoolNetworkTestDestroyInstantiate (UnityEngine.GameObject target) (at Assets/Scripts/Managers/PoolManager.cs:282)
PlayerNetworkActions.CmdAdminRespawn (UnityEngine.GameObject toRespawn) (at Assets/Scripts/Player/PlayerNetworkActions.cs:750)
PlayerNetworkActions.CallCmdAdminRespawn (UnityEngine.GameObject toRespawn) (at <18a947598b4544ef8b655e0d3501b528>:0)
CustomNetTransform.AdminRespawn () (at Assets/Scripts/Transform/CustomNetTransform.cs:691)
RadialMenu.Update () (at Assets/Scripts/UI/Right click/RadialMenu.cs:225)
```

chairbender:
I am quite sure the synclist error is many months old, I've been getting it constantly.

Will look into the others

UnityProject/Assets/Scripts/Managers/PoolManager.cs
chairbender: I actually want this to work as well rather than doing nothing. I'll add logic so if it's not pooled, it will respawn using the non-pooled approach when you select that option.

chairbender:
Hm, I wrote that only some objects use object pooling, but from what I'm seeing almost all objects use object pooling. I need to update the wiki on that.

chairbender:
@fomalsd updated wiki with my findings. Basically found that all objects use pooling (unless they have no ObjBehavior). So pretty much all objects should implement the new hooks I added (though I think most consoles don't have object behavior).

Added better instructions for thoroughly testing if objects can be spawned, despawned, and reused.

And pushed a fix so that the respawn thing still works even when object is not pooled. Should be ready to look at again

DooblyNoobly:
I'm getting One ID error report on steam client, so looks like its unrelated to this PR. But this PR could be exacerbating it:
image

fomalsd:
It's an old thing, I get it like every time when I test multiplayer

DooblyNoobly:
yeah I also got synclist error when restarting a round from calling the emergency shuttle

fomalsd:
Thanks, Respawn is really handy for debugging, this is good to merge!

fomalsd (approved):

PR #2167, opened 2019-09-22
Moses/2137 body bags

### LINKED ISSUE #2137: TODO: Body bags B:70
# Description
- Dead bodies can't just be left around on the station to rot. Put them in body bags!

## Criteria:

- Create the Bodybag prefab and add a bunch of body bags in one of the closets in Medbay
- Add the ability to open and close body bags
- Add the ability to put a body into a body bag
- Hide the body when the body bag has been zipped up
- Allow body bag to be dragged around on the station
- Make sure all actions work of multiplayer and test in a mp game with 2 clients

Contributor will receive a bounty of 70usd on a successful merge of a solution that satisfies the criteria above.

StrawberryMoses:
### Purpose
Fixes #2137. Adds Body Bags and Bluespace Body Bags (which is normally a research item). Body bags essentially function as placeable closets, with one prefab of them folded and one open. 

### Notes:
I did not add any body bags to the scene, as someone else is doing scene work at the moment.
This is my first major feature, so another set of testable hands would be appreciated.

### DIFF
--- a/UnityProject/Assets/Scripts/Medical/BodyBag.cs
+++ b/UnityProject/Assets/Scripts/Medical/BodyBag.cs
@@ -1,8 +1,13 @@
 using UnityEngine;
 
-public class BodyBag : Interactable<MouseDrop>
+public class BodyBag : Interactable<MouseDrop>, IOnStageServer, IRightClickable
 {
 	public GameObject prefabVariant;
+
+	public void GoingOnStageServer(OnStageInfo info)
+	{
+		GetComponent<ClosetControl>().ToggleLocker(false);
+	}
 
 	protected override bool WillInteract(MouseDrop interaction, NetworkSide side)
 	{
@@ -11,12 +16,16 @@ public class BodyBag : Interactable<MouseDrop>
 			return false;
 		}
 
+		var cnt = GetComponent<CustomNetTransform>();
+		var ps = interaction.Performer.GetComponent<PlayerScript>();
+
 		var pna = interaction.Performer.GetComponent<PlayerNetworkActions>();
 
 		if (!pna
 			|| interaction.Performer != interaction.TargetObject
 		    || interaction.DroppedObject != gameObject
-			|| pna.GetActiveHandItem() != null)
+			|| pna.GetActiveHandItem() != null
+			|| !ps.IsInReach(cnt.RegisterTile, side == NetworkSide.Server))
 		{
 			return false;
 		}
@@ -31,15 +40,15 @@ public class BodyBag : Interactable<MouseDrop>
 		var closetControl = GetComponent<ClosetControl>();
 		if (!closetControl.IsClosed)
 		{
-			ChatRelay.Instance.AddToChatLogClient("You wrestle with the body bag, but it won't fold while unzipped.",
-				ChatChannel.Examine);
+			UpdateChatMessage.Send(interaction.Performer, ChatChannel.Examine,
+				"You wrestle with the body bag, but it won't fold while unzipped.");
 			return;
 		}
 
 		if (!closetControl.IsEmpty())
 		{
-			ChatRelay.Instance.AddToChatLogClient("There are too many things inside of the body bag to fold it up!",
-				ChatChannel.Examine);
+			UpdateChatMessage.Send(interaction.Performer, ChatChannel.Examine,
+				"There are too many things inside of the body bag to fold it up!");
 			return;
 		}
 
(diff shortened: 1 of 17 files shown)


Bod9001:
you can do 
`#2137`
to get github to link it automatically to the issue
#2137

Bod9001:
>I did not add any body bags to the scene, as someone else is doing scene work at the moment.

You can add them to locker contents? That's a way of bypassing not Editing the scene?

StrawberryMoses:
Added to the CMO lockers

fomalsd:
looks good in editor, the only thing that looks a bit strange is predictive placement that isn't tile snapped and gets corrected later, I think it's better off without any prediction.
that and maybe there's special zipper sound for body bags? atm they sound like lockers

astralenigma:
There's a "zip" sound file in the code.

UnityProject/Assets/Scripts/Medical/BodyBag.cs
```diff
@@ -0,0 +1,48 @@
+using Mirror;
+using UnityEngine;
+
+public class BodyBag : NetworkBehaviour, IInteractable<MouseDrop>
```
fomalsd: wait, it looks like you're running this code on client side only, like this:
https://github.com/unitystation/unitystation/wiki/Implementing-Interactions#i-dont-need-any-networking-i-just-want-client-side-interaction-logic
it's only suitable if you call Cmd or send a special message (a niche use case).
I think you should extend `Interactable<MouseDrop>` here instead, move validation part to to `WillInteract` and call TryFoldUp within `ServerPerformInteraction`.

fomalsd (changes requested):
Folding code won't work on client, please move it to serverside

UnityProject/Assets/Scripts/Medical/BodyBag.cs
fomalsd: actually I fell in the same trap the first time I made an item using IF2

StrawberryMoses: Updated to use this, need to test and will push up in new commit. Will remember this for next time :)

StrawberryMoses:
@fomalsd Hm, I didn't add anything with prediction- is this an issue with other objects being placed? I am mostly noticing this with body bags that were previously placed being placed down again, so I'm going to take a look at how it gets destroyed and placed back into the pool. Maybe straight up destroying it is needed, but we'll see. Do you have any breadcrumbs that might guide me in the right direction on this? Were you only noticing it with re-placed body bags or fresh ones as well? (I was always spawning folded body bags with admin controls when I was testing)

fomalsd:
I had this when placing freshly spawned folded ones (didn't know they could be folded back then). They appeared on clicked position for a frame (if placed within interaction range), then blinked and snapped to nearest tile center. Don't really know what exactly is causing this without digging, but might be able to take a closer look later

StrawberryMoses:
Without being able to test until later (and having no real idea of what I'm talking about)-

`// Place the opened body bag in the world
		PoolManager.PoolNetworkInstantiate(prefabVariant, interaction.WorldPositionTarget);`

My guess is that since I'm spawning it at just a world position (where the mouse is clicked) it needs to move the body bag to the center of the tile? idk I'll play later

fomalsd:
yeah, you could do `interaction.WorldPositionTarget.RoundToInt()` instead and it should fix it

UnityProject/Assets/Scripts/Medical/BodyBag.cs
StrawberryMoses: Pushed up!

StrawberryMoses:
Yep, RoundToInt fixed it.
Also added zip sound effect.
Additionally now just destroying the object when placing into inventory, pooling it was causing weird behaviors.

fomalsd:
What kind of weird behaviour are you talking about? 
Absolute majority of objects (including yours) are poolable and should be destroyed using `PoolNetworkDestroy`, simple `Destroy` is just for clientside GameObjects as far as I know. Check this link:
https://github.com/unitystation/unitystation/wiki/Object-Lifecycle-Best-Practices
If there are any bugs then they should be looked at and fixed, because there are many other situations when `PoolNetworkDestroy` will be called on it anyway (like, it burning in a fire)
Most likely you just have to implement `IOnStageServer` and reset bag state to default in `GoingOnStageServer`.
You can test it by right clicking bag and choosing Respawn

StrawberryMoses:
Alright- will try later. I was just seeing some weird stuff with bags acting like they were on a different tile

fomalsd:
add a TransformGizmoDrawer component to bags and enable gizmos when you get this weird stuff again – colored boxes will help determine what’s going on

StrawberryMoses:
Added a parent to the pool instantiate and now it works. Added a right click menu option. Fixed issue where the messages on folding would be sent to the server player.

fomalsd (approved):
Works really well! Thanks for taking time to polish it

PR #2181, opened 2019-09-29
Added Paper Bins

StrawberryMoses:
### Purpose
Starting to add pointless bureaucracy to the station, starting with Paper Bins. Paper bins can hold an unlimited amount of paper, dispensing user placed paper first as a stack. Can also hold a single pen on top. Can be picked up on right click.

### Notes:
I did not implement drag to pick up. It seems with the radial right click menu, it might be more standardized than having a few different behaviors depending on the item. I wanted to add more bureaucratic stuff, but I wasn't sure if I was making questionable design decisions with this or not, so wanted to have this in before having to debug a bunch of different features.

### DIFF
--- a/UnityProject/Assets/Scripts/Input System/InteractionV2/CoreComponents/InteractionComponentUtils.cs
+++ b/UnityProject/Assets/Scripts/Input System/InteractionV2/CoreComponents/InteractionComponentUtils.cs
@@ -1,5 +1,6 @@
 
 using System;
+using UnityEngine;
 
 /// <summary>
 /// Utilities used in IF2 for various core interaction components
@@ -35,10 +36,10 @@ public static class InteractionComponentUtils
 	/// <returns>If validation succeeds and coordinator sends the interaciton msg, invokes
 	/// onValidationSuccess and returns STOP_PROCESSING. Otherwise returns CONTINUE_PROCESSING.</returns>
 	public static bool CoordinatedInteract<T>(T info, InteractionCoordinator<T> coordinator,
-		Action<T> onValidationSuccess)
+		Action<T> onValidationSuccess, string specificComponent = null)
 		where T : Interaction
 	{
-		var validated = coordinator.ClientValidateAndRequest(info);
+		var validated = coordinator.ClientValidateAndRequest(info, specificComponent);
 		if (validated)
 		{
 			//success, so do client prediction if not server player
--- a/UnityProject/Assets/Scripts/Items/Bureaucracy/PaperBin.cs
+++ b/UnityProject/Assets/Scripts/Items/Bureaucracy/PaperBin.cs
@@ -212,8 +212,8 @@ namespace Items.Bureaucracy
 			// First, get the papers that players have put in the bin
 			if (storedPaper.Count > 0)
 			{
-				paper = storedPaper[0];
-				storedPaper.RemoveAt(0);
+				paper = storedPaper[storedPaper.Count - 1];
+				storedPaper.RemoveAt(storedPaper.Count - 1);
 				SyncBlankPaperCount(blankPaperCount);
 			}
 			else // Otherwise, take from blank paper stash
--- a/UnityProject/Assets/Scripts/Items/Bureaucracy/PaperBin.cs
+++ b/UnityProject/Assets/Scripts/Items/Bureaucracy/PaperBin.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using Mirror;
 using UnityEngine;
 
@@ -10,14 +11,13 @@ namespace Items.Bureaucracy
 		public int initialPaperCount = 20;
 		public GameObject blankPaper;
 
-		[SyncVar (hook = nameof(SyncBlankPaperCount))]
-		private int blankPaperCount;
-
-		private readonly SyncListPaper storedPaper = new SyncListPaper();
+		[SyncVar (hook = nameof(SyncPaperCount))]
+		private int paperCount;
 
 		[SyncVar (hook = nameof(SyncStoredPen))]
 		private GameObject storedPen;
 
+		private List<GameObject> storedPaper = new List<GameObject>();
 		private SpriteRenderer binRenderer;
 		private SpriteRenderer penRenderer;
 		private Sprite binEmpty;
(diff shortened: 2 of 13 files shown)


DooblyNoobly (changes requested):
Initial paper stack isn't being added to the tray on round start (this is as host in the editor):
image

UnityProject/Assets/Scripts/Items/Pickupable.cs
```diff
@@ -145,7 +145,7 @@ public RightClickableResult GenerateRightClickOptions()
 	private void RightClickInteract()
 	{
 		//trigger the interaction manually, triggered via the right click menu
-		Interact(HandApply.ByLocalPlayer(gameObject));
+		Interact(HandApply.ByLocalPlayer(gameObject), "Pickupable");
```
fomalsd:
```suggestion
		Interact( HandApply.ByLocalPlayer(gameObject), nameof(Pickupable) );
```

UnityProject/Assets/Scripts/Items/Bureaucracy/PaperBin.cs
```diff
@@ -0,0 +1,250 @@
...
+	public class PaperBin : NBHandApplyInteractable, IOnStageServer
+	{
+
+		public int initialPaperCount = 20;
+		public GameObject blankPaper;
+
+		[SyncVar (hook = nameof(SyncBlankPaperCount))]
+		private int blankPaperCount;
+
+		private readonly SyncListPaper storedPaper = new SyncListPaper();
```
fomalsd: Why should all clients know contents of each paper bin on the server? (syncvars are public for all clients)
Look at how closet keeps its contents secret from player when it's closed. 
Syncvars are okay for visual appearance, but not for hidden stuff

fomalsd (changes requested):
~~For some reason I get loads of exceptions on this branch, restarting Unity doesn't help~~

fomalsd:
woops, sorry, exceptions aren't present on different machine

UnityProject/Assets/Scripts/Input System/InteractionV2/CoreComponents/InteractionComponentUtils.cs
```diff
@@ -35,10 +36,10 @@ public static bool ServerProcessCoordinatedInteraction<T>(T info, InteractionCoo
 	/// <returns>If validation succeeds and coordinator sends the interaciton msg, invokes
 	/// onValidationSuccess and returns STOP_PROCESSING. Otherwise returns CONTINUE_PROCESSING.</returns>
 	public static bool CoordinatedInteract<T>(T info, InteractionCoordinator<T> coordinator,
-		Action<T> onValidationSuccess)
+		Action<T> onValidationSuccess, string specificComponent = null)
```
chairbender, line 39: I notice you modified the core interaction classes, adding a string parameter to identify a specific component type. This shouldn't be necessary (and if it was, we should use an ID uint instead of a string so the message size doesn't become needlessly large), and this was a feature in place for security reasons as well - we don't want clients to be able to decide which component should fire on the server, otherwise clients would be able to have control over server logic such as interacting with things in certain ways when it should technically not be possible. 

What was the issue you were having that required this? If there are multiple components on the object which handle HandApply, you can edit the order in which they are checked by rearranging those components on the prefab. You should most likely have PaperBin above whatever other components it was conflicting with. You may want to revisit https://github.com/unitystation/unitystation/wiki/Implementing-Interactions

StrawberryMoses: This was the questionable decision I was worried about. This came about specifically because of the right click menu. The paper bin has the two handapply interactions- first being the new functionality by picking up and placing paper, and the second being Pickupable. When using the right click menu and selecting pick up, going through the list of hand applys will incorrectly select the paper bin functionality instead of picking it up. 
This will likely be a problem for some other objects in the future, so I'm open for any suggestions on what we can do about this

chairbender: Oh, I think you only added this so you could fix the right click Pickup interaction. It should be fixed in another way, the way right click pickup was implemented is not viable as you discovered (it should directly invoke the interaction logic rather than simply creating a HandApply interaction).

@fomalsd I'm open to just leaving this as is but we should create a bug so I can figure out a better way to achieve this. I can't think of a good way to do address this right now off the top of my head.

EDIT: And security isn't really a huge concern right now with this, I can't think of a situation when there would be multiple handapply components on a single object and a client would be able to cheat by calling the "wrong one", but my fix for this will ensure that this will not be a security issue. I'll probably add capability to the right click menu to have server-side logic.

chairbender: ah, I only saw your message after I posted mine. Yep, it will be a problem for other objects so I'll just make a bug to fix it and we can leave it as is.

fomalsd:
Ok, since `specificComponent` is acceptable as a temp solution, the only thing that worries me here is extra info (synclist of actual paper sheets) shared with all clients.
While count syncvars are ok (syncing a visual state enum like Empty/Half/Full would be best though),
I don't get why it's needed to broadcast all contents – not like they're of any immediate use to clients? 
I feel like the best solution would be to compose examine message on serverside and send it back to player, while keeping its contents secret and only broadcasting its visual state. 
However I just checked and looks like we don't have any convenience code for requesting server-generated examine strings yet, so for the time being two SyncVars – paper count and pen presence – should be enough for clientside inspection text. What do you think?

StrawberryMoses:
Sounds good to me. I'll refactor to use only the 2 syncvars.
As for the reason I did the sync list... didn't know any better :)

UnityProject/Assets/Scripts/Items/Pickupable.cs
StrawberryMoses: Done

UnityProject/Assets/Scripts/Items/Bureaucracy/PaperBin.cs
StrawberryMoses: No longer being synced

UnityProject/Assets/Scripts/Input System/InteractionV2/CoreComponents/InteractionComponentUtils.cs
StrawberryMoses: Issue is: https://github.com/unitystation/unitystation/issues/2188

StrawberryMoses:
Fixed to fill paper tray at round start. No longer syncing edited paper.

fomalsd (approved):

PR #2193, opened 2019-10-05
added shuttle fuel system and cleaned up a few  sprites

### LINKED ISSUE #1329: TODO: Shuttle fuel system B: 70
### Now that #1327  #1328 has been completed this job can begin.

# Criteria

### The shuttles can only operate if there is sufficient fuel available:
 
- Monitor fuel levels before and during space flight
- Stop the shuttle if fuel runs out
- Update Fuel Meter GUI to reflect current fuel levels
- Subtract plasma gas from the canister while flying the shuttles
- Test in multiplayer with 2 clients

-------------------
### Suggested approach:

Scripting:
_We need a component that can be added to the parent of the shuttle matrix that can monitor and control the fuel levels in the shuttle_
1) Create a new script and name it ShuttleFuelSystem.cs and place it in the Assets/Scripts/Shuttles/ folder
2) Derive the script from *ManagedNetworkBehaviour and add the UnityEngine.Networking using directive

_*ManagerNetworkBehaviour has a special update override that is cheaper to run then MonoBehaviour.Update() and it also derives from NetworkBehaviour. Make sure that you call the base.OnEnable and base.OnDisable methods if you choose to override them in your script (Or else UpdateMe() won't work)_

3) Create some fields that we will use to make the fuel system work:
  - Add a Private MatrixMove field (Get the component to assign to this field in Awake())
  - Add a Public AtmosConnectorSlot field ( drag the atmos connector slot for this shuttle to this field in via the inspector (Don't forget to save the scene after you do this!))
  - Add a Public Float named FuelLevel (we will need to sync this variable with the one from the server. You need to determine how you will do that, you can use a NetworkMessage and send every second or on every event or you can turn the FuelLevel field into a SyncVar and create a hook for it and only assign a value to it on the server every second.

4) Open the MatrixMove.cs script and add a field for the ShuttleFuelSystem. 
5) On the MatrixMove script add some conditions so that flight cannot start if the FuelLevel from the ShuttleFuelSystem component is 0. Also monitor the levels so that if the level ever reaches 0 in flight then stop the movement
6) On the GUI_ShuttleScript you now have access to the ShuttleFuelSystem through the MatrixMove reference so use the FuelLevel data to update the FuelLevel meter that was created in #1327 
7) On the ShuttleFuelSystem script create the logic to monitor the gas level in the connected Plasma Tank. You do this by having the AtmosConnectorSlot.cs component reference the connected Plasma Canister (Canister script is being completed currently by @krille90  and is coming in the Atmos merge soon)
8) Do all your calculations in the UpdateMe override (You need to override it: `public void override UpdateMe()` <-- This now replaces your Update() method and is a lot cheaper on the CPU)
9) Add some validation checks to make sure the gas in the canister has enough plasma
10) Start a game, connect a plasma tank and try to fly the shuttle!

Contributor will receive a bounty of 70usd on a successful merge of a solution that satisfies the criteria above.

Bod9001:
### Purpose
Fixes #1329
Adds connector for plasma tank for the Shuttle Fuel
shuttles now require fuel Pacificly plasma
Can be optimised with a oxygen mix
bool Override for running without Fuel, Default on
Moves over a few sprites that were still on the old DMI Sprite sheets


Bod9001:
Currently got a problem when I wrench down the canisters and fly off the canister stay where they are?
Not sure if it's a matrix issue Since a going onto the Command escape shuttle  your Character Would think that they are in space for second and take damageThat might be related?

krille90:
@Bod9001 I presume you couldn't fix that yet? Putting a DNM label on it for now

Bod9001:
oh sorry, I did fix it , Should have been more clear in the last push xD
It was a mapping Issue where the station was underneath the shuttle, and the canister was sticking to the station instead of the shuttle @krille90

fomalsd:
I think that matrix bug can be easily fixed by looping matrices in reverse order. at the moment station matrix is always at index 0 therefore is always picked first when there are multiple tiles on the same world coordinates

UnityProject/Assets/Scripts/Objects/Canister.cs
```diff
@@ -128,6 +137,20 @@ protected override void ServerPerformInteraction(HandApply interaction)
 						return;
 					}
 				}
+				var foundeFuelConnectors = MatrixManager.GetAt<ShuttleFuelConnector>(registerTile.WorldPositionServer, true);
+
+				foreach (var bob in foundeFuelConnectors)
+				{
+					var conn = bob;
```
fomalsd: - there's a typo in `foundeFuelConnectors`
- I think canister should look for connectors on its own matrix only, it doesn't make much sense to look for them on others: `registerTile.Matrix.Get<ShuttleFuelConnector>(registerTile.LocalPositionServer, true);`
- why an extra var inside foreach? why call var `bob`?

UnityProject/Assets/Scripts/Objects/Canister.cs
```diff
@@ -128,6 +137,20 @@ protected override void ServerPerformInteraction(HandApply interaction)
...
+					SoundManager.PlayNetworkedAtPos("Wrench", registerTile.WorldPositionServer, 1f);
+					isConnected = true;
+					connectorFuel = bob;
+					conn.ConnectCanister(this);
+					SetConnectedSprite(connectorSprite);
+					objectBehaviour.isNotPushable = true;
+					ServerOnConnectionStatusChange.Invoke(true);
+					return;
```
fomalsd: why use `return` in a foreach loop? can't you just do all things to the first one in the list instead?
`return` in the previous Connector loop made sense because there was an `if` clause inside

UnityProject/Assets/Scripts/Shuttles/MatrixMove.cs
```diff
@@ -358,15 +372,18 @@ public void ToggleMovement()
 	[Server]
 	public void StartMovement()
 	{
-		//Setting speed if there is none
-		if (serverTargetState.Speed <= 0)
+		if (IsFueled || !RequiresFuel)
```
fomalsd, line 375: does that mean that if fuel runs out while shuttle is still moving it won't stop?

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/GasMix.cs
```diff
@@ -133,6 +133,7 @@ public float GetMoles(Gas gas)
 		}
 
 		public void ChangeVolumeValue(float value){
+			var TemperatureCash = Temperature;
```
fomalsd: you know, this variable is local so its results are never used. that "cash" thing will never work here this way
if you used a proper IDE it would have warned you about that:
img
instead of local variables you should've made a `TemperatureCache` field in this class

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/GasMix.cs
```diff
@@ -179,16 +180,25 @@ public void MergeGasMix(GasMix otherGas)
 
 		public void AddGas(Gas gas, float moles)
 		{
+			var TemperatureCash = Temperature;
```
fomalsd: use field instead of local variable. then you don't need to pass it to RecalculateTemperatureCache() method

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/GasMix.cs
```diff
@@ -179,16 +180,25 @@ public void MergeGasMix(GasMix otherGas)
...
 			Gases[gas] += moles;
 
-			Recalculate();
+			RecalculateTemCash(TemperatureCash);
 		}
 
 		public void RemoveGas(Gas gas, float moles)
 		{
+			var TemperatureCash = Temperature;
```
fomalsd: same here

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/GasMix.cs
```diff
@@ -179,16 +180,25 @@ public void MergeGasMix(GasMix otherGas)
...
 			Gases[gas] -= moles;
+			RecalculateTemCash(TemperatureCash);
+		}
 
-			Recalculate();
+		public GasMix RemoveGasReturn(Gas gas, float moles)
+		{
+			var TemperatureCash = Temperature;
```
fomalsd: same here

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/GasMix.cs
```diff
@@ -211,5 +221,12 @@ private void Recalculate()
 		{
 			Pressure = AtmosUtils.CalcPressure(Volume, Moles, Temperature);
 		}
+
+
+		//Used to change the pressure instead of temperature when removing/Adding gas
+		private void RecalculateTemCash(float TemperatureCash )
```
fomalsd: please rename to `RecalculateTemperatureCache( float TemperatureCache )`

UnityProject/Assets/Scripts/UI/Shuttles/GUI_ShuttleControl.cs
```diff
@@ -174,56 +196,68 @@ private void PowerOff()
...
+		if (State == TabState.Off)
 		{
 			yield break;
 		}
 		EntryList.RefreshTrackedPos();
-		yield return WaitFor.Seconds( 2f );
+		//Logger.Log((MatrixMove.shuttleFuelSystem.FuelLevel * 100).ToString());
+		if (MatrixMove.shuttleFuelSystem == null)
+		{
+			this["FuelGauge"].SetValue = (100).ToString();
```
fomalsd: not really efficient to send this value again and again each couple of seconds even if it's never changed, set it once on start and forget about it

UnityProject/Assets/Scripts/Shuttles/ShuttleFuelSystem.cs
```diff
@@ -0,0 +1,132 @@
...
+
+		var CMassconsumption = 1 / Ratio;
+		if (CMassconsumption > 6)
+		{
+			CMassconsumption = 6;
+		}
+		//Logger.Log("Ratio3 > " + Ratio);
+		CalculatedMassConsumption = (CMassconsumption * OptimumMassConsumption * FuelConsumption);
+
+		if ((Plasma > (CalculatedMassConsumption)  * (7 / 10)) && (Oxygen > (CalculatedMassConsumption)  * (3 / 10)))
```
fomalsd: use 0.7f and 0.3f in this file, diving 7 and 10 **integers** results in 0, see for yourself:
https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYCMBYAUHjAZgAIBvPYy4jdMiqh1TATgApiB2AekwAZiAlAG56lAL54xQA=
again, proper IDEs warn you about that but it's actually basic theory

UnityProject/Assets/Scripts/Shuttles/ShuttleFuelSystem.cs
```diff
@@ -0,0 +1,132 @@
...
+
+	}
+
+	void FuelCalculations()
+	{
+		if (IsFuelledOptimum())
+		{
+			//Logger.Log("CalculatedMassConsumption > " + CalculatedMassConsumption*massConsumption);
+			Connector.canister.container.GasMix = Connector.canister.container.GasMix.RemoveGasReturn(Gas.Plasma, CalculatedMassConsumption * massConsumption  * (7f / 10f));
+			Connector.canister.container.GasMix = Connector.canister.container.GasMix.RemoveGasReturn(Gas.Oxygen, CalculatedMassConsumption * massConsumption  * (3f / 10f));
```
fomalsd: use 0.7f and 0.3f, no need to calculate division by ten

fomalsd (changes requested):
It would've been nice if there was a visual indication of canister connection, and an almost empty plasma tank in mining shuttle that's already connected for convenience.
Also look at things that need to be addressed below

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/GasMix.cs
Bod9001: that was just leftovers from another equation That I was planning To implement, It would take all of the energy inside of the volume and then put it back into the Resized volume, getting you the correct temperature and pressure

UnityProject/Assets/Scripts/Shuttles/MatrixMove.cs
Bod9001: What stops it is inside of the fuel update , https://github.com/unitystation/unitystation/pull/2193/files#diff-b9effcd4f5e1eb418c85d6ae953d3c1cR54

Bod9001: what stops it should be inside of the fuel update
code

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/GasMix.cs
```diff
@@ -23,6 +23,8 @@ public float Temperature
 			set => Pressure = AtmosUtils.CalcPressure(Volume, Moles, value);
 		}
 
+		public float TemperatureCash { get; private set; }
```
fomalsd: since you've added this field you'll also need to initialise it in GasMix constructor, currently code does not compile:
```cs
		private GasMix(float[] gases, float pressure, float volume = AtmosConstants.TileVolume)
		{
			Gases = gases;
			Pressure = pressure;
			Volume = volume;
			TemperatureCache = 0f;
		}
```

fomalsd: also there's still a typo, it should be called `TemperatureCache`

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/GasMix.cs
```diff
@@ -211,5 +222,12 @@ private void Recalculate()
 		{
 			Pressure = AtmosUtils.CalcPressure(Volume, Moles, Temperature);
 		}
+
+
+		//Used to change the pressure instead of temperature when removing/Adding gas
+		private void RecalculateTemperatureCache(float _TemperatureCash )
```
fomalsd: `float _TemperatureCash` argument is not needed as you always pass TemperatureCache field anyway

fomalsd (changes requested):
Code doesn't compile, you have to initialise temperature cache field in GasMix constructor

Bod9001:
sorry not too use to working with structs, but it should be fixed now

fomalsd (approved):
Yeah, works well now

PR #2194, opened 2019-10-05
Add renamable capability

StrawberryMoses:
### Purpose
Add Renameable component. Attached Renameable to paper. Objects can be renamed by right clicking and hitting "Rename". Dialog pops up, it's a squashed down paper background (I'm not an artist).

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Renameable.cs
+++ b/UnityProject/Assets/Scripts/Items/Renameable.cs
@@ -5,9 +5,16 @@ using WebSocketSharp;
 public class Renameable : NBHandActivateInteractable, IRightClickable
 {
 	public NetTabType NetTabType = NetTabType.Rename;
+	[SerializeField]
+	private string originalName;
+	[SerializeField]
+	private string customName;
 
-	public string originalName;
-	public string customName;
+	public string CustomName
+	{
+		get => customName;
+		private set => SetCustomName( value );
+	}
 
 	private ItemAttributes attributes;
 
@@ -23,7 +30,7 @@ public class Renameable : NBHandActivateInteractable, IRightClickable
 		base.OnStartServer();
 	}
 
-	public void SetCustomName(string custom, string original)
+	public void SetCustomName(string custom)
 	{
 		var itemName = originalName;
 
@@ -38,10 +45,6 @@ public class Renameable : NBHandActivateInteractable, IRightClickable
 		if (!string.IsNullOrEmpty(custom))
 		{
 			itemName += " - '" + custom + "'";
-		}
-		else
-		{
-			itemName = original;
 		}
 
 		attributes.SetItemName(itemName);
--- a/UnityProject/Assets/Scripts/Items/Renameable.cs
+++ b/UnityProject/Assets/Scripts/Items/Renameable.cs
@@ -6,66 +6,50 @@ public class Renameable : NBHandActivateInteractable, IRightClickable
 {
 	public NetTabType NetTabType = NetTabType.Rename;
 
-	[SyncVar(hook = nameof(SyncOriginalName))]
 	public string originalName;
-
-	[SyncVar(hook = nameof(SyncCustomName))]
 	public string customName;
 
 	private ItemAttributes attributes;
 
-	private void SyncOriginalName(string original)
-	{
-		originalName = original;
-	}
-
-	private void SyncCustomName(string custom)
-	{
-		customName = custom;
-
-		var itemName = originalName;
-
-		if (itemName.IsNullOrEmpty())
-		{
-			itemName = attributes.itemName;
-			SyncOriginalName(itemName);
-		}
-
-		if (!string.IsNullOrEmpty(custom))
-		{
-			itemName += " - '" + custom + "'";
-		}
-
-		attributes.SetItemName(itemName);
-	}
-
 	public override void OnStartClient()
 	{
-		SyncEverything();
+		attributes = gameObject.GetComponent<ItemAttributes>();
 		base.OnStartClient();
 	}
 
 	public override void OnStartServer()
 	{
-		SyncEverything();
+		attributes = gameObject.GetComponent<ItemAttributes>();
 		base.OnStartServer();
 	}
 
-	public void SetCustomName(string msg)
+	public void SetCustomName(string custom, string original)
 	{
-		SyncCustomName(msg);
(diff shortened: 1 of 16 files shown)


UnityProject/Assets/Scripts/Items/Renameable.cs
```diff
@@ -0,0 +1,80 @@
...
+
+	private void OpenRenameDialog(GameObject player)
+	{
+		TabUpdateMessage.Send(player, gameObject, NetTabType, TabAction.Open);
+	}
+
+	public void SetCustomName(string msg)
+	{
+		var attributes = gameObject.GetComponent<ItemAttributes>();
+		var customName = attributes.itemName.Split('-')[0].Trim();
```
fomalsd: won't this mess with items that contain `-` in their name?

UnityProject/Assets/Scripts/UI/PopUps/GUI_Rename.cs
```diff
@@ -0,0 +1,104 @@
...
+	//Request an edit from server:
+	public void OnTextEditEnd()
+	{
+		var customName = textField.text;
+		if (customName.Length > MAX_NAME_LENGTH)
+		{
+			customName = customName.Substring(0, MAX_NAME_LENGTH);
+		}
+
+		PlayerManager.LocalPlayerScript.playerNetworkActions.CmdRequestRename(Provider.gameObject, customName);
```
fomalsd, line 90: ~~It's a bit weird: using NetTabs, but running clientside code here only while they're designed for server authoritative NetElements.
Well, it works as is, but generally it's expected to use NetSubmitButton for sending text from your InputField to some serverside method (you can check TabComms – AnnouncementPage)~~
...
I just realized that you copied Paper code, never really looked at it before. It's acceptable to leave it as is, papers are a bit of special case. NetTabs are generally more machine-oriented.

UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
```diff
@@ -660,6 +660,17 @@ public void CmdRequestPaperEdit(GameObject paper, string newMsg)
...
+	[Command]
+	public void CmdRequestRename(GameObject target, string customName)
+	{
+		var rename = target.GetComponent<Renameable>();
+
+		if (rename != null)
+		{
+			rename.SetCustomName(customName);
+		}
+	}
```
fomalsd: Add some validation here, at least checking length and removing <size=x></size> tags from input

fomalsd (commented):
It actually works nicely, just needs some serverside validation

UnityProject/Assets/Scripts/Items/Renameable.cs
StrawberryMoses: Assumably, yes! This was supposed to get rid of the existing custom name, and most likely this component won't be on anything that does have - in it's name, but for future-proofing and doing things safely, the custom name and original name are stored here and synced.

UnityProject/Assets/Scripts/UI/PopUps/GUI_Rename.cs
StrawberryMoses: I did just copy paper code :) Was not aware that NetTabs were more for machines. For future UI popups (file folders, inventory stripping and such), is there a different system to use?

UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
StrawberryMoses: Done did

UnityProject/Assets/Scripts/UI/PopUps/GUI_Rename.cs
fomalsd: I'm afraid there aren't any other systems for that aside pure Unity UI. Inventory stripping should work really nicely with NetTabs though

UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
```diff
@@ -660,6 +661,27 @@ public void CmdRequestPaperEdit(GameObject paper, string newMsg)
...
+		{
+			return;
+		}
+
+		if (customName.Length > 42)
+		{
+			customName = customName.Substring(0, 42);
+		}
+
+		customName = Regex.Replace(customName, "<size=\"(.*)\">", "");
```
fomalsd: woops, bypassed :)
image
it has to be case insensitive

UnityProject/Assets/Scripts/Items/Renameable.cs
```diff
@@ -0,0 +1,119 @@
...
+
+public class Renameable : NBHandActivateInteractable, IRightClickable
+{
+	public NetTabType NetTabType = NetTabType.Rename;
+
+	[SyncVar(hook = nameof(SyncOriginalName))]
+	public string originalName;
+
+	[SyncVar(hook = nameof(SyncCustomName))]
+	public string customName;
```
fomalsd: I don't quite understand this – why make these two syncvars?
from what I understand, `ItemAttributes.SetItemName()` is only supposed to be called on server, why share contents of these fields with all clients? server can do it on its own

fomalsd (commented):
sorry for not reviewing it sooner, had a flu
I really think that two syncvars in Renameable aren't required and are good as server-only fields
And regex check should be case insensitive
Otherwise it's good!

UnityProject/Assets/Scripts/Items/Renameable.cs
StrawberryMoses: So it looks like these might be needed for late joiners renaming an object- they need to be able to know the original name in case they remove the rename. Setting the custom name comes from a [Command] which I think is client side?

UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
StrawberryMoses: Added case insensitive to the regex

UnityProject/Assets/Scripts/Items/Renameable.cs
fomalsd: Clients don't have to know original name. You can make server interpret empty string in Cmd as a request to restore original name and set syncvar from that field on server

fomalsd (approved):

PR #2201, opened 2019-10-09
Show transparent overlay if item can or cannot fit in an item slot

StrawberryMoses:
### Purpose
Green or red transparent item shows in slot when hovering over it with an item in players hand. Does this for pockets, worn inventory, backpack slots


fomalsd (commented):
very nice!  
there's one thing that newbies will find confusing – adjacent hand slot also lights up as green but just clicking it doesn't transfer item to another hand and changes slot instead (need to drag it over instead of clicking, and that's ok):
Scr 2019-10-12 в 15 46 36 mov
I don't know how this works in byond right now, but maybe just disable these overlays for hand slots?
Also I think byond overlays have lower transparency, maybe decrease it to like 70%? (~178 of 255)

Bod9001:
yeah, It doesn't have it for Hand slots, the way you would swap with hands is swap to the other hand then click on the item in the other hand

StrawberryMoses:
Changed to not display for hand slots. And 92% is correct, but i dunno if byond or anything changes it someplace else: 
Screen Shot 2019-10-09 at 1 08 36 PM

fomalsd:
Sweet!
Got a feeling that it might be 92/255 or 92/127 in byond, that would explain higher transparency

StrawberryMoses:
Looks like you were correct, changed it and it looks just about the same.

fomalsd (approved):

PR #2202, opened 2019-10-13
Add racks and rack parts

StrawberryMoses:
### Purpose
Adds racks. Racks function similarly to tables, but all items placed go into the center. They can be deconstructed by using a wrench in a non-help intent, then deconstructed again into a piece of metal. Clicking the ground while holding rack parts will build a new rack at that spot.


chairbender:
@StrawberryMoses nicely done! 

Few things I noticed, but I'll still merge this because these should be for other tickets.
1. When you try to construct a rack and then drop the item in hand, it glitches out and spams you with construction messages. I think I'll make a bug for this since I think most progress actions need to stop if you drop the thing you are using to perform the action.
1. When right clicking and choosing "Respawn", the rack disappears. It should maintain its position. Probably a bug in how respawn works.

Issue #2204, opened 2019-10-15
Progress Actions need to be interrupted in more situations


chairbender:
# Description
Currently they are interrupted if you move. But they aren't interrupted by dropping your in hand item, which can cause glitches in several cases (try constructing a Rack and drop the item before it's complete).

Think of other cases where progress needs to be interrupted and try to add the interrupt call for those. Please add to this list if you think of something. Otherwise just try to implement interrupts for these things:

- [ ] When dropping the item being used to perform the action
- [ ] When moving the item being used to perform the action to a different inventory slot
- [ ] When doing something with bare hands and picking up an item in the active hand
- [ ] When doing something with bare hands and moving an item into the active hand from an inventory slot
- [ ] When downed due to falling over or going crit
- [ ] When pushed or pulled (this might already interrupt)

fomalsd:
welding should be interrupted when welder is turned off as well, but it's a specific case

Bod9001:
I think it interrupts when you change direction it's not supposed to do that I think?

chairbender:
yup, we have another bug for that problem in #1780

chairbender:
Needs to be interrupted if thing being worked on (such as a tile position, or a movable object) moves. Also needs to support cross matrix - if standing in space and working on a shuttle tile from the outside, progress should stop if shuttle starts moving.

chairbender:
Actually as far as active hand goes, the action should probably always be interrupted if active hand changes in any way - either swapping active hand or putting something into (if empty) or out of (if occupied) it.

chairbender:
Several of these cases will be completed as part of #2245  - anything related to inventory slot changes has logic in place for cancelling the progress

DooblyNoobly:
@chairbender can this ticket be closed ?

chairbender:
It's close to being done, but the stuff mentioned needs to be tested to make sure, and I don't know if welder had custom logic added to it either for interrupting when turned off.

PR #2213, opened 2019-10-20
Progress Bar Improvements

### LINKED ISSUE #1778: ProgressAction cancels on character turning
# Description
If your character turns, the progress action will stop.

## Category
BUG

## Steps to Reproduce
1.Find anything that uses a progress action, in this case for example, a mop
2.Start cleaning the floor
3.Click behind you
4.Your character will turn and stop the progressaction bar, stopping you from cleaning the floor tile

### LINKED ISSUE #1780: Can't clean more than one tile at the same time
# Description
Probably related to ProgressAction, when using the mop, I can't clean other tiles next to me at the same time unless I cancel the current progress action. I should be able to clean all at the same time

## Category
BUG

## Steps to Reproduce
1.Click on a tile with a mop to clean it
2.Click on the tiles next to it to clean them at the same time
3.Nothing happens unless my character turns and the progress stops (covered in another issue report)

chairbender:
### Purpose
Fixes #1780 
Fixes #1778 
Partially implements #2204 

Improves usability of progress bar API so it "just works" in the way you would expect and you don't have to write as much boilerplate.

- [x]  Any new / changed components have been [tested using Respawn](https://github.com/unitystation/unitystation/wiki/Object-Lifecycle-Best-Practices)

### Notes:
Wow, what a slog this was. Main difficulty was the networking and matrix move interactions. Didn't use PoolNetworkInstantiate and register tile / other networking stuff because progress bars only need to be known by one client and one server, don't even want to let other players know a progress bar was instantiated (even if we wouldn't send them the update messages). PoolNetworkInstantiate would've created the object on all clients, even if we never plan on sending any progress update messages it still would've exposed extra info to each client.

Hopefully that's okay that I didn't use registerTile / CNT / objectbehavior. These are more analogous to bullets, which also don't use those things. If it REALLY seems like a problem I can change it to use those, but I have tested it pretty thoroughly.

On the plus side it shows a way that such use cases can be supported if it will be needed for other things that should only be known by server + one client.


fomalsd:
Haven't looked at this thoroughly yet, but at this point it doesn't fix #1780 – you can't clean more than one tile. Welding multiple tiles simultaneously no longer works, too (while it was possible before but with visual glitches)
GIF0

chairbender:
I definitely tested that, must have regressed something

chairbender:
@fomalsd try again

DooblyNoobly:
- fixed the merge conflicts related to the new chat system

DooblyNoobly:
looking alright. I just need to do some quick fixed with the chat when I merge this

PR #2227, opened 2019-10-25
IF2 Enhancement

### LINKED ISSUE #2188: IF2 + Right Click improvement, remove "specificComponent" from IF2
# Description
Now that people have been using IF2 for a bit, a few problems are emerging:
1. Dealing with IC precedence is proving challenging. People are having trouble figuring out how to create / place ICs in order to ensure their particular interaction fires under the right circumstances.
2. Amount of components on each object is growing as people are implementing one component per interaction in some cases. Not necessarily an issue, but it might be better if we can consolidate these somehow.
3. Want ability to trigger interactions at will, which IF2 doesn't really support (as documented below) #2181 added "specificComponent", which was needed for the right click -> Pickup option to work properly. Otherwise, it would end up triggering the wrong component when pickup was attempted, because pickup sent a HandApply message rather than directly triggering the pickup interaction logic. We should refactor so this string isn't necessary.

See what can be done to fix the above issues.

chairbender:
### Purpose
Fixes #2188 

1. Refactors the core if2 classes to be interfaces instead of abstract base classes. No more NBBlahBlah crap! Multiple inheritance now possible!
2. Add Interaction logging category and TRACE level messages to make it really easy to see the order of precedence / which components are being checked during a given interaction.
2. Consolidates the messages into a single net message, RequestInteractMessage, to simplify organization / maintenance.
3. Provide a way to manually trigger interactions for a specific component (useful for right click menu). Removed the string param which was originally being used for this in favor of a smaller data type.
3. Fix some broken drag and drop interactions.
4. Updated the documentation to reflect these changes

I know it's a bit of a big PR but look at the lines changed, I've removed more code than I've added.

- [x]  Any new / changed components have been [tested using Respawn](https://github.com/unitystation/unitystation/wiki/Object-Lifecycle-Best-Practices)


chairbender:
One sec, going to compress the net message even more for some values that can be bytes rather than ushort

chairbender:
alright, it's ready to review now.

fomalsd:
code looks good!

UnityProject/Assets/Scripts/Consoles/CommsConsole.cs
```diff
@@ -3,7 +3,7 @@
 using UnityEngine;
 using UnityEngine.Events;
 
-public class CommsConsole : NBHandApplyInteractable
+public class CommsConsole : MonoBehaviour, ICheckedInteractable<HandApply>
```
StrawberryMoses, line 6: This was a network behavior before, does it still need to be one?

UnityProject/Assets/Scripts/Shuttles/ShuttleConsole.cs
```diff
@@ -8,7 +8,7 @@
 /// <summary>
 /// Main component for shuttle console
 /// </summary>
-public class ShuttleConsole : NBHandApplyInteractable
+public class ShuttleConsole : MonoBehaviour, ICheckedInteractable<HandApply>
```
StrawberryMoses, line 11: Also a network behaviour here

StrawberryMoses (approved):

UnityProject/Assets/Scripts/Consoles/CommsConsole.cs
fomalsd: I don't think so, it doesn't have any `[Server]` methods or syncvars or anything else NetworkBehaviourish

UnityProject/Assets/Scripts/Shuttles/ShuttleConsole.cs
fomalsd: Same here, I think it should be fine. otherwise editor would give an error straight away

chairbender: Nope. You really only need nb if you are using syncvars or other nb features in a component. Quite a few components had it despite not needing it.

fomalsd (approved):

PR #2275, opened 2019-11-14
Small Boxes, Inventory UI Improvements, Spatial Relationships

chairbender:
### Purpose
Fixes #2080 
Fixes #1698 (pretty sure...)

Also adds an X to close the currently open storage.

Lots of polish / improvements to inventory UI. It should be way easier to use now. Now there is no more having to click the sprite exactly in the slot. You can click the item or the area around and you can still drag and drop / interact like normal.

Tweaks to how the InventoryApply interaction works. Now you can InventoryApply between 2 arbitrary slots and aren't restricted to only applying from the active hand. Adds some validation to allow enforcing the interaction to only work when it's from the active hand (which was how the old apply worked by default).

Adds examine messages in  more cases to explain what's happening when something doesn't fit. Also improves the logic of ExpensiveName to handle empty string stuff better.

Adds a new system - [Spatial Relationships](https://github.com/unitystation/unitystation/wiki/Spatial-Relationships%3A-Checking-When-Stuff-Moves-Away). Just a small tool that may be useful in other situations.

Fixes a bug where players were showing as naked to other players. PlayerAppearanceMessage had a race condition with the various SyncVars in Clothing.

### Open Questions and Pre-Merge TODOs

- [x]  This fix is tested on the same branch it is PR'ed to.
- [x]  I correctly commented my code
- [x]  My code is indented with tabs and not spaces
- [x]  This PR does not include any unnecessary .meta, .prefab or .unity (scene) changes
- [x]  This PR does not bring up any new compile errors
- [x]  This PR has been tested in editor
- [x]  Any new / changed components have been [tested using Respawn](https://github.com/unitystation/unitystation/wiki/Proper-Spawning-and-Despawning-%28Object-Lifecycle%29)
- [x]  This PR has been tested with round restarts.
- [x]  This PR has been tested on moving / rotating / rotated-before-joining matrices (if applicable)
- [x]  This PR has been tested in multiplayer (with 2 clients and late joiners, if applicable)

PR #2279, opened 2019-11-15
Attempt 3 - Fix large tab updates exceeding max packet size

### LINKED ISSUE #2251: Clients can't use ID Console
# Description
When a client tries to use the ID console and inserts the target card, a Mirror error message shows up on server side complaining that the packet size is too large. This is probably because the amount of data in the net tab is huge.

Expected behavior is that client should receive the necessary data and the ID console should populate with all the various options for modifying the target card.

chairbender:
### Purpose
Fixes #2251 

Note- also includes #2275 so please merge that first.

- [x]  This PR has been tested with round restarts.

### Notes:
@fomalsd may want to see this since I think he was the main person working on tabs.

### DIFF
--- a/UnityProject/Assets/Scripts/Messages/Client/TabInteractMessage.cs
+++ b/UnityProject/Assets/Scripts/Messages/Client/TabInteractMessage.cs
@@ -68,10 +68,10 @@ public class TabInteractMessage : ClientMessage
 		}
 	}
 
-	private TabUpdateMessage FailValidation( ConnectedPlayer player, GameObject tabProvider, string reason="" ) {
+	private void FailValidation( ConnectedPlayer player, GameObject tabProvider, string reason="" ) {
 
 		Logger.LogWarning( $"{player.Name}: Tab interaction w/{tabProvider} denied: {reason}",Category.NetUI );
-		return TabUpdateMessage.Send( player.GameObject, tabProvider, NetTabType, TabAction.Close );
+		TabUpdateMessage.Send( player.GameObject, tabProvider, NetTabType, TabAction.Close );
 	}
 
 	public static TabInteractMessage Send( GameObject tabProvider, NetTabType netTabType, string elementId, string elementValue = "-1" )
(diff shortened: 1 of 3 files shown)


fomalsd (changes requested):
Unfortunately this won't work well. Client expects Open message to contain information about all elements, otherwise split Open messages are treated like separate ones.
For example, I've set up splitting into 3 element chunks:
`[NetUI] [TabUpdateMessage Provider: 1, Type: IdConsole, Action: Open, ElementValue: [ScreenBounds=0]; [IdNameButton=-1]; [IdNameText=********]]`
`[NetUI] [TabUpdateMessage Provider: 1, Type: IdConsole, Action: Open, ElementValue: [LoginButton=-1]]`
And result was
`[NetUI] 'TabIdConsole(Clone)' wonky value import: can't find 'IdNameText'.
 Expected: ScreenBounds/LogoutButton`
So it forgets about first values as if these were two separate tabs.

I'd not attempt splitting at this stage and instead try lowering bandwidth of that ID tab as a temp fix. It shouldn't really have to send background colour info, for example.

chairbender:
@fomalsd I don't know if that's doable (lowering bandwidth of this tab enough). It seem pretty shortsighted to have a fundamental system like this just break because there's too many things on the tab. Why not just fix the logic so that the broken up messages work? Surely it's possible to make it doable to break the open message up into a few messages.

chairbender:
Basically, if we get an Open message and the tab is already open, client can treat it as an update.

chairbender:
Also, I know it shows that warning but does it actually result in some error? It seemed to work fine for me, and when I was looking at the code for Open on client side it seemed like it did exactly what it should for this case - when it's already open it just updates the already open tab. The warning message may be a false positive.

I can test more but it would help to see what problem it actually has rather than there just being a warning message.

chairbender:
alright, talked a bit in Discord, gonna close this and instead optimize the tab itself.

Issue #2032, opened 2019-07-18
Handcuffs stop player movement B: 35


Jibbajabbafic:
# Description
Players should still be able to move when handcuffed, but any hand interactions should be disabled (in inventory and in the world). 

The only time they can't move is if the player is being dragged by someone else as well. The handcuffed player shouldn't be able to break pulls either.

Edit: 
When Doobly made #1935 he specified that cuffs should stop movement which isn't the case in /tg/

## Category
BUG

## Steps to Reproduce
Please enter the steps to reproduce the bug or behaviour:
1. Get handcuffed
2. Try to move

JaytleBee:
For clarity, this is intended behaviour that Doobly specified in #1935:

> Stop player movement input from being allowed if [the player is cuffed]

Emma3DSS:
Fixing this

Brizzzle1738:
Is this a bug? May have found a fix, however it seems that when handcuffs were created it was supposed to stop player movement input.  Has the design of handcuffs changed to where a player is able to move when handcuffed?

chairbender:
@Brizzzle1738 the description in the ticket is still correct about what needs to be done. Note that I have already made it impossible to interact with things when cuffed in #2281 (which is not yet merged but should be soon), so all that's left to do is prevent player movement and pull breaking when cuffed, but allow them to move otherwise.

chairbender:
@Brizzzle1738 just an update, I would suggest holding off on this until #2281 is merged as I'm making some changes to interaction / movement validation

DooblyNoobly:
Contributor will receive a bounty of 35usd on a successful merge of a solution that resolves the issue above.

PR #2281, opened 2019-11-16
Attempt 2 - Adds player inventory interaction / stripping

### LINKED ISSUE #2144: Inventory inspection, Cloth stripping/stealing functionality
# Description
Would be nice to be able to see a list of other person's equipment with ability to take it off,
because at the moment there's no way to retrieve dead person's belongings

chairbender:
### Purpose
Removed the broken attempted tab update fixes.

Fixes #2144 

Also the "Networking > Respawn Player" command should work now regardless of game mode.

Can click drag a non-conscious player to yourself to view their inventory. A button allows dropping all of their items on the ground, or you can interact with their inventory slots just like it was a storage you are looking into.

It's a bit "quick and dirty" as far as the UI goes but at least now you can do it and items aren't lost forever in their inventory.

- [X]  Any new / changed components have been [tested using Respawn](https://github.com/unitystation/unitystation/wiki/Proper-Spawning-and-Despawning-%28Object-Lifecycle%29)
- [X]  This PR has been tested on moving / rotating / rotated-before-joining matrices (if applicable)

### DIFF
--- a/UnityProject/Assets/Scripts/Input System/InteractionV2/Validations.cs
+++ b/UnityProject/Assets/Scripts/Input System/InteractionV2/Validations.cs
@@ -70,12 +70,13 @@ public static class Validations
 	/// <param name="player">player gameobject to check</param>
 	/// <param name="side">side of the network the check is being performed on</param>
 	/// <param name="allowSoftCrit">whether interaction should be allowed if in soft crit</param>
+	/// <param name="allowCuffed">whether interaction should be allowed if cuffed</param>
 	/// <returns></returns>
-	public static bool CanInteract(GameObject player, NetworkSide side, bool allowSoftCrit = false)
+	public static bool CanInteract(GameObject player, NetworkSide side, bool allowSoftCrit = false, bool allowCuffed = false)
 	{
 		if (player == null) return false;
 		var playerScript = player.GetComponent<PlayerScript>();
-		if (playerScript.playerMove.IsCuffed || playerScript.IsGhost || playerScript.canNotInteract() && (!playerScript.playerHealth.IsSoftCrit || !allowSoftCrit))
+		if ((!allowCuffed && playerScript.playerMove.IsCuffed) || playerScript.IsGhost || playerScript.canNotInteract() && (!playerScript.playerHealth.IsSoftCrit || !allowSoftCrit))
 		{
 			return false;
 		}
@@ -336,20 +337,23 @@ public static class Validations
 	public static bool IsDeadCritStunnedOrCuffed(GameObject player, NetworkSide side)
 	{
 		if (player == null) return false;
+		var playerScript = player.GetComponent<PlayerScript>();
+		if (playerScript == null) return false;
+
 		if (side == NetworkSide.Client)
 		{
 			//we don't know their exact health state and whether they are slipping, but we can guess if they're downed we can do this
-			var registerPlayer = player.GetComponent<RegisterPlayer>();
-			var playerMove = player.GetComponent<PlayerMove>();
+			var registerPlayer = playerScript.registerTile;
+			var playerMove = playerScript.playerMove;
 			if (registerPlayer == null || playerMove == null) return false;
 			return registerPlayer.IsDown || playerMove.IsCuffed;
 		}
 		else
 		{
 			//find their exact conscious state, slipping state, cuffed state
-			var playerHealth = player.GetComponent<PlayerHealth>();
-			var registerPlayer = player.GetComponent<RegisterPlayer>();
-			var playerMove = player.GetComponent<PlayerMove>();
+			var playerHealth = playerScript.playerHealth;
+			var registerPlayer = playerScript.registerTile;
+			var playerMove = playerScript.playerMove;
 			if (playerHealth == null || playerMove == null || registerPlayer == null) return false;
 			return playerHealth.ConsciousState != ConsciousState.CONSCIOUS || registerPlayer.IsSlippingServer || playerMove.IsCuffed;
 		}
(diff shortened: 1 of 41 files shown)


chairbender:
oops, one sec, forgot about foma's request:
"how about also checking for "stunned" (registerPlayer.IsSlippingServer)
and "cuffed" (apparently playerMove.IsCuffed)?"

chairbender:
still got a few minor things to fix but here's a gif of it in action
transform_mime

chairbender:
it's ready now

I had to add event hooks so player storage will become unobservable when they are, for example, no longer cuffed, and did it by creating a new spatial relationship that can have all of the logic related to when storage should no longer be observed.

chairbender:
gah, one more thing I realized, sorry

chairbender:
Alright, now it should be good. Fun fact - if you look at a player's storage and drag one of their storage items over your character, like their backpack, you can view it right there without removing it from their storage, and (thanks to this fix) you will stop being able to see it once the player becomes un-observable.

Fixed an unrelated NRE that happened once when server spawned their player while a client was joining. I imagine we have a few of these sorts of latent join-time issues btw.

Added the hand slots to the other player inventory display. I thought they wouldn't be needed since you always drop items when you become observable, but I think it's best to include them just in case. Now you can plant things in people's hands.

I also made it so you can't generally interact when you're cuffed, and you drop your hand items when cuffed.

UnityProject/Assets/Scripts/Input System/InteractionV2/Validations.cs
```diff
@@ -350,4 +332,26 @@ public static bool CanFit(ItemSlot itemSlot, Pickupable toCheck, NetworkSide sid
...
+			//we don't know their exact health state and whether they are slipping, but we can guess if they're downed we can do this
+			var registerPlayer = player.GetComponent<RegisterPlayer>();
+			var playerMove = player.GetComponent<PlayerMove>();
+			if (registerPlayer == null || playerMove == null) return false;
+			return registerPlayer.IsDown || playerMove.IsCuffed;
+		}
+		else
+		{
+			//find their exact conscious state, slipping state, cuffed state
+			var playerHealth = player.GetComponent<PlayerHealth>();
```
fomalsd: Could reduce `GetComponent` calls here by getting just the `PlayerScript`, it has references to all the scripts you're trying to get here

UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
```diff
@@ -81,6 +81,28 @@ public void CmdDropItem(NamedSlot equipSlot)
...
+
+		//are we an observer of the player to disrobe?
+		if (!itemStorage.ServerIsObserver(gameObject)) return;
+
+		//disrobe
+		foreach (var itemSlot in itemStorage.GetItemSlots())
+		{
+			//skip slots which have special uses
+			if (itemSlot.NamedSlot == NamedSlot.handcuffs) continue;
+			Inventory.ServerDrop(itemSlot);
```
fomalsd: would be nice to convert this into a progress action with a coroutine down the road, taking like .2s for each emptied slot 
as this "insta naked" button seems a bit overpowered at the moment

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/SpatialRelationship/ObserveStorageRelationship.cs
```diff
@@ -0,0 +1,92 @@
...
+			observedPlayerHealth.OnConsciousStateChangeServer.AddListener(OnConsciousStateChangeServer);
+		}
+	}
+
+	private void OnConsciousStateChangeServer(ConsciousState oldState, ConsciousState newState)
+	{
+		//stop the relationship if observed player becomes conscious
+		if (newState == ConsciousState.CONSCIOUS)
+		{
+			SpatialRelationship.ServerEnd(this);
```
fomalsd: shouldn't this check other factors sustaining relationship before breaking it? like, if player is cuffed all the time it should probably remain unbroken regardless of conscious/slipped state change
but if you intentionally skipped on this to avoid code getting overly complex then it's understandable

fomalsd:
this worked nicely with a cuffed player, but not with a temporarily slipped one.
also while it mat not be fully relevant to this PR, it seems weird that you can't move while being cuffed but are able to hug and beat up others

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/SpatialRelationship/ObserveStorageRelationship.cs
chairbender: yeah I thought it was just easier for now to just break it once any factor changes, but actually the logic you're talking about should be quite simple so I'll add that

UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
chairbender: I'll do this. It should be easy with the new Progress action system.

chairbender:
Oh you aren't supposed to be able to do either of those things, I think it's because the hug / melee actions are bypassing the default interaction checks. I can add that. The not being able to move thing should be handled in #2032 

EDIT: There are probably quite a few things in PNA which don't use the "proper" validation methods before performing the action so I'll take a look over them. I have quite a few comments throughout the codebase I think suggesting to convert the PNA stuff to IF2 stuff, which would solve this but I think that's too big for this PR.

chairbender:
And I realized lots of stuff is not actually using the correct validation logic in many cases. going to get rid of canNotInteract and replace all calls with the correct method, Validation.CanApply or Validation.CanInteract depending on the situation. Not just in PNA / WNA.

chairbender:
Alright @fomalsd , fixed all your suggestions and more. Validation logic is way more consistent now and added overloads for calling the validations passing a PlayerScript so no GetComponent is needed.

fomalsd (approved):
awesome

PR #2282, opened 2019-11-17
Fixed issue #2166 where lockers could drag vents and scrubbers

### LINKED ISSUE #2166: Lockers can pickup vents and scrubbers
# Description
Pulling a locker on a square with a vent or a scrubber, then closing, will bring the vent or scrubber with.

## Category
BUG

## Steps to Reproduce
1. Start pulling a locker.
2. Pull the locker to a tile with a vent or a scrubber object.
3. Close the locker so the vent or scrubber is inside.
4. Pull the locker to a new tile.
5. The vent or scrubber will move with the locker.

dillonfurey:
### Purpose

chairbender edit: Fixes #2166 

- [ ]  Any new / changed components have been [tested using Respawn](https://github.com/unitystation/unitystation/wiki/Proper-Spawning-and-Despawning-%28Object-Lifecycle%29)

### Notes:


chairbender:
hi @dillonfurey , thanks for the PR!

I notice you are checking for the item name, which I'm sure works for this case, but might not work for other items which are on the floor, and would break if someone were to change the name even slightly. To make this a bit more general, instead of checking their name, you could check if something is attached to the floor using this logic:

```csharp
var pipe = item.GetComponent<Pipe>();
if (pipe != null && pipe.anchored) continue;
``` 

Also, if you want to put comments on an individual statement (rather than methods or classes), you can just use the normal `//comment` rather than `///<summary>comment</summary>` You only need to use the triple slash comments on methods, classes, and fields.

PR #2284, opened 2019-11-18
Optimize ID console so it is usable by client

### LINKED ISSUE #2251: Clients can't use ID Console

chairbender:
### Purpose
Fixes #2251 (and does not depend on my other current PR, I've learned my lesson)

See notes for how I did it.

Also improved IDCard encapsulation. There was a lot of public stuff, including SyncVars, which I made private. I also changed all the stuff which was using ints rather than enums to just use enums. Added tooltips and comments as always.

- [X]  Any new / changed components have been [tested using Respawn](https://github.com/unitystation/unitystation/wiki/Proper-Spawning-and-Despawning-%28Object-Lifecycle%29)
- [X]  This PR has been tested with round restarts.
- [X]  This PR has been tested on moving / rotating / rotated-before-joining matrices (if applicable)

### Notes:
So as (I think) was suggested, I just built the UI out completely and didn't rely on any dynamic lists. Considering that all of the configuration was already basically hardcoded in IDConsoleManager, and it was only used by ID console, I don't think this really makes much impact on maintainability. I removed IDConsoleManager as well - it's more straightforward to have the ID console configuration live in the Id console tab prefab.

All of the old stuff is renamed to include "Old" in the prefab name / component so you can use it still for stress testing nettab system.


fomalsd (approved):
thanks a lot! really nice to see it load instantly. cleaned up structure is sweet, too

PR #2289, opened 2019-11-19
Add Spawnables, convert lockers to use them for initial contents

### LINKED ISSUE #2216: UniCloth doesn't work
# Description
img
Should UniCloth be removed? Since we apparently use prefab-per-cloth now?

## Steps to Reproduce
1. Go to mining
2. Open mining closet
3. See questionmark clothes

### LINKED ISSUE #2255: Make inventory populators able to spawn things in the world (not just in inventory). Refactor closets to use them for default contents
# Description
This should probably be done after or alongside #2247 , and assumes that #2245 is merged.

Currently, default closet contents are defined in the Prefab and can be overridden in the scene simply by mapping items on top of the closet. One limitation of this is that closet default contents cannot contain cloths, because it only allows assigning a prefab.

Instead, we should be able to use the inventory populators system to define a closet's default contents. It might require a little reworking, either to make it so that populators can just spawn items rather than needing to put them into a slot, or modifying lockers to have ItemStorage and putting the items into storage when the locker is closed.

This way, any population logic built for the inventory system can be repurposed for closet contents.

chairbender:
### Purpose
Fixes #2255 
Fixes #2216 

Adds spawnables, [documented here](https://github.com/unitystation/unitystation/wiki/Proper-Spawning-and-Despawning-%28Object-Lifecycle%29#spawnables) . Restructured inventory slot populators to use these and move the assets into a different folder. Inventory wiki page was updated to mention this.

Set up the initial contents of lockers to use these assets (via a SpawnableList SO). Also closets will correctly spawn their initial contents when spawned via dev spawner.

I also fixed an issue where some faulty logic was causing PreRoundStart to be invoked twice when loading via lobby. This may have been the cause of some other issues? I only noticed it because closet contents were being doubled.

Some changes to the internal of lifecycle system to utilize spawnables, which I think makes it more flexible if we want to be able to spawn things in specific configuration in the future (like reagent containers with predefined liquids or whatever). I also think the internal lifecycle system code is a bit cleaner now - particularly SpawnInfo.

Mapping change on outpost station is just to fix the stuff on the 4 mining lockers (the messed up unicloths). Easily redone if there's a conflict.

The large amount of changes is just due to moving the various slot populator assets into the new spawnable folder.


UnityProject/Assets/Scripts/Lifecycle/Spawnable/SpawnableResult.cs
```diff
@@ -0,0 +1,88 @@
...
+
+	/// <summary>
+	/// True iff this was only a single spawned object
+	/// </summary>
+	public bool IsSingle => GameObjects.Count() == 1;
+
+	/// <summary>
+	/// True iff this was for spawning multiple objects.
+	/// </summary>
+	public bool IsMultiple => !IsSingle;
```
fomalsd, line 41: looks like this will return true if GameObjects.Count() is 0

UnityProject/Assets/Scripts/Lifecycle/Spawn.cs
```diff
@@ -253,7 +272,8 @@ public static SpawnResult ClientPrefab(string prefabName, Vector3? worldPosition
 	public static SpawnResult ServerClone(GameObject toClone, Vector3? worldPosition = null, Transform parent = null,
 		Quaternion? rotation = null)
 	{
-		return Server(SpawnInfo.Clone(toClone, worldPosition, parent, rotation));
+		return Server(
+			SpawnInfo.Clone(toClone, SpawnDestination.At(toClone)));
 	}
```
fomalsd: hm, so this method now ignores `worldPosition`, `parent` and `rotation` parameters and instead assumedly always takes that info from `toClone` gameObject. are you sure this is desired? what if someone wants to clone this object at a different position?

UnityProject/Assets/Scripts/Lifecycle/Spawn.cs
```diff
@@ -211,7 +222,11 @@ public static SpawnResult ClientPrefab(GameObject prefab, Vector3? worldPosition
 	public static SpawnResult ServerPrefab(string prefabName, Vector3? worldPosition = null, Transform parent = null,
 		Quaternion? rotation = null, int count = 1, float? scatterRadius = null, bool cancelIfImpassable = false)
 	{
-		return Server(SpawnInfo.Prefab(prefabName, worldPosition, parent, rotation, count, scatterRadius, cancelIfImpassable));
+		return Server(
+			SpawnInfo.Spawnable(
+				SpawnablePrefab.For(prefabName),
+				SpawnDestination.At(worldPosition, parent, rotation),
+				count, scatterRadius));
```
fomalsd: `cancelIfImpassable` flag might be ignored here, unless I misunderstood something

UnityProject/Assets/Scripts/Lifecycle/Spawn.cs
```diff
@@ -167,7 +170,11 @@ public static List<GameObject> SpawnablePrefabs()
 	public static SpawnResult ServerPrefab(GameObject prefab, Vector3? worldPosition = null, Transform parent = null,
 		Quaternion? rotation = null, int count = 1, float? scatterRadius = null, bool cancelIfImpassable = false)
 	{
-		return Server(SpawnInfo.Prefab(prefab, worldPosition, parent, rotation, count, scatterRadius, cancelIfImpassable));
+		return Server(
+			SpawnInfo.Spawnable(
+				SpawnablePrefab.For(prefab),
+				SpawnDestination.At(worldPosition, parent, rotation),
+				count, scatterRadius));
 	}
```
fomalsd: `cancelIfImpassable` flag might be ignored here as well

fomalsd:
Liking the changes in code! (yet to playtest tho), especially the SpawnDestination abstraction

chairbender:
oops, good catch. Fixed now.

fomalsd (approved):

PR #2296, opened 2019-11-21
Account Creation Error Handling

MrLeebo:
- Unhandled Exceptions left the Create Account screen locked at "Please Wait..."
  - Firebase Exceptions are wrapped in a "AggregateException" class
  - Do the necessary yak shaving to get legible error messages in the UI


DooblyNoobly:
Thanking you

PR #2342, opened 2019-11-23
Moses/toolbelt

StrawberryMoses:
### Purpose
Adds toolbelts and the most basic toolbox ever.
A few toolbelts were added to a few jobs that needed them. I only added the blue toolbox.

### Open Questions and Pre-Merge TODOs

- [x]  This fix is tested on the same branch it is PR'ed to.
- [x]  I correctly commented my code
- [x]  My code is indented with tabs and not spaces
- [x]  This PR does not include any unnecessary .meta, .prefab or .unity (scene) changes
- [x]  This PR does not bring up any new compile errors
- [x]  This PR has been tested in editor
- [ ]  Any new / changed components have been [tested using Respawn](https://github.com/unitystation/unitystation/wiki/Proper-Spawning-and-Despawning-%28Object-Lifecycle%29)
- [ ]  This PR has been tested with round restarts.
- [ ]  This PR has been tested on moving / rotating / rotated-before-joining matrices (if applicable)
- [ ]  This PR has been tested in multiplayer (with 2 clients and late joiners, if applicable)

### Notes:
_Please enter any other relevant information here_

Issue #2340, opened 2019-11-23
Better Examine System


Bod9001:
# Description
We need the ability to let components on an object customize that object's examine text. For example, this is used often during construction to give hints about what you need to do next or to see what state an object is in for objects that don't have different sprites for each construction state.

We need to support components adding arbitrary lines to the examine message. For tooltips on mouseover I think it would be fine to just display the item attributes name + description, but when you actually perform the examine action on it and see the message in chat it should fire off this examine system logic so you can display more information about the object.

This also needs to be supported on tiles, though I think each tile can pretty much just have a name + description and might not need any dynamic logic at the moment (since each construction stage of a tile is its own tile asset).

The examine logic should most likely be server-side, as there are things the client wouldn't be able to know since it's not synced to them.

I might suggest just adding another interface hook, IExaminable, which has a method that allows the component to "add on" to the examine message. It would be called on server side when client requests to examine.

And as a proof of this system, please add examine logic for reinforced wall construction / deconstruction (for tile-based) and computer construction / deconstruction (for object-based)

fomalsd:
they did have proper descriptions when dm data was a thing, maybe there’s still some way to automate it

chairbender:
I'm doing this for all the cloths as part of #2312

chairbender:
I will eventually be migrating the itemattributes component users to the new and improved "ItemAttributesV2" so I'll also just look up the names / descriptions when I do that and that should fix this ticket, but that won't be until a later PR.

EDIT: So I'll assign this to me.

chairbender:
going to make this more general

StrawberryMoses:
This should also support hovering over items that are in inventory as well

Issue #2283, opened 2019-11-18
Refactor everything to use prefab variants


chairbender:
# Description
Depends on #2386 
Somewhat overlaps with #2247 and #2246 

Due to the tilemap changes this would require, it may be best to hold off on this until doing #2191 , but even then I think it would still require lots of mapping changes, so it can probably be done without.

Now that we have somewhat better idea of the structure of our prefabs, it would save effort to refactor them to be prefab variants.

For example, items (generally things which are Pickupable) should have a base "Item.prefab" which has all the components that any item should have. There should also probably be a base "Object.prefab" for things which aren't pickupable (chairs, machines, etc..).

Furthermore, use additional child prefab variants to put common prefab structure into appropriate parent prefabs, just like was suggested with guns, lockers, and crates. Other things which might be appropriate for this - Food, healing packs, consoles, etc...

For example, you could have Item > Food > Meat Steak, and Item > Food > Cookie.

The end goal is to save time so that changes to the fundamental way certain classes of things work will not require massive edits to lots of different prefabs, and so it's even easier to not screw up and forget necessary components when creating new things.

This will most likely require lots of remapping and tilemap changes, there's not an easy way to avoid that since you can't really turn an existing prefab into a variant of another prefab. But once things are within this prefab variant structure it will become increasingly easier to deal with prefabs and create new ones.

I suggest doing this in a piecemeal fashion, some incremental PRs. Each time you convert a prefab to be a variant, you will have to remap it in the scene and adjust its tilemap entry. So we can kinda go prefab by prefab.

chairbender:
One issue with variants is you can only add to the bottom of components in the parent, you can't rearrange parent components or insert things in the middle of them in a variant.

A simple fix for this is that we should change the IF2 interaction checking logic to check components in reverse order, from bottom up. Makes sense since variants that add ICs will want to have their logic fire first, and their components will always show up at the bottom. A feewexisting prefabs will need to reorder their components but other than that it should be simple

chairbender:
This is basically in place now. There are still some things which aren't migrated but it's documented and there are quite a few things already using prefab variants.

PR #2343, opened 2019-11-24
Cloth Killer + Belts

### LINKED ISSUE #2315: Consider Converting Cloths to be prefabs (and use prefab variants)
# Description
For many things, we've had to implement custom logic specifically for handling cloths. For example, prefabs have the capability to do prefab variants, but cloths don't unless we build it.

I personally have spent many hours dealing with cloth-specific logic due to them not using prefabs. We should decide if we're really going to stick to this system or refactor them to use prefabs / prefab variants.

IMO we would save a lot of trouble if we just moved them to prefabs / prefab variants. So many features we want to develop are hindered by having to account for unicloths.

EDIT: As far as space usage goes, it should be noted that prefab variants only store their modifications upon their base prefab (as you would expect), so they don't duplicate unnecessary data, see https://github.com/unitystation/unitystation/blob/develop/UnityProject/Assets/Resources/Prefabs/Objects/Canisters/PlasmaCanister.prefab

Also worth noting that the new Spawnables at least provide a standardized way to spawn a prefab or a cloth, but I view this only as a temporary stopgap measure as all the difficulties I mentioned in understanding and creating the cloths still apply.

In fact, by using the prefab variant system, we could do away with the need for spawnables, which adds a bit of convenience into the editor configuration due to not needing to create a Spawnable asset. Basically removing an unnecessary layer of indirection, thus making it easier to configure things which need to be able to reference other objects. Currently it's a little tedious having to create a SpawnablePrefab, reference the prefab, then reference the Spawnable in whatever you wanted it for (like an inventory populator). With prefab variants, you'd just directly reference the prefab you wanted to populate the slot with, and if you wanted any custom initial state, you could make a variant of that prefab and modify the state in the components and use that variant to populate the slot.

chairbender:
### Purpose
EDIT: New wiki page on creating objects which somewhat covers the changes in this pr is here: https://github.com/unitystation/unitystation/wiki/Creating-Items-and-Objects%3A-Now-With-Prefab-Variants

I also updated the inventory system as slot populators are slightly changed (to hopefully be easier to use) https://github.com/unitystation/unitystation/wiki/Inventory-System

Includes @StrawberryMoses belt PR #2342, which adds belts, refactored to use the new cloth variant system.
Fixes #2315 

Partially implements #2283 because I created a base "Item" prefab and further variants which can be used as the parent for different items (used for different types of clothing)

Adds name / description / resistances / flags to ALL cloths based on tg values, partially implementing #2340 

Things which were referencing Spawnables can now just directly reference prefabs, and all spawnable assets were removed. It's now MUCH easier to define what prefabs should go in things. And due to using prefab variants, if you want any custom logic for a particular prefab, you can simply create a variant and add extra config / component to that variant to handle its particular initialization logic.

### Notes:
This requires some explanation.

First of all, all of the cloths are now in a prefab variant structure, under ClothKiller folder (to protect them from the upcoming Jibby PR). It very nicely aligns with the DM structure of inheritance.

Second of all, the remnants of the previous way of doing things are not fully eliminated yet! This is to try to avoid potential merge issues. In future PRs, this stuff will gradually be eliminated. Most notably:

- ClothingV2 was added to be a "better" version of Clothing, and a temporary interface IClothing was built to simplify the migration. The old clothing and interface will eventually be removed
- Same goes for ItemAttributesV2, I will need to migrate the prefabs that are currently using ItemAttributes to use the new ItemAttributesV2
- The ONLY purpose of ClothData and the other "data" assets now is defining their sprites. ItemAttributes and other stuff is all ignored, except for headset key. These are all now defined on the prefab variants. All of the extra fields are still left on these SOs however, to prevent potential merge issues. These will be removed in future PRs, and it's possible that the different subclasses of BaseClothData will be merged into one if they no longer have a need to have lots of different fields.
- The various SOs which referenced spawnables / slot populators still have those fields defined to prevent potential merge issues, but most of these will eventually be removed. I say "most" because I am planning to maybe remove the custom slot populators like AutoIDPopulator so that a prefab could  have slot population logic in it instead in a component using some sort of hook interface, as I really like the simplicity of all prefabs having self-contained logic of how to init themselves, and using prefab variants to achieve that. And it declutters the storage populator SOs so there's no confusion about whether you need to pick a prefab or pick a slot populator asset.

There are a number of additional changes planned in upcoming PRs, but wanted to get the minimum stuff merged ASAP so as not to block other's work. As it stands it should function correctly. Here's the stuff I'm planning in addition to the removals mentioned above:

- Rename the NamedSlot.exosuit to NamedSlot.outerwear and NamedSlot.uniform to NamedSlot.suit to try to better match what those slots actually represent
- Fix the belt setups, will get @StrawberryMoses to help with this. At the moment we have janitor, CE, and security belts but I am not sure that they have the correct structure, capacity and initial items. I think some of moses' changes may not have ended up in the PR.
- Delete all the code relating to cloth lifecycle stuff, as it shouldn't be needed anymore - no more Spawn.ServerCloth, no more cloth dictionaries, cloths in resource tracker, Uni-item prefabs
- Add a plugin which allows searching for things that reference a particular asset. This will be ESSENTIAL as we add more assets to the game, as it allows us to see what might be affected when changing or migrating a particular asset. Unity doesn't provide a built in way to do this!
- Move the ClothKiller stuff into its correct place after Jibby's PR

Wanted to get this PR up ASAP, I am in the process of updating the wiki with the new structure. I also want to flesh out information on how to create objects / cloths so it's a lot easier for people to do, coupled with the changes in this PR.


StrawberryMoses:
I spawned in as the captain and I am naked

Hovering over the belts doesn't seem to show a name or description

The belts also all seem to have the default belt sprite

Final minor nitpick for now, i would rename the belts to be just, for example JanitorBelt and JanitorBeltFull, since imo the full is just a child/variant of the belt itself which doesn't need to specify it's state

chairbender:
some stuff didn't get pushed, I'll tweak those things

chairbender:
@StrawberryMoses should be fixed now. Make sure you do it with a clean branch - close unity, git reset --hard, reopen unity.

chairbender:
EDIT: New wiki page on creating objects which somewhat covers the changes in this pr is here: https://github.com/unitystation/unitystation/wiki/Creating-Items-and-Objects%3A-Now-With-Prefab-Variants

I also updated the inventory system as slot populators are slightly changed (to hopefully be easier to use) https://github.com/unitystation/unitystation/wiki/Inventory-System

fomalsd (approved):
Looking good!

fomalsd:
alright, going to merge this and port janitor tweaks to the new system

PR #2372, opened 2019-11-28
Resources Folder Re-Structure

Jibbajabbafic:
### Purpose
Cleaned up some stray folders and made everything a bit more consistent and re-saved the OutpostStation scene which was causing warnings/errors.

I tested each commit as I went and I noticed a couple errors but I don't think they were related to these changes, let me know if that isn't the case.

There are still a lot of stray one off Resource.Loads (mostly on legacy items such as guns) which should be replaced with GameObject references so we can more easily move away from the Resources system in the future.

Also removed some stray textures and sprites from the resources folder which weren't needed anymore.

- []  My code is indented with tabs and not spaces
- [ ]  Any new / changed components have been [tested using Respawn](https://github.com/unitystation/unitystation/wiki/Proper-Spawning-and-Despawning-%28Object-Lifecycle%29)


chairbender:
works for me, thanks for getting this done

PR #2395, opened 2019-11-29
Adds descriptions for ItemTrait instances

### LINKED ISSUE #2351: Add description to ItemTrait
# Description
Super easy. As we add more item traits it might become hard to get a clear idea of what an ItemTrait represents. 

Just want to add a description field to the ItemTrait SO and go through and fill it in for each ItemTrait if you have the motivation (at least fill it in for a few). The purpose of this field is simply to allow designers / devs to see what an ItemTrait represents if the name is not clear enough.

For readability I would suggest making description a text area rather than a text field so you don't need to scroll the field to read the full description.

Ilysen:
Hello! This is my first time (and possibly only time, depending on if I can catch onto the code) contributing here. I may have broken some guidelines by accident. Let me know if I've done anything wrong and I'll do what I can to fix it.

### Purpose
This MR adds a TextArea string to the base `ItemTrait` class that's meant to describe what each trait does, and fills out that field on each current existing instance of that class. Each of the descriptions is based on my currently-limited understanding of how objects are implemented and interact with each other, so feel free to correct me if I've made a mistake.

Any trait that didn't seem to be implemented yet was given a description stating that fact.

I also renamed `cultivator.asset` to `Cultivator.asset` to stay in line with the naming scheme for the other assets.

Fixes #2351.

Since this change is purely for editor/developer use, it has no in-game changes, so I haven't done any testing beyond booting and making sure traits still work properly, which they do.

### Notes:

### DIFF
--- a/UnityProject/Assets/Resources/ScriptableObjects/Traits/Magazine/bucket.asset
+++ b/UnityProject/Assets/Resources/ScriptableObjects/Traits/Magazine/bucket.asset
@@ -12,3 +12,4 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 4ccfe138752d4b66bcb7c56a0c9769c7, type: 3}
   m_Name: bucket
   m_EditorClassIdentifier: 
+  traitDescription: Describe me!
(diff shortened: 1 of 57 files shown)


UnityProject/Assets/Resources/ScriptableObjects/Traits/Magazine/bucket.asset
```diff
@@ -12,3 +12,4 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 4ccfe138752d4b66bcb7c56a0c9769c7, type: 3}
   m_Name: bucket
   m_EditorClassIdentifier: 
+  traitDescription: Describe me!
```
fomalsd, line 15: well, it’s a bucket

fomalsd (commented):
Thanks for contributing! 
It’s all good, just missing a description for bucket, which is no biggie
Just so you know, unless you’re on case sensitive file system (linux), simply changing filename case without adding/removing symbols won’t make git register the change. So from my experience it requires two commits: `cultivator -> CultivatorProper -> Cultivator`

PR #2397, opened 2019-11-29
Adds some messages in chat for the hydroponics tray and player death

Ilysen:
### Purpose
In SS13, chat is a crucial part of gameplay and the main way that visual feedback is conveyed, but due to examining currently being very barebones, there's no real way to see the state of things without guessing.

This PR adds the iconic death message, which appears whenever someone dies (`(player) seizes up and falls limp, his/her/its eyes dead and lifeless...`) and adds some messages to the hydroponics tray.

### Notes

I've also added a public string function to character settings, `PossessivePronoun()`, which returns a the possessive word for that player's gender - currently either "his", "her", or "its".


UnityProject/Assets/Scripts/Health/Living/HumanHealth/PlayerHealth.cs
```diff
@@ -76,6 +76,8 @@ protected override void OnDeathActions()
 			if (isServer)
 			{
 				EffectsFactory.BloodSplat(transform.position, BloodSplatSize.large, bloodColor);
+				string descriptor = player.Script.characterSettings.PossessivePronoun();
+				Chat.AddLocalMsgToChat($"<b>{playerName}</b> seizes up and falls limp, {descriptor} eyes dead and lifeless...", (Vector3)TransformState.HiddenPos);
```
chairbender: When you add a local message you need to tell the chat system where it's happening at, which is what that last parameter is for. You should pass registerPlayer.WorldPositionServer here which represents the position the server thinks this player is at, instead of HiddenPos. The message wasn't showing up for me when testing with killing dummy player. HiddenPos is a special position used for things that aren't supposed to be visible to the players, which shouldn't be passed to this method.

chairbender (changes requested):
good stuff! Good idea to add a pronoun method as well. Just need to fix the position of the local msg when they die.

UnityProject/Assets/Scripts/Health/Living/HumanHealth/PlayerHealth.cs
Ilysen: oh, alrighty, thanks. I'll patch it up once I get a minute!

Ilysen:
Apologies for taking a while - work has had me hugely busy since we just had a major release. I haven't forgotten about this.

Bod9001:
is ok we've had PRs open for much longer : P

Ilysen:
Should be ready to go now. I changed the possessive word for neuter genders to "their" instead of "its", and changed the position to what should be correct. I haven't been able to test it myself, but I can if you need me to.

chairbender:
@Ilysen works for me now. One thing to call out, though I'll still merge it, is that if you do Networking > Spawn Dummy Player and kill the dummy you will get an NRE because they have no characterSettings. It's more an issue with our dummy player, they should have some placeholder character settings.

Ilysen:
oh, sorry about that. I'll put in a sanity check once I can.

PR #2401, opened 2019-11-30
2261 Add prices to item attributes, add cargo scanner

### LINKED ISSUE #2261: Cargo: Add ability to check prices and sell things
# Description
In order to be able to make money from mined ores so you can buy fun stuff, we need to be able to sell things by sending them back on the cargo shuttle.

As part of that, we should add a scanner item that tells you the sell price of something (if it has a value).

At the very least, you should be able to put things onto the shuttle and when the shuttle docks at centcomm it should despawn them and add their sell price to the current money (not sure where this is tracked).

This will require a way to define a sell price of things, which most likely would be a good fit for ItemAttributes.

TODO: Someone who knows more can fill in any important details for how this should work

StrawberryMoses:
Fixes #2261 

### Purpose
Adds the cargo export scanner, and adds export fields to item attributes.
In the future, the fields will need to be added to objects as well, to handle crates.

### DIFF
--- a/UnityProject/Assets/Scripts/Cargo/ExportScanner.cs
+++ b/UnityProject/Assets/Scripts/Cargo/ExportScanner.cs
@@ -1,0 +1,33 @@
+using UnityEngine;
+
+public class ExportScanner : MonoBehaviour, ICheckedInteractable<HandApply>
+{
+	public bool WillInteract(HandApply interaction, NetworkSide side)
+	{
+		if (!DefaultWillInteract.Default(interaction, side))
+		{
+			return false;
+		}
+
+		if (interaction.HandObject != gameObject)
+		{
+			return false;
+		}
+
+		return true;
+	}
+
+	public void ServerPerformInteraction(HandApply interaction)
+	{
+		var attributes = interaction.TargetObject.GetComponent<ItemAttributesV2>();
+		var price = attributes ? attributes.GetExportCost() : 0;
+		var exportName = interaction.TargetObject.ExpensiveName();
+		var message = price > 0
+			? $"Scanned { exportName }, value: { price } credits."
+			: $"Scanned { exportName }, no export value.";
+
+		// TODO #2400 if it has contents say contents included.
+
+		Chat.AddExamineMsg(interaction.Performer, message);
+	}
+}
--- a/UnityProject/Assets/Scripts/Cargo/ExportScanner.cs
+++ b/UnityProject/Assets/Scripts/Cargo/ExportScanner.cs
@@ -20,13 +20,13 @@ public class ExportScanner : MonoBehaviour, ICheckedInteractable<HandApply>
 	public void ServerPerformInteraction(HandApply interaction)
 	{
 		var attributes = interaction.TargetObject.GetComponent<ItemAttributesV2>();
-		var price = attributes ? attributes.GetExportCost() : 0;
+		var price = attributes ? attributes.ExportCost : 0;
 		var exportName = interaction.TargetObject.ExpensiveName();
 		var message = price > 0
 			? $"Scanned { exportName }, value: { price } credits."
 			: $"Scanned { exportName }, no export value.";
 
-		// TODO #2400 if it has contents say contents included.
+		// TODO #2400 if it has contents add " (contents included)"
 
 		Chat.AddExamineMsg(interaction.Performer, message);
 	}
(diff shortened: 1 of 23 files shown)


UnityProject/Assets/Scripts/Items/ItemAttributesV2.cs
```diff
@@ -159,6 +161,16 @@ public IEnumerable<string> ServerAttackVerbs
 		set => attackVerbs = new List<string>(value);
 	}
 
+	[Tooltip("How much does one of these sell for when shipped on the cargo shuttle?")]
+	[SerializeField]
+	private int ExportCost; // Use GetExportCost to obtain!
+
+	[Tooltip("Should an alternate name be used when displaying this in the cargo console report?")]
+	public string ExportName;
```
chairbender: This and ExportMessage should be private SerializeFields with public get-only properties (just like initialName above). A little verbose, but it protects against people thinking they can use it to change the value dynamically. If this was changed server-side, the change would not be synced to the client. If it was changed client-side, the change would not be propagated to the server.

chairbender: ...and thank you for adding tooltips, really helps a lot with usability.

UnityProject/Assets/Scripts/Items/ItemAttributesV2.cs
```diff
@@ -159,6 +161,16 @@ public IEnumerable<string> ServerAttackVerbs
 		set => attackVerbs = new List<string>(value);
 	}
 
+	[Tooltip("How much does one of these sell for when shipped on the cargo shuttle?")]
+	[SerializeField]
+	private int ExportCost; // Use GetExportCost to obtain!
```
chairbender: this should be called exportCost since it's private (add FormerlySerializedAs["ExportCost"] so you don't have to re-add the prices).

UnityProject/Assets/Scripts/Items/ItemAttributesV2.cs
```diff
@@ -357,5 +369,15 @@ public void ServerSetSize(ItemSize newSize)
 		SyncSize(newSize);
 	}
 
+	public int GetExportCost()
```
chairbender: It's a pretty insignificant thing, but if you make the above change to make exportCost lowercase, you can change this to be a public get-only property with custom get logic so it's a little nicer to access, a user of this component can call `itemAttributes.ExportCost` rather than `itemAttributes.GetExportCost()`

Just change it to look like this:
```csharp
	public int ExportCost
	{
		get
		{
			var stackable = GetComponent<Stackable>();

			if (stackable != null)
			{
				return exportCost * stackable.Amount;
			}

			return exportCost;
		}
	}
```

UnityProject/Assets/Scripts/Cargo/ExportScanner.cs
```diff
@@ -0,0 +1,33 @@
...
+	public void ServerPerformInteraction(HandApply interaction)
+	{
+		var attributes = interaction.TargetObject.GetComponent<ItemAttributesV2>();
+		var price = attributes ? attributes.GetExportCost() : 0;
+		var exportName = interaction.TargetObject.ExpensiveName();
+		var message = price > 0
+			? $"Scanned { exportName }, value: { price } credits."
+			: $"Scanned { exportName }, no export value.";
+
+		// TODO #2400 if it has contents say contents included.
```
chairbender: thanks for creating a TODO and corresponding ticket for it. I updated that ticket with the info on how to get the contents but you can do that in future PR if you don't want to do it now.

chairbender (changes requested):
Thanks for this, very glad to have the ability to earn money properly, will be essential for cargo to be fun! Just a few small changes related to naming / access. They should be quick to make.

UnityProject/Assets/Scripts/Items/ItemAttributesV2.cs
StrawberryMoses: Good point- changed them

UnityProject/Assets/Scripts/Items/ItemAttributesV2.cs
StrawberryMoses: Ehh this wasn't even merged yet, so I just redid the prices (at least, I hope I got all of them)

UnityProject/Assets/Scripts/Items/ItemAttributesV2.cs
StrawberryMoses: yep that looks better

UnityProject/Assets/Scripts/Cargo/ExportScanner.cs
StrawberryMoses: I'll do it in a future PR, don't want to hold this one up too much so bod can get base attribute stuff in smoother

PR #2405, opened 2019-12-02
2191 Generate Tile Palette Semi-Automatically

### LINKED ISSUE #2191: Automatically add new object prefabs to Tile Palette
# Description
After sprite refactor most object-containing tile palettes became empty.
image
Adding objects to scene manually without palette is a pain in the ass;
Adding newly created item to tile palette is a bigger pain in the ass and no one wants to do it manually (which is totally understandable).
We need to ensure that all spawnable objects can be found in tile pallettes

StrawberryMoses:
### Purpose
Fixes #2191. Using a button in the toolbar, tile palettes can be generated given a list of folders to pull from per palette. In the future I would love to have this additionally run through every prefab, grab a list of tags from BaseAttributes (IE, for the chefs knife, "Kitchen", "Chef", "Melee Weapon") and generate more palettes per tag. Thoughts on that?

- [ ]  Any new / changed components follow the [Component Development Checklist](https://github.com/unitystation/unitystation/wiki/Component-Development-Checklist)
- [ ]  Any new objects / items follow the [Creating Items and Objects Guide](https://github.com/unitystation/unitystation/wiki/Creating-Items-and-Objects%3A-Now-With-Prefab-Variants) (especially concerning the use of prefab variants)


Bod9001:
hummm Could add a component that runs when you're editing the prefab variant that automatically generates the tile asset and checks against a SO to make sure it is not already generated?

fomalsd (approved):
good stuff, thanks!
are palettes gitignored, by the way? had to generate them myself and don't really see relevant created/changed files in my changelist

re: more palettes per tag
I suspect it will be easier to lose palettes like Walls if there will be too many of these, but it depends on total amount of palettes

StrawberryMoses:
Palettes and their generated tiles are gitignored- they both go in relevant /Generated folders. The idea was that we could still make custom palettes and these would be good for bulk amounts of items.
However, the preview icons aren't ignored, I was using code from Prefab2Tile to do that

fomalsd:
I don't really see why they should be gitignored – people will expect to see filled palettes when they first discover Tile Palette tab.
And modified palettes could be supplied when new items are added, so that other people don't have to do it unless they add something new themselves

StrawberryMoses:
Makes sense to me. Will either webedit today or will do after work later

PR #2442, opened 2019-12-08
Computer construction + generic object construction helpers

### LINKED ISSUE #2249: Construction System
# Description
EDIT: #1870 needs to be done first as construction relies on stacks of things

We want a system for defining the steps for constructing / deconstructing things, so construction can be implemented / changed without code changes (assuming the necessary materials / prefabs are already defined). This is already partially implemented in ConstructionHandler to show the sorts of features the system should have, but I think it would be best to define this using SOs so that the steps of construction aren't hardcoded into a particular prefab.

We haven't fleshed out the design of this, but some general capabilities would be:
1. Construction consists of multiple stages, each which has some requirements to be met in order to advance to the next stage. I'm not sure if each stage would be a separate Prefab, or if there would be one Prefab and each stage would simply change the sprite. Probably you could have a mix - a given construction chain might go through several stages, each of which is just a sprite change, and end on despawning the in-progress object and replacing it with an actual prefab of the thing that was constructed. IDK
1. Probably each stage / state would have an optional associated hook (such as how a state machine might work), so that arbitrary logic can be run when the stage is reached - it could be changing a particular aspect of the object (like making door impassable) or just despawning and spawning a new prefab in its place.
1. The requirements for each stage would be defined using ItemTraits to describe which materials must be applied to the object and which tool should be used to advance it to the next stage
1. I'm not sure what other requirements there might be in order to advance to a given stage, so it would be good if the way requirements are defined is easily extensible to support those possibilities
1. An in-progress construction stage can use the inventory system to store the materials that have been applied to that stage.
1. Should be possible to construct / deconstruct stuff.
1. It should be documented.

### LINKED ISSUE #2451: Make computers eject contents upon disassembly
# Description
Computers should eject anything they might have inside (like IDs) upon disassembly, currently you just lose them

## Steps to Reproduce
1. Put your id into comms console
2. Screwdriver it
3. Your id is gone forever

chairbender:
### Purpose
Fixes #2249 
Fixes #2451 

It doesn't fully implement all construction logic, but at this point I feel that the various logic I have added to support the currently implemented constructable things is enough of a base to build off of and provides working examples of the major kinds of construction use cases, so I'm marking the ticket as complete. I also think we've implemented the bare minimum of what would be needed for 0.4.

Adds full construction logic for consoles. Adds computer circuit boards for each console.

Cable coils are now stackable.

Fixes various issues with construction, such as being able to build or anchor things on top of each other.

Adds a fair amount of supporting logic to make it easier to write construction code for components. Adds more convenience methods to ToolUtils to reduce code duplication. Adds more methods to Validations, also adds ServerValidations for common validation logic that is only done server side.

Welder now uses reagent container and can't be used for construction when it's off.

Fixes expensivename.

Fixes cloning console to auto-detect adjacent cloning pod and dna scanner.

Fixes shuttle console to not be openable if it's on a matrix which can't be moved.

Fixes the issue where you couldn't stack things that were mapped.

Makes grills wire-cuttable again. Grills don't yet have the full construction logic of tg (being able to unanchor and drag them) but the existing construction logic should be a good reference for how to do it.


UnityProject/Assets/Scripts/Construction/ComputerFrame.cs
```diff
@@ -0,0 +1,245 @@
...
+				{
+					//stick in the circuit board
+					Chat.AddActionMsgToChat(interaction, $"You place the {interaction.UsedObject.ExpensiveName()} inside the frame.",
+						$"{interaction.Performer.ExpensiveName()} places the {interaction.UsedObject.ExpensiveName()} inside the frame.");
+					Inventory.ServerTransfer(interaction.HandSlot, circuitBoardSlot);
+				}
+				else if (Validations.HasUsedItemTrait(interaction, CommonTraits.Instance.Screwdriver) && circuitBoardSlot.IsOccupied)
+				{
+					//screw in the circuit board
+					Chat.AddActionMsgToChat(interaction, $"You place the {interaction.UsedObject.ExpensiveName()} inside the frame.",
```
fomalsd:
```suggestion
					Chat.AddActionMsgToChat(interaction, $"You screw {circuitBoardSlot.ItemAttributes.ArticleName} into place.",
```

UnityProject/Assets/Scripts/Construction/ComputerFrame.cs
```diff
@@ -0,0 +1,245 @@
...
+						$"{interaction.Performer.ExpensiveName()} places the {interaction.UsedObject.ExpensiveName()} inside the frame.");
```
fomalsd:
```suggestion
						$"{interaction.Performer.ExpensiveName()} screws {circuitBoardSlot.ItemAttributes.ArticleName} into place.");
```

fomalsd (approved):
very nice!
when disassembling a computer, wires you get back become SingleCableCoil, even if you put CableCoil before, therefore they no longer stack.
have a couple more observations but I'll make an issue for them

chairbender:
@fomalsd good to go. Fixes #2451 as well

chairbender:
also fixed an unrelated issue where adding plasma to generator consumed entire stack

PR #2445, opened 2019-12-08
Bugfixes - Round Restart, ID card logic

### LINKED ISSUE #2444: ID weirdness on headless
# Description
There are a few things wrong with IDs that we have encountered while testing build 3930 on dev server:

- Any ID picked up by captain became all access (!), and even started looking golden when on floor, while name and role name remained the same
img

- Any picked up ID became recolored to player's own initial ID while in local inventory:

img

- After round restart players spawned without IDs (but that might be a round restart-related bug)

Also: if you have an ID in your ID slot and another one in your hand slot, ID slot one is used for door access check. Not sure if this is right.

chairbender:
### Purpose
Fixes #2444 

There also was an embarassingly bad issue caused by a flaw in lifecycle system. It didn't clear the pools on round restart, causing all sorts of problems. Now round restart seems to work a lot better. Sorry.

### DIFF
--- a/UnityProject/Assets/Scripts/Items/IDCard.cs
+++ b/UnityProject/Assets/Scripts/Items/IDCard.cs
@@ -96,6 +96,7 @@ public class IDCard : NetworkBehaviour, IServerInventoryMove, IServerSpawn
 			//auto init if being added to a player's inventory
 			if (info.ToPlayer != null)
 			{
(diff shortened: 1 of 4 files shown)


chairbender:
going to merge now since it's quite small and seems to fix a lot of problems with round restarts.

PR #2474, opened 2019-12-20
Fixed Issue #2032; Players can now move when cuffed

### LINKED ISSUE #2032: Handcuffs stop player movement B: 35

JohnHandley:
Removed the check for the handcuffs when attempting a move.

### Purpose
Fixes #2032

### Notes:
Don't care about any bounty, just wanted to try out contributing if it's useful.

### DIFF
--- a/UnityProject/Assets/Scripts/Player/PlayerMove.cs
+++ b/UnityProject/Assets/Scripts/Player/PlayerMove.cs
@@ -147,7 +147,9 @@ public class PlayerMove : NetworkBehaviour, IRightClickable, IServerSpawn
 			// If player attempts to move
 			if (KeyboardInputManager.CheckMoveAction(moveList[i]))
 			{
-				if(allowInput && !IsBuckled){
+				bool beingDraggedWithCuffs = IsCuffed && PlayerScript.pushPull.PulledBy != null;
+				
+				if(allowInput && !IsBuckled && !beingDraggedWithCuffs){
 					actionKeys.Add((int)moveList[i]);
 				}
 				else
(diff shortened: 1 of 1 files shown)


DooblyNoobly (changes requested):
The cuffed player shouldn't be allowed to move if he is being dragged by another player or is buckled to a chair or bed

JohnHandley:
Ah, I see, will try to correct it.

JohnHandley:
I believe the latest commit resolves the requested change.

DooblyNoobly:
cool thanks

DooblyNoobly:
Reviewed it again and it seems to introduce a player sync bug.

Too recreate:
1) Make a build and host a game
2) connect with editor
3) get some handcuffs while playing as the editor player and handcuff the other player
4) start dragging the player
5) switch to build and run away
6) notice that you can run away while cuffed and that it also breaks the player sync

A gif of the bug:
movebug

fomalsd:
thing is, client thinks it's allowed to move predictively.
need to check for handcuffs on clientside, too, and don't allow predictive movement if cuffed and pulled by someone.
it's something among these lines: 
https://github.com/unitystation/unitystation/blob/a7eae55556775256a378949378a95d22c2379bbb/UnityProject/Assets/Scripts/Player/PlayerSync.cs#L486-L526
should avoid calling DoAction() in that particular case, either by setting `didWiggle` to true or otherwise

JohnHandley:
Thanks, I'll try to fix that. Thanks for the tip!

JohnHandley:
I switched to use the what I believe is the client version of the beingPulled check which seems to resolve the syncing issues. I could be approaching this the wrong way. Still figuring out the client and server relationship but this seems to work so far.

DooblyNoobly:
Thank you, its all working now

PR #2478, opened 2019-12-20
Added more stuff to get from cargo

Bod9001:
### Purpose
addes more items to the cargo, still could use more items


StrawberryMoses:
Can you up the prices on these? Assuming they come in crates, the crates can be immediately returned for 500 credits :)

Bod9001:
oh ok Let me do that

PR #2482, opened 2019-12-22
"Fixes" sprite Handler

Bod9001:
### Purpose
also another poke to see if git will properly track the admin folders

this should get rid of the error when dropping items

### Notes:
it's kind of a ghetto way of fixing it but it would work

tested in editor
haven't tested in headless 

I don't think this is worth a bounty so you could just put it back in the pool or add it to my next Bounty

### DIFF
--- a/UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
+++ b/UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
@@ -43,7 +43,8 @@ public class SpriteHandler : SpriteDataHandler
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


DooblyNoobly:
whats this about the admin folders? `also another poke to see if git will properly track the admin folders`

Bod9001:
when I updated my fork, the folders for the admin.txt weren't generated so it errored and broke everything

DooblyNoobly:
yikes, merging now

Bod9001:
it's only differences so it might be needed to be tested a bit more

PR #2497, opened 2019-12-29
Window knocking 

### LINKED ISSUE #2469: Knocking on Windows B: 35
# Description

- Add the ability to knock on windows
- Make sure the soundfx is network synced and varies in pitch on each knock

Contributor will receive a bounty of 35usd on a successful merge of a solution that resolves the issue above.

Bod9001:
### Purpose
Fixes #2469
and some auto format
### Notes:
 I've had to Change CmdRequestMeleeAttack  to Except the intent
so There might be some funny Behaviours caused because of this

This could be implemented in a better way but no one said anything about it on Discord so I went ahead with it

Tested in editor
Will test in build, editor Multiplayer /
Renamed functions to appropriate names


chairbender:
@Bod9001 this isn't quite the right way to do this as far as the interaction logic is concerned.  This is a tile interaction, so it should use the Tile Interaction system: https://github.com/unitystation/unitystation/wiki/Implementing-Interactions#how-can-i-create-tile-interactions

This ticket shouldn't require changing the CmdRequestMeleeAttack or TilemapDamage. All that should be needed is to 
  * modify Meleeable, moving LayerType.Windows into harmIntentOnlyAttackbleLayers so you'll only bash windows on harm intent
  * Create a new TileInteraction SO which performs the tapping validation / logic and add it to the Window tile asset's tileInteractions field.

There shouldn't be any other changes than that (except the ones related to sound manager)

chairbender (changes requested):
@Bod9001 this isn't quite the right way to do this as far as the interaction logic is concerned.  This is a tile interaction, so it should use the Tile Interaction system: https://github.com/unitystation/unitystation/wiki/Implementing-Interactions#how-can-i-create-tile-interactions

This ticket shouldn't require changing the CmdRequestMeleeAttack or TilemapDamage. All that should be needed is to 
  * modify Meleeable, moving LayerType.Windows into harmIntentOnlyAttackbleLayers so you'll only bash windows on harm intent
  * Create a new TileInteraction SO which performs the tapping validation / logic and add it to the Window tile asset's tileInteractions field.

There shouldn't be any other changes than that (except the ones related to sound manager)

Bod9001:
All fixed up

UnityProject/Assets/Scripts/Tilemaps/Behaviours/TilemapDamage.cs
```diff
@@ -24,12 +24,12 @@ public class TilemapDamage : MonoBehaviour, IFireExposable
```
chairbender, line 24: Are there any changes in this file other than formatting changes? Looks like only formatting changes, which is fine.

chairbender (changes requested):
Wow didn't take you long to figure out and fix. Had an issue with missing script error probably due to the naming of the window knock, also added an interaction message for it. See my PR: https://github.com/Bod9001/unitystation/pull/4

Once that's merged it's good to merge.

UnityProject/Assets/Scripts/Tilemaps/Behaviours/TilemapDamage.cs
Bod9001: yeah, Probably is because I copied it over from the raw on github to reverse the changes I did

Bod9001:
yeah the Hard part is finding the right systems to use 99% of the time : P

PR #2500, opened 2019-12-29
Makes fire spread work now

### LINKED ISSUE #2438: Large fires no longer happen?
# Description
Fires no longer expand regardless of plasma/oxygen amounts in the room.
Reproducible locally.

## Steps to Reproduce
1. Spawn in atmosia
2. Open plasma and oxygen tanks
3. Turn on a welder or artificially hotspot some item via right click menu
4. Observe a small fire that dies down in a second, while it should spread like hell and take over the station

Bod9001:
### Purpose
It kept on going out before it could spread properly so I reduce the temperature it Goes out at
Fixes #2438

Tested in editor
No need for multiplayer since it's just changing some Default variables

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/Reactions.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/Reactions.cs
@@ -5,7 +5,7 @@ namespace Atmospherics
 	public static class Reactions
 	{
 		public const float T0C = 273.15f;
-		public const float PLASMA_MINIMUM_BURN_TEMPERATURE = T0C + 100;
+		public const float PLASMA_MINIMUM_BURN_TEMPERATURE = 200;
 		public const float FIRE_MINIMUM_TEMPERATURE_TO_SPREAD = T0C + 150;
 		public const float PLASMA_UPPER_TEMPERATURE = T0C + 1370;
 		public const float OXYGEN_BURN_RATE_BASE = 1.4f;


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/Reactions.cs
```diff
@@ -5,7 +5,7 @@ namespace Atmospherics
 	public static class Reactions
 	{
 		public const float T0C = 273.15f;
-		public const float PLASMA_MINIMUM_BURN_TEMPERATURE = T0C + 100;
+		public const float PLASMA_MINIMUM_BURN_TEMPERATURE = 200;
```
Jibbajabbafic, line 8: This feels like a bit of a hacky fix since 200K is a very low temp for a fire.

The old value is from the original atmospherics as well so this would change the behaviour of plasma quite a bit:
image

chairbender: I agree, I think the ticket is more about figuring out why the fires don't seem to spread anymore and if that's actually a problem, considering it seemed to be possible before. Apparently Bod is already able to cause large fires somehow anyway without this fix?

Bod9001: yeah, you have to be persistent with the hot Spot until the room gets hot enough to cause it, I think one of the issues is that the game is so fast that it can't have time to build up heat inside of the area it's on fire so It goes out, Whilst in the byond version it's very slow and Has plenty of time to heat up the tiles the fire is on

Bod9001: It seems to work roughly the same as byond ss13 with the 200k Setting,
apart from the massive fire explosion but I think that was there before The issue

chairbender: then maybe the issue is heat dissipating too quickly?

chairbender: heck, maybe it's not even an issue? I don't have a good sense of how easy it is to cause fires in us vs tg and whether it even should be changed.

Bod9001: TG is Easy py, Only time I haven't seen fires spread with a Welding torch when I think there was too much plasma and too little oxygen And the torch was being dragged around as well, 99%  of time One Spark will set a massive fire off It's kind of that balance, like detective walks in with lit cigarette, One other issue is that Currently everything gets broken way to quickly From fire, like on TG doors never break to fire and Glasse as well, but Glass only breaks when there is a high pressure difference, and I find everything Burns up too quickly Something a super hot fire should last about 30 Seconds That's for items and  a lot of stuff doesn't even break if it's a gentle fire like Machines but this can be addressed later on

Bod9001: *never breaks, I think doors only break when there's fire inside of itself

chairbender: okay, I think this issue should focus on fixing fires so they don't go out so quickly, without changing the temperature constants.

Bod9001: if so, I should wait until foma has finished there atmospheric stuff since it might fix it

chairbender: I think it's fine to work on now, the atmos stuff foma is working on is NaN infection, which I don't think happens consistently with fires on OutpostStation, and shuttle crashing which also is an occasional thing, so this fix would already be on the happy path which should be protected from those bugs.

Bod9001: If it is Strange behaviour, Since I remember it working before It might be fixed in one of those bugs

Jibbajabbafic:
Based on the discussion there will need to be a better fix for this issue, so this PR can be closed for now.

PR #2521, opened 2020-01-02
Made combat more smooth

### LINKED ISSUE #2053: Accidentally triggering Player drag to when fighting people
# Description

so Since they drag player to thing mechanic, In combat your mouse moves a lot And of course that accidentally triggers the drag scenario instead of actually the intent you want to do this can be frustrating for Disarming and attacking since Since that blocks the attack
I did highlight this would be an issue With input

https://cdn.discordapp.com/attachments/312454684021620736/603207113325543447/Desktop_2019.07.23_-_13.49.16.28.mp4

example video

## Category
BUG

## Detailed Bug Report

Should have a bigger Dead zone so it doesn't trigger accidentally in combat

## Steps to Reproduce
1. get Two clients
2. Try running around and spamming Disarm on On the other client
3. What the drag UI Pop-up for a tiny second Instead of hitting/Disarming

Bod9001:
### Purpose
Got rid of the input Lock When attacking
Made it so you can't Trigger their drag to when in Harm intent
Fixes #2053
Tested with built client Connecting With another instance On the same computer

### note
humm Should I remove them Instead of commenting them out?

### DIFF
--- a/UnityProject/Assets/Scripts/Weapons/WeaponNetworkActions.cs
+++ b/UnityProject/Assets/Scripts/Weapons/WeaponNetworkActions.cs
@@ -152,7 +152,7 @@ public class WeaponNetworkActions : ManagedNetworkBehaviour
 				{
 					SimpleAnimal attackTarget = victim.GetComponent<SimpleAnimal>();
 					RpcMeleeAttackLerp(attackDirection, weapon);
-					playerMove.allowInput = false;
+					//playerMove.allowInput = false;
 					attackTarget.Harvest();
 					SoundManager.PlayNetworkedAtPos("BladeSlice", transform.position);
 				}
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Weapons/WeaponNetworkActions.cs
```diff
@@ -152,15 +152,15 @@ public void CmdUnloadWeapon(GameObject gunObject)
 				{
 					SimpleAnimal attackTarget = victim.GetComponent<SimpleAnimal>();
 					RpcMeleeAttackLerp(attackDirection, weapon);
-					playerMove.allowInput = false;
+					//playerMove.allowInput = false;
```
daalta, line 155: I think it's OK to remove those lines. Or you could add a comment to explain why the lines are commented.

chairbender:
seems to be working fine on moving matrices as well

Issue #1745, opened 2019-04-16
Spamming left-right-left-right movement keys behaviour


Aranclanos:
# Description
If you move left and right fast, the game seems to miss or ignore some keypresses, making you move twice on the same direction.
The idea is to stay on an area of two turfs while constantly moving

## Category
BUG

## Steps to Reproduce
Please enter the steps to reproduce the bug or behaviour:
1.try to move back and forward in the same two turfs/tiles by only pressing left and right
2.Notice how the game doesnt respect your keypresses and moves you out of these two turfs

if this doesnt make much sense tell me, I'll try to explain it better

fomalsd:
Ahh, I think it's due to `MoveCooldown` in PlayerSync.Client that filters out move requests that happen faster than a decisecond. Was designed to avoid spamming server with move requests

coolcat1711:
> 
> 
> Ahh, I think it's due to `MoveCooldown` in PlayerSync.Client that filters out move requests that happen faster than a decisecond. Was designed to avoid spamming server with move requests

My gut response would be to implement a small move queue and update the server when off cooldown. The only problem would be if it would make movement feel artificially unresponsive if you filled the queue with too many 'In advance' commands.

chairbender:
I like the queue idea, I think it's a common practice to allow "buffering" a single move in many games. I think that would make sense here - just allow them to buffer a single move.

chairbender:
not a good first issue because movement is involved

daalta:
A movement buffer could definitely make sense, but I would clarify this to make clear *how* it's a bug. (Because I wasn't sure at first)

1. The player presses `D` and their character starts moving to the right.
2. Shortly before the the character stops moving the player presses `W`. 
3. The character finishes their movement to the right.
4. The character does not move left - the `D` input was dropped.

Instead, what should have happened:

4. The characters moves to the left because `D` was pressed within the last x milliseconds.

===

Also, here's how I would not consider a bug: When spamming.

1. The player presses `A`.
2. The character starts moving to the right.
3. The player presses `D`.
4. The player presses `A`.
5. The player presses `D`.
6. The player presses `A`.
8. The characters stops moving.
9. The player presses `D`.
10. The character moves to the right.

Even though the player was spamming `D` and `A` they moved to the right twice. This is because it was the most recently pressed movement key. Even if we added a buffer this would not be fixed, unless the player paces their `D` and `A` presses perfectly.

MaxIsJoe:
[...]

PR #2523, opened 2020-01-03
Various microwave improvements, and player movement can cancel itself

### LINKED ISSUE #2527: Holding opposite movement keys does not move players but produces footsteps
# Description

Holding two different movement keys (UP and DOWN, LEFT and RIGHT, or UP, DOWN, LEFT and RIGHT) continuously produces footstep sounds, even though the player is not moving.

## Steps to Reproduce
1. Hold two opposite (or all four) directional movement keys, either by using WASD or the arrow keys.
2. Listen to the footsteps.

## Suggestions

* Make cancelled movement not produce footstep sounds. 

Alternatively...

* Implement "Null Movement". (Might alleviate #1745)
  With "Null Movement", when a player holds two opposite directional keys (e. g. LEFT and RIGHT) the most recently pressed key will be used, moving the player. If one of the two keys is released, movement continues as normal.
   For example: A player is going LEFT down a corridor. The player now decides to go RIGHT while still (clumsily) holding LEFT. Instead of coming to a standstill they would move to the right. Thanks Null Movement!

daalta:
### Purpose

- Added comments!
- Microwaves now count down instead of up.
- Players can interact with the microwave to see:
  - How much cooking time is remaining
  - If the microwave is empty
  - If the microwave rejects their held item
- RegisterTile is now used to spawn the cooked meal (instead of transform.position)
- The microwave 'ding' sound now plays 1.5 seconds before the microwave is finished. The soundfile contains some 'warm-up' noises at the start and did previously not sync up with the microwave being done cooking
---
- Opposite player movement directions (e.g. left and right) now cancel each other
- Fixes #2527
- _Kind of_ addresses #1745?
- Autoformatted Chat.cs

### Self checks:

  - One client in editor, one in a Development Build
  - The behaviour of the microwave is largely unchanged, so multiplayer checks were not extensive.
  - Player movement is only changed client-side

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Objects/Kitchen/Microwave.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Objects/Kitchen/Microwave.prefab
@@ -189,8 +189,10 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   syncMode: 0
   syncInterval: 0.1
-  cookTime: 10
-  onSprite: {fileID: 21300000, guid: 47022b2620601ec4e87ddd6e2f22e370, type: 3}
+  COOK_TIME: 10
+  microwaveTimer: 0
+  meal: 
+  SPRITE_ON: {fileID: 21300000, guid: 47022b2620601ec4e87ddd6e2f22e370, type: 3}
 --- !u!114 &114643057465573826
 MonoBehaviour:
   m_ObjectHideFlags: 0
@@ -217,8 +219,8 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   syncMode: 0
   syncInterval: 0.1
-  layer: {fileID: 0}
-  ObjectType: 1
+  matrixDebugLogging: 0
+  objectType: 1
   AtmosPassable: 1
   Passable: 1
 --- !u!114 &114348264163667782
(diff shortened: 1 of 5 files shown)


Bod9001:
By gum, We don't have all the features of TG yet but We have the most advanced microwave code : P

daalta:
Added my player movement changes to this pull request. Hope that's OK. When a player presses two opposite movement keys they are now interpreted as not moving at all.

chairbender (approved):
Working nicely, thanks! Good job for a first PR. This is probably one of our most stale components haha, and now it's much more robust. I'm going to approve and merge this since it's working but here's a few suggestions anyway.

Thank you as well for the code formatting changes.

General comments:
* Naming convention for editor assignable fields should be UppercaseCamelCase, not SNAKE_CASE. If our code style guide indicated otherwise I probably need to update it, I think the naming convention stuff might not be that clear or wrong because I wrote it a very long time ago.
* I don't actually think there's a need for Microwave and InteractableMicrowave to be separate anymore, the interaction logic from InteractableMicrowave can most likely be merged into Microwave and deleted. Don't think we plan on having any microwaves you can't interact with.

daalta:
Thanks chairbender. Nice to hear that I did OK. 😙

The snake case is definitely a mistake for SPRITE_ON and COOK_TIME. I was toying with making those variables static and did not end up correcting the names. Oops.

Issue #2425, opened 2019-12-05
Make Roller Chairs draggable B: 50


DooblyNoobly:
# Criteria:

- Make roller chairs draggable and also play the chair rolling sfx periodically
- Change the direction of the chair so it faces the direction of dragging
- Allow players buckled to the chair to be dragged along and also face the same direction as the chair being dragged
- Ensure all of it works over multiplayer with 2 clients

Contributor will receive a bounty of 50usd on a successful merge of a solution that resolves the task request above.

chairbender:
I think even the normal metal chairs are supposed to be draggable as well, though maybe not when someone is in them. Last I checked on tg you could drag at least empty metal chairs.

JaytleBee:
> 
> 
> I think even the normal metal chairs are supposed to be draggable as well, though maybe not when someone is in them. Last I checked on tg you could drag at least empty metal chairs.

Just tested it on /tg/ and nope, **metal chairs can't be dragged** (however they can be picked up by dragging them onto your sprite. The same is true for (bar) stools and wooden chairs. Comfy chairs and shuttle seats can be neither picked up nor pulled. Please refer to [chair.dm](https://github.com/tgstation/tgstation/blob/6e238392bffac435bc6c2a4d614bfff0e998de24/code/game/objects/structures/beds_chairs/chair.dm) in the /tg/station github (look for the "**ANCHORED**" variable).

I think I'll try my hand at this issue.

chairbender:
What should happen if you drag roller chairs on top of each other, or try to drag a roller chair into the same space as a metal chair?

JaytleBee:
> I think I'll try my hand at this issue.

Due to [unforeseen circumstances](https://user-images.githubusercontent.com/7951887/70396285-b27f4580-1a07-11ea-8a83-56a2f3d818d3.gif), I won't be able to work on this anytime soon (currently don't have a good machine at my new apartment.)

> What should happen if you drag roller chairs on top of each other, or try to drag a roller chair into the same space as a metal chair?

Just tested this, two chairs (of any type, as far as I can tell) can be in the same spot (even if people are buckled to it, though you have to avoid collision by crawling into the tile).

chairbender:
understood, thanks for providing this info

JesterX666:
[...]

PR #2524, opened 2020-01-03
Make Roller Chairs draggable B: 50 #2425

JesterX666:
Make Roller Chairs draggable B: 50 #2425

The chair can now be pushed when occupied
Ctrl-Clicking anywhere on the occupant makes the chair be pullable

There is still a delay between the occupant and the chair movement to be solved.


UnityProject/Assets/Scripts/Furniture/OccupiableDirectionalSprite.cs
```diff
@@ -50,8 +50,10 @@ public class OccupiableDirectionalSprite : NetworkBehaviour
 	[Tooltip("sprite renderer on which to render the front sprites")]
 	public SpriteRenderer spriteRendererFront;
 
-	[SyncVar(hook=nameof(SyncIsOccupied))]
-	private bool isOccupied;
+	[SyncVar(hook=nameof(SyncOccupant))]
+	private uint occupant;
```
chairbender, line 54: can you rename these to occupantNetId, SyncOccupantNetId, and OccupantNetId?

UnityProject/Assets/Scripts/Furniture/OccupiableDirectionalSprite.cs
```diff
@@ -50,8 +50,10 @@ public class OccupiableDirectionalSprite : NetworkBehaviour
...
+
+	public uint Occupant => occupant;
```
chairbender, line 56: Exposing the actual net ID of the occupant seems like exposing a bit too much of the internals of this. How about changing this to public GameObject Occupant and having it return null if there's no occupant, otherwise have it look up the game object using the net ID and NetworkIdentity.spawned (but use TryGetValue version so it doesn't throw exception if not found)?

UnityProject/Assets/Scripts/Furniture/OccupiableDirectionalSprite.cs
```diff
@@ -138,26 +140,27 @@ private void UpdateFrontSprite()
 	}
 
 	/// <summary>
-	/// Set whether this object should render itself as if it is occupied or vacant.
+	/// Set the occupant of this object (also indicate if the object should render itself as if it is occupied or vacant).
 	/// </summary>
 	[Server]
-	public void RenderOccupied(bool renderOccupied)
+	public void SetOccupant(uint occupant)
```
chairbender, line 150: Instead of passing a NetId, this can just accept a GameObject and look up the NetId internally. And if null is passed it should set the net ID to 0 (add a note on this to the summary - pass null to remove occupant)

UnityProject/Assets/Scripts/Furniture/OccupiableDirectionalSprite.cs
```diff
@@ -113,7 +115,7 @@ private void UpdateFrontSprite()
 	{
 		if (spriteRendererFront)
 		{
-			if (isOccupied)
+			if (occupant > 0)
```
chairbender, line 122: instead of 0, use NetId.Empty whenever dealing with net ids, this applies to all other places you are using 0.

UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
```diff
@@ -331,16 +332,31 @@ public static List<PushPull> GetPushableAt(Vector3Int worldOrigin, Vector2Int di
...
 		Vector3Int worldTarget = worldOrigin + dir.To3Int();
 		List<PushPull> result = new List<PushPull>();
+
 		foreach ( PushPull pushPull in GetAt<PushPull>(worldTarget, isServer) )
 		{
+			PushPull pushable = pushPull;
+
 			if ( pushPull && pushPull.gameObject != pusher && (isServer ? pushPull.IsSolidServer : pushPull.IsSolidClient) )
 			{
+				// If the PushPull parent object is a player, and that player is buckled, we should use the pushPull object that the player is buckled to.
```
chairbender: in this section, I think you can use GetComponent in both places instead of GetComponentInParent

UnityProject/Assets/Scripts/Transform/CustomNetTransform.cs
```diff
@@ -542,6 +545,22 @@ private void AdminRespawn()
 	{
 		PlayerManager.PlayerScript.playerNetworkActions.CmdAdminRespawn(gameObject);
 	}
+
+	// Checks if the object is occupiable and update occupant position if it's occupied (ex: a chair)
+	private void UpdateOccupant()
+	{
+		// Since we are at the transform level, we go up one level to get the parent
```
chairbender: Shouldn't need to go up a level, GetComponent should be fine. CustomNetTransform lives on the same level as all the other components

UnityProject/Assets/Scripts/Transform/CustomNetTransform.cs
```diff
@@ -542,6 +545,22 @@ private void AdminRespawn()
...
+		OccupiableDirectionalSprite occupiableDirectionalSprite = GetComponentInParent<OccupiableDirectionalSprite>();
```
chairbender: For performance reasons it's best to avoid GetComponent in update loops, which I THINK this might occur in. It creates garbage.

Instead, you should just call `GetComponent<OccupiableDirectionalSprite>()` and save it to a private field in CustomNetTransform's Start method. Then in this method you can check if the field is null to decide if this needs to update its occupant.

UnityProject/Assets/Scripts/Transform/CustomNetTransform.cs
```diff
@@ -542,6 +545,22 @@ private void AdminRespawn()
...
+		if ((occupiableDirectionalSprite != null) && (occupiableDirectionalSprite.Occupant > 0))
+		{
+			NetworkIdentity networkIdentity = NetworkIdentity.spawned[occupiableDirectionalSprite.Occupant];
+			if (networkIdentity != null && networkIdentity.TryGetComponent<PlayerMove>(out var playerMove))
```
chairbender: You can avoid a GetComponent call here by caching PlayerScript of the buckled player in OccupiableDirectionalSprite whenever the buckled player net ID changes (just do it in OccupiableDirectionalSprite.syncOccupant, check if the game object with that net ID has a PlayerScript component and save it to a private field like buckledPlayerScript with a read-only public accessor like BuckledPlayerScript). Then you can access that via occupiableDirectionalSprite.BuckledPlayerScript and don't need TryGetComponent.

And PlayerScript already has the ability to cache and retrieve all of the other components on player via public accessors so that's pretty much the only component you need to have in order to avoid GetComponent calls.

UnityProject/Assets/Scripts/Transform/CustomNetTransform.cs
```diff
@@ -542,6 +545,22 @@ private void AdminRespawn()
...
+			{
+				//sync position to ensure they buckle to the correct spot
+				playerMove.PlayerScript.PlayerSync.SetPosition(occupiableDirectionalSprite.gameObject.TileWorldPosition().To3Int());
```
chairbender:
```suggestion
				playerMove.PlayerScript.PlayerSync.SetPosition(registerTile.WorldPosition);
```

TileWorldPosition can cause GC, this should avoid that. Getting registered position through a cached register tile is the way to avoid TileWorldPosition() and other similar extension methods that cause GC.

UnityProject/Assets/Scripts/Player/PlayerMove.cs
```diff
@@ -314,8 +314,9 @@ public void ServerBuckle(GameObject toObject, Action unbuckledAction = null)
 			playerDirectional.FaceDirection(playerDirectional.CurrentDirection);
 		}
 
-		//force sync direction to current direction
-		playerDirectional.TargetForceSyncDirection(PlayerScript.connectionToClient);
+		//force sync direction to current direction (If it is a real player and not a NPC)
+		if (PlayerScript.connectionToClient != null)
```
chairbender, line 349: nice catch

chairbender (changes requested):
Nice improvements, pushing and pulling people in chairs is quite usable now.

I have some refactorings I suggested but the only reason I'm not merging this now is the GC inducing stuff.

I also noticed a bug where if you strap someone into the chair and try to push them off into space, the chair flies off without them. I'll update the ticket. They also move out of sync if they get blown by wind or newtonian movement possibly? Didn't test that too thoroughly. I think the issue is that we need a generic way to "weld" 2 objects together as far as movement is concerned. I'm planning on looking into that and the movement lag as my next task, I'll probably end up working off your branch and PRing to yours.

UnityProject/Assets/Scripts/Furniture/OccupiableDirectionalSprite.cs
JesterX666: Good idea

UnityProject/Assets/Scripts/Furniture/OccupiableDirectionalSprite.cs
JesterX666: Done.  Look at that: 


	/// <summary>
	/// NetId of the occupant.  0 (NetId.Empty) means there is no occupant.
	/// </summary>
	[SyncVar(hook=nameof(SyncOccupantNetId))]
	private uint occupantNetId;

	/// <summary>
	/// Occupant.  Null means there is no occupant.
	/// </summary>
	public GameObject Occupant => NetworkIdentity.spawned.TryGetValue(occupantNetId, out NetworkIdentity networkIdentity) ? networkIdentity.gameObject : null;

Way cleaner! :)

UnityProject/Assets/Scripts/Transform/CustomNetTransform.cs
JesterX666: So... Basically, a "registerTile" is the tile (with coordinates) an object is in?

JesterX666:
I've corrected the requested changes.

Also included :
Correction for bug #2525 
Removed some unused "usings"
Added some comments

UnityProject/Assets/Scripts/Transform/CustomNetTransform.cs
chairbender: Not quite, registerTile is a component that lives on each object (not on each tile) which TRACKS which position and matrix the object is in. It's a bit of an unfocused component TBH and it's quite old.

chairbender:
Great, thanks for fixing that stuff up 👍

PR #2530, opened 2020-01-03
Interrupt Progress + Progress Action Improvements

### LINKED ISSUE #2204: Progress Actions need to be interrupted in more situations

chairbender:
### Purpose
Fixes #2204 

Refactors progress actions to provide a more usable and flexible API, decoupling the logic of progress from the UI components. It should now be capable of supporting a wider range of use cases without having to rework the internals of progress actions.
Adds an ActionTarget class which should be a useful way to abstract out the differences between actions which target tiles and actions which target specific objects. This is intended to be used for more than just progress actions, for example some interaction logic and other systems which involve targeting something could benefit from not having to distinguish between tile targets and object targets.


UnityProject/Assets/Scripts/UI/ProgressBar/StandardProgressAction.cs
```diff
@@ -0,0 +1,279 @@
...
+			.Where(pb => pb.RegisterPlayer.gameObject == info.Performer)
+			.Where(pb => pb.transform.parent == targetParent)
+			.FirstOrDefault(pb => Vector3.Distance(pb.transform.localPosition, info.Target.TargetLocalPosition) < 0.1);
+		if (existingBar != null)
+		{
+			//progress already started by this player at this position
+			return false;
+		}
+
+		//is this cross matrix? if so, don't start progress if matrix is moving
```
fomalsd: I assume this is temporary?

fomalsd (approved):
Looking very good code wise! not able to playtest though

UnityProject/Assets/Scripts/UI/ProgressBar/StandardProgressAction.cs
chairbender: Not really, there's basically no reason I can think of to allow progress to be started when a crossed matrix is moving.

There is a bug here where it doesn't check if the player's own matrix is moving, though.

chairbender:
I'll fix the cross matrix thing now

chairbender:
good to review again

Issue #2311, opened 2019-11-21
EPIC: Usable Sprite Animation System (USAS) 


chairbender:
# Epic Description
This is an epic with multiple subtasks.

The goal of this epic is to create a convenient / usable system for defining, changing, and playing animations on a particular object , which includes non-animated / static sprites, and syncing them over the network. We also want this system to be flexible enough to support possible future use cases, such as having cool effects or particle emitters on clothes.

The current system uses SpriteHandler / SpriteDataHandler and while it does support various use cases we've encountered some difficulties trying to actually use and make sense of it.

Subtasks:
- [x] #2590 Simple World Icons
- [x] #2591 Composite World Icons
- [x] #2592 Simple Inventory Icons
- [x] #2593 Composite Inventory Icons
- [ ] #2594 Inhands
- [x] #2595 Clothing
- [x] #2596 Uniform Suit and Skirt Variants
- [x] #2597 Full Migration

Some guidance on the design and implementation:
- Try to get feedback early and often so you don't do a bunch of work that ends up having to be re-done.
- As far as flexibility is concerned (to support future or missed use cases), the general idea is that it should be set up so that we won't "code ourselves into a corner" and create a situation where lots of effort is needed to add capabilities to the system due to the design. Take a look at TileInteraction and its subclasses for an example of a somewhat flexible system. Inheritance / containment / interfaces can help here.
   - For example, in the future we might want to be able to apply a particular effect, shader, or material to something. Like a glowing effect. Or have it emit particles, such as a glittering crown.
- To help it coexist alongside the old system until it's completely migrated, you might want to make "V2" or "V3" versions of components you changed along with a common interface that the new / old components share (which would eventually be removed after migration) to make this easier.
- Re-evaluate the overall design of the system as we work on subtasks.
- Editor-assignable fields should almost always be private, lowercaseCamelCase, and use [SerializeField].
- Take care when naming classes / assets to maximize usability. Try to be precise but not too wordy.
- Be precise when naming fields. Longer field names are acceptable because they show up okay in the editor.
- Leverage scriptable objects. You can leverage inheritance and containment along with ScriptableObject classes to allow for lots of flexibility.
- Avoid using array/list fields where each index has a special meaning (like 0 is for inhands, 1 is for icon, etc...). This is difficult to make sense of. It's acceptable to use these fields for situations where the meaning of the order is obvious, such as defining a sequence of sprites to animate through.
- Try to use named fields rather than dictionary fields.
[...]

PR #2534, opened 2020-01-04
Cleaned up and formatted SpriteHandler.cs

### LINKED ISSUE #2221: BUG: SpriteHandler error when dropping items B: 25
# Description

Sometimes when dropping items from your hand SpriteHandler throws an ArgumentOutOfRangeException.

Full error log:

```
ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
System.ThrowHelper.ThrowArgumentOutOfRangeException (System.ExceptionArgument argument, System.ExceptionResource resource) (at <599589bf4ce248909b8a14cbe4a2034e>:0)
System.ThrowHelper.ThrowArgumentOutOfRangeException () (at <599589bf4ce248909b8a14cbe4a2034e>:0)
System.Collections.Generic.List`1[T].get_Item (System.Int32 index) (at <599589bf4ce248909b8a14cbe4a2034e>:0)
SpriteHandler.PushTexture () (at Assets/Scripts/Sprite Handler/SpriteHandler.cs:63)
ClothingItem.PushTexture () (at Assets/Scripts/Items/ClothingItem.cs:169)
ClothingItem.SetReference (UnityEngine.GameObject Item) (at Assets/Scripts/Items/ClothingItem.cs:100)
EquipmentSpritesMessage+<Process>d__7.MoveNext () (at Assets/Scripts/Messages/Server/EquipmentSpritesMessage.cs:34)
```

## Steps to Reproduce
1. Start game from latest develop in the editor
2. Go to a locker with items inside, open it and pick up an item
3. Drop the item using shortcut key. Do a few times if it doesn't error right away it will happen pretty quickly

dargonforce:
### Purpose
- refactored some methods for simpler logic w/ fewer scopes
- made variables private and added SerializeField where expected
- possible helps #2221, or at least makes it easier to diagnose/understand

### DIFF
--- a/UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
+++ b/UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
@@ -12,7 +12,7 @@ public class SpriteHandler : SpriteDataHandler
 	[SerializeField]
 	private int spriteIndex;
 
-	[SerializeField]
+	[SerializeField] [FormerlySerializedAs("VariantIndex")]
 	private int variantIndex;
 
 	private SpriteJson spriteJson;
--- a/UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
+++ b/UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
@@ -12,7 +12,8 @@ public class SpriteHandler : SpriteDataHandler
 	[SerializeField]
 	private int spriteIndex;
 
-	[SerializeField] [FormerlySerializedAs("VariantIndex")]
+	[SerializeField]
+	[FormerlySerializedAs("VariantIndex")]
 	private int variantIndex;
 
 	private SpriteJson spriteJson;
--- a/UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
+++ b/UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
@@ -72,7 +72,7 @@ public class SpriteHandler : SpriteDataHandler
 				TryInit();
 			}
 
-			if (spriteIndex < Infos.List[spriteIndex].Count &&
+			if (spriteIndex < Infos.List.Count &&
 			    variantIndex < Infos.List[spriteIndex].Count &&
 			    animationIndex < Infos.List[spriteIndex][variantIndex].Count)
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
```diff
@@ -1,40 +1,53 @@
...
-	public SpriteRenderer spriteRenderer;
+	[SerializeField]
+	private SpriteRenderer spriteRenderer;
 
-	public int spriteIndex;
+	[SerializeField]
+	private int spriteIndex;
 
-	public int VariantIndex;
+	[SerializeField]
```
chairbender, line 15:
```suggestion
	[SerializeField] [FormerlySerializedAs("VariantIndex")]
```

Need to add another attribute [FormerlySerializedAs("VariantIndex")] otherwise any components on prefabs which had this value set will lose their setting.

UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
```diff
@@ -58,139 +71,91 @@ public void PushTexture()
...
-					}
-				}
-				else
-				{
-					spriteRenderer.sprite = null;
-					TryToggleAnimationState(false);
-				}
-			}
-			else
+			if (spriteIndex < Infos.List[spriteIndex].Count &&
```
chairbender: I get an ArgumentOutOfRangeException here, due to Infos.List[spriteIndex]. Reproducible by going into botany as host player, opening the crate, and picking up  some seed packets.

It's because that first if was removed, if (spriteIndex < Infos.List.Count) still needs to be checked

chairbender (changes requested):
Thanks for the cleanups. Looks like there's just one exception it introduces which is easily fixed.

FYI we are planning on significantly overhauling how animations / sprites work in #2311

UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
dargonforce: Oops, typo on my part.  Sorry about that.

chairbender:
nicely done, good to merge

Issue #2535, opened 2020-01-04
Some crate descriptions do not appear when hovering over them


daalta:
# Description

Hovering with the mouse cursor over certain crates does not display their object attribute description.

This seems to affect the crates in Resources/Prefabs/Objects/Crates (except crates_59). 

## Detailed Bug Report

Discovered this while going over item descriptions. I will update the issue if I notice any other crates that misbehave. Not all crates are affected, so I suspect there is an issue with a copy & pasted prefab.

## Steps to Reproduce
1. Hover your mouse over a radiation crate while playing the game. 
2. The object name is displayed in the lower-left of the screen ("radiation crate"), but the description is not shown ("A crate with a radiation sign on it.")

StrawberryMoses:
I refactored crates not too long ago to inherit from a base prefab / each other- I'd start by looking at the base crate prefab

daalta:
You're right @StrawberryMoses , I think. The various crates inherit from "Crate", and "Crate" inherits from "Object". I think the issue is that there's something wrong with "Object".

Lockers, for example, do not inherit from anything.

StrawberryMoses:
I'll take a look at object then

(if I don't complete within the week feel free to take me off)

daalta:
I've compared crate and locker, and they seemed very similar. It's also possible that I missed something. I'm sure that if we recreated crates the issue would disappear, but... ugh

StrawberryMoses:
The thing is that all new object prefabs should inherit from Object to try and recreate the similar structure that tg has. If something is up with the base prefab then it'll affect much more down the line

StrawberryMoses:
closing since merged

PR #2538, opened 2020-01-04
Added names and descriptions to most hallway objects

daalta:
### Purpose

- Many objects still lack a proper name or description. When hovering over them to see their tooltip they often display names_such_as_these_59. Players new to Unitystation might get confused.
- This PR adds names and descriptions to most objects and machines that can be found in public hallways. This was done by adding ObjectAttributes.cs to those objects.
- This PR also adds names to some of the most common objects around the station, such as doors, closets and vending machines.
- Proper integrity values were added to some objects (like doors and lockers). They now have different armor, integrity and heat resistance values.
- Attributes.cs was changed so that the first letter of an object name is automatically capitalized in the tooltip.

### Notes
- Some descriptions were taken from tgstation, but many were rewritten or newly written.
- Crates can not be named until #2535 is resolved. Crates are very common and should also receive proper names, descriptions and integrity values.
- The heat resistance of some metal object was set to 1370 °C (the actual melting point of steel). This value should be lowered if it turns out that using realistic values is not suitable for Unitystation.

### Known issues

- The last two words of vending_16's tooltip are cut off, likely due to vending machine's the long name & description. However, objects should be allowed to have long names and descriptions. This feature should be implemented in the future.

### Self checks:

- [x] The PR does not bring up any new compile errors (that I could notice)

### DIFF
--- a/UnityProject/Assets/Scripts/Health/Objects/Integrity.cs
+++ b/UnityProject/Assets/Scripts/Health/Objects/Integrity.cs
@@ -55,9 +55,9 @@ public class Integrity : NetworkBehaviour, IHealth, IFireExposable, IRightClicka
 	public Resistances Resistances = new Resistances();
 
 	/// <summary>
-	/// Below this temperature, the object will take no damage from fire or heat and won't ignite.
+	/// Below this temperature (in degrees Celsius) the object will take no damage from fire or heat and won't ignite.
 	/// </summary>
-	[Tooltip("Below this temperature, the object will take no damage from fire or heat and" +
+	[Tooltip("Below this temperature (in degrees Celsius) the object will take no damage from fire or heat and" +
 	         " won't ignite.")]
 	public float HeatResistance = 100;
 
--- a/UnityProject/Assets/Resources/Prefabs/Objects/Closets/Closet.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Objects/Closets/Closet.prefab
@@ -310,12 +310,12 @@ MonoBehaviour:
   Resistances:
     LavaProof: 0
     FireProof: 0
-    Flammable: 1
+    Flammable: 0
     UnAcidable: 0
     AcidProof: 0
     Indestructable: 0
     FreezeProof: 0
-  HeatResistance: 600
+  HeatResistance: 1370
   initialIntegrity: 200
 --- !u!114 &114434305287704188
 MonoBehaviour:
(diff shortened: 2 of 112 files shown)


daalta:
Reverted my most recent resistance (and integrity changes) as it would have made objects instantly flammable. I'll put the integrity changes back in at some other time.

UnityProject/Assets/Scripts/Health/Objects/Integrity.cs
```diff
@@ -55,9 +55,9 @@ public class Integrity : NetworkBehaviour, IHealth, IFireExposable, IRightClicka
 	public Resistances Resistances = new Resistances();
 
 	/// <summary>
-	/// Below this temperature, the object will take no damage from fire or heat and won't ignite.
+	/// Below this temperature (in degrees Celsius) the object will take no damage from fire or heat and won't ignite.
```
chairbender, line 58: It's actually Kelvin. All temps in the game are kelvin unless otherwise stated. The reason it was having no effect (despite this being extremely low) was that this field currently only gets used when an object is in an exposed hot spot (where there's a visible burning sprite on the tile). Even if the heat on a tile is really high the object doesn't currently take any damage. We haven't yet implemented temperature exposure, only fire exposure.

There is an equivalent field in tg used by a few objects such as canisters, temperature_resistance, and a "temperature_expose" method.

For now let's just change this comment to say Kelvin. The default of 100 was just a random guess, so when we implement temperature exposure I will most likely refactor to make it more sane.

chairbender (changes requested):
Thanks for adding all this stuff! Just one comment about kelvin instead of celsius

UnityProject/Assets/Scripts/Health/Objects/Integrity.cs
chairbender: edit: also remembered the reason for this field was so I could give canisters a higher value so they wouldn't all pop like balloons when they are in a fire

chairbender: IMO this field should actually be removed so we match TG logic, but keep it in for now. I'll make a ticket.

chairbender:
actually I'll just edit it here and merge

UnityProject/Assets/Scripts/Health/Objects/Integrity.cs
```diff
@@ -55,9 +55,9 @@ public class Integrity : NetworkBehaviour, IHealth, IFireExposable, IRightClicka
...
 	/// </summary>
-	[Tooltip("Below this temperature, the object will take no damage from fire or heat and" +
+	[Tooltip("Below this temperature (in degrees Celsius) the object will take no damage from fire or heat and" +
```
chairbender, line 60:
```suggestion
	[Tooltip("Below this temperature (in Kelvin) the object will take no damage from fire exposure" +
```

daalta:
Glad you like my contribution. Thanks for the review and the edit, things are much clearer now.

I'll just keep adding integrities, names and descriptions in the future.

PR #2542, opened 2020-01-05
Variable viewer now Can change, int strings floats, enum and bools

Bod9001:
And now looks better with text mesh pro

### Purpose
VV's are fun

yeah, need anything added added am good  To do it

Tested in editor


Bod9001:
need to quickly get the input focus stuff done

Bod9001:
It's strange the input fields aren't being selected properly in builds idk What to do what to do

chairbender:
We have some global stuff that affects UI focus, in UIManager I believe.

Bod9001:
is good now

chairbender:
yeah, VV Is looking nicer. Thanks!

PR #2561, opened 2020-01-08
Chat bubble scaling (+ shouting and whispering)

daalta:
# Purpose
- This PR makes it so chat bubbles always stay the same size regardless of the current zoom. They are no longer unreadable when too far zoomed out / in. 
- Chat bubbles are increased or decreased in size if players whisper or shout (see notes).
- Chat bubble max / min size is clamped (https://github.com/unitystation/unitystation/pull/2561/commits/1e08d61607b831c49087db23d9b4de68e01cce86)
- Fixes #2560
- Fixes #2567
- Also adresses #2269, but only for text bubbles. The font sizes for speaking, whispering and shouting were manually checked to be readable, though there's still aliasing.

_Update: Chat bubbles look slightly different from the video with a smaller border, tail and bug-fixed opacity  (https://github.com/unitystation/unitystation/pull/2561/commits/1e08d61607b831c49087db23d9b4de68e01cce86):_
vlc_2020-01-09_01-46-51

### Notes:
- Bubbles stay the same size by scaling themselves according to the camera zoom.
- Chat bubbles are not visible in darkness. This was previously not an issue because chat bubbles were much smaller and were rarely far away from the player. However they can now reach into occluded parts of the station when zoomed out very far.
- Players can shout by typing something in capital letters or ending their sentence with two exclamation marks ("!!")
- Players can whisper by starting their sentence with a number sign ("#"). The number sign at the start of the message is removed.
- The detection of the player whispering / shouting is **not** consistent with Chat.Process.cs and does not use ChatModifier. The syntax for shouting also currently differs from the chat window (awaiting feedback from reviewers). PlayerChatBubble.cs could be rewritten to make use of them in the future.
- Added placeholder comments for clowns having their own chat bubble.
- Feedback appreciated.

### Please make sure you have followed the self checks below before submitting a PR:

- [x] Code is sufficiently commented (But the rest PlayerChatBubble.cs wasn't / isn't well commented)
- [x] Code is indented with tabs and not spaces
- [x] The PR does not include any unnecessary .meta, .prefab or <b>.unity (scene) changes</b>
- [x] The PR does not bring up any new compile errors
- [x] The PR has been tested in editor
- [x] The PR has been tested in multiplayer (Locally on one editor and one regular build)

### What changes to chat bubbles I'd like to see in the future:
- Rich text with TextMeshPro for more interesting chat messages (Comic Sans font for clowns)
- ~~ Smaller "bubble" around the text ~~
[...]

Issue #2569, opened 2020-01-09
Chat Bubbles Offset on Rotated Matrix


chairbender:
Get on a shuttle and turn once. Turn on chat bubbles. Say something. See that the chat bubble is a bit off from where it should be. In the below picture it's too far to the right of my player.

![image](https://user-images.githubusercontent.com/1329062/72042151-6facc980-3262-11ea-8b4b-5e72cf7dc592.png)

Additionally, there's a TODO comment in PlayerChatBubble.Update saying the euler angle stuff can be removed.  This logic actually ensures that chat bubbles stay upright when on a rotating matrix so it actually is necessary, so this comment should be removed.

daalta:
I'll check it out later, feel free to assign me.

We might be able to solve this by changing the bubble's pivot, actually. Then we can take the rotation code out of Update.

daalta:
Fixed it by changing the pivot and the position in the player prefab. Will include the fix in my next PR.
https://i.imgur.com/Xoe5NAS.mp4

fomalsd:
still the case as of build 3982
![unknown-13](https://user-images.githubusercontent.com/10403536/73126589-b1a96f80-3fc5-11ea-9948-f92550bcc4a2.png)

daalta:
You're right, thanks for bringing this up.

The chat bubbles containing text still work, just tested it. I've created a new issue.

daalta:
Fixed in https://github.com/unitystation/unitystation/pull/2739

daalta:
Has since been merged.

PR #2571, opened 2020-01-09
Chat bubbles use TextMeshProUGUI, which is now supported by UI themes

daalta:
### Purpose
* Chat bubbles now use TextMeshPro, increasing text readability.
* Chat bubble size was changed to be similar to the chat font size (on 16:9 monitors).
* Themes can now apply text themes to TextMeshProUGUI components!
* Chat bubble font changed to a font similar to the regular chat. The old font (arialbd) is not currently compatible with TextMeshPro.

Unitystation_2020-01-10_00-17-27

### Notes
* Minimum and maximum bubble sizes https://github.com/unitystation/unitystation/pull/2561 have been temporarily removed. They did not work on all resolutions and *should* depend on zoom level.

### Self checks:
* I only changed the necessary prefabs, C#-files and assembly definitions.
* Only few code changes were made, and they should be self-explanatory.
* Chat bubbles should behave the same as before and were not extensively tested in multiplayer.


daalta:
### To-Do
The following is a list of things I will plan to change about chat bubbles at some point in the future.

* The current font shader has needless outlines
* Maxmimum and minimum chat bubble sizes should be based on zoom level
* Rich text for whispering and shouting
* Fixing https://github.com/unitystation/unitystation/issues/2569
* Changing the font back to arialbd (unless you're happy with the new font 😉).

HeadClot:
Hey got a question @daalta in unity 2019.3 there will be a new UI System called UI Builder. It will take advantage of a CSS styled language called unity style sheets or .uss. Do you or anyone on the dev team plan on taking advantage of that? It should greatly reduce the amount of time to build UI elements :)

DooblyNoobly:
unity style sheets sound cool. I'm sure we will convert to it when we move to a stable 2019.3 version

Issue #2544, opened 2020-01-06
Set text bubbles Default on


Bod9001:
# Description
I never Saw them until someone told me you could turn them on

chairbender:
you just need to turn them in options

daalta:
I agree with @chairbender that it's not a bug, but I also think they should be on by default. Maybe.

Though I feel like there are some kinks that could be worked out. There's some heavy aliasing going on, which I might take a look at.

Here's a picture after zooming out a bit.
![2020-01-06_23-52-57](https://user-images.githubusercontent.com/9169275/71854840-b1613700-30df-11ea-8fde-0b543cb3d78e.png)

(Edit: We could also call that a feature, hehe. Can't hear people who are far away, right? 😉)

fomalsd:
it only starts looking good at 2x zoom. nothing will look good at 1.5x zoom (see #2269) and fonts are a bit too big to look good at 1x
same applies to status display text

fomalsd:
<img width="99" alt="Scr 2020-01-07 в 1 59 48" src="https://user-images.githubusercontent.com/10403536/71855291-ae6f4200-30f1-11ea-9880-b40a2b31df22.png">
<img width="110" alt="Scr 2020-01-07 в 2 00 16" src="https://user-images.githubusercontent.com/10403536/71855298-b333f600-30f1-11ea-85d3-248537d0a715.png">
<img width="128" alt="Scr 2020-01-07 в 2 00 56" src="https://user-images.githubusercontent.com/10403536/71855306-ba5b0400-30f1-11ea-86b3-bb6c5db07078.png">

daalta:
Well, we could fix that with texture filtering. We'd lose the pixel perfect look though!

Unfiltered:
![Unity_2020-01-07_00-06-57](https://user-images.githubusercontent.com/9169275/71855622-d656a980-30e1-11ea-8333-b8aaf48e0109.png)

Bilinear filter:
![2020-01-07_00-07-02](https://user-images.githubusercontent.com/9169275/71855623-d787d680-30e1-11ea-923c-cddd684322b4.png)

It's definitely an improvement I think. Honestly I'm not too attached to the pixel font anyway - I wouldn't mind replacing it with something more readable.

StrawberryMoses:
Would this be changed with the other filtering issue anyway?

daalta:
@StrawberryMoses I don't think the other issue mentions changing how text is rendered. Maybe I'm misunderstanding something.

[...]

PR #2578, opened 2020-01-11
Chat bubble styles & curves, changed zoom handler range, and other improvements

### LINKED ISSUE #2269: Certain zoom values are not pixel perfect
# Description
Zoom slider (that multiplies zoom factor) currently goes  from 1 to 10.
Certain commit changed base zoom of 1x down to to 0.5x, so instead of `1, 2, 3... ` it now goes `0.5, 1, 1.5...`
Need a custom zoom slider sequence, preferably modifiable through inspector, that would have a following set of values:
1, 2, 4, 6, 7, 8, 10, 12, 14, 16
instead of a simple 1 to 10

### LINKED ISSUE #2569: Chat Bubbles Offset on Rotated Matrix

daalta:
# Purpose
## Chat bubbles
  

* Size relative to the zoom level is now determined by an animation curve. This makes it easy to set the bubble size for each zoom level.
* Changed size in the new curve.
  * Chat bubbles are smaller when zoomed out very far.
  * Chat bubbles are bigger when zoomed out very close.
* Styles: Whispered text in italics, shouted text is bold.
  * Note: This change would be incompatible with the previously used `arialbd` font.
* All chat bubble font sizes for different vocalizations (whispering, talking, shouting) are now more similarly sized (Previously shouting was twice as large as talking).
* Shader changed to a mobile TMP shader (and removed emission)
* Changed anchor to be centered on player
  * Fixes #2569 by removing the problematic zoom levels

## Zoom handler
zoom
* Zoom now only goes from 2 to 10 in increments of 2 (2, 4, 6, 8, 10)
  * Fixes https://github.com/unitystation/unitystation/issues/2269
  * Increments of 1 are allowed when changing zoom through the options screen.

## Other changes
* Added names and descriptions to crates
* Corrected some other object names / descriptions
* Added integrity values to chests
* Fixed ship ambience sound file (was not seamlessly looping)

# Self checks


chairbender:
excellent, nicely done 👍 IMO we should default chat bubbles on now that these are in place. I think I'll start a vote over in that ticket

daalta:
Thanks for the review.

Good idea! Here's a link to the vote. https://github.com/unitystation/unitystation/issues/2544#issuecomment-573283810

PR #2599, opened 2020-01-13
Fixes and improves butchering of animals

### LINKED ISSUE #2309: Butchering
# Description

Add ability to butcher hostile mobs by attacking a dead mob with a knife in hand and configure their drop tables for what they can drop. Each mob's drop table should be an SO with a list of prefabs and a drop rate for each prefab. Do NOT hardcode the drop table into a prefab, because it is highly likely we want to be able to easily swap out drop tables!

Configure a drop table for whatever mobs we already have (Xenos only at the time of writing).

ziiplock:
### Purpose
Partially fixes #2309 
Fixes butchering and moves it into the meleeable component. Also makes the interaction take some time. Currently only works on animals.

### Notes:
To butcher a dead animal, interact with it while holding a knife and being in harm intent.


JesterX666 (approved):
Looks good to me.  Tried it, and it works.

UnityProject/Assets/Scripts/Weapons/Meleeable.cs
```diff
@@ -7,6 +7,19 @@
 /// </summary>
 public class Meleeable : MonoBehaviour, ICheckedInteractable<PositionalHandApply>
 {
+	[SerializeField]
+	private ItemTrait butcherKnifeTrait;
+
+	[SerializeField]
```
chairbender, line 13: @ziiplock no need to put SerializeField on static fields. They can't be changed in editor.

PR #2602, opened 2020-01-15
Cooldown System

### LINKED ISSUE #2515: Players can spam click with help, disarm and grab intents
# Description
Help/disarm/grab intent clicks need a cooldown like with harm intent clicks (melee attacks).

Ideally all intents will share the same cooldown so you can't still spam clicks by changing intents really quickly.

## Steps to Reproduce
1. Select help/grab/disarm intent
2. Spam click window
3. Notice it's really fast
4. Select harm intent
5. Spam click window again
6. Notice how there's a reasonable cooldown between clicks

chairbender:
### Purpose
Fixes #2515 
Adds a "cooldown system", intended to be a general way to allow definition and tracking of cooldowns. Main API for it is Cooldowns.cs. Cooldowns can be defined using SOs and tracked on each object using HasCooldowns component. Cooldowns can also be defined one-off for small interactable components without needing to create an asset (see KnockWindow.cs for example).

This system is also intended to be able to support other kinds of cooldowns, like abilities. HasCooldowns should work properly even if put on any game object, not just on players. Could be used for AI abilities for example.

Fixes annoying "interaction request will not be sent because it's clientside" warning.


UnityProject/Assets/Scripts/Cooldowns/HasCooldowns.cs
```diff
@@ -0,0 +1,101 @@
...
+/// Allows a game object to have cooldowns which can be triggered on, take a defined amount of time to
+/// go off, and can be checked if they are on/off.
+/// Intended to be used for a wide variety of things, such as interaction logic and abilities. A given cooldown
+/// can be used across multiple components as well, such as for various melee-related actions that should not
+/// be possible to spam all together.
+///
+/// Cooldowns are primarily defined using Cooldown assets, but for convenience there are also ways to use cooldowns
+/// for one-off situations without needing to create an asset.
+///
+/// At the moment, there is currently NO networking logic for cooldowns. The clients (incl. host player) and server track cooldowns totally
```
daalta, line 18: Do you think there could be networking issues when a player joins mid-cooldown?

DooblyNoobly: I think the last part of the comment is sufficient enough "We can add networking logic for future cooldown use cases if it becomes necessary."

Most (if not all) cooldowns would be less then a second so it won't affect new players.

DooblyNoobly (approved):

PR #2606, opened 2020-01-15
Removed AnnounceNewPlayer method

### LINKED ISSUE #2598: New player announcement is played when you leave your dead body
## Description
As per title:
img
That TTS welcome announcement is legacy code which never worked well with more than one player and should be disabled until we come up with a local TTS solution

### Steps to Reproduce

1.  Host from editor, spawn
2.  Die
3.  Fly out of your dead body

nikopetr:
### Purpose
Fixes #2598 
### Notes:
- Removed the announcement code when a player joins from the RequestSyncMessage class
- Not sure if it it was supposed to be completely removed or add a player dead check, so just tell me if you want the check to be added instead !


DooblyNoobly:
I think the decision is to remove it completely so will merge now. Thanks

Issue #2264, opened 2019-11-11
Atmos breaks due to oxygen NaN infection


fomalsd:
# Description
Oxygen levels can become NaN:
![GIF](https://user-images.githubusercontent.com/10403536/68569453-07a25880-0456-11ea-9966-2acfea38feaf.gif)
![image](https://user-images.githubusercontent.com/10403536/68569889-310fb400-0457-11ea-99b6-1ce16b0e02d7.png)

## Steps to Reproduce
Please enter the steps to reproduce the bug or behaviour:
Exact steps are unknown. Try exploding stuff and perhaps crashing matrices
Note that pressure is unaffected by NaNs

chairbender:
@fomalsd are you still able to reproduce this? We are having trouble getting this to happen.

fomalsd:
No longer seems to happen on my branch – no idea why because I don't recall atmos code being changed. Will close this unless it decides to reemerge

fomalsd:
It's still a thing, but exact steps to reproduce are unclear. Very nasty

fomalsd:
Crash escape shuttle into station to reproduce

chairbender:
Adding a bounty of 50

chairbender:
Drakelor in Discord identified the cause of this in the atmos channel, caused by CalcMeanGasMix targetCount being 0 and causing a Nan.

JesterX666:
I'll take a look at this tonight to see if I can reproduce and fix it.  I'll keep the team informed if I can fix it.

JesterX666:
This is caused by some tiles with z coordinates of -1 after a crash with another matrix.
The tile has no "neighbours" and it causes a division of 0 by 0 (NaN) for Pressure and Moles of Gazes.

First of all, I'll find out what a Z coordinate of -1 means and fix it if possible.
I'll also protect the code so it would not be possible to happen.

Issue #1902, opened 2019-05-28
Support for airtight objects is unrobust


Aranclanos:
Our support for airtight ObjectBehaviours is quite unrobust – airtight flag for tile is only assigned at server start (but not during runtime spawn) and is never gone afterwards, and doesn't follow the object if it's pushable.
See this comment: https://github.com/unitystation/unitystation/issues/1902#issuecomment-575571921

StrawberryMoses:
I think this may have been fixed with the new mapping changes

daalta:
foma has fixed this issue on his branch. Atmos caused the knife to suck in the monkey.

fomalsd:
This is related to #1856 as actual knife hit happens because of a pressure difference when Dinnerware vendor (which is not atmos passable) is initialized, and such tiles (like walls) initialize without air.
Therefore initializing non-atmos passable tiles with air will fix both issues:
https://github.com/fomalsd/unitystation/blob/43d794cf520bb39f3228b863f19c60b3cc2ccd0e/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/AtmosSystem.cs#L15

But that's not all – our support for airtight ObjectBehaviours is quite unrobust – airtight flag for tile is only assigned at server start (but not during runtime spawn) and is never gone afterwards, and obviously doesn't follow the object if it's pushable:

_dragging airtight object around:_
![GIF (1)](https://user-images.githubusercontent.com/10403536/72605465-b9597c00-3914-11ea-9b7b-8af395d7e052.gif)

_runtime spawning another dinnerware (airtightness doesn't init), then forcefully updating tile with a special command:_
![GIF (2)](https://user-images.githubusercontent.com/10403536/72605605-04738f00-3915-11ea-9633-e9547430d5eb.gif)

so on my branch I just removed support for airtight objects, which isn't really a solution (shuttles now count as space because engines were airtight objects)

fomalsd:
updated title and description

Issue #1856, opened 2019-05-14
Destroying windows/walls sucks in items


Jibbajabbafic:
# Description
Destroying window:
![iMRB940WlM](https://user-images.githubusercontent.com/4493127/57724228-2fb92680-7682-11e9-8815-923fa2acb69b.gif)

Destroying wall:
![pax3XUGwSE](https://user-images.githubusercontent.com/4493127/57724233-321b8080-7682-11e9-8873-3b7aaa553f4a.gif)

Expected behaviour would be for items to stay where they are

## Category
BUG

## Steps to Reproduce
Please enter the steps to reproduce the bug or behaviour:
1. Gather items around a window/wall
2. Smash window/weld wall
3. Watch items get sucked in

Aranclanos:
The turfs spawn with no gas at all so they suck from the ones next to it
First time it happened it spooked me

fomalsd:
I kinda like it to be honest

Jibbajabbafic:
It's kinda cool, but also makes the game seem a bit unfinished, like we couldn't be bothered to implement a fix

PR #2609, opened 2020-01-16
Protection against NaN in Atmos

JesterX666:
Protection against NaN in Atmos

Atmos breaks due to oxygen NaN infection B: 50 #2264

Fixes the possibility of NaN in CalcMeanGasMix

I was not able to reproduce a big NaN propagation into the Atmos System, it seems that some protections were already put in place to prevent that.

I crashed a lot of matrixes into the station and it seems I don't get NaN anywhere.


fomalsd (changes requested):
Something else still causes NaNs:
mov
And so far I'm only able to detects NaNs post factum when checking what gas mix mobs are going to inhale:
https://github.com/fomalsd/unitystation/blob/aacdd44032ce0153e7beafa884fe427a81c1d427/UnityProject/Assets/Scripts/Health/RespiratoryHealth/RespiratorySystem.cs#L86-L90

JesterX666:
That's weird... I believe it's only on your branch, @fomalsd ... Or related to a different setting?!

NAN

fomalsd:
woops, might be just my branch indeed

fomalsd:
ok, since it works on develop this can be merged – but I'd advise adding measures suggested in #2610, apparently it removes possibility of situations where `targetCount` is 0
but keep in mind that whenever someone starts fixing #1902 or #1856 oxygen NaNs might appear again, and for a different reason – that's what happens in my branch

JesterX666:
Agreed.  I will not be able to incorporate #2610 tonight but I'll assign myself on it and put it "In progress".

PR #2613, opened 2020-01-16
Fixed missing item names with steal objective

### LINKED ISSUE #2565: Traitor steal objectives unclear and unreliable B: 40
## Description

```
[Antags] Created new antag. Made Sydney J Zadovsky a Traitor with objectives:
1. Steal 1 
2. Steal 1 
3. Stay alive until the end!
```
what exactly are you supposed to steal?

EDIT: Additionally, the CheckCompletion method only checks the article name and doesn't work properly for stackables. The logic should check the prefab of the items they picked up, and in the case of stackables should check "StacksWith" and also respect the count of items in the stack when checking the count.

DooblyNoobly:
As per the title. This fixes #2565 

The item name was being taken from IA2.ArticleName which is the name given to an item after it is spawned on the server. The objective asset actually looks at the prefab in asset resources where it has not spawned yet. I created a readonly property to read the actual Item name and I also use that for the green text validation of the stolen item on round end.

Tested in MP

### DIFF
--- a/UnityProject/Assets/Scripts/Antagonists/Objectives/Steal.cs
+++ b/UnityProject/Assets/Scripts/Antagonists/Objectives/Steal.cs
@@ -57,7 +57,7 @@ namespace Antagonists
 
 			// Pick a random item and add it to the targeted list
 			var itemEntry = possibleItems.PickRandom();
-			ItemName = itemEntry.Key.Item().ArticleName;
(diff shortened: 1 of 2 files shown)


fomalsd:
I'd still like a check for empty name for future proofing

DooblyNoobly:
ok will add to next pr

PR #2614, opened 2020-01-16
Prevent users from logging in twice on release servers

### LINKED ISSUE #2540: Strange auth/body assignment issues
# Description
Here’s what I encountered on build 3954:
 
Joined game for the first time using my main `foma` account, and was surprised to see that character wasn’t mine, instead it was a body of someone else called Chuck Howard, floating in space.
Asked Bod to come check and he joined alright.
Meanwhile I made a new account `foma2` and joined alright (second window) while keeping original window open. However I had a shit connection so both windows got disconnected soon. I closed the first game window and rejoined from the second one — I got into wrong Chuck Howard‘s body again (this time with a different account). 
Then I started another game instance from the same foma2 account and joined proper foma2’s body this time:
4CC66C92-FD0F-4831-9FFE-1D837041E20C

So the issues are:
— Being handed the wrong body
— Being able to join multiple times from the same account, simultaneously

DooblyNoobly:
As per title. This should fix #2540 

Silenced some minor editor errors as well

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
+++ b/UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
@@ -140,6 +140,26 @@ public partial class PlayerList
 			return false;
 		}
 
+		//Do not allow users to log in twice for release servers:
+		if (BuildPreferences.isForRelease)
+		{
+			if (GetByUserID(userid) != null)
+			{
+				if (GetByUserID(userid).Connection != null)
+				{
+					if (playerConn.Connection != GetByUserID(userid).Connection)
+					{
+						StartCoroutine(
+							KickPlayer(playerConn, $"Server Error: You are already logged into this server!"));
+						Logger.Log($"A user tried to connect with another client while already logged in \r\n" +
+						           $"Details: Username: {playerConn.Username}, ClientID: {clientID}, IP: {playerConn.Connection.address}",
+							Category.Admin);
+						return false;
+					}
+				}
+			}
+		}
+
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
```diff
@@ -140,6 +140,26 @@ private async Task<bool> CheckUserState(string userid, string token, ConnectedPl
...
+		//Do not allow users to log in twice for release servers:
+		if (BuildPreferences.isForRelease)
+		{
+			if (GetByUserID(userid) != null)
+			{
+				if (GetByUserID(userid).Connection != null)
+				{
+					if (playerConn.Connection != GetByUserID(userid).Connection)
+					{
+						StartCoroutine(
```
fomalsd, line 152: won't it cause problems if someone loses connection but will still be treated as "in-game" so won't be able to join again (at least for a while)?
I think multiplayer games traditionally kick existing player when new one joins

DooblyNoobly: Nope, once a player disconnects their NetworkConnection gets destroyed

fomalsd: Well then if someone steals your account you won't be able to stop him by joining the game

DooblyNoobly: they can reset their password and an admin can kick the player, or wait until next round to get it back

PR #2617, opened 2020-01-16
Fixed annoying occupations meta change

### LINKED ISSUE #2374: Fix Occupation assets showing as changed
# Description
They keep showing that the case of one of the fields is changing even when you haven't touched or viewed them.

Probably just need to remove the FormerlySerializedAs field and set all the Occupation assets up to make sure they still have the correct job type. I think this might be caused by the name overlap between the formerlyserializedas and the public get-only property.

Macoron:
### Purpose
Fixes #2374 - occupations meta doesn't rewrite again


chairbender:
thanks, was annoying!

PR #2622, opened 2020-01-17
Made Posters to no longer be pullable

### LINKED ISSUE #2618: Posters can be pulled
## Description
You shouldn't be able to pull posters.

I think you just need to edit their ObjectBehaviour component in the editor.

image

### Steps to Reproduce
1.  Find poster
2.  Pull it with Ctrl click or right click menu
3.  lol

nikopetr:
# Pull Request Template

### Purpose
In order to make posters not pullable, Fixes #2618.

### Notes:
Removed the ObjectBehaviour component from the Poster prefab. 
Tested after removed, works fine


chairbender:
Honestly did not expect the object would function properly without object behavior, but it does seem to be working correctly in MP. Thanks!

PR #2625, opened 2020-01-17
Fixed blood spots to not cause gravity

### LINKED ISSUE #2582: Blood spots cause gravity
If you go out into space and bleed, the tile that has blood on it is treated as having gravity. Blood splats should have absolutely no effect on movement logic.

nikopetr:
### Purpose
Fixed blood spots to not cause gravity, fixes #2582.

### Notes:
Followed @chairbender guidance and added a public method in PushPull, called CausesGravity()  for it.


UnityProject/Assets/Scripts/Objects/PushPull.cs
```diff
@@ -491,6 +492,17 @@ public RightClickableResult GenerateRightClickOptions()
 			}
 		};
 	}
+	/// <summary>
+	/// Return true if the object causes gravity, otherwise false.
+	/// <returns>bool that represents if the object must cause gravity or not.</returns>
+	/// </summary>
+	public bool CausesGravity()
+    {
+		if (isNotPushable == true || floorDecal!= null)
```
chairbender: Not quite right, this causes lockers to stop when pushed out into space.
```suggestion
		if (!isNotPushable || floorDecal!= null)
```

The previous logic is - if an object is not pushable, it causes gravity, thus returning true from this method. Currently it would return false from this method if an object is not pushable. The suggested change fixes that.

chairbender (changes requested):
Thanks for looking into this @nikopetr . The blood splat issue is fixed, however there's a regression introduced by this. 
Try pushing an object such as a locker into space. It should keep drifting, but it stops. This issue doesn't happen on latest develop.

I believe that something may be inverted in the boolean logic in your changed lines. See my comment.

UnityProject/Assets/Scripts/Objects/PushPull.cs
nikopetr: Thanks for the review @chairbender! You are right. I just wonder why does that happen on lockers since CausesGravity() method is only used by MetaTileMap.IsNoGravityAt() at the moment. Did you try using that method on other objects as well?

nikopetr:
Thanks for the review @chairbender! You are right. I have commited the requested changes. Just wondering why does that happen on lockers since CausesGravity() method is only used by MetaTileMap.IsNoGravityAt() at the moment. Did you try using that method on other objects as well?

chairbender:
Nice, fixed 👍 

It would happen on any impassable but pushable object, such as canisters.

PR #2632, opened 2020-01-17
More fixes

Bod9001:
Makes mining satchel  be able to pick up diamonds and blue space Crystals ,
Make the food and ice cream carts be able to be pushed around

Tested a little bit in editor


chairbender:
thanks 👍

PR #2636, opened 2020-01-18
Adds GameData.OfflineMode

chairbender:
### Purpose

Adds a special field on GameData called OfflineMode. Checking this will allow you to skip login and allow the game to work even if you can't connect to the auth server. By default it is disabled - it's only to be used in those uncommon circumstances.

Also disables login skip (which doesn't work anyway) unless OfflineMode is enabled.


chairbender:
this is small and doesn't affect anything unless you enable it (and defaults to disabled) so going to merge this myself

PR #2670, opened 2020-01-19
feat(admin): added a quick add-admin console command

Jibbajabbafic:
Adds a quick and dirty command to add admins from the console

### DIFF
--- a/UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
+++ b/UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
@@ -8,6 +8,7 @@ using UnityEngine;
 using Mirror;
 using UnityEditor;
 using Random = UnityEngine.Random;
+using DatabaseAPI;
 
 namespace IngameDebugConsole
 {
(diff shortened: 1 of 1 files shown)


Bod9001:
I can get this to work  do you add the long jumble of numbers? 
like add-admin oi435lkfd-riglihjkndn- so and so?

Bod9001:
it said it was adding it but,  the admin  text didn't change , and I didn't get admin but that might be Another issue

Bod9001:
This is using the Build by the way

Jibbajabbafic:
You can find your account ID in the console (there's a message when you connect like this: [Admin] XXX logged in successfully. userid: XXXXXXXXXXXX).

Bod9001:
does it also show a bunch of stuff like, job n stuff?

Jibbajabbafic:
Nope that's a different ID, no idea why it doesn't use the same ID

Bod9001:
ok, Will try to use the correct ID this time

UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
```diff
@@ -440,5 +441,21 @@ private static void PlayStrandedEnding()
...
+		public static void AddAdmin(string userIDToPromote)
+		{
+			if (CustomNetworkManager.Instance._isServer == false)
+			{
+				Logger.LogError("Can only execute command from server.", Category.DebugConsole);
+				return;
+			}
+
+			var player = PlayerList.Instance.GetAdmin(ServerData.UserID, PlayerList.Instance.AdminToken);
+			if (player != null)
```
DooblyNoobly: No need to get admin here, if you can do server only console commands you should be consider an admin by default. Also this will prevent hosts from making admins if they haven't added themselves to the admin list yet

DooblyNoobly (changes requested):

UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
Jibbajabbafic: Good point

Bod9001:
Just say I haven't been able to get this working on builds, it doesn't add anything to the admins text maybe it's just builds ( done inside of the build with the 5f menu
 )

PR #2673, opened 2020-01-19
Wall Protrusion hit box fixed

### LINKED ISSUE #1258: Bug: Light hit boxes are too big
# Description

The light hitbox is as big as a tile. It needs to fit the size of the light itself. Also consider adding the effects of the light when it breaks (sparks and broken tube sprite)

Macoron:
### Purpose
Fixes #1258 - hit boxes of wall protrusion (light bulb, camera, etc) now rotates correctly.
Also lights are breakable now again

### Notes:
Introduced new DirectionalRotatesParent. It checks rotation of Directional and rotate only parent.
Scene was changed due to scene overrides of local sprite rotations.


Bod9001:
have you considered the network part of it, (I would test but the authentication system is messing with it) currently, I'm not sure if client interacts with the hit boxes, but if it does you make sure it's properly set on the client

chairbender:
I'm in progress on a code review but I have some testing I'll need to do that I can't for a few hours. I like the approach but I have concerns about changing the root transform as our rotation system currently has a baked-in assumption that all objects always have a Quaternion.identity localRotation on their root transform, and this violates that assumption. That's not to say we can't modify the rotation logic to no longer require this assumption! But such a structural change would make me concerned for 0.4 release so I'd like to postpone it. 

Still, I'll test it out and see if it's breaking anything. I suspect this might break on rotated / moving matrices, particularly on clientside.

chairbender:
Actually, considering that all wall protrusions are always fixed in place and it's not possible to construct them...THIS might actually be the good short-term solution and we can re-evaluate it when it comes time to implement wall protrusion construction.

chairbender:
Okay, so this actually does seem to work fine. I forgot that wall protrusions don't really move around much.

Just to inform you, the reason I was concerned is that, back when I was working on rotation sync issues, I decided for the rotation logic to have a convention where objects should always have Quaternion.identity for their localRotation. This violates that convention. That's why all the other rotation components only rotate children.

I can't remember why I decided I needed that convention, I think it was just so all objects would have a common reference point. It's possible the need for it can be eliminated with some refactoring. 

I actually like this approach though, it's simpler and doesn't require messing with children and now that I see it in action I can think of how we can eliminate the need to always have Quaternion.identity for local rotation. I might eventually switch the other rotation logic over to this after 0.4, and then we can remove all the "rotates children" components and just use this one.

PR #2675, opened 2020-01-19
pushed the unity changes that unity likes to make every so often, and

Bod9001:
Fixed a bug with tile preview with on validate that caused it to loop Infinitely,
and and added some catch Statements for an issue that was happening on the EU test
And pushed prefab changes

tested in editor


DooblyNoobly:
what is with all of the prefab changes?

chairbender:
We've tried pushing the prefabs in the past and it re-emerged...I'm thinking it might have something to do with minor differences in version? Or maybe unity was always like this. Anyway, I"m not sure pushing the prefabs helps.

PR #2681, opened 2020-01-20
Fixes a Server start Not being called issue

Bod9001:
If there is invalid IDs for objects it causes an error that, causes the start server not to run
so Remember to open your Outpost scene and save it


DooblyNoobly:
Not sure this will fix it but hey lets give it a go

Issue #2639, opened 2020-01-18
Shuttles Don't Repressurize (Fix AirVent / Scrubbers?)


chairbender:
## Description

Some (all?) shuttles don't repressurize after opening the doors into space and letting them depressurize. For example, command escape shuttle has this problem. I think all shuttles should recover their atmosphere after venting.

Maybe this is just as simple as mapping in an AirVent and Scrubber to each shuttle?

EDIT: For this issue it's fine to just fix it so that they repressurize as long as atmos is working. You don't need to fix the issue where atmos dies.

chairbender:
Excuse my indulgence, but I'm putting this into 0.4 because it makes shuttles annoying to use if you have to always wear a hardsuit.

fomalsd:
fyi neither AirVent nor Scrubber work at the moment

Bod9001:
Maybe you could bring a canister with you as a Stand in?

chairbender:
Ah, so maybe fixing airvent / scrubbers would be a good solution, and mapping them in

JesterX666:
On a server for which the atmos system is working (not stopped), the pressure is getting back.  However when I examined the gas mix, it was my plasma canister leaking even if it was secured to the fuel connector.  Not sure it's supposed to leak like that when secured.

Also, the fuel connector flew right in my face when I reopenned the door and lost pressure again.  Fuel connectors are not supposed to move from the floor arn't they?

chairbender:
Ah ok. So this is probably just #2628 

I believe canister shouldn't leak unless you actually open it. It's not required to open the valve when connecting it to the fuel port. When piping stuff is more thought out we need to revisit how this works.

chairbender:
@JesterX666 I'm talking about when the shuttle is away from the station...I'm not seeing it repressurizing in those circumstances. Even when atmos is working fine. I tested this on mining shuttle btw, not sure if we are talking about the same thing.

EDIT: And there's nothing to my knowledge that would cause it to repressurize unless you map in a scrubber / air vent

chairbender:
TODO: I also need to ensure shuttle connector is anchored and not pullable / able to be blown around

JesterX666:
@chairbender It was just an observation that my shuttle was getting repressurized but it was caused by plasma leaking.  (Which imo is another bug, that I will log on my lunch break with video proof).  Plasma pressure isn't good since it's toxic :p

If O2 and N2 air mix would be injected, it would repressurize the shuttle.

PR #2712, opened 2020-01-23
Restore AirVent / Scrubber functionality

chairbender:
### Purpose
Part of #2639 . The mapping change needs to be made, however, to map airvents and scrubbers into each shuttle. I will do that in a different PR.

This also simplifies the logic of/removes stuff for pipenets in Pipe.cs that wasn't fully implemented and wasn't working anyway. Also simplifies logic / sprites for attaching / detaching them so it's clear what state it is.

Also makes connectors un-pushable.

AirVents now revert to their old behavior (turn tile into GasMix.Air) and Scrubbers were changed to remove a fraction of the gases from their tile (the old behavior for scrubbers was to turn their tile into Space gasmix which causes really strong wind). In my testing this combination produced the best results for gameplay. 

There is probably room for improvement but this should do for now. Trying to change the way that they worked too much was becoming a time sink, would rather focus on more pressing issues as long as this stuff is in a playable / working state.

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Objects/Atmos/AirVent.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Objects/Atmos/AirVent.prefab
@@ -22,6 +22,7 @@ GameObject:
   - component: {fileID: 114798627279048580}
   - component: {fileID: 114511699214781296}
   - component: {fileID: -678374216968888500}
+  - component: {fileID: 8341462466351120766}
   m_Layer: 10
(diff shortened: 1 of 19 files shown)


fomalsd (changes requested):
thanks!
pre-mapped scrubbers/vents work alright, but spawned ones don't work at all:
GIF
also when a scrubber is destroyed in wrenched/working state, its last tile continues sucking air:
GIF

Issue #2661, opened 2020-01-19
Microwave Doesn't Spawn Meal


chairbender:
## Description

On headless, we went to the microwave in the lobby and tried cooking some meat. It played the animation but when it was done, it didn't spawn the cooked meal.

daalta:
Huh, odd. Inserting meat into the microwave yields the 'The microwave is empty' interact prompt. Normally this shows the remaining time.

daalta:
Yeah tried this myself in the editor (with 2 clients), it really is a headless-only issue. Bummer, wouldn't really know how to investigate this.

I made the most recent microwave changes, but I don't think I touched anything network-related.

chairbender:
@daalta it is possible to run a server in headless mode locally by enabling GameData.TestServer. However, some of these headless issues might be caused by other things happening on our headless server which only happen after a bit of gameplay / round restarts, such as breaking atmos / UpdateManager.

daalta:
@chairbender Thanks! Good to hear that it might not be the fault of whoever worked on microwaves! Wink wink

DooblyNoobly:
still doesn't work on stable headless

daalta:
~~https://github.com/unitystation/unitystation/pull/2732 supposedly fixed this, but... I think microwaves are still broken on headless?~~

Nevermind! It's a graphical issue with the sprite not resetting. @JesterX666 fixed the issue nicely though, my bad 👍

![Unitystation_2020-01-26_00-43-21](https://user-images.githubusercontent.com/9169275/73128810-51222e80-3fd5-11ea-8cfa-a36f25b67cd8.png)

JesterX666:
The now don't turn off.  Repairing.

PR #2732, opened 2020-01-25
Fix microwave spawning meals in headless

JesterX666:
Fix microwave spawning meals in headless
The timer variable was not set on server, only on client.

Fixes
Fix microwave spawning meals in headless #2732

### DIFF
--- a/UnityProject/Assets/Scripts/Machines/Microwave.cs
+++ b/UnityProject/Assets/Scripts/Machines/Microwave.cs
@@ -15,9 +15,10 @@ public class Microwave : NetworkBehaviour
 
 	/// <summary>
 	/// Time left until the meal has finished cooking (in seconds).
+	/// I don't see the need to make it a SyncVar.  It will economize network packets.
 	/// </summary>
 	[HideInInspector]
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Machines/Microwave.cs
```diff
@@ -15,9 +15,10 @@ public class Microwave : NetworkBehaviour
 
 	/// <summary>
 	/// Time left until the meal has finished cooking (in seconds).
+	/// I don't see the need to make it a SyncVar.  It will economize network packets.
 	/// </summary>
 	[HideInInspector]
-	public float microwaveTimer = 0;
+	public float MicrowaveTimer = 0;
```
DooblyNoobly, line 21: Fields should be camelCase. Use PascalCase for properties. Not a big issue so will merge this.

daalta:
I don't quite understand what was changed in this PR, all I can see it a change in capitalization... 😗
EDIT: But it *did* fix https://github.com/unitystation/unitystation/issues/2661, so it must do something 😅

PR #2742, opened 2020-01-26
Fix the microwave not turning off

JesterX666:
I accidently removed the timeout assignation on client side *oops* Sorry


daalta:
Link for reference https://github.com/unitystation/unitystation/pull/2732

PR #2765, opened 2020-01-27
Collision atmos

### LINKED ISSUE #2291: Shuttle crashing into station doesn't update atmospherics properly B: 100
# Description
It seems that atmospherics isn't properly notified about the destruction of tiles resulting in with behaviour like

image
image
image
image

## Category
BUG 

## Steps to Reproduce
1.Get on mining shuttle 
2.smash in to station
3.Open up Matrix checker
4 Observe strange space values In atmospherics

krille90:
### Purpose
Fixes #2291, where crashing shuttles and then created space tiles wouldnt be updated properly.
Problem was how the neighbors of a node were chosen.
Previously, if a neighboring tile was space, there was a check if there is a neighboring matrix and if there is a tile.
Now, this check doesn't trigger anymore, when the current tile is also space, thus choosing the proper neighboring tiles.

Extended the matrix check for neighboring nodes to draw a white line between two neighboring nodes.

### Notes:
Multiplayer test wasn't done, since the fix doesn't touch anything network-related.

### DIFF
--- a/UnityProject/Assets/Scripts/Editor/Matrix/Utils/GizmoUtils.cs
+++ b/UnityProject/Assets/Scripts/Editor/Matrix/Utils/GizmoUtils.cs
@@ -103,6 +103,21 @@ public static class GizmoUtils
 		Gizmos.DrawWireCube(position + (local ? HalfOne: Vector3.zero), Vector3.one * size);
 	}
 
+	public static void DrawRay(Vector3 pos, Vector3 direction, bool local=true)
+	{
+		if (direction == Vector3.zero)
+		{
+			return;
+		}
+
+		if ( local )
+		{
+			pos += HalfOne;
+		}
+
+		Gizmos.DrawRay(pos, direction);
+	}
+
 	public static void DrawArrow(Vector3 pos, Vector3 direction, bool local=true, float arrowHeadLength = 0.25f, float arrowHeadAngle = 20.0f)
 	{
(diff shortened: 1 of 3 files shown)


krille90:
will add some comments on the MetaDataSystem class to make it more understandable.

krille90:
done commenting

JesterX666:
On it.

JesterX666:
Code looks clean enough.  Just gonna test it before merging it.

JesterX666:
Moles + Connected to space

Connected to Space and moles are still behaving weird. See that row that contains those 2 x "6.829" moles tiles? That row is not loosing gaz component at all.

JesterX666:
Is it normal that the red tiles are both "Connected to space" :

image

And "Space" at the same time ? :

image

I thought that space would have "no moles" instead of 0.

Please see with a pro at atmos if this is acceptable.

krille90:
> Is it normal that the red tiles are both "Connected to space" And "Space" at the same time ?

Yes, the check is defined like that:
`node.IsSpace || node.Neighbors.Any(n => n != null && n.IsSpace)`

> I thought that space would have "no moles" instead of 0.

Moles in GasMix are always the sum of all gases (with is a float array):

`public readonly float[] Gases;`
`public float Moles => Gases.Sum();`

JesterX666:
Since it doesn't seem to cause troubles and everything seems to work fine, I'm merging it.

PR #2772, opened 2020-01-28
Fixed MetaDataNodes being able to have more than 4 neighbors

fomalsd:
# Pull Request Template

### Purpose
As per title
Before:
b4
After:
after


fomalsd:
merging since it's stable anyway

PR #2774, opened 2020-01-28
Fixed TTS not being triggered by player speech

### LINKED ISSUE #2766: TTS stopped speaking & causes exceptions in the editor
## Description

* TTS is no longer triggered because of recent chat log changes.
* TTS causes editor exceptions in certain cases.

These two issues are not related, but could both be fixed together.

Related: https://github.com/unitystation/unitystation/issues/1525

### Detailed Bug Report
* 90508f1 changed the colons in the chat log to commas. TTS relies on detecting the commas.
* TTS causes exceptions when a message contains no pronouncable characters (such as "!!!").

### Steps to Reproduce
1. Enable TTS.
2. Speak - you'll hear no TTS.
3. Say "!!!" and observe the exceptions.

### Proposed solutions

* The quick fix: Replace the commas with colons.
  * The previous changes to colons / commas can easily be undone but...
  * Not all messages contain commas or colons. This is intentional for messages such as emotes. 
  * However, players may be able to insert their own symbols in chat to trigger TTS!
  * The exceptions would require their own fix.
* The more robust fix: Format messages in Chat.Process.cs for TTS. Only try to use TTS when the message is on the correct channel, instead of relying on detecting certain characters.
  * This would probably be considered a refactor.

daalta:
### Purpose
* Fixes https://github.com/unitystation/unitystation/issues/2766 (TTS was no longer being triggered)
  * TTS is now triggered by commas instead of colons. And more reliably, too.
  * Fixed some TTS exceptions.
* Various formatting changes, added comments and autoformatting.
### Known Issues:
* If a chat message contains a comma it be read by TTS. This also includes interact messages.
  * Could be fixed by looking for invisible special characcters instead of commas (eww)
  * Could be fixed by looking at the chat modifiers. If there are none, discard the message.
* TTS still causes exceptions occasionally. Requires further investigation.

### Self checks:

### DIFF
--- a/UnityProject/Assets/Scripts/Chat/ChatRelay.cs
+++ b/UnityProject/Assets/Scripts/Chat/ChatRelay.cs
@@ -17,6 +17,13 @@ public class ChatRelay : NetworkBehaviour
 	private ChatChannel namelessChannels;
 	public List<ChatEvent> ChatLog { get; } = new List<ChatEvent>();
 
+	/// <summary>
+	/// The character indicating that the following text is speech.
+	/// Currently the comma indicates speech.
+	/// For example: Player says, "ALL CLOWNS MUST SUFFER"
+	/// </summary>
+	private char saysChar = ',';
+
 	private void Awake()
 	{
 		//ensures the static instance is cleaned up after scene changes:
@@ -34,7 +41,7 @@ public class ChatRelay : NetworkBehaviour
 	public void Start()
 	{
 		namelessChannels = ChatChannel.Examine | ChatChannel.Local | ChatChannel.None | ChatChannel.System |
-		                   ChatChannel.Combat;
+						   ChatChannel.Combat;
 	}
 
 	[Server]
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Chat/ChatRelay.cs
```diff
@@ -205,4 +190,27 @@ private void UpdateClientChat(string message, ChatChannel channels)
...
 		}
 	}
+
+	/// <summary>
+	/// Sends a message to TTS to vocalize.
+	/// Only messages that are spoken by player characters are currently vocalized.
+	/// Messages must contain at least one letter from the alphabet.
+	/// </summary>
+	/// <param name="message">The message to try to vocalize.</param>
+	private void trySendingTTS(string message)
```
DooblyNoobly, line 200: Methods should be PascalCase. Remember for next time

daalta: Whoops! Will do.

Feel free to request changes next time I forget 😙

Issue #2753, opened 2020-01-27
A server only method was called on a client in chat.cs


DooblyNoobly:
When I clicked the trashbag in my hand I got this error in the client. Build 3984:

```
[Chat] A server only method was called on a client in chat.cs
UnityEngine.DebugLogHandler:Internal_Log(LogType, LogOption, String, Object)
UnityEngine.DebugLogHandler:LogFormat(LogType, Object, String, Object[])
UnityEngine.Logger:Log(LogType, Object)
UnityEngine.Debug:LogError(Object)
Logger:TryLog(String, LogLevel, Category, Object[]) (at /home/unitystation/UnityProject/Assets/Scripts/Debug/Logger.cs:163)
Logger:LogError(String, Category) (at /home/unitystation/UnityProject/Assets/Scripts/Debug/Logger.cs:117)
Chat:IsServer() (at /home/unitystation/UnityProject/Assets/Scripts/Chat/Chat.Process.cs:409)
Chat:AddExamineMsgFromServer(GameObject, String) (at /home/unitystation/UnityProject/Assets/Scripts/Chat/Chat.cs:388)
InteractableStorage:Interact(HandActivate) (at /home/unitystation/UnityProject/Assets/Scripts/Clothing/BackPack/InteractableStorage.cs:296)
InteractionUtils:CheckInteractInternal(IBaseInteractable`1, HandActivate, NetworkSide, Boolean&) (at /home/unitystation/UnityProject/Assets/Scripts/Input System/InteractionV2/InteractionUtils.cs:76)
InteractionUtils:ClientCheckAndTrigger(IEnumerable`1, HandActivate) (at /home/unitystation/UnityProject/Assets/Scripts/Input System/InteractionV2/InteractionUtils.cs:52)
UI_ItemSlot:TryItemInteract() (at /home/unitystation/UnityProject/Assets/Scripts/UI/UI Bottom/UI_ItemSlot.cs:346)
UI_ItemSwap:OnClick() (at /home/unitystation/UnityProject/Assets/Scripts/UI/UI Bottom/UI_ItemSwap.cs:33)
UI_ItemSwap:OnPointerClick(PointerEventData) (at /home/unitystation/UnityProject/Assets/Scripts/UI/UI Bottom/UI_ItemSwap.cs:23)
[...]

PR #2775, opened 2020-01-29
Corrects many NRE with trash bag and mining satchel

JesterX666:
Correction for : A server only method was called on a client in chat.cs #2753

Corrected a NRE that was caused by calling AddExamineMsgFromServer (a server only method) by the client

Corrected a NRE that was caused by triggering OnClothingEquiped since mining satchels / trash bags are not clothing


chairbender:
thanks!

PR #2785, opened 2020-01-29
Menu music: Replaced low quality tracks, normalized peak volume, and more

daalta:
Unity_2020-01-29_12-26-53
# Purpose
* Various improvements to menu music, such as higher quality music files, removed silences and peak normalization.
* Reduced import quality *slightly* to decrease music file sizes
### Notes:
* All tracks: Normalized peak amplitude to -1 dB, removed DC offset, reduced Unity import quality from 100 to 81
* Clown title: Replaced file with higher quality version, added fadeout
* Flipflap: Completely re-rendered from the original mod file
* Robocop: Replaced file with higher quality version
* Space Asshole: Shortened fadeout by half
* Tintin: Removed 9 second silence at the start

### Self checks
* [x] Music sounds much better now
* [x] Volume is consistent across all OGG tracks
* [x] HONKs are now in HD


fomalsd:
very nice!
> Flipflap: Completely re-rendered from the original mod file

there is a .mod file for it? why not playback a mod file using sunvox?

daalta:
@fomalsd Very good idea! I've not touched sunvox yet, but that should be possible. I found the mod here, if you'd like to do it yourself: https://archive.org/details/XCeedFlipFlap

Issue #2407, opened 2019-12-02
Dead/crit players get back up when stun is over


fomalsd:
# Description
![image](https://user-images.githubusercontent.com/10403536/69956888-32357d80-14f9-11ea-8b89-91dfb992210e.png)

## Steps to Reproduce
Please enter the steps to reproduce the bug or behaviour:
1. Beat yourself up a bit
2. Explode large tanks nearby to get stunned and hit by a few objects to crit/die
3. See your dead/unconscious player get back up after a while

chairbender:
nvm though I saw this happening in other situations but I see this ticket covers them

chairbender:
would like to add, on client side, I see lots of cases where dead players never even show up laying down, they just die standing up.

fomalsd:
This is a thing again

fomalsd:
![i](https://user-images.githubusercontent.com/10403536/73135721-99cdfc00-4056-11ea-9dd9-cc1aa535e052.gif)

chairbender:
ah, so the reproduction is getting knocked down and then getting killed. It's making you stand back up again due to the timer that was set when you first were knocked down which stands you back up when timer is done. I think all that's needed here is adding a check for those various things that can knock you down so that whatever they are calling to stand you back up doesn't occur if player is dead.

JesterX666:
Fixed by PR #2801

PR #2801, opened 2020-01-31
Do not raise up a dead body!

JesterX666:
Fixes Dead/crit players get back up when stun is over #2407

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterPlayer.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterPlayer.cs
@@ -195,7 +195,13 @@ public class RegisterPlayer : RegisterTile, IServerSpawn
 	{
 		var oldVal = IsSlippingServer;
 		IsSlippingServer = false;
-		SyncIsLayingDown(false);
+
+		// Do not raise up a dead body
+		if (playerScript.playerHealth.ConsciousState != ConsciousState.DEAD)
+		{
+			SyncIsLayingDown(false);
+		}
+
 		OnSlipChangeServer.Invoke(oldVal, IsSlippingServer);
 
 		if ( playerScript.playerHealth.ConsciousState == ConsciousState.CONSCIOUS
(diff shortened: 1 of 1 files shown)


daalta (commented):
Looks good, trying it in the editor now.

fomalsd:
Scr 2020-02-01 в 1 19 33
does the job.
one thing I've noticed though is that player also gets up in crit, while not supposed to.
so maybe extend that fix to crit players, too?

daalta:
@fomalsd Hmmm... Maybe? Instead of checking if a player is dead we could check if they're not ConsciousState.CONSCIOUS. Though I feel like that might introduce problems with players not being able to stand up once the stun wears off.

We could make a separate issue for that. @JesterX666 's fix is probably good enough to fix dead players standing up..?

fomalsd:
#2407 clearly reads "Dead/Crit..." while this fix only addresses dead ones.
I think there's an event being invoked when conscious state is changed (that eventually updates upright state, too), so checking for ConsciousState.CONSCIOUS makes sense to me

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterPlayer.cs
```diff
@@ -195,7 +195,13 @@ private void ServerRemoveStun()
 	{
 		var oldVal = IsSlippingServer;
 		IsSlippingServer = false;
-		SyncIsLayingDown(false);
+
+		// Do not raise up a dead body
+		if (playerScript.playerHealth.ConsciousState != ConsciousState.DEAD)
```
fomalsd:
```suggestion
		if (playerScript.playerHealth.ConsciousState == ConsciousState.CONSCIOUS)
```

JesterX666:
Oh!  I forgot about "crit".  I will do it as soon this evening.

JesterX666:
Yup, the == Concsious does take care of not getting up while "crit"
Thanks @fomalsd

fomalsd (approved):

PR #2803, opened 2020-02-01
Smoother Shuttle Movement

### LINKED ISSUE #2423: Shuttle lerping is jerky
# Description
We need a better formula for shuttle lerping as it doesn't behave well at low speeds: ship doesn't even appear to move for a couple of seconds while it's supposed to start, and then proceeds with speed fluctuations:
GIF

It needs to be smooth (no jerking or rubberbanding) and working at all speeds

## Steps to Reproduce
1. Get in the mining shuttle and make sure it has a plasma tank connected
2. Make sure speed is set to lowest
3. Start the ship

chairbender:
### Purpose
Fixes #2423 

Server-side and client side shuttle movement was "discarding" small parts of movement. This
ensures the movement is no longer discarded while still running all the logic that's supposed
to be run at each exact tile position. Shuttle movement appears to be a lot smoother and less jerky and all the tile-based logic still seems to work properly.

### DIFF
--- a/UnityProject/Assets/Scripts/Shuttles/MatrixMove.cs
+++ b/UnityProject/Assets/Scripts/Shuttles/MatrixMove.cs
@@ -486,7 +486,7 @@ public class MatrixMove : ManagedNetworkBehaviour
 		else if (IsMovingClient)
 		{
 			//Only move target if rotation is finished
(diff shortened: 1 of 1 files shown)


fomalsd (approved):
working nicely locally. merging to see how it behaves on a live server

PR #2804, opened 2020-02-01
Automatic admin for editor

Bod9001:
### what this do?
Automatic off-line mode for editor, can be turned off if needed in game data prefab
### why you do this?
so  you have admin tools when you need to on the editor and Don't make a prefab dirty

### what you tested?
Editor and builds

### Potential Issues?
Might cause login issues not to be shown on editor When in general use, so to test you can disable it in game data

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/SceneConstruction/NestedManagers/GameData.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/SceneConstruction/NestedManagers/GameData.prefab
@@ -44,5 +44,5 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   offlineMode: 0
-  editorofflineMode: 1
+  editorAdmin: 1
   testServer: 0
--- a/UnityProject/Assets/Resources/Prefabs/SceneConstruction/NestedManagers/GameData.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/SceneConstruction/NestedManagers/GameData.prefab
@@ -44,4 +44,5 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   offlineMode: 0
+  editorofflineMode: 1
   testServer: 0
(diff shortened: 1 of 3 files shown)


chairbender (changes requested):
Hm, I'm concerned with how this is defaulting to OfflineMode when running via Editor unless a dev is aware of this flag and manually turns it off. I think the default behavior of the Editor should be to closely match how it will be for clients / headless server. OfflineMode doesn't just grant admin privs - it also skips talking to the auth server, thus would cause some critical code paths to be missed and might increase the number of bugs introduced which don't appear locally but do appear on headless / built client.

Instead of this, I suggest host player always be granted admin privs by default, regardless of whether the host is in editor or in the standalone build. This should be done without using OfflineMode. Any time someone is hosting a game themselves, they have full control of the server anyway so there's no reason to restrict their permissions in any way.

EDIT: If you want to see how you can implement this logic, find usages of OfflineMode, that'll show you where the "is admin" checks are being done.

Bod9001:
added Asked changes

chairbender (changes requested):
@Bod9001 please make it so that host player is always admin, regardless of if it's launched in editor

Bod9001:
idk, How I would detect that 
if (isServer && isClient)  
Would this trigger on headless Server?

Bod9001:
Logger.Log(NetworkBehaviou.isClient.ToString() + "c1");
Logger.Log(NetworkBehaviou.isServer.ToString() + "S1");
Logger.Log(NetworkBehaviou.isServerOnly.ToString() + "F1");
Logger.Log(NetworkBehaviou.isClientOnly.ToString() + "D1");
Logger.Log(NetworkBehaviou.isLocalPlayer.ToString() + "E1");

none of these are true in Init()

Bod9001:
I can't think of a way of implementing it currently

chairbender:
you need to look beyond network behavior fields. There is probably already a way to tell if current game is being hosted with host player. And if not, add one. Just set a flag if they click the host button or something. Shouldn't be tricky.

DooblyNoobly:
stagnating. closing this for now

PR #2821, opened 2020-02-02
Server NREs

chairbender:
Fixes a few server NREs and other issues:

- SpaceCleaner NRE
- NRE that can happen when trying to uncuff
- Chat NRE that can happen when there's a joined viewer that hasn't spawned
- Vendor_GUI NRE
- Fix Vendor gui missing references
- Tried to fix vendor GUI scrolling right on client side occasionally when buying something but it didn't seem to take. Will make an issue, I think there's an issue with the layout.


chairbender:
oops, included unity change

Issue #2660, opened 2020-01-19
Fire Resist Action B:20


chairbender:
## Description

Currently when on fire there's no way to resist it. Add a way so you can resist and remove fire stacks by dropping and rolling or however it works on tg.

Bod9001:
Doesn't need the animation, Just need to have cool down if you're doing it yourself  added to the resist button to do it yourself, And if you do Help intent on someone who is on fire you put them out as well, It's quicker than doing it yourself

daalta:
Yep, a fire resist button is needed.

... Though I can think of another way to remove fires from a player?
How about players being able to pat fires, either on themselves or on other people.

~~We might even be able to wrangle it into the hug code we already have.~~
```
if (lhb != null && lhbOther != null && (lhb.FireStacks > 0 || lhbOther.FireStacks > 0))
{
    lhb.ApplyDamage(huggedPlayer, 1, AttackType.Fire, DamageType.Burn);
    lhbOther.ApplyDamage(gameObject, 1, AttackType.Fire, DamageType.Burn);
    Chat.AddCombatMsgToChat(gameObject, $"You give {hugged} a fiery hug.", $"{hugger} has given {hugged} a fiery hug.");
}
```
~~Lines 543-548 of PlayerNetworkActions.cs already checks if the player is on fire. By doing the reverse we could remove fire stacks.~~
EDIT: ...Aaaaand @Bod9001  did a quicker writeup than me 😁

daalta:
I'm investigating the option I just mentioned. I'll leave resisting fires via the resist button to someone else.
So another person can feel free to assign themselves to this issue! 🎉🎉🎉

daalta:
The PR above introduces **FIRE PATTING**! Pat yourself, pat yourself and watch those flames **disappear**.

Though it's not fire resisting. Dropping and rolling might be (should be?) easier and more effective than patting out yourself. Buuuuut now other players can help you.

daalta:
https://github.com/unitystation/unitystation/pull/2830 has been closed (for now), so this feature still needs to be implemented. I'd need some more PR feedback before reopening it.

PR #2830, opened 2020-02-02
Patting out players who are on fire (new interaction)

daalta:
### Purpose
* Previously there was not much a player could to about being on fire. https://github.com/unitystation/unitystation/issues/2660 requested resisting fires.
* This PR lets players extinguish the flames on other players by patting them. Players can also pat out themselves (which is less effective).

### Notes:
* Patting out players is a new interaction.
* CPR can no longer be performed if the patient is on fire. (They will be patted instead)
* Patting decreases the fire stacks on that player. Patting other players is more effective than patting yourself. The number may need to be tweaked in case patting out fires is too easy.
* Added a function for getting a player's reflexing pronoun ("herself", "himself," ...)
* Made the function for changing fire stacks public. (Or is there another method..?)
* Players can still give each other fiery hugs - but only when the person performing the hug is the one on fire. (Sorry @chairbender !)
* Did a tiny bit of autoformatting.

- [x] The PR has been tested in multiplayer (**Editor only**)

### DIFF
--- a/UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
+++ b/UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
@@ -98,7 +98,7 @@ public abstract class LivingHealthBehaviour : NetworkBehaviour, IHealth, IFireEx
 	//how on fire we are, sames as tg fire_stacks. 0 = not on fire.
 	//It's called "stacks" but it's really just a floating point value that
 	//can go up or down based on possible sources of being on fire. Max seems to be 20 in tg.
-	[SyncVar(hook=nameof(SyncFireStacks))]
+	[SyncVar(hook = nameof(SyncFireStacks))]
 	private float fireStacks;
 
 	/// <summary>
@@ -231,9 +231,19 @@ public abstract class LivingHealthBehaviour : NetworkBehaviour, IHealth, IFireEx
 		SyncFireStacks(0);
 	}
 
-	public void SyncFireStacks(float newValue)
+	/// <summary>
+	/// Sets the serverside fireStacks of a LivingHealthBehaviour to a new value.
+	/// </summary>
+	/// <param name="newValue">The new value, clamped to values 0 and above.</param>
+	[Server]
+	public void ServerSetFireStacks(float newValue)
 	{
-		this.fireStacks = Math.Max(0,newValue);
+		this.fireStacks = Math.Max(0, newValue);
+	}
+
+	private void SyncFireStacks(float newValue)
+	{
+		this.fireStacks = Math.Max(0, newValue);
 		OnClientFireStacksChange.Invoke(this.fireStacks);
 	}
 
@@ -241,7 +251,8 @@ public abstract class LivingHealthBehaviour : NetworkBehaviour, IHealth, IFireEx
 	/// PUBLIC FUNCTIONS: HEAL AND DAMAGE:
 	/// ---------------------------
 
-	private BodyPartBehaviour GetBodyPart(float amount, DamageType damageType, BodyPartType bodyPartAim = BodyPartType.Chest){
+	private BodyPartBehaviour GetBodyPart(float amount, DamageType damageType, BodyPartType bodyPartAim = BodyPartType.Chest)
+	{
 		if (amount <= 0 || IsDead)
 		{
 			return null;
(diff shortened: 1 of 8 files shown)


UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
```diff
@@ -231,7 +231,7 @@ public void Extinguish()
 		SyncFireStacks(0);
 	}
 
-	private void SyncFireStacks(float newValue)
+	public void SyncFireStacks(float newValue)
```
chairbender: This is a SyncVar hook, so I'd prefer this to be kept private and instead make a public [Server] annotated method which calls this internally (like [Server] public void ServerSetFireStacks).

When you make SyncVar hook public, it could result in another dev mistakenly calling it from clientside code, which would not work correctly.

UnityProject/Assets/Scripts/Player/ExtinguishByHand.cs
```diff
@@ -0,0 +1,29 @@
...
+		if (!PlayerManager.LocalPlayerScript.IsInReach(interaction.WorldPositionTarget, false) ||
+			UIManager.CurrentIntent != Intent.Help ||
+			interaction.HandObject != null ||
+			performerPlayerHealth.ConsciousState != ConsciousState.CONSCIOUS ||
+			performerRegisterPlayer.IsLayingDown)
+		{
+			return false;
+		}
+
+		PlayerManager.LocalPlayerScript.playerNetworkActions.CmdRequestPatFire(PlayerManager.LocalPlayerScript.playerName,
```
chairbender: You don't need to pass the player name into this, playerNetworkActions already lives on the same object that the local player script does. You can just access playerScript from within playerNetworkActions to retrieve player name. The way commands work, the local player is only allowed to call commands on their local player object. PNA lives on the local player object.

UnityProject/Assets/Scripts/Player/ExtinguishByHand.cs
```diff
@@ -0,0 +1,29 @@
+using UnityEngine;
+
+/// <summary>
+/// Allows an object to be patted when on fire.
+/// </summary>
+public class ExtinguishByHand : MonoBehaviour, IClientInteractable<PositionalHandApply>
+{
+	public bool Interact(PositionalHandApply interaction)
```
chairbender: This should be doable without needing to add a player network action cmd, by using ICheckedInteractable instead of IClientInteractable. You don't have to refactor it to do this, but I just wanted to make sure you're aware of that, it's usually preferable to adding to PNA. Maybe it doesn't work due to precedence issues (which are planned to be fixed but will probably be a bit before that happens)? Precedence shouldn't be a problem as long as you put this below Meleeable.

UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
```diff
@@ -13,6 +13,16 @@ public partial class PlayerNetworkActions : NetworkBehaviour
 	private static readonly StandardProgressActionConfig CPRProgressConfig =
 		new StandardProgressActionConfig(StandardProgressActionType.CPR);
 
+	/// <summary>
+	/// Fire stacks removed when patting fires (on other people)
+	/// </summary>
+	private readonly float patFireStacksRemoved = 4f;
```
chairbender: this stuff is really specific to fire-patting so it'd be preferable to define these in the fire patting component

chairbender (changes requested):
Nice to have a way to deal with fires better. Good that you added a method for reflexive pronoun as well. Just a few small changes to suggest.

UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
daalta: Thank you for letting me know. Would this work?
```

/// <summary>
/// Sets the serverside fireStacks of a LivingHealthBehaviour to a new value.
/// </summary>
/// <param name="newValue">The new value, clamped to values 0 and above.</param>
[Server]
public void ServerSetFireStacks(float newValue)
{
    this.fireStacks = Math.Max(0, newValue);
}
```

UnityProject/Assets/Scripts/Player/ExtinguishByHand.cs
daalta: I see... that makes sense.

Huggable does things in a similar way, which is why I implemented it like that:
```
PlayerManager.LocalPlayerScript.playerNetworkActions.CmdRequestHug(PlayerManager.LocalPlayerScript.playerName,
    gameObject);
```
Would you like me to change Huggable as well?

UnityProject/Assets/Scripts/Player/ExtinguishByHand.cs
daalta: Ah, well I'll try to make it a ICheckedInteractable.

I was hoping to implement ExtinguishByHand by taking Huggable as an example. But it seems like that wasn't the way to go.

daalta:
Thank you for the review @chairbender ! A lot of these changes are because I re-used parts of the hugging code.

I've tried implementing all your changes and I'll push them in a moment. I haven't had the time yet to test them extensively. The tests that I've tried were interrupted by unrelated bugs, such as players spawning without items and not being able to pick up items.

Even at first glance there seem to be some issues, such as not always being able to extinguish other players. So no need for another review yet - though I'd appreciate if you could reply to my questions above :)

EDIT: Oh, I'd be totally fine with this issue being closed if there are inherent issues in my implementation. I'd rather spend time fixing high-priority bugs than getting this PR to work.

Issue #2647, opened 2020-01-19
Re-logging breaks player sprites for all equipment.


Bod9001:
# Issue Template

### Steps to Reproduce
Please enter the steps to reproduce the bug or behaviour:  

1.  join headless 
2.  put on mask and tank
3.  take it off
4. relog, and put it on again and see the problem

JesterX666:
Doesn't seem to break the oxygen toggle UI anymore but I'm gonna fix this :

![Gas mask](https://user-images.githubusercontent.com/32009105/73612847-845c5300-45bd-11ea-8aa0-90cbea32eb01.gif)

JesterX666:
In fact, the problem is reproductible with ANY clothing item.

1. Log on a headless server
2. Remove your suit
3. Drop it on the ground
4. Log out
5. Log back on
6. You have 1 suit on the ground, 1 suit on your hand and appear fully clothed!

JesterX666:
The clothing was not even removed from clothingitem equipment.  It was merely hidden.  Making it reappear when re-logging.

PR #2843, opened 2020-02-03
Fix Re-logging breaks player sprites for all equipment. #2647

JesterX666:
Clothing items were never removed from equipment slots, they were merely hidden, making them reappearing when you re-logging.  Even if you dropped them on the floor!

### DIFF
--- a/UnityProject/Assets/Scripts/Items/ClothingItem.cs
+++ b/UnityProject/Assets/Scripts/Items/ClothingItem.cs
@@ -93,8 +93,6 @@ public class ClothingItem : MonoBehaviour
 		UpdateReferenceOffset();
 		if (Item == null)
 		{
-			GameObjectReference = null; // Remove the item from equipment
-
 			if (spriteHandler != null) //need to remove
 			{
 				spriteHandler.spriteData = null;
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Items/ClothingItem.cs
```diff
@@ -93,6 +93,8 @@ public virtual void SetReference(GameObject Item)
 		UpdateReferenceOffset();
 		if (Item == null)
 		{
+			GameObjectReference = null; // Remove the item from equipment
```
chairbender: this causes an nre when you remove your clothing after spawning:

```
var unequippedClothing = GameObjectReference.GetComponent<ClothingV2>();
```

Need to ensure this hook is still fired before nulling out GameObjectReference.

chairbender (changes requested):
getting an NRE

UnityProject/Assets/Scripts/Items/ClothingItem.cs
JesterX666: Was the NRE in client or server?

chairbender: Both. I hosted in editor, joined as client, then as client removed my suit and dropped it on the ground.

JesterX666: Looking at it.  Thanks for finding it.

JesterX666: Corrected by moving the null assignation bellow where the reference is needed.

chairbender:
Thank you 👍

PR #2845, opened 2020-02-03
adds Tool belts tp the map, Adds multitool, Fixes a few bedsheets

### LINKED ISSUE #2648: Make Toolbelts Easier to Acquire
## Description
Toolbelts are handy, but you can't get them easily unless you spawn with one.

The tool crate you can buy comes with a lot of tools but no belt. I would suggest that the crate should instead just contain a full toolbelt.
Also suggest mapping some toolbelts into engineering so they can be picked up even if you didn't spawn as engie.

Bod9001:
### what this do?
adds Tool belts to map 
Fixes  #2648
Adds multitools
Fixes a few bedsheets
### why you do this?
yay Bug fixing and features
### what have you changed?
Outpost station Seen
### what you tested?
Tested in editor


UnityProject/Assets/Scripts/Tool/Multitool.cs
```diff
@@ -0,0 +1,48 @@
...
+	public void ServerPerformInteraction(PositionalHandApply interaction)
+	{
+		if (!Validations.IsTarget(gameObject, interaction))
+		{
+			APCPoweredDevice PoweredDevice = interaction.TargetObject.GetComponent<APCPoweredDevice>();
+			if (PoweredDevice != null)
+			{
+				if (APCBuffer != null)
+				{
					PoweredDevice.SetAPC(APCBuffer);
```
chairbender: I see a ^M character here, can you remove it?

UnityProject/Assets/Scripts/Tool/Multitool.cs
```diff
@@ -0,0 +1,48 @@
...
+public class Multitool : MonoBehaviour, ICheckedInteractable<PositionalHandApply>
+{
+
+	public APC APCBuffer;
+
+	public bool WillInteract(PositionalHandApply interaction, NetworkSide side)
+	{
+		// Use default interaction checks
+		if (!DefaultWillInteract.Default(interaction, side)) return false;
+		if (!Validations.IsTarget(gameObject, interaction))
```
chairbender, line 17: interactions don't always have a target object, such as when clicking space / floor. Add
`Validations.hasTarget(interaction)` to ensure it does. This causes NRE when clicking floor otherwise.

UnityProject/Assets/Scripts/Tool/Multitool.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class Multitool : MonoBehaviour, ICheckedInteractable<PositionalHandApply>
```
chairbender, line 7: This needs some interaction messages so that players (and nearby players) can understand what is happening when the tool is being used.

chairbender (changes requested):
Thanks for fixing the belts and adding more to the station 👍 
Also thank you for formatting the code.

I'm not quite sure what multitool does so I didn't test that yet, but I do have some requested fixes.

DooblyNoobly:
multitool does nothing as client. Bod will fix after release. And we need to have a discussion about its use

PR #2862, opened 2020-02-03
Bugsquash and polish №1

### LINKED ISSUE #2831: Grilles became airtight
At some point grilles became airtight, while they shouldn't be.
Can be reproduced by breaking a window – the room won't be decompressed until you cut/destroy the grille.

fomalsd:
### Purpose

- Added support for forced offline mode (if you want to test a github build with a firewall): 
use `--offlinemode` argument
- Tweaked chat bubble fonts
- Minor fixes to lobby
- Fixed #2831


fomalsd:
Merging this so that new PRs can be tested in offline mode

PR #2876, opened 2020-02-04
Reduce nesting in CI build output.

RedlineTriad:
### Purpose
_This should make the build output less confusing and easier to use._<br>

### Notes:
_The space in the output zip was difficult to work with and the output had an unnecessary folder level._

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,6 +8,7 @@ on:
     paths:
     - 'Tools/**'
     - 'UnityProject/**'
+    - '.github/workflows/main.yml'
   pull_request:
     branches:
     - develop
@@ -15,6 +16,7 @@ on:
     paths:
     - 'Tools/**'
     - 'UnityProject/**'
+    - '.github/workflows/main.yml'
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


fomalsd:
there's still a Unitystation.exe folder inside, only zip name has changed
img

RedlineTriad:
@fomalsd there is one less level, before it was 
`StandaloneLinux64 Build.zip/StandaloneLinux64/Unitystation`
Now it is 
`StandaloneLinux64.zip/Unitystation` 
the reason i can't remove that last level of nesting is that on windows the last folder is `Unitystation.exe`. 
Yes, it is a folder called `Unitystation.exe` and i can't easily remove it if they are called different things on different platforms.
Maybe @Jibbajabbafic can help.

Jibbajabbafic:
Ok the directory having the file extension is to do with the Unity Build Action. It looks like at some point when the action was updated they added the file extension to the build file name which is also used for the directory above. 
image

I'll raise an issue with the action itself, and I'll try to fix it if the owner doesn't beat me to it.

Jibbajabbafic:
Ok made a quick fix for this until the action gets updated in #2879

Jibbajabbafic (approved):

PR #2892, opened 2020-02-05
Removed tiles equalize their pressure

### LINKED ISSUE #2890: Mined rocks create wind
## Description
When you go to an asteroid and mine a rock, it creates wind and moves stuff around and disrupts you.

When you mine a rock it shouldn't create any wind.

chairbender:
### Purpose
Fixes #2890 and the more general case of any removed tile ensuring it doesn't cause sudden wind.

Any time an atmos-impassable tile is removed, the gasmix on that tile is changed to be the average of its atmos-passable neighbors.

I haven't touched atmos too much and I have a bit of a sleep debt so I want someone else to take a look at this to make sure it seems alright.

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Layer.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Layer.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections;
+using Atmospherics;
 using UnityEngine;
 using UnityEngine.Tilemaps;
 
@@ -13,6 +14,7 @@ using UnityEngine.Tilemaps;
 		private const bool ROTATE_AT_END = true;
 
 		private SubsystemManager subsystemManager;
+		private MetaDataLayer metaDataLayer;
 
 		public LayerType LayerType;
 		protected Tilemap tilemap;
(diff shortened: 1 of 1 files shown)


chairbender:
In particular, make sure shuttle crashing is not breaking atmos. I wasn't involved in that fix so I'm not sure how to spot the symptoms of a problem with Matrix Check.

Bod9001:
just saying that this isn't thread safe, but stuff like fire isn't thread safe as well so don't worry about it, and technically since the atmospheric thread is limited, it should do all the calculation before the main thread gets to this spot

chairbender:
Ah good point. I can make it thread safe.

Bod9001:
Doesn't matter too much of it is not thread safe

Bod9001:
humm  best thing to do make it not explode if it does become entangled in atmospheric thread and game thread

chesse20:
make it thread safe fire already makes horrible lag its not a good example

krille90:
wouldn't this be adding atmos gases to the system out of nothing?

chairbender:
so i'm gonna close this. After having more of a look through atmos I see problems with how we are using it in a non-thread-safe manner so I'd rather fix those first

PR #2900, opened 2020-02-05
Bugsquash and polish №2

fomalsd:
Fixes multiple exceptions from #issues channel (they are marked with green tick)
Made airlocks buildable from just metal (for now)
Fixes vending tab icons:
GIF
Made status displays visible from everywhere


DooblyNoobly:
nice, thanks foma

Issue #2829, opened 2020-02-02
Can't place item on a rack


Macoron:
## Description
By left clicking on a rack you can't place any item on them. Need to add similar to open closet or table behaviour.

fomalsd:
It used to work, so it's possibly something with script execution order

PR #2906, opened 2020-02-06
Fix for Can't place item on a rack #2829

JesterX666:
Fix for Can't place item on a rack #2829

Commit 01da2ed (Jan 14) moved the Cooldowns.TryStartServer inside RequestInterractMessage.   So, trying to start a second timer causes the problem (it's already started!)

Also, ServerDrop expects to receive a target vector, not an absolute position.

Setting the Z coordinate isn't supported by ServerDrop, it's a Vector2, not Vector3

Also removed unused variables.


DooblyNoobly:
touches some input code, will review this one after release

DooblyNoobly:
Tested thoroughly in mp and works fine. Thanks

PR #2929, opened 2020-02-06
Makes bloods splats only spawn if If there is not an existing blood splats

Bod9001:
### what this do?
Makes bloods splats only spawn if If there is not an existing blood splats
### why you do this?
Too many spawn on a dead body
### what you tested?
Editor

### DIFF
--- a/UnityProject/Assets/Scripts/Factories/EffectsFactory.cs
+++ b/UnityProject/Assets/Scripts/Factories/EffectsFactory.cs
@@ -96,8 +96,12 @@ public static class EffectsFactory
 
 		if (chosenTile != null)
 		{
-			Spawn.ServerPrefab(chosenTile, worldPos,
-				MatrixManager.AtPoint(Vector3Int.RoundToInt(worldPos), true).Objects);
+			var matrix = MatrixManager.AtPoint(Vector3Int.RoundToInt(worldPos), true).Matrix;
+			if (!(matrix.Get<FloorDecal>(worldPos.ToLocalInt(matrix), true).Count() > 0))
+			{
+				Spawn.ServerPrefab(chosenTile, worldPos,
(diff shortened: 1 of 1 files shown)


daalta:
I think ideally blood would spread out over the floor instead of vanishing...

But this is still better than the current situation of blood getting stacked to the ceiling!! 😁

daalta (commented):
Looks good, testing it in the editor now.

Bod9001:
yeah, it's meant to be a temporary solution until they are probably imprinted into tile effects, instead of individual game objects

UnityProject/Assets/Scripts/Factories/EffectsFactory.cs
```diff
@@ -96,8 +96,12 @@ public static void BloodSplat(Vector3 worldPos, BloodSplatSize splatSize, BloodS
 
 		if (chosenTile != null)
 		{
-			Spawn.ServerPrefab(chosenTile, worldPos,
-				MatrixManager.AtPoint(Vector3Int.RoundToInt(worldPos), true).Objects);
+			var matrix = MatrixManager.AtPoint(Vector3Int.RoundToInt(worldPos), true).Matrix;
```
fomalsd: don't call AtPoint twice, just cache `MatrixManager.AtPoint(Vector3Int.RoundToInt(worldPos), true)` here once:
```suggestion
			var matrixInfo = MatrixManager.AtPoint(Vector3Int.RoundToInt(worldPos), true);
```

daalta:
Here's what it looks like now. After implementing foma's solution I think we can merge this, though we'll need to add more blood in the future :)

Unity_2020-02-06_22-23-07

fomalsd:
blood should be moved to tile effects anyway in order to be cheaper

PR #2930, opened 2020-02-06
Added hunger and basic MetabolismSystem

ziiplock:
### Purpose
Added a basic MetabolismSystem that is intended to handle hunger and general intake of substances.

Added an Anti Steak, which decreases the nutrients in the MetabolismSystem for debug purposes and fun.

Added AddWarningMsgFromServer to Chat.cs and gave warnings a color.

Also fixed a bug where ClientCheckAndRequestInteract in InteractionUtil would always return false.

### Notes:
The nutrition level in MetabolismSystem starts at 125. It is capped at 200 and when it falls to 0 the player will be limited to walking speed. Edible now has a field for the nutrients it will add.
MetabolismSystem can currently handle nutrient and toxin intake.


Bod9001:
is this integrated with the grown food from botany?

ziiplock:
@Bod9001 Well it should be if they have the Edible component :)

Bod9001:
humm, Good point they don't to be fair the reason was because they were colliding with the turning to seeds option, I'll have to make the machine for it

Bod9001:
will review tomorrow, will need some input of some of the over devs as well

fomalsd (commented):
Wow, impressive!
I'm not a big fan of adding a lot of syncvars though, because they are a public knowledge. Other players shouldn't really know your hunger state, for example. Same goes for speed – players should only receive a message with result of movement with certain speed, anything more than that is a meta knowledge I'd like to avoid if possible.

ziiplock:
@fomalsd Ah okay thanks for the feedback. I will try to revise some things then.

UnityProject/Assets/Scripts/Player/PlayerMove.cs
```diff
@@ -118,9 +118,10 @@ public bool IsSwappable
 
 	[HideInInspector] public PlayerNetworkActions pna;
 
-	[FormerlySerializedAs("speed")] public float RunSpeed = 6;
-	public float WalkSpeed = 3;
-	public float CrawlSpeed = 0.8f;
+	[FormerlySerializedAs("speed")] public float InitialRunSpeed = 6;
+	[HideInInspector][SyncVar] public float RunSpeed = 6;
+	[SyncVar] public float WalkSpeed = 3;
+	[SyncVar] public float CrawlSpeed = 0.8f;
```
Bod9001: It's better to make these into hooks so when you update the value it call the Function on the client with the new value, Without the networking having to check constantly If it has changed, 
https://unitystation.github.io/unitystation/development/SyncVar-Best-Practices-for-Easy-Networking/

fomalsd:
Please don't hesitate to ask the questions if you have any trouble, really eager to have this in soon

UnityProject/Assets/Scripts/Player/PlayerMove.cs
fomalsd: as I mentioned above, these shouldn't really be syncvars in the first place

Bod9001:
yeah,  if you got no time to work on it we can merge it and then fix it up later

ziiplock:
Oh ya sorry guys I was caught up in some RL stuff lately, should have communicated better. I think I've fixed the issues now. I will push my commit shortly.

fomalsd (approved):
Awesome! Very impressed to see netmessages done right at first attempt. Keep on rocking!

Issue #2884, opened 2020-02-04
RconManager requires hub credentials.


RedlineTriad:
## Description
RconManager only works if hub credentials are supplied.

### Detailed Bug Report
It loads the `config.json` and then calls `SendServerStatus`, however if `HubUser` or `HubPass` is empty it will set `config` to `null`.
https://github.com/unitystation/unitystation/blob/f6b6803b3963a2a3fe44bb9aac442c3b700cf030/UnityProject/Assets/Scripts/Database/ServerData.ServerStatus.cs#L74-L79
This means that when `RconManager` tries to get `ServerConfig` it just gets `null` so it destroys itself.
https://github.com/unitystation/unitystation/blob/f6b6803b3963a2a3fe44bb9aac442c3b700cf030/UnityProject/Assets/Scripts/RCON/RconManager.cs#L60-L64

Issue #2520, opened 2020-01-02
Docker image for a dedicated server host


dargonforce:
# Description

A Docker image for hosting the server would be nice.  /tg/station provides one that sets up dream daemon.  Assuming the application can be run in a "headless" mode that would be pretty easy.

jnms-me:
I'm new here but quite experienced in web backend. The /tg/station Dockerfile builds the server from source, then sets up the server environment.
I doubt we can build it in docker with Unity requiring a licence and going through the editor to press build.
If we had release builds being published or maybe even CI (don't know if we have this) this would be much easier.
Right now the best i can do is building the latest release/commit and creating instructions for doing so yourself (build for linux to build/ss13, run docker build).

RedlineTriad:
@Reavershark we previously had Travix CI, and building Unity in docker isn't a problem.
If we add Docker images it would most likely be using Github Actions that build the project and then creates a Docker image with the build artifacts.
I don't think it is worth worrying about it yet though.

Jibbajabbafic:
We used to have Travis setup but it became too much of a pain to maintain, and it used a Docker image of Unity (using https://gitlab.com/gableroux/unity3d) and an encrypted licence file (Unity_v2018.x.ulf.enc) to run.

Now that we have GitHub Actions available it could be worthwhile switching over to that as there's a collection of Unity related ones which can handle the licence file, building and testing: https://github.com/webbertakken/unity-actions

Old Travis setup if you want to check it out: https://travis-ci.com/Jibbajabbafic/unitystation

jnms-me:
Made a working github action. Last run has a linux artifact [here](https://github.com/Reavershark/unitystation/commit/7ac4f17d0a2f1f7f517e33d762df34929e372001/checks?check_suite_id=383762469). The entire process takes 22 mins.

The license file is stored in a Github secret.
I tried to use https://github.com/webbertakken/unity-actions but the build process is kinda(very) broken so i rewrote that using the Travis scripts as an example. The license actions are great though and it's using that.

Still needs some work like removing unity-return-license as it's only for pro, adding test jobs, building for multiple platforms and figuring out how to make a headless build.

The issue should probably be renamed if i'm allowed to continue with this.

RedlineTriad:
It may be possible to speed it up using the cache action: https://github.com/marketplace/actions/cache.
Also, it would be nice if your build actions took the version of Unity to use as a parameter.
Is using docker necessary btw? Why not do it directly in the runner?
[...]

PR #2947, opened 2020-02-07
Add automated docker server builds.

RedlineTriad:
### Purpose
Adds:  #2520
This automates building docker server image and pushing it to dockerhub.

### Notes:
@DooblyNoobly  already set up creds for upload in main repository.
Buildnumber isn't currently set automatically.
Rcon doesn't work because of #2884 unless hub creds are set (only @DooblyNoobly has valid ones)


Jibbajabbafic:
Got pushing to docker working, but haven't tested the image itself. Will merge for now since I trust you tried it out already.

PR #2950, opened 2020-02-07
Fixed some HUD elements preventing 'Enter' key from opening the chat.

### LINKED ISSUE #2897: Hitting enter causes the last pressed button on the UI to be pressed again
# Issue Template

### Detailed Bug Report
Anytime you hit a UI button when you hit the enter key afterwards it will press that same button again once.

Here's a ~200KB gif on **_GER_** showing it off. First I press enter a couple times, and hit the Toggle Light button a couple times just to show those work as normal. Then I hit the Toggle Light button once, hit enter, and repeat that a few times. Notice that it toggles off again as soon as I hit enter.
Unitystation -- Enter key presses UI button last pressed #8
_Reproduction_

I've confirmed this same bug happens with the turning buttons on the shuttle console too.
Unitystation -- Shuttle Turning Keys
_Shuttle turning keys_

### Steps to Reproduce

For sake of ease-of-reproducing I went with this specifically. As said earlier it seems to work on atleast the shuttle U.I. turning buttons too. Ping has no effect on this, it seems entirely client-side.
1.  Die, and ghost.
2.  Hit the mining lamp button 'Toggle Light'.
3.  Hit enter, see that the button is toggled again.

daalta:
### Purpose
* Fixes https://github.com/unitystation/unitystation/issues/2897
* The issue above may be incomplete - please let me know if more HUD elements block input.

### Notes:
* Some HUD elements had their Navigation set to "Automatic". This allowed navigating that interface with the keyboard and pressing that button with the enter key. Opening the chat was no longer possible immediately after pressing that button.
* This PR changes the navigation of those buttons to "None". A possible side effect is that the buttons can no longer be selected with the arrow keys. However I don't think this behaviour was intended in the first place.
* There are likely other HUD elements which block the chat. This should be easy to address if this PR's method of fixing the issue is valid. 

### Self checcks:

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/SceneConstruction/NestedManagers/UIManager/GhostHud.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/SceneConstruction/NestedManagers/UIManager/GhostHud.prefab
@@ -89,7 +89,7 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   m_Navigation:
-    m_Mode: 3
+    m_Mode: 0
     m_SelectOnUp: {fileID: 0}
     m_SelectOnDown: {fileID: 0}
     m_SelectOnLeft: {fileID: 0}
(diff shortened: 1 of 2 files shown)


Bod9001:
ok, Seems good

PR #2952, opened 2020-02-07
Atmos Fixes

### LINKED ISSUE #2841: Freezing cold after round restart
## Description

After a round restart, atmos fills up but it's really cold, like -243C. Somehow temperature is not being restored. We don't currently simulate cold damage so it doesn't cause any serious gameplay issues, but it should still be fixed. It is probably related to how atmos initializes the initial state of the game - it may not be correctly "resetting" itself after a round restart so the normal init logic doesn't complete properly.

I've reproduced this locally so I don't think it is caused by server breaking.

krille90:
### Purpose
Fixes some problem with Atmos I noticed.
This might fix #2841, but I couldn't reproduce that one.

* Fixed bug where Gas.Space would be changed in runtime and get moles, resulting in moles of all updates nodes steadily increasing
* Fixed scrubber-airvent loop due to floating point inaccuracies.


Bod9001:
ok, Looks good no errors and This atmospheric stuff is a bit hard to test so it seem to work okay so let's merge

PR #2953, opened 2020-02-07
Atmos Analyser Values in Percentage

### LINKED ISSUE #2770: Atmos Values dont show in percentage
### Detailed Bug Report
When using the at Atmos Analyser, the values aren't shown in percentage, but as a value between 0 and 1. Easy fix is to multiply these values by 100.

image

### Steps to Reproduce

1.  Take Atmos Analyser
2.  Click Atmos Analyser
3.  Wonder, why the values don't sum to 100, but 1.

krille90:
### Purpose
Fixed #2770 

* changed atmos analyser to show actual percentages
* all values are capped to 3 decimal points


Bod9001:
hopefully this should work, but I trust you

Issue #2942, opened 2020-02-07
Try Mirror's network proximity


fomalsd:
We should try simple distance-based network culling for higher-level networking stuff like SyncVars, that might help with #2917 and #2941

Misterixmax:
Also suggest updating Mirror version.

fomalsd:
Yeah, see #2462

ThatDan123:
Closing in favour of #6084

Issue #2462, opened 2019-12-14
Upgrade Mirror to latest asset store release


chairbender:
# Description
A new asset store release of mirror just came out, we should upgrade as these releases are supposed to be stable.

I will probably take this on after my next PR since I've already done a fair bit of work with previous mirror upgrade attempts.

Bod9001:
yeah, Mirror uses a lot of performance Checking network objects to see if they've changed, So I want to see what it's like in the new version, To see if it's been sorted,  if it's still an issue then Solutions can Start being developed and implemented

chairbender:
doin it!

chairbender:
notes https://github.com/unitystation/unitystation/wiki/Mirror-Upgrade-Customization-Notes

PR #2960, opened 2020-02-08
Fixed atmos not resetting state on round restart

Scorr:
Atmos thread takes roughly 1ms (on my machine) and would add another 1ms for each round restart. This results in roughly 4ms after 3 restarts as seen in bottom right below:

image

Fixed by clearing updateList on round restart.

Also added profiling for the atmos thread.

Does not conflict with #2952


Emma3DSS:
wow this is going to be a really great pull to get in it might even fix some other optimization issues such as atmos and fires getting very laggy after a server has been up for a long period of time

Misterixmax:
@Emma3DSS no it won't. #2942 and #2462 should help.

Bod9001:
oh nice, Will quickly test this

Bod9001:
Tested works well I like the new profiling things

PR #2976, opened 2020-02-09
Diagnosis to find what causes our server crashing

JesterX666:
Diagnosis to find what causes our server crashing
Yes, I did modify Mirror.  But this is a temporary measure with the time being.


chairbender:
You should know, I am working on mirror upgrade right now, there's a lot that's changed as the latest asset store release is 4 days ago. You may have found a bug in mirror that's now fixed. That said as this is only diagnostic it should be okay

JesterX666:
Nice to know for the mirror upgrade.  I'm happy about it and put a lot of hope into that :)

Issue #1525, opened 2019-02-08
Improve upon the current TTS system


JaytleBee:
# Description
The way TTS currently works is that every client connected to a server with TTS enabled sends a request to the server at unitystation.org to request an audio file of the chat line. This is not only a system with a single point of failure, but it's also not scalable. The more developers and players we get, the worse the situation becomes. This is obviously not ideal.

I talked about this with chairbender in the discord and we came up with the idea of using MaryTTS (which we seem to be using already) locally somehow. Of course, this opens up the can of worms called "Java", but I think that kind of solution is still better than what we have right now. It seems @DooblyNoobly and @fomalsd set this up so please comment here if I'm missing something.

## Category
TODO

fomalsd:
we tried doing something with IKVM.net, but it didn't work out: it's discontinued and javax.sound.* is not supported at all:
https://sourceforge.net/p/ikvm/wiki/Limits/

chairbender:
Yeah, we were thinking that we shouldn't need IKVM. Just run the server as a separate java process using the actual JRE. Will need some logic to handle initial setup / startup, and checking if Java is installed or bundling the JRE with the game, etc...but seems like it should be doable and cross platform.

fomalsd:
Things to keep in mind:
"TTS is not really used much atm and bundling java+marytts would add like +100-150mb while compressed game is 60mb, plus it won't work for android+ios obviously"
Of course I'd **much** rather prefer a local solution, but please consider priorities and convenience, too

chairbender:
seems there isn't a free cross platform TTS, you'd need to use an approach that worked for each supported platform. This guy outlines how it could be done:
https://stackoverflow.com/a/42576071/765419

It seems like a lot of work, but the bulk of the work is really just hooking up the TTS for each platform - most of the work of actually implementing TTS is probably already done for each platform, just have to hunt it down.

And then once it's implemented we get to have the honor of being the first to have an open source, cross platform client-side TTS in Unity!

fomalsd:
meh
I've seen and analyzed a paid asset that does exactly as you described but the dull part is all the voices would be different for each system, that sucks
MaryTTS provides several tweakable voices that sound good, and I really want them to be consistent regardless of OS

chairbender:
I think those are nice to have features, but it sounds like you didn't think marytts is an option because it doesn't work on Android and ios and would increase the game size. So I'm trying to think of an alternative.

fomalsd:
[...]

PR #2980, opened 2020-02-09
Replace TTS with Client-Side Espeak

Elijahrane:
PR Purpose: Replaces the old server-side MaryTTS with the compact Espeak system. This is just the initial implementation and soon there will be new voices and modifications of them based on player-controlled in-game factors.

Addresses Issue #1525

What was tested and works:
Linux Client
Windows Client
2 Player Multiplayer with Late Joiners
Input Sanitation

What was not tested:
OSX. 

Additional work:
I couldn't find anyone to compile the OSX espeak system. It should work though, as I've put the code in place in the Unity side already for OSX. Just add the OSX is a folder mirroring the Linux and Windows ones in Assets/StreamingAssets


UnityProject/Assets/StreamingAssets/Espeak/Linux/include/espeak-ng/encoding.h
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, see: <http://www.gnu.org/licenses/>.
+ */
+#ifndef ESPEAK_NG_ENCODING_H
```
fomalsd: is it necessary to include all these C files?

Elijahrane: Probably not. I haven't pruned the install folder too much

daalta:
Looks cool, a client-side TTS would really be something.

It looks like you've included a lot of TTS languages, too! Is that intentional? Or is it just easier to include them all?

Elijahrane:
They're a few MB and useful for a few things (accent filters, localization). If you want to remove everything but the English I wouldn't be opposed

Bod9001:
just  saying it doesn't work in editor
image

Got that in build
image


All windows

Elijahrane:
Packaging issue. If you find the library it will work. I didn't make the windows build of espeak nor do I know if we're allowed to distribute those libs

Elijahrane:
https://github.com/espeak-ng/espeak-ng/blob/master/docs/building.md

It seems that that redistributable is a dependency for Windows. I guess it's a matter of including it.

https://www.microsoft.com/en-us/download/details.aspx?id=48145

Elijahrane:
What's left:
1.Have someone compile and hook up the Mac Folder.
2.See if there is any advantage in using speak over espeak, or figure out if we could use the shared library folder.
3. Instead of the program running via it's own audio output system, have it give stdout into the old MarryTTS synthesize subroutine for more consistent audio output. Depending on how the mac compiler deals with PCAudioLib this may or may not be required for it to work on mac. Also we can further modulate it that way which we can't do otherwise.

As it is right now it's fully functional though. Macs just go through the old MaryTTS temporarily.

Bod9001:
image
I'm getting this in editor

Elijahrane:
When it checks Application_platform it looks for the player client, not editor.

Elijahrane:
It should work now. At least, every case should be given a file name.

To debug, look at any Win32 exception "File not found". Compare the place it looked and change the filename under startEspeak to correct it.

Bod9001:
humm One last note can you push the metafiles that it changes since unity uses that to track the folders and it regenerates it 
every time I open it

Bod9001:
image
More errors?

Elijahrane:
I thought the Windows binary might have relative pathing because it worked with no arguments when I tested it in Wine. Strange it'd look in a unix directory by default. It needs the --path argument passed like Linux and Mac. I'll fix it after work.

Elijahrane:
The issue is that `espeak.StartInfo.Arguments = "--path=" + Application.dataPath + relative_Path + message;` generates arguments Linux parses correctly, but Windows parses as one giant string to attempt to read and speak. If anyone knows why ProcessStartInfo.Arguments behaves differently on Linux and Windows and how to fix it, I'd appreciate your help.

Bod9001:
humm Maybe try passing \\" and experiment with that a bit?

Elijahrane:
Can you uncomment the Debug.Log I commented out, update scripts, then run and copy + paste the "This was passed to espeak: " in the log and see what it passed?

Bod9001:
ok will do

Bod9001:
~~humm Do this in what script?~~ Forgot to save it xD

Bod9001:
This was passed to espeak: --path=Q:/Fast programmes/Unity station_unity station/unitystation/UnityProject/Assets/StreamingAssets/Espeak/Windows/  "yo this work"

You might need quotation marks around path Since there are spaces in it

hecksadecimal:
Oh it's just one directory away from being correct, I think.
Edit: Disregard. It's probably just that the path needs to be wrapped in quotes.

Elijahrane:
Try now on Windows. Also for Mac testers notice that the legacy TTS Is only flagged for the mac player and not editor.

hecksadecimal:
So two things. 
One, it's still giving that same error. 
Two, when it tries to launch espeak, it opens up a windows console that very briefly gets in the way of Unitystation.

Elijahrane:
I managed to get the threading working properly and implemented Windows parsing and unintended windows popping up with help from hecksadecimal. In addition to fixing the Windows hang (I hope, at least) threading should also lessen performance impact.

Bod9001:
ok cool, Seems to be working I can hear it, only issue now is to test it to make sure you can't put characters into run stuff on people's computers

Elijahrane:
Screenshot from 2020-02-13 07-56-51
😎

fomalsd:
Tried a mac github build and TTS used is still the serverside MaryTTS. Maybe it didn't build properly for some reason...

hecksadecimal:
I think the mac stuff might've been commented out until a mac build of the TTS was made available.

fomalsd:
I really think TTS should be consistent on all platforms. Have you tried espeak-ng?

Elijahrane:
I asked around for someone to compile for mac and no one took up the offer. I could cross compile but there'd be no one to test.

Elijahrane:
To be clear, the Linux and Windows binaries were manually compiled.

fomalsd:
While I really appreciate the effort, espeak simply doesn't sound good enough (you can listen to examples in discord), and slapping reverb on it to try to make it sound less harsh doesn't sound like a proper solution. I'm going to close this one.

daalta:
This is a good place to test out what espeak sounds like.

https://eeejay.github.io/espeak/emscripten/espeak.html

I don't use TTS, so it's hard for me to judge how distracting or useful espeak is compared to our current system. MaryTTS definitely sounds more natural but eSpeak runs client side. I feel like neither solution is perfect. Should we really rule out eSpeak?

fomalsd:
Rane provided examples of something else what sounded quite alright, I think it was called Festival

daalta:
Yes, I've stumbled upon festival too. It sounds more natural than espeak.

It looked like it's not being developed as actively though. At least from a first glance.

PR #2988, opened 2020-02-09
Mirror Upgrade + Mirror Event Problem Fix

### LINKED ISSUE #2146: TODO: Mirror Migration B: 160
# Description
- Remove UNET, Replace with Mirror

## Criteria:

- Follow guides: https://mirror-networking.com/xmldocs/articles/General/Deprecations.html and https://mirror-networking.com/xmldocs/articles/General/Migration.html
- Remove UNET completely and Implement Mirror Lib
- Review all networking methods, properties and base methods to make sure that they work with mirror
- Test thoroughly in all network environments with multiple clients

Contributor will receive a bounty of 160usd on a successful merge of a solution that satisfies the criteria above.

### LINKED ISSUE #2462: Upgrade Mirror to latest asset store release

### LINKED ISSUE #2986: Mirror Event Problem - Audit All Mirror Events - Mirror Events Called Before Unity Start / Awake / OnEnable
## Description
Per this: https://forum.unity.com/threads/onstartclient-called-before-scene-is-fully-loaded.427952/#post-5453898

Mirror events - SyncVar hooks, OnStartClient, OnStartServer, have a chance of being called BEFORE Start / Awake / OnEnable. It's also depending on all sorts of things, so it won't happen the same way on all systems or between editor / standalone build

We need to go through EVERY usage of these events and ensure that any things they reference which are initialized only in Start / Awake are actually initialized. For example, can create a EnsureInit() method that simply checks if init logic has not yet been done (for example check if one of the component fields are null) and do the init logic if not, then call this at the start of every Mirror event and in Awake.

The SyncVar best practices docs also need to be updated to ensure they call this out very clearly. Also they need to be modified due to changes in Syncvar hook signature.

chairbender:
### Purpose
Fixes #2462 
Fixes #2986 - I went through and added proper init-insurance to every user of Mirror events except for a few special cases where it's obvious it's not needed because syncvar doesn't touch anything and components that are very complex and known to work fine (movement sort of things) whose precious init logic I didn't want to disturb.

Might fix some other things? Won't claim it does though.

Need to update SyncVar best practices document, will do it tmrw.


DooblyNoobly:
Character reclaim on rejoin doesn't work anymore. I think the changes were made to mirror itself in the last version. I'll see if I can hunt it down

DooblyNoobly:
merging to a temp branch so I can apply some quick fixes

PR #2987, opened 2020-02-09
Fixes the health doll on the right of the screen

FijiWatr:
### Purpose

The PR fixes #2966, the health doll on the right side of the screen is no longer mirrored to the health doll at the bottom.
### Notes:
All I did was flip the sprite, I don't know if that fix might create problems later for some reason. Check out PR #2990  that might fix some underlying the issue with the assignment of limbs.

PR #2990, opened 2020-02-09
flipped health doll damage

### LINKED ISSUE #2966: The health doll on the right hand side of the screen is mirrored to the health doll that the player can select to heal.
Reproduce:
- Get shot in left leg.
- Select left leg.
- Try to heal left leg.
- "Your left leg is not injured."
- Do the same for the opposite leg, heals up just fine and dandy.

Haven't been able to test this with arms yet, will comment here if I can find anything.

myrmecoleonn:
fixes #2966 ; both the Player_v3 and PlayerHealth_UI_Hud prefabs had their right and left arms/legs swapped. this has hopefully been rectified

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Player/Resources/Player_V3(uNet).prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Player/Resources/Player_V3(uNet).prefab
@@ -58,7 +58,7 @@ MonoBehaviour:
   GrayDamageMonitorIcon: {fileID: 21300000, guid: ef9c15b1dae2be64ab9cc67afa230913,
     type: 3}
   MaxDamage: 200
-  Type: 3
+  Type: 2
   isBleeding: 0
   livingHealthBehaviour: {fileID: 0}
   Severity: 0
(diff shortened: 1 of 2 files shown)


daalta:
Hi @myrmecoleonn ! Thanks for taking a look at this, I would never have noticed, haha.

I think it's a bit weird that a body part such as `HumanBodyPartLeftLeg` now has the type `RightLeg`.

(EDIT: Image redacted - it was outdated!)

The solution https://github.com/unitystation/unitystation/pull/2987 currently also has a problem, but I think it avoids mislabeling the limbs. What a coincidence that both you @FijiWatr attempted to fix the same problem! 😅

myrmecoleonn:
woah, what? that was the exact thing I was trying to fix in the first place - where did you get that `HumanBodyPartLeftLeg` from? I couldn't find a prefab of the any of the humanbodyparts with sprites(or a script) assigned!

daalta:
Ooooh that makes absolute sense! I guess my player was outdated, so you fixed the right thing. Give me a moment and I'll merge it :)

daalta:
Perfect, now it's all as it should be! Nice job 👍

PR #2992, opened 2020-02-09
Added simple chat filter for stopping spam and walls of text

### LINKED ISSUE #2957: Players can spam the chat.  Create a protection against that.
Players can spam the chat with the same shit again and again.

We can put a protection against that.

daalta:
*Example of triggering the filter by copy-pasting text.*

### Purpose
* All player messages from the chat input field now go through `ChatFilter.cs`, which limits the frequency and the length of player messages client-side.
* Players see a message if their message got caught by the filter.
* Fixes https://github.com/unitystation/unitystation/issues/2957
* Limits the maximum length of player input to 250 characters (not directly related to the filter).

### How it works
* Players are limited when chatting by:
  * The maximum amount of text they can send every minute.
    * This value is currently at 500 characters per minute, which equals to 100 WPM. 100 WPM is considered very high should only rarely be achieved by non-spammers.
    * If the message would go over the 500 CPM limit it gets cut short.
  * The maximum number of messages they can send every 5 seconds.
    * This value is currently at 4 and is hard to trigger by a human without spamming very short messages
    * If the message triggers the filter it is not sent at all.

### Notes
The filter could be used to automatically notify admins about players who spam repeatedly.

Or, if the chat filter need to be replaced, it can easily be removed in the future by changing two lines in `ChatUI.cs`. Just in case we decide to implement a server-side check or another spam prevention method.
```
	private void PlayerSendChat(string sendMessage)
	{
		// Selected channels already masks all unavailable channels in it's get method
		chatFilter.Send(sendMessage, SelectedChannels);
		// The filter can be skipped / replaced by calling the following method instead:
		// PostToChatMessage.Send(sendMessage, SelectedChannels);
		InputFieldChat.text = "";
	}
```

### Known issues
* The messages informing the player that they have triggered spam prevention shows up late. Indead of being shown immediately it is displayed along with any *following* message. I suspect that I did not use `Chat.AddExamineMsgToClient()`. This issue is non-intrusive and does not impede the filter's functionality, though I would appreciate help in identifying the issue.


JesterX666:
Review in progress

JesterX666:
Code is looking great!
Can you help me find a way to flood the text chat very quickly to test it at maximum speed?
Nevermind.  Found a way.  Works great with "paste" large block of text too.

Bod9001:
Is it something for OOC?

JesterX666:
@Bod9001 It would work for all channels, including OOC.
It's kinda "goofy" to see a "you must catch your breath" while speaking OOC... but it gets the job done.

Trying it with headless + 2 clients right now.  It's my last test before merging it.

daalta:
This is for in-game and ooc chat. Players are limited equally on all channels. The values are currently too high to be triggered by 'human' players (unless they type insanely fast), thus only triggering on spam or copy pasting. 

In the future we could decrease the limits to prevent hand-typed spam, though I expect that to be less of an issue.

OOC (and ghosts) have do have a single difference though: The warning for getting caught by the filter is slightly different .

JesterX666:
Indeed... OOC and Ghost is different!  Well done!

daalta:
By the way - OOC should *not* display the "catch your breath" message, there's a different message for OOC and ghosts. If that's a bug I can check it out later

EDIT: Ah, nevermind, you noticed. Thanks for the review :)

JesterX666:
It does work as intended with a different message.
"Your messages have been too long..."
Excellent.

PR #2996, opened 2020-02-09
Expanded construction for Glass Windows and Grills

Necromunger:
### Purpose

Adds the ability to place grills using rods on the base ground layer.
Adds the ability to place windows on grills.
Adds capacity for the script PlaceableTile to define the amount of resources required to place a tile.
Add's 1 rods to be spawned when using wirecutters on grills.
Removed debris/effect when destroying grill.
Added some capitalisation to proper nouns in the current metal construction items.
Changed 'canister' to 'Oxygen Canister' in the construction listing.

### Notes:

You can now build and repair previously destroyed grills and windows.

The main code work in the pull request is expanding PlaceableTile and updating damage code to allow for repairing tiles.

Reason for destroyed debris removal is there was currently no lifecycle to the effect. No way to clean the destroyed debris.

It has been replaced by 1 spawned rods.

Balance notes:
New PlaceableTile behaviour added a cost of 2 rods to place a grill.
On destroying a grill you now get returned 1 rods.

- [x] Any new / changed components follow the [Component Development Checklist]
- [x] Any new objects / items follow the [Creating Items and Objects Guide]


Necromunger:
Forgot to mention.
Added some capitalisation to proper nouns in the current metal construction items.
Changed 'canister' to 'Oxygen Canister'.

fomalsd (changes requested):
> Removed debris/effect when destroying grill.

not a fan of this, debris should stay if grille is destroyed and not cut cleanly with wirecutters

Necromunger:
I agree, i tried to replace the effect with dropping a rod.

This was an attempt for consistency to try and do the same thing smashed windows are currently doing dropping glass.

Presenting destroyed grille tiles a tile layer does not look very good at the moment because its just a small amount of an image at the bottom of the tile that remains forever even after a bomb.

Would this change request be satisfied with some sort of additional debris gameobject in the same capacity as glass?

Or any other suggestion for the effect that has an actual life cycle? current one can't be removed and always shows even for explosions.

Necromunger:
I have concluded a new process that could resolve all my issues with debris and @fomalsd change request.

1. When grilles are destroyed by damage, it is destroyed and 1 rods remain with no debris effect, this resolves issue of no cleanup for grilles even with great damage.

2. When grills are destroyed using wirecutters the debris effect remains and does not give any rods.
    a. Health is given to debris so that it can be destroyed with damage to produce 1 rods.

    b. New interaction is created cutting torch can remove the debris and produce 1 rods.

    This cleanly resolves the problem with debris not being able to removed.

    It also solves the problem of having no evidence that wirecutters were used, debris effect can be removed with additional work with torch or damage.


Can i get a confirmation that this change would satisfy change request before implementation @fomalsd?

fomalsd:
> Presenting destroyed grille tiles a tile layer does not look very good at the moment because its just a small amount of an image at the bottom of the tile that remains forever even after a bomb.

It shouldn't be like that in the first place, that destroyed grille effect can be removed rather easily.
Here's how grill debris effect is created:
https://github.com/unitystation/unitystation/blob/062afe96fe56f2bdcd6c82ec119ed59af0b71ee1/UnityProject/Assets/Scripts/Tilemaps/Behaviours/TilemapDamage.cs#L516
And then it can (most likely) be removed by using
`tileChangeManager.RemoveTile(cellPos, LayerType.Effects);`

How does this interaction work in byond? I think we should make whole construction system match /tg/. I don't really remember how it worked in byond, but it seems quite illogical that bashing grill would cut it cleanly and cutting it with a specialized tool would leave remains.

Necromunger:
https://tgstation13.org/wiki/Guide_to_construction#Grille

```
Use wirecutters to neatly cut the grille, or
...Use a screwdriver to unfasten and drag it away, or
...Smash it down.
```

Current pull request behaviour is apparently correct according to /tg/

As for what should remain after an explosion or using wire cutters i will have to load into tg and see.

Necromunger:
Ready for change review.

fomalsd (approved):
Thanks, working fine. The only thing that might need tweaking is that grille is now 2x less durable, way too easy to break. Is it like that in tg?

Necromunger:
The server I tested on needed 10 hits with crowbar to get to passable, and we need a bit over 20 crowbar hits at the moment.
Can do another balance pass as needed soon.

fomalsd:
Ok, can play with balance later

PR #3013, opened 2020-02-10
Fixed having no hub config stopping rcon config from running.

### LINKED ISSUE #2884: RconManager requires hub credentials.

Charlese2:
# Pull Request Template

### Purpose
Fixes #2884 
If you didn't have hub credentials set in the config but you had rcon set up you it would just skip the code to look at the rcon config.

### Notes:
Without HubUser set in the config, the hub communication loop code will not run.

### DIFF
--- a/UnityProject/Assets/Scripts/Database/ServerData.cs
+++ b/UnityProject/Assets/Scripts/Database/ServerData.cs
@@ -96,8 +96,10 @@ namespace DatabaseAPI
 		void Update()
 		{
 			if (config != null)
-			{
-				MonitorServerStatus();
+				if (!string.IsNullOrEmpty(config.HubUser))
+				{
+					MonitorServerStatus();
+				}
 			}
 		}
 
--- a/UnityProject/Assets/Scripts/Database/ServerData.cs
+++ b/UnityProject/Assets/Scripts/Database/ServerData.cs
@@ -96,6 +96,7 @@ namespace DatabaseAPI
 		void Update()
 		{
 			if (config != null)
+			{
 				if (!string.IsNullOrEmpty(config.HubUser))
 				{
 					MonitorServerStatus();
--- a/UnityProject/Assets/Scripts/Database/ServerData.cs
+++ b/UnityProject/Assets/Scripts/Database/ServerData.cs
@@ -97,7 +97,7 @@ namespace DatabaseAPI
 		{
 			if (config != null)
 			{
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Database/ServerData.cs
```diff
@@ -96,8 +96,10 @@ void OnDisable()
 		void Update()
 		{
 			if (config != null)
-			{
-				MonitorServerStatus();
+				if (!string.IsNullOrEmpty(config.HubUser))
```
RedlineTriad: In order to replicate previous behavior, it should probably check the same things.
```suggestion
				if (!string.IsNullOrEmpty(config.HubUser) && !string.IsNullOrEmpty(config.HubPass))
```

Charlese2: I think I messed up the formatting.

RedlineTriad: It looks like it does not have enough curly braces.

DooblyNoobly:
@RedlineTriad you are still reviewing this?

RedlineTriad:
@DooblyNoobly no.

Charlese2:
I'm considering changing to this. If only one of the config variables is missing it would still be caught by the check with logging. (In `ServerData.ServerStatus.cs`) The logging would repeat every 10 seconds instead of every frame.
```C#		
void Update()
{
	//If there is a config loaded, check whether HubUser and/or HubPass are set.
	if (!string.IsNullOrEmpty(config?.HubUser) || !string.IsNullOrEmpty(config?.HubPass))
	{
		MonitorServerStatus();
	}
}
```

PR #3016, opened 2020-02-10
Mirror Upgrade Followup Fixes + Disable Stranded Zoom Out

chairbender:
Following up from #2988 -

- Doors init correctly on headless server
- APCs as well
- Stranded zoom out doesn't happen any more, trying to prevent perf issues / crashing

FYI the issues with Doors / APCs were due to an apparently new behavior of Mirror regarding syncvars and hooks - on clientside, the value of the syncvar field is already set BEFORE the hook fires, and you can access the before / after value using the 2 params. Before, on clientside the syncvar field was not set unless you assigned it in the hook (or maybe after the end of the hook).


DooblyNoobly:
working pretty good now. I don't think I can test absolutely everything but everything I have tested has worked fine

PR #3024, opened 2020-02-10
Added Unity Test runner for editor tests

Macoron:
### Purpose
This PR add unity test runner to current CI process. It runs editor test before PR and after develop merge. The test results are added to artifacts after the run.
With more test coverage it will become great tool for early PR validation. In case if unit test fails, test runner returns error, but doesn't block builds or merge.

Right now it doesn't support playmode test and need more research how to implement them.

### Notes:
Fixed some old chemistry unit tests to run smoothly


Jibbajabbafic:
Good stuff.
Is there a nice way to parse/view the test results XML file?
Does it correctly fail the build when the tests have errors?

Jibbajabbafic:
So looks like the tests fail correctly, but it's not very easy to view the results. It's a good start though.

Scorr:
I like this. If it actually fails the check on github when a test fails that'd be very useful for improving reliability.
Why does it take 16 minutes to perform the test, though? That's longer than creating a build. Is it creating a build and then running tests, and if so, is there a way to avoid that?

Macoron:
@Jibbajabbafic yeah, nice html preview would be great. I tried to setup several report generators, but they all failed for some reasons. I'll make separate issue after this PR and write all my research there.

@Scorr Unity starts editor in docker, load/generate cache and sequentially run tests. The actual test execution is matter of seconds, but there is no way to execute without running editor environment. So yeah. overhead is bad.

Macoron:
Damn, I just realized that renaming Build Project to something else will ruin our glorious building badge. Moved testing to separate file and added separate badge to it.
Tests Build Project

Jibbajabbafic:
Awesome ok!

PR #3051, opened 2020-02-11
Adds Lockable Lockers to the Game

Robustin:
This PR adds the ability to lock/unlock lockers if you have the appropriate access level.

Notes:

- I am not a coder by training or profession and this is my first time working with C# or Unity so I will appreciate any feedback because there's a lot I don't understand. 

- It behaves largely as it does for /tg/code with one exception: Lockers will automatically lock if closed by someone with sufficient access. 

- Since 'Resist' or its behavior are not implemented, avoid letting someone lock you into a locker that you do not have access to, or you better hope someone gets you out before that locker ends up in an airlock. 

- With the exception of the job-oriented "keypad" lockers that I edited (Warden, Security, Engineering, HOP, Hydroponics, CMO, Mining), lockers remain unlocked by default, locks can be enabled for a prefab closet by:
1) Adding a LockLight asset
2) Under the ClosetControl component, tick the "isLockable" boolean to true and update the LockLight reference 
3) Adding an "access restriction" component

[Here's an example of what the Lock-Lights Look-Like](https://i.imgur.com/Si9ihoI.png)


ghost:
Automatic locking of lockers is a feature that I never knew I needed in BYOND code SS13. This will make locking people in lockers extremely powerful however, especially when we won't have resists for a while! Since you can't resist out of them, this could easily become the new meta for efficient arrests as security: Stun someone and close a locker on them.

Robustin:
Well it would work until the victim reveals they have all-access, stolen sec ID, emag, etc.

Definitely the sort of gimmick I would expect from SS13 players though, handcuffs use an action button to break free so maybe the same could be implemented here.

Misterixmax:
.Without working resist function It's a full round ban for any player. Also locker must have a function to be emaged. And yeah, adding lockers must be done after brig will have a working cells for criminals, otherwise security will use your lockers.

Bod9001:
humm I've had a look through it, And a quick playtest is it possible to be able to toggle the lock, If you have an ID in hand?

Robustin:
Lock toggling independent of openining/closing isn't in this, largely because I wasn't sure if we wanted to:

1) Alt-toggle
2) ID-toggle
3) Use the unity "hold click" menu to provide toggling

I'm most inclined to (1), (2) was always a source of various bugs and issues in /tg/code because of all the ways ID's could be stored.

Bod9001:
ID-toggle 
If you want you can do as well Alt-toggle, 
Basically so you can just make it unlocked, Since one part of the game is not remembering to lock your locker

fomalsd:
@Robustin any news? lockable lockers are a cool addition and I'm looking forward to merging it when it's ready

Robustin:
@fomalsd 

Was out for the holiday weekend, I plan on adding the ID lock toggle today.

Robustin:
@Bod9001 @fomalsd 

Ok, ID cards (in-hand) can now toggle locks independent of opening/closing the locker door. To avoid adding a new function I changed an existing function with the exact behavior I needed to public (in AccessRestrictions.cs) I hope that isn't an issue.

fomalsd (approved):
Thanks! It's working really well!
The only thing that I'd like to change is that no one should be able to open locked lockers (even if you have access) unless you unlock it first.
Otherwise no one with access will ever bother to unlock them with a card, and it was a gameplay feature of people forgetting to lock their lockers
GIF

Robustin:
If that's the preference then I should probably just remove auto-locking altogether. Any other opinions on this?

Bod9001:
humm yeah That sounds good

Fishfish458:
I really think it would be a good option to keep in for super secure stuff, though should be mainly toggle by id for majority

fomalsd:
I'd say leave it to manual locking and don't bother separating "super secure stuff" from "not so secure stuff"

Robustin:
Ok it should just be manual locking/unlocking only now.

Bod9001:
all good

PR #3053, opened 2020-02-11
Prevents dead or unconscious people from removing handcuffs.

### LINKED ISSUE #2993: You can remove your handcuffs when dead
## Description

When you're dead or unconscious, you can remove handcuffs. 

### Steps to Reproduce

1.  Die handcuffed.
2.  Stay in your body
3.  Uncuff yourself normally.

Ruzihm:
### Purpose
Fixes #2993

### Notes:
Problem was caused by a reference to `PlayerHealth.serverPlayerConscious` which doesn't seem to actually do anything.

### DIFF
--- a/UnityProject/Assets/Scripts/Clothing/RestraintOverlay.cs
+++ b/UnityProject/Assets/Scripts/Clothing/RestraintOverlay.cs
@@ -98,9 +98,11 @@ public class RestraintOverlay : ClothingItem
 
 	bool CanUncuff()
 	{
-		if (!thisPlayerScript.playerHealth.serverPlayerConscious ||
+		PlayerHealth playerHealth = thisPlayerScript.playerHealth;
+		if (playerHealth.ConsciousState == ConsciousState.DEAD ||
+			playerHealth.ConsciousState == ConsciousState.UNCONSCIOUS ||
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Clothing/RestraintOverlay.cs
```diff
@@ -98,9 +98,11 @@ IEnumerator UncuffCountDown(CancellationToken cancelToken)
 
 	bool CanUncuff()
 	{
-		if (!thisPlayerScript.playerHealth.serverPlayerConscious ||
+		PlayerHealth playerHealth = thisPlayerScript.playerHealth;
```
DooblyNoobly, line 101: Pretty sure PlayerHealth is null if you are a ghost. Is there anyway CanUncuff() might get called if you are a ghost? Maybe you should check that playerHealth exists and if it doesn't then return false. To be safe.

DooblyNoobly (changes requested):

UnityProject/Assets/Scripts/Clothing/RestraintOverlay.cs
Ruzihm: Great point. `playerHealth` was already being dereferenced before this pull request but now is as good a time as any to have that check.

PR #3056, opened 2020-02-12
made plasma from plasma canisters colder so it won't catch on fire

Bod9001:
### what this do?
made plasma from plasma canisters colder so it won't catch on fire, and A lot of players are complaining about people who rush atmospherics to get the plasma ccanisters to burn everything down
### why you do this?
so it won't catch on fire
### what have you changed?
 plasma canisters
### what you tested?
editor

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Objects/Canisters/PlasmaCanister.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Objects/Canisters/PlasmaCanister.prefab
@@ -36,6 +36,11 @@ PrefabInstance:
       propertyPath: ReleasePressure
       value: 300
       objectReference: {fileID: 0}
+    - target: {fileID: 114104253718565984, guid: e9620f26b24f33c4cbe56a85ae95865d,
+        type: 3}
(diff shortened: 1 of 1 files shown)


AnimatedTanker:
> 
> ### what this do?
> 
> made plasma from plasma canisters colder so it won't catch on fire
> ### why you do this?
> 
> so it won't catch on fire
> ### what have you changed?
> 
> plasma canisters
> ### what you tested?
> 
> editor

This is very needed as it can stop plasmagriefers until we can get true moderation

urbanawakening:
For the love of god, I beg you merge this.

hecksadecimal:
Please. Please merge this.

DooblyNoobly:
This is not the way to address the problem of plasma fires. The canisters need to be in a more secure lock up. The plan is to move atmos to a more secure location in the south west corner of outpost. Maybe even have a security post to guard it as well. We also need to reduce the amount of plasma canisters on the station which will help curb the massive station fires.

Misterixmax:
@dooblynoobly +++++ agree

PR #3059, opened 2020-02-12
Throwing over tables

### LINKED ISSUE #1733: Items aren't thrown onto selected tile
# Description
As throwing works now, thrown items are either thrown past where you select or are unable to land on the surface you throw them on, such as a table. Items thrown towards tables hit the tables rather than land on them or past them.
## Category
BUG or TODO?
BUG
## Detailed Bug Report
GIF2

## Steps to Reproduce
1.Take item in hand
2.Throw item to target tile
3.Observe item's destination

### LINKED ISSUE #2388: More traditional throwing
# Description
At the moment thrown items always miss the clicked position, sliding farther than expected. 
It's impossible to make things stop closer than like 2.5 tiles away (with standard throw range and speed).
Also while technically more realistic, rotated items aren't too pretty and are harder to pick up.

Thrown items should: 

- land on the designated tile (if within range) with minimal residual movement
- if rotated during flight, they should become upright after landing (in a nice looking way)
- if flight is abruptly stopped by a wall, item should slide along the wall for a bit
- perhaps a hitbox should be increased from just a central point to something like 0.5x0.5 square to avoid items partially covering the walls:

image

Ruzihm:
### Purpose
Adds a new collision type for thrown ("airborne") objects, makes tables not block airborne objects

Fixes #1733

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
@@ -13,7 +13,8 @@ public enum CollisionType
 {
 	None = -1,
 	Player = 0,
-	Shuttle = 1
+	Shuttle = 1,
+	Airborne = 2
 };
 
 /// Matrix manager keeps a list of matrices that you can access from both client and server.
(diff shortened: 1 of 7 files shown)


Bod9001:
oh noGIF

 seems like there's some client side prediction that is getting messed up can you have a look into it?

Ruzihm:
@Bod9001 yes I will look into that. Thank you for the feedback

Ruzihm:
@Bod9001 Fixed that issue. Please take another look when you get a chance

PR #3092, opened 2020-02-13
Action system with free networking  V1

### LINKED ISSUE #2298: Action System B: 100
# Description
in ss13 You can have spells and Stuff that You can select by panels on top left
This is used for magic, race abilities, item abilities, E.g Triggering Spell, Something, Jetpack settings, Stuff like that 

These are called "actions".

Add the ability to have "actions", which show up somewhere on the UI (like the top left) and can be dynamically added / removed based on various things. The actions can be clicked to trigger them, and have mouseover text explaining what they do. They might also have some sort of cooldown indicator?

For example, certain pieces of equipment come with actions when you equip them, like a helmet lamp.

Actions should be able to be dynamically added / removed regardless of WHY that happens. You can gain / lose them in various ways. Such as being a wizard.

I suggest that all actions are defined using an SO, so you can set their text, icon, and other behavior. They should have a callback that can be dynamically configured to be invoked at runtime.

Example
image

Another example
Screen Shot 2019-11-21 at 11 31 01 AM

Bod9001:
Is bounty fixes
Fixes #2298

Need a little bit of work still cleaning up just doing before I go off to university

### what this do?
Simple way of managing UI actions
### why you do this?
Bounty and making it a lot simpler to use the action system

### what you tested?
Editor currently
Need to test multiplayer

### What do you need to add
Remove test components
And add anything if multiplayer test reveals anything


fomalsd:
there are some conflicts.
are there code examples on how to use this?

Bod9001:
You can see examples in IActionGUI

Bod9001:
dam it, Damn merge conflicts

UnityProject/Assets/Scripts/Clothing/BackPack/InteractableStorage.cs
```diff
@@ -417,8 +424,13 @@ public void OnInventoryMoveClient(ClientInventoryMove info)
 			var pna = PlayerManager.LocalPlayerScript.playerNetworkActions;
 			var showAlert = pna.GetActiveHandItem() == gameObject ||
 							pna.GetOffHandItem() == gameObject;
-			UIManager.AlertUI.ToggleAlertPickupMode(showAlert);
+
+			UIActionManager.Instance.SetAction(this,showAlert);
```
chairbender: From an API usability perspective, I think it would be nice to call it like this:
```
UIAction.Show(this)
or
UIAction.Hide(this)
or
UIAction.Toggle(this,true or false for on or off)
```

I.e. wrap the manager calls with a static class with static methods and add overloads which omit the boolean (while still keeping the one with the boolean)

UnityProject/Assets/Scripts/Clothing/BackPack/InteractableStorage.cs
```diff
@@ -51,6 +53,11 @@ enum PickupMode
 	/// </summary>
 	private PickupMode pickupMode = PickupMode.All;
 
+
+	[SerializeField]
+	private ActionData actionData;
```
chairbender, line 58: ah I like this approach, didn't think about having a get-only property as part of the interface for exposing config.

UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
```diff
@@ -238,6 +278,31 @@ public void SetInfo(SpriteData _Info, int _spriteIndex = 0, int _variantIndex =
 		}
 	}
 
+
+	/// <summary>
+	/// Used to generate A buffer from a list of sprites
```
chairbender:
```suggestion
	/// Set this sprite handler to be capable of displaying the elements defined in _Info and sets it  to display as the element in _Info indicated by _spriteIndex and _variantIndex
```

UnityProject/Assets/Scripts/UI/Action/ActionControlInventory.cs
```diff
@@ -0,0 +1,34 @@
...
+		{
+			UIActionManager.Instance.SetAction(_IActionGUI, showAlert);
+		}
+	}
+
+	void Start()
+	{
+		var ActionGUIs = this.GetComponents<IActionGUI>();
+		foreach (var ActionGUI in ActionGUIs) {
+			if (!ActionGUI.ActionData.PreventBeingControlledBy.Contains(ActionControllerType)) {
```
chairbender, line 28: not sure on the use case for this logic

UnityProject/Assets/Scripts/UI/Action/IActionGUI.cs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+
+/// <summary>
+/// Simply implement this to Implement your Networked screen action
+/// </summary>
+public interface IServerActionGUI : IActionGUI
+{
+	void CallActionServer(ConnectedPlayer SentByPlayer); //Requires validation in this
+	NetworkIdentity GetNetworkIdentity(); //so since it needs Something to relate scripts
```
chairbender: Shouldn't be needed. See my comment in the message classes

UnityProject/Assets/Scripts/UI/Action/RequestGameAction.cs
```diff
@@ -0,0 +1,107 @@
...
+				var IServerActionGUI = (IActionGUIs[ComponentLocation] as IServerActionGUI);
+				IServerActionGUI.CallActionServer(SentByPlayer);
+			}
+		}
+	}
+
+
+	public static RequestGameAction Send(IServerActionGUI iServerActionGUI )
+	{
+		var netObject = iServerActionGUI.GetNetworkIdentity();
```
chairbender: You don't need this GetNetworkIdentity method in the interface. Cast iServerActionGUI to Component, then call GetComponent<NetworkIdentity>()

UnityProject/Assets/Scripts/UI/Action/SetActionUI.cs
```diff
@@ -0,0 +1,165 @@
...
+
+	private static SetActionUI _Send(GameObject recipient,
+									 IServerActionGUI iServerActionGUI,
+									 SetActionUIActions ProposedAction,
+									 bool _showAlert = false,
+									 int location = 0)
+	{
+		if (!(iServerActionGUI is UIActionScriptableObject))
+		{
+			var netObject = iServerActionGUI.GetNetworkIdentity();
```
chairbender: You don't need this GetNetworkIdentity method in the interface. Cast iServerActionGUI to Component, then call GetComponent<NetworkIdentity>()

UnityProject/Assets/Scripts/UI/Action/RequestGameAction.cs
```diff
@@ -0,0 +1,107 @@
...
+		var _ComponentType = iServerActionGUI.GetType();
+		var iServerActionGUIs = netObject.GetComponentsInChildren(_ComponentType);
+		var _ComponentLocation = 0;
+		bool Found = false;
+		foreach (var _iServerActionGUI in iServerActionGUIs) {
			if ((_iServerActionGUI as IServerActionGUI) == iServerActionGUI) {
```
chairbender: This allows it to have multiple action components on a single game object, right? Nice approach.

EDIT: Also remove the ^M

UnityProject/Assets/Scripts/UI/Action/RequestGameActionSO.cs
```diff
@@ -0,0 +1,30 @@
...
+using UnityEngine;
+using Utility = UnityEngine.Networking.Utility;
+using Mirror;
+
+public class RequestGameActionSO : ClientMessage
+{
+	public static short MessageType = (short)MessageTypes.RequestGameActionSO;
+	public string soName;
+
+	public override IEnumerator Process()
```
chairbender, line 11: Oh cool, this allows you to request an action via an SO instead of via a component, so it looks like you added multiple ways to trigger UI actions.

One issue, sending a string name over the network is pretty expensive in terms of bandwidth. Ideally it should be a numeric ID. There are a few suggestions online for generating unique IDs for scriptable objects, such as doing it when the asset is saved.

EDIT: In fact, I saw that you already have setup logic for these SO's so actually generating the ID's is really straightforward! You can use a similar approach as with the component IDs - just sort the SO assets alphabetically by name, and us an auto-incrementing ushort as the id. You can build the dict that way instead of using string as the key.

UnityProject/Assets/Scripts/UI/Action/UIActionSOSingleton.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+[CreateAssetMenu(fileName = "UIActionSOSingleton", menuName = "Singleton/UIActionSOSingleton")]
+public class UIActionSOSingleton : SingletonScriptableObject<UIActionSOSingleton>
+{
+	public List<UIActionScriptableObject> uIActions = new List<UIActionScriptableObject>();
+
+	public static Dictionary<string, UIActionScriptableObject> actions = new Dictionary<string, UIActionScriptableObject>();
```
chairbender: see my earlier comment about doing this using ints instead of strings

docs/development/How-to-use-UIActions.md
```diff
@@ -0,0 +1,38 @@
+# How to use UIActions
+
```
chairbender, line 2: nice, you went above and beyond by adding some documentation on how to use this

UnityProject/Assets/Scripts/UI/Action/ActionData.cs
```diff
@@ -0,0 +1,24 @@
...
+[CreateAssetMenu(fileName = "ActionData", menuName = "ScriptableObjects/ActionData")]
+public class ActionData : ScriptableObject
+{
+	public bool CallOnClient;
+	public bool CallOnServer;
+
+	public List<SpriteSheetAndData> Sprites = new List<SpriteSheetAndData>();
+	public List<SpriteSheetAndData> Backgrounds = new List<SpriteSheetAndData>();
+
+	public List<ActionController> PreventBeingControlledBy = new List<ActionController>();
```
chairbender, line 16: I'm having trouble understanding the use case for this field. Can you clarify?

chairbender (changes requested):
Not bad! I can tell you were willing to put up with whatever difficult internal logic you had to write in order to make the API simple and straightforward. There are a few public things I didn't mention that I'd want to have more comments or more clarity in the comments but I can take that on later, don't worry about it.

Still need to actually test it out in local MP more, but I have to sleep for the night. I reviewed all the code and that should give you some stuff to do in the meantime.

UnityProject/Assets/Scripts/UI/Action/ActionControlInventory.cs
Bod9001: so, Let's say if you have a action implemented on your Item E.G The spell Call to yourself, When you drop it you want to keep the spell because The spell is Summoning the item to you so it wouldn't make sense if it would disappear when you drop it, but however It has another Component for delinking The spell and That's getting turned off and on by the inventory script, so you wouldn't want the summons spell to be turned off  by the inventory Script

UnityProject/Assets/Scripts/UI/Action/RequestGameAction.cs
Bod9001: The reason I decided to do this is because they could be calling it from component without a network behaviour in the game object, so if that was the case then this would fail, so It's a bit of extra logic so they can specify the correct network identity to use if there Isn't one on the same game object

UnityProject/Assets/Scripts/UI/Action/RequestGameAction.cs
Bod9001: Multiple of the same class

UnityProject/Assets/Scripts/UI/Action/SetActionUI.cs
```diff
@@ -0,0 +1,165 @@
+﻿using System.Collections;
+using UnityEngine;
+using Utility = UnityEngine.Networking.Utility;
+using Mirror;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
```
daalta, line 9: I think having a short class description for your new files would be useful for future development.

Bod9001:
am just waiting on  chairbender

UnityProject/Assets/Scripts/UI/Action/RequestGameAction.cs
chairbender: @Bod9001 it shouldn't fail in that case. Whether the component implementing the logic is a NetworkBehavior or not shouldn't affect this system - what matters is that the object itself has a NetworkIdentity. Pretty much all objects in the game have a NetworkIdentity, even if none of their components are NetworkBehaviors.

UnityProject/Assets/Scripts/UI/Action/ActionControlInventory.cs
chairbender: This sounds like a complex way to solve that unique problem, and it's pulling this complexity into a very fundamental part of the action system - the ActionData, just to solve this unique problem. Can't this be solved without such a complicated field, such as a flag? Maybe a "Don't Remove When Dropped" field (something like that). PreventBeingControlledBy is hard to make sense of so I wouldn't want it to be part of the main API for working with actions.

UnityProject/Assets/Scripts/Clothing/BackPack/InteractableStorage.cs
chairbender: Please also get rid of the need for "Instance". We have plenty of other static util classes which show how to accomplish this.

UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
chairbender: I think you resolved this without applying the suggestion. Please fix the comment as indicated so it's clear what it's doing

chairbender:
Still have some changes to request. See my follow up messages.

UnityProject/Assets/Scripts/Clothing/BackPack/InteractableStorage.cs
Bod9001: https://github.com/unitystation/unitystation/pull/3092/commits/afacb5c6adafe747776c9d1f682f5deabb7cbe96
Already done, You're viewing outdated code

UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
Bod9001: Outdated code as well

UnityProject/Assets/Scripts/UI/Action/ActionControlInventory.cs
Bod9001: I don't feel like it's a complicated Thing, like it's going to be hardly used,
and it's just a list of stuff saying that you should ignore me if you're this

fomalsd:
@chairbender could you take another look at this please?

Ruzihm:
There's a NRE that occurs if you equip an item with a paletteable sprite but this is my fault and was fixed with 3f1c4edc. When I merge this (a5271c5) with unitystation/develop (d3dad25), I don't see any issues with paletteable stuff 👍

DooblyNoobly:
Testing failed on the staging server so had to revert this PR: #3301 please see #2298 for a list of the issues

Issue #2851, opened 2020-02-03
Automatically check for missing prefab references


chairbender:
## Description

It would be useful, especially if it can be part of a build process or at least an automated script to begin with.

We would want to scan all assets and prefabs and locate any which have missing Prefab references.

This will help prevent issues where someone rebuilt a prefab and didn't update all references to it.

chairbender:
Asset Usage Detector we already have might even be able to do this

PR #3099, opened 2020-02-13
Adds test for missing component in asset references.

Ruzihm:
Also, changes asset finder to only care if it starts with "Assets/" and ends with ".asset"

### Purpose
Addresses (Maybe fixes) #2851

### Notes
There are some missing components in unitystation/unitystation@6a71c84e996767df99d8656912d9e5f8208591cb that this test (correctly) fails on. I don't know how best to fix them~

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Objects/Atmos/BentPipe.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Objects/Atmos/BentPipe.prefab
@@ -99,7 +99,6 @@ GameObject:
   - component: {fileID: 114185907784239616}
   - component: {fileID: 114612731264727302}
   - component: {fileID: 1739955761982081149}
-  - component: {fileID: 114253643887694602}
   - component: {fileID: -753721034424592532}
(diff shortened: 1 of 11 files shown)


Bod9001:
Could try manually removing them if they don't look crucial?

chairbender:
That's awesome. I wouldn't have expected in a million years someone would add more unit tests. I'll take a look at what's missing and see if it's okay to remove.

chairbender:
This is great, I didn't even realize we had the ability to run normal unit tests outside of Unity's really slow test runners now.

Yes, removing them is fine

Ruzihm:
@chairbender sadly it's still in unity's slow unit testing ;)

chairbender:
ah, I was able to just tell rider to run all tests in solution and it seemed pretty fast, at least it's less cumbersome now, last time I was messing with unit tests was over a year ago and it was annoying to do, had to do it through editor

good to merge now though!

Ruzihm:
Testing a fix for this working improperly. Will submit as a new PR

PR #3102, opened 2020-02-14
Added checklist to template request.

h3killa:
Kind of self explanatory.


chairbender:
Believe it or not we used to have this as a checklist! We decided to move it to a simple bullet list a few months ago because people were complaining about how long and intimidating it was, and feeling they had to go in and check each and every box and being annoyed by that. Personally I liked it as a checklist but I respect the wishes of the people who weighed in on it so I want to keep it this way for now.

PR #3105, opened 2020-02-14
Adds Outpoststation 2

Ranzarth:
### Purpose
Because just one isn't enough

### Notes:
Adds Outpoststation copy for mapping changes
Includes new bridge addition in Outpostsation02
Include some fixes for Brig in Outpostsation02


Bod9001:
why copy th map?

Ranzarth:
Per Doobly's request, refer to mapping channel for details

PR #3114, opened 2020-02-14
Fixed #2909 (Latency letting players go over job limits)

### LINKED ISSUE #2909: With lag, multiple people can select the same role.
## Description
If multiple people click roles like Captain in the beginning of a round, if there's enough lag you get 2+ 
### Steps to Reproduce

1.  Have a lot of people online
2.  Have multiple people click captain at round start.
3.  Observe captains.

daalta:
BigReadyIvorybilledwoodpecker-small
*This is what players now see when they failed to spawn for any reason (like their job already being taken).*
*Screenshot is slightly outdated. The wait time is shorter and the player is shown a message:
"Please wait, checking job availability..."*
*Thanks to Necromunger for the feedback.*
### Purpose
* Fixes #2909 : Players could go over job limit when they select their role at the same moment (depending on latency). For example: More than 1 captain could spawn. This is now fixed.

### Notes:
* This PR introduces a serverside check which prevents the player from being spawned if their selected job is no longer available.
* The job select UI now (partially) hides itself when a job is selected. It then waits for ~~6~~ 4.5 seconds, awaiting the player's spawn.
  * If the player spawns then the UI is fully hidden.
  * If the timer runs out then it is assumed that the player did not spawn. The job menu is then fully revealed again. 
  * The length of the timer (~~6~~ 4.5 seconds) is probably much longer than necessary, but should ensure that jobs do not exeed their slot limit. We might want to shorten the timer in case players get confused, or add something to the UI.

### Self checks below before submitting a PR:

- [x] The PR has been tested in multiplayer
  - With 1 server and 2 clients both trying to get the same role and failing. 1200ms latency
  - With 1 server and 1 user both forced to select the same job (by keeping job buttons always on). ~0ms latency.
  - It's important the the UI shows up again! In my tests this never happened, but I'm concerned about it.

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/SceneConstruction/NestedManagers/UIManager/Windows (Put all window UI's in here).prefab
+++ b/UnityProject/Assets/Resources/Prefabs/SceneConstruction/NestedManagers/UIManager/Windows (Put all window UI's in here).prefab
@@ -900,6 +900,83 @@ MonoBehaviour:
     m_VerticalOverflow: 0
     m_LineSpacing: 1
   m_Text: Starting in...
+--- !u!1 &6966413391240333405
+GameObject:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  serializedVersion: 6
+  m_Component:
+  - component: {fileID: 1825598347150382488}
+  - component: {fileID: 3832769593240119352}
+  - component: {fileID: 7583201979419768367}
+  m_Layer: 5
+  m_Name: WaitText
+  m_TagString: Untagged
+  m_Icon: {fileID: 0}
+  m_NavMeshLayer: 0
+  m_StaticEditorFlags: 0
+  m_IsActive: 0
+--- !u!224 &1825598347150382488
+RectTransform:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 6966413391240333405}
+  m_LocalRotation: {x: -0, y: -0, z: -0, w: 1}
+  m_LocalPosition: {x: 0, y: 0, z: 0}
+  m_LocalScale: {x: 1, y: 1, z: 1}
+  m_Children: []
+  m_Father: {fileID: 8530562246598747092}
+  m_RootOrder: 4
+  m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
+  m_AnchorMin: {x: 0.5, y: 0.5}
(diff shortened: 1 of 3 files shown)


Scorr:
Nice, but couldn't the timer be avoided by just sending a message back from the server with a succeed/fail?

daalta:
@Scorr Hmm, I suppose so! Though I'm not familiar enough with Unitystations's networking to create that.

The timer solution should suffice for the moment though. Players don't always pick the same role within ~300ms, which is when this bug usually occurs. If the 6s confirmation gets too annoying we could shorten it - I don't think a ping of 6000 ms happens often.

We might also want to consider a maximum ping we allow our players to have. I was surprised that I could test my changes with a ping of 1200 ms. That would be hard to tolerate in a real game.

Necromunger:
Hiding the UI from the player should show another message about what is happening. At the same time you toggle off the UI for roles, we can show another message about how it's checking if role is free.

daalta:
@Necromunger Sure, give me a moment.

daalta:
Alright, the player now sees a message informing them that the job availability is being checked.

I also shortened the wait time to a maximum of 4.5 seconds, and added more comments.

fomalsd:
There should be a better way than relying on time delays. Otherwise it will never 100% guarantee lack of job duplicates, if server happens to lag hard, for example. And extra wait time will annoy players and will give an impression of hanging.

Evan-R-Rogers:
I think for now this is a good idea and should merged to live. The timer is a bit too long, however, and in my opinion should be around 2 seconds at most.

Though you could go ahead and incorporate profession preferences into character creation ;)

daalta:
@fomalsd Thanks you for taking a look!

The job limit **is** actually checked serverside. So jobs should no longer go over their limit after this PR is merged (unless I missed something).

```
[Command]
public void CmdRequestJob(JobType jobType, CharacterSettings characterSettings)
{
	int slotsTaken = GameManager.Instance.GetOccupationsCount(jobType);
	int slotsMax = GameManager.Instance.GetOccupationMaxCount(jobType);
	if (slotsTaken >= slotsMax)
	{
		return;
	}

...

	PlayerSpawn.ServerSpawnPlayer(...
}
```


I've not implemented a message for the server to inform the client about the spawn failure. Instead the client waits for `PlayerManager.HasSpawned` to be true after choosing a job, hence the 4.5 second delay.

So duplicate jobs should no longer happen, but one thing I *could* image happening is players being spawned twice.

1. Player reqests to spawn as a certain role
2. Server processes the request and spawns the player.
3. Job selection times out because `PlayerManager.HasSpawned` has not yet updated client-side (server response took more than 4500ms)
4. Player is selects another job
5. ???

I'm testing this right now to see what happens. If the player *does* get spawned then more work needs to be done, maybe in `ServerSpawn.cs`.

daalta:
Alright, so even with extremely high ping players cannot spawn multiple times via the job buttons, even if they choose several. They always receive the first (available) job they picked. Yaay 🙌

Fixed a different issue though where the job window doesn't close when players spam different job.

EDIT: Cleared things up with foma, removed the DNM.

Issue #2915, opened 2020-02-06
Local events (chat, sounds) can be heard outside of range B: 120


fomalsd:
Sometimes you get event messages and sounds from the station while you're e.g. far away mining on a shuttle. This should be addressed.

fomalsd:
uh, still a thing

fomalsd:
I really think this needs a bounty since it's much trickier than it looks for whatever reason

DooblyNoobly:
This is a complicated bug and is hard to recreate locally. Its urgent as well so placed a bounty of 120 on it

PR #3141, opened 2020-02-16
Fixes a possible cause of issue #2915

goman0:
A List can't safely remove items while it's being iterated over. Also, this should make chat code less expensive.

### Purpose
Fixes a code issue in chat code.

- [ ] The PR has been tested in editor

### DIFF
--- a/UnityProject/Assets/Scripts/Chat/ChatRelay.cs
+++ b/UnityProject/Assets/Scripts/Chat/ChatRelay.cs
@@ -150,7 +150,7 @@ public class ChatRelay : NetworkBehaviour
 		
 			if (players[i] == null)
 			{
-				continue
+				continue;
 			}
 			
 			ChatChannel channels = chatEvent.channels;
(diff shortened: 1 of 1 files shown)


DooblyNoobly:
did you test this in multiplayer?

goman0:
Not yet @DooblyNoobly. I still need to set up Unity.

Ruzihm:
ooh nice catch @goman0

DooblyNoobly:
I'm testing it now. No errors so far. I might put it into production and see what happens

DooblyNoobly:
Can't hear pete gasping from cargo so thats a good sign

Scorr:
you can remove while iterating safely if you loop backwards.

I don't know what the exact intention of this code is (maybe removal is not the best here either?), but just setting list values to null doesn't sound like the right way to do it.

DooblyNoobly:
Can't hear shuttles crashing into the otherside of the station so thats good too. Lets give it a go. Iterating backwards or setting them to null achieves the same result. Iterating backwards might of been better but it is what it is now. Lets see what it happens in a build

PR #3143, opened 2020-02-16
Canister holding tank

### LINKED ISSUE #2297: Canister Holding Tank B: 80
# Description
Need to implement the holding tank functionality / UI so you can fill things from a canister (such as an oxygen tank). This will require adding to the canister UI - a label for the name of the tank, an indicator of the pressure, and an eject button. This stuff should only show up when a tank is inserted. The sprite of the canister should also change when a holding tank is inserted.

Looks like this:
image

DeeJayhX:
# Pull Request Template

### Purpose
_Describe the problem the PR fixes or the feature it introduces_<br>
_Don't forget to use "Fixes #issuenumber" to select issues and auto close them on merge_
Fixes #2297 

### Notes:
-Updated Canister GUI to reflect ability to refill tanks
-Added ability to click on canister with tank to insert the tank
--Will not allow if a tank is already inserted
-Added ability to eject tank from canister
--Will do nothing if no tank is inserted
-Added CanisterFillable item trait
--Should only be applied to items that can fit into and attach to a canister.


UnityProject/Assets/Resources/ScriptableObjects/Traits/CommonTraitsSingleton.asset
```diff
@@ -34,4 +35,4 @@ MonoBehaviour:
   PlasteelSheet: {fileID: 11400000, guid: debdeac43ef6b474f934cecc7f43a273, type: 2}
   Cable: {fileID: 11400000, guid: e39c048ec864b47a495a40f9c76af764, type: 2}
   Welder: {fileID: 11400000, guid: aaed69546e98a47d8b488a046d00ebe7, type: 2}
-  Shovel: {fileID: 11400000, guid: 35ecd7ae92ffc9f4d9e62613c2ad8cf8, type: 2}
+  Shovel: {fileID: 11400000, guid: b336a4abd9d45924cb48b49a49fdea2b, type: 2}
```
fomalsd: is shovel reference change intentional?

UnityProject/Assets/Scripts/UI/Canister/GUI_Canister.cs
```diff
@@ -326,16 +360,108 @@ public void ServerEditReleasePressure(string newValue)
...
+								   "insert a tank, and then open the tank valve.");
 		}
 	}
 
 	public void CloseTab()
 	{
 		ControlTabs.CloseTab(Type, Provider);
 	}
+
+	public void EjectExternalTank()
```
fomalsd, line 458: Looks like this code is executed on clientside rather than serverside. NetTab UI logic is supposed to run on serverside and pass information back to clients using NetElements.
`PlayerManager.LocalPlayer` won't work on a headless server since there's no local player. Also there can be multiple tab 'peepers' at any given time, so keep tat in mind. They can be accessed from `Peepers` hashset, but it's highly advisable to display all tab related warnings within a tab (updating some label) instead of resorting to chat notifications for UX reasons.
Since you don't know who exactly initiated the ejection, just make ejected container appear on top of canister.

UnityProject/Assets/Scripts/Util/SweetExtensions.cs
```diff
@@ -26,7 +26,7 @@ public static ItemAttributesV2 Item(this GameObject go)
 	public static string ExpensiveName(this GameObject go)
 	{
 		var item = go.Item();
-		if (item != null && !String.IsNullOrWhiteSpace(item.ArticleName)) return item.ArticleName;
+		if (item != null && !String.IsNullOrWhiteSpace(item.ArticleName)) return item.ArticleName.Replace("(Clone)","");
```
fomalsd: hmm, I think this is more of a bandaid for a deeper issue. ArticleName isn't supposed to have `(Clone)` in its name in the first place. maybe you could keep it as is for the time being

fomalsd (changes requested):
Thanks for contribution!
UI's looking good, but ejection won't work on a real server due to it being clientside while it should've been serverside. Wrote suggestions below.

UnityProject/Assets/Resources/ScriptableObjects/Traits/CommonTraitsSingleton.asset
DeeJayhX: To my knowledge, I didn't change it, Unity did. So I would classify it as Unintentional, but possibly correct?

UnityProject/Assets/Scripts/Util/SweetExtensions.cs
DeeJayhX: @fomalsd 
> hmm, I think this is more of a bandaid for a deeper issue. ArticleName isn't supposed to have (Clone) in its name in the first place. maybe you could keep it as is for the time being

The code is as follows before the fix:
```csharp
public static string ExpensiveName(this GameObject go)
{
    var item = go.Item();
    if (item != null && !String.IsNullOrWhiteSpace(item.ArticleName)) return item.ArticleName

    var player = go.Player();
    if (player != null && !String.IsNullOrWhiteSpace(player.Name)) return player.Name;

    return go.name.Replace("NPC_", "").Replace("_", " ").Replace("(Clone)","");
}
```

Any GasContainer is considered an item.

The way I'm displaying the name is:
`externalTankName.SetValue = insertedContainer.ExpensiveName();`

which is displaying (Clone) at the end, because anything that returns as an item doesn't filter out (Clone).

I see a few solutions to this:
-Add a replace for (Clone) for items, which was my original solution
-Revert the replacement for (Clone) on items and use `netLabel.SetValue = item.Item().InitialName;`
instead of `netLabel.SetValue = item.ExpensiveName();`

Would the latter be preferred? Seems logical to me, and works in testing, but I'm curious if there's an issue I'm missing with it.

UnityProject/Assets/Scripts/UI/Canister/GUI_Canister.cs
DeeJayhX: Fixed.

UnityProject/Assets/Scripts/Util/SweetExtensions.cs
DeeJayhX: Fixed:
Reverted the replacement for (Clone) on items in SweetExtensions.cs and use `netLabel.SetValue = item.Item().InitialName;`
instead of `netLabel.SetValue = item.ExpensiveName();`

UnityProject/Assets/Resources/ScriptableObjects/Traits/CommonTraitsSingleton.asset
DeeJayhX: Reverted and pushed.

UnityProject/Assets/Scripts/UI/Canister/GUI_Canister.cs
fomalsd: `Chat.AddExamineMsgToClient()` only works on clientside

UnityProject/Assets/Scripts/UI/Canister/GUI_Canister.cs
```diff
@@ -326,16 +360,107 @@ public void ServerEditReleasePressure(string newValue)
...
+			Chat.AddExamineMsgToClient("There is no tank inside this canister.");
+		}
+	}
+
+	private void VentContainer(bool isOpen)
+	{
+		container.Opened = isOpen;
+		if (isOpen)
+		{
+			Chat.AddExamineMsgToClient($"Canister releasing at {container.ReleasePressure}");
```
fomalsd: here, too

UnityProject/Assets/Scripts/UI/Canister/GUI_Canister.cs
```diff
@@ -326,16 +360,107 @@ public void ServerEditReleasePressure(string newValue)
...
+			foreach (Gas gas in Gas.All)
+			{
+				canisterGas.SetGas(gas, updatedCanisterGases[gasIteration]);
+				tankGas.SetGas(gas, updatedTankGases[gasIteration]);
+				gasIteration++;
+			}
+		}
+		else if (usingTank && externalTank == null)
+		{
+			Chat.AddExamineMsgToClient("Tank valve lockout occured! You should close the valve, " +
```
fomalsd: here as well. note how `AddLocalMsgToChat` was used here before

fomalsd (changes requested):
Something is wrong: canister pressure never drops (or just never gets refreshed) while filling tank certainly gets filled with gas. See the gif.
A pressure counter for inserted tank would be really useful as well!
GIF
Also that inserted tank icon is barely visible, perhaps you could just use a generic red tank icon for that, like one you showed in early screenshots?

UnityProject/Assets/Scripts/UI/Canister/GUI_Canister.cs
DeeJayhX: Fixed.

UnityProject/Assets/Scripts/UI/Canister/GUI_Canister.cs
DeeJayhX: Fixed.

UnityProject/Assets/Scripts/UI/Canister/GUI_Canister.cs
DeeJayhX: Fixed.

DeeJayhX:
@fomalsd 
> Something is wrong: canister pressure never drops (or just never gets refreshed) while filling tank certainly gets filled with gas. See the gif.

This was working on a previous iteration of the call, but broke when I decided to change that. Fixed. 
(Older commit, ignore outdated GUI)


> A pressure counter for inserted tank would be really useful as well!

Planned on adding this as well as ways to depressurize tanks and other nifty things in the near future, but as I see it was in the original SS13 canister GUI, it seems only fair to include it now for an issue with a bounty attached. Added.

latestgui

> Also that inserted tank icon is barely visible, perhaps you could just use a generic red tank icon for that, like one you showed in early screenshots?

This is working as intended, it's meant to be a "behind thick glass" look. It used to be even more transparent, but I upped the opacity significantly for this exact reason. Will leave as-is, especially now that the GUI has a flashing display to indicate changes.

> Chat.AddExamineMsgToClient() only works on clientside
> here, too
> here as well. note how AddLocalMsgToChat was used here before

All resolved. Messages that don't directly have to do with the interaction have been moved to be displayed on the GUI.

DeeJayhX:
Fixed levers going off-center when turning:
yay

fomalsd (approved):
sweet, thatks!

Issue #3072, opened 2020-02-12
Make Unity Test Runner generate human-readable reports


Macoron:
## Description
Right now Unity Tests runs on a github actions CI. It works nice and smooth, but the generated report is non-human readable. It's NUnit3 .xml output that looks like this:
```
...    
<test-case id="1007" name="CheckChannelAndClearMessage(&quot;&quot;,None,&quot;&quot;)" fullname="Tests.ChatParseInputTests.CheckChannelAndClearMessage(&quot;&quot;,None,&quot;&quot;)" methodname="CheckChannelAndClearMessage" classname="Tests.ChatParseInputTests" runstate="Runnable" seed="1775714569" result="Passed" start-time="2020-02-10 21:38:55Z" end-time="2020-02-10 21:38:55Z" duration="0.000645" asserts="0">
<properties />
</test-case>
...
```

There are several way to implement this:

- Use third-party test reporter like Jenkins, Circle CI, Travis, etc. Sounds like overkill, because our current CI fit all needs, easy to setup and free.
- Use [Github Checks API](https://developer.github.com/v3/checks/) to generate report after actions execution. Sounds nice, but hard to implement ([example of Travis check report](https://github.com/Baystation12/Baystation12/runs/439813912))
- Generate .html by unity actions and download this as artifact. I really like this approach, fits nice to current CI. [Check this forum topic for more details.](https://forum.unity.com/threads/unity-test-runner-xml-report-to-html.826545/)

Feel free to write your idea and suggestions. I will make some research about how to run .NET apps in github actions environment.

NoooneyDude:
Looks like that PR #3146 fixes this issue and so this should be closed.

PR #3146, opened 2020-02-16
Human-readable test reports

Macoron:
### Purpose
Implements #3072 

- Added [Extent Reporter](https://github.com/extent-framework/extentreports-dotnet-cli) support - now you can download human-readable html test report via artifacts. Check screenshots with reports examples.
- Editor.log will be also saved to artifacts. It can be really heavy though (8 mb right now)
- Fixed a small error in test.yml hook

Снимок экрана 2020-02-16 в 15 29 09
Снимок экрана 2020-02-16 в 15 29 29


JesterX666:
Did not test, but looked at the changes.  Looks good.  I thing that's enough of a check for code that doesn't go in production per se.

PR #3155, opened 2020-02-16
Update manager refactor

Scorr:
Updated the update manager. The main objective of this PR is to add profiling functionality to the update manager so we can actually see where frame time is going. Before this PR only "UpdateManager.Update()" shows up in profiler, but now the individual methods are visible.
Sorry about the text, the text in this PR is probably longer than the code I changed, I just wanted to be thorough.

**Refactor**
* UpdateManager add/remove is now faster, but uses slightly more memory.
* UpdateManager methods now show up properly in profiler with where it's being called from. Excluded from build to prevent overhead.
* Now catches exceptions. Excluded from build to prevent overhead, even though the performance impact is very small (less than 10% of what one iteration costs). One exception in updatemanager flow will still break entire build, but in editor keeps working. We might want to extend this to builds as well, but that is up for debate.
* Moved priority in the script execution order so it will always execute (and be available) before everything else. This is now an assumption you can make rather than having it be random wether updates before or after.

**Fixes**
* Updated existing calls to add/remove to use new static method.
* Removed calls and reflection that were used in case user did not unsubscribe/tried to double subscribe, user now needs to always unsubscribe properly. This revealed the following issues that were fixed:
* Fixed invalid remove that happened during UprightSprites during editor (non-playmode) time.
* Fixed duplicate add that happened when waking up CustomNetTransform before sleep timer had completed.

**Changes**
UpdateManager.Instance.Add(Action) has been changed to UpdateManager.Add(CallbackType, Action). This is to make it easier for the user (as in programmer), because they should not have to think about instances. Same for Remove.
Old methods were left functional and marked obsolete to prevent breaking code that people are currently working on.

**Benchmarks**
**Updating 10k calls**
Before commit: 0.25ms
After commit: 0.25ms
After commit, profiling enabled: ~2ms

**Adding 10k**
Before commit: 1000+ ms, 383.9 mb gc alloc
After commit: 12.87ms, 2.9 mb gc alloc
After commit, profiling enabled: 108ms, 12.1 mb gc alloc

**Removing 10k**
Before commit: 1000+ ms, 383.9 mb gc alloc
After commit: 16.80ms, 2.1 mb gc alloc
After commit, profiling enabled: 20ms, 2.1 mb gc alloc

I don't want to spam this entire issue with screenshots, but here's the last 3:
image
image
image

**Conclusion**
Updating is the same performance in build. Adding/removing is a lot faster. Profiling reduces performance. I think the overhead on profiling is acceptable because now we can optimize what's actually in the update loop by seeing what takes the most time, and it is excluded from build.

### DIFF
--- a/UnityProject/Assets/Scripts/Botany/hydroponics tray/HydroponicsTray.cs
+++ b/UnityProject/Assets/Scripts/Botany/hydroponics tray/HydroponicsTray.cs
@@ -49,7 +49,7 @@ public class HydroponicsTray : NetworkBehaviour, IInteractable<HandApply>
 	public override void OnStartServer()
 	{
 		EnsureInit();
-		UpdateManager.Instance.Add(ServerUpdate);
+		UpdateManager.Add(CallbackType.UPDATE, ServerUpdate);
 		IsServer = true;
 		if (isSoilPile)
 		{
@@ -79,7 +79,7 @@ public class HydroponicsTray : NetworkBehaviour, IInteractable<HandApply>
 	{
 		if (IsServer)
 		{
-			UpdateManager.Instance.Remove(ServerUpdate);
+			UpdateManager.Remove(CallbackType.UPDATE, ServerUpdate);
 		}
 	}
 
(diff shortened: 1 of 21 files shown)


chairbender:
Excellent work, thank you!

PR #3157, opened 2020-02-16
Job information on job selection screen

daalta:
CanineForkedGrison-small

### Purpose
* Introduces a window to the right of the character selection screen, displaying basic information about that job.
* When hovering over a job the information updated (even if that job in unavailable)
### Notes:
New players previously didn't have a good way of determining which job they should pick and how they should play it. And veteran players couldn't tell at a glance which roles are fully implemented yet. This PR should fix that.

The info window contains:
* The job's title
* A preview sprite
* How difficult that role is to play
* How 'completete' the role is
  * These are estimated which I crowdsourced through a poll on Discord.
* Three bulletpoints describing the job
  * Many of these are taken from the tgstation wiki. (Is that ok..?)
* A more elaborate description of the job.
  * Only the assistant currently has a description. The others have none.

The job info UI is modular and can be re-used if we ever change how job selection works.

### Self checks

- [x] The PR has been tested in multiplayer (1 client, 1 server)


Evan-R-Rogers:
There seems to be a grammar error in the 2nd to last paragraph of the assistant profession:

"Not sure what do so?"

chairbender:
Works nicely, thank you for adding this, people are clearly excited for it 👍

daalta:
@Evan-R-Rogers Thank you for pointing that out. Most of the bulletpoints are placeholders - leaving them completely empty just felt wrong. Maybe someone could go through them and add full descriptions in the future?

@chairbender Thanks for the review!

Issue #2378, opened 2019-11-28
Set up all the belts


chairbender:
# Description
Set up all the belts so they have the correct structures, capacities, and initial items (in their Full variants).

You should use prefab variants for each different type of belt. Use the existing belt prefabs as a reference. You can find the folder with all the belt sprites / existing belt data assets at Assets/Textures/clothing/belt/. Set up the other belts that we have sprite sheets for, create an empty and full variant for each one and make sure any occupations which start with those belts will spawn with the correct belt. Make sure any prefabs you create follow the guidelines here: https://github.com/unitystation/unitystation/wiki/Creating-Items-and-Objects%3A-Now-With-Prefab-Variants

NoooneyDude:
It seems all belts are variants of the Beltwear prefab, and the only belt that doesn't have a populated prefab version is the medical belt.

Update: actually, the medical belt is populated, there is just no empty variant.

PerfectTangent:
This has been done for quite some time, correct? I'm going to close it until otherwise noted.

PR #3172, opened 2020-02-17
Fix for building metal rods, missing trashbags

CodeKalias:
### Purpose
Fixes being unable to build metal rods
Puts trash bags back in janitor lockers, janitor toolbelt populator

### Notes:
Previously broken by prefab IDs changing and .asset files losing their reference to them as a result.

Janitor belts cannot currently accept any janitor items (SpaceCleaner, Mop, TrashBag, Wet floor sign) as they are not whitelisted in Assets/Resources/ScriptableObjects/Inventory/Structure/Belts/JanitorBeltCapacity.asset
Additionally TrashBag is set to be Initial Size Large, so it wont currently fit in the Medium capacity belt even when whitelisted.
As a result both purchased and starter Janitor belts remain empty at this time. (related issue #2378)


fomalsd:
thanks!

PR #3210, opened 2020-02-19
Fixes Syndicate spawns

Ranzarth:
Syndicate spawns were set to the wrong job type, which meant no syndies would spawn


DooblyNoobly:
this was the actual problem breaking the servers! thanks

PR #3221, opened 2020-02-19
Added basic weldability to doors.

### LINKED ISSUE #3020: Add door welding
## Description
You should be able to weld and unweld doors shut with a lit welder. This includes regular doors and airlocks.
Make sure you use progress actions and play the welding sounds during the action, and change the door sprites to be welded/unwelded as needed.

Tjarr737:
## Purpose
Fix of #3020.
Doors can no be welded shut and unwelded again with an activated welder using a non-help-intent.

## Notes:
First time using SyncVars, and I heard they are quite fickle.
### Some gripes I have with my code:
- the sprite renderer for the new overlay has to be inserted into the DoorController. It would be nice to do something like in AirLockAnimator, but I couldn't get such a thing to work
- not really sure how sensible the weld-handling in DoorController is (i.e. can a welded door be forced open by some event?) due to insufficient knowledge of the project
- unsure whether the handling of the graphical aspects might be better in AirLockAnimator - though it has to be said, that such a toggle is not really an animation (and I also do not want to mess with the DoorMessages)
 - welding sound should possibly loop while action is being performed, though a general solution for this kind of problem may be more desirable
### Other:
Many Airlock-Prefabs had non-(0,0,0) standard position-coordinates, which caused a bit of confusion, since added new sprites got this offset as well. All of them now have (0,0,0) as position-coordinates.

### Performed self-checks:

[X] Code is sufficiently commented
[X] Code is indented with tabs and not spaces
[X] The PR does not include any unnecessary .meta, .prefab or <b>.unity (scene) changes</b> [I think a new welded-overlay for all the doors is reasonable.]
[X] The PR does not bring up any new compile errors
[X] The PR has been tested in editor
[X] Any new files are named using PascalCase (to avoid issues on case sensitive file systems) [no new files]
[ ] Any new / changed components follow the [Component Development Checklist](https://github.com/unitystation/unitystation/wiki/Component-Development-Checklist) [I think i followed the guides but this is the first time i use SyncVars, so take a better look at them]
[X] Any new objects / items follow the [Creating Items and Objects Guide](https://github.com/unitystation/unitystation/wiki/Creating-Items-and-Objects%3A-Now-With-Prefab-Variants) (especially concerning the use of prefab variants) [No new items]
[X] The PR has been tested in multiplayer (with 2 clients and late joiners, if applicable) [local]
[X] The PR has been tested with round restarts.
[X] The PR has been tested on moving / rotating / rotated-before-joining matrices (if applicable)

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Objects/Doors/AirLock.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Objects/Doors/AirLock.prefab
@@ -37,11 +37,12 @@ Transform:
   m_PrefabAsset: {fileID: 0}
   m_GameObject: {fileID: 1000011462531918}
   m_LocalRotation: {x: -0, y: -0, z: -0, w: 1}
-  m_LocalPosition: {x: 1978, y: 1131, z: -0.1}
+  m_LocalPosition: {x: 0, y: 0, z: 0}
   m_LocalScale: {x: 1, y: 1, z: 1}
   m_Children:
   - {fileID: 4183345251610980}
   - {fileID: 4000013698625186}
+  - {fileID: 7322139021859764651}
   - {fileID: 4300014015526388}
   - {fileID: 4390488788653686}
   m_Father: {fileID: 0}
@@ -109,6 +110,8 @@ MonoBehaviour:
   DoorCoverSpriteOffset: 49
   DoorLightSpriteOffset: 0
   DoorSpriteOffset: 2
+  weldOverlay: {fileID: 2943727638702044015}
+  weldSprite: {fileID: 21300000, guid: a02b4f763fa778d4ba5c9664ee50c42c, type: 3}
   doorType: 13
   damageOnClose: 0
   ignorePassableChecks: 0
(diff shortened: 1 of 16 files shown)


Fishfish458:
> non-help-intent

Personally I really don't think this action should be housed under specific intents, i don't think it was on ss13

Tjarr737:
@Fishfish458
See https://github.com/tgstation/tgstation/blob/master/code/game/machinery/doors/airlock.dm line 953 
`if(user.a_intent != INTENT_HELP)`
with the code for welding following.
I may be mistaken however, I haven't actually tried on TG to see.

Robustin:
The only way you can undo welding in SS13 is by unwelding (with a welder) or destroying the door. The fact it overrides/ignores every other aspect of door usage has always made it kind of clunky so I think having it in door controller makes sense. 

@Fishfish458 

It wasn't originally but in the last few years a lot of tools got intent-specific treatment so that you could repair a door on help intent, weld it on harm intent, etc.

Fishfish458:
Ahh wasn't aware of the intent changes

DooblyNoobly:
you know what lets try it in staging

PR #3231, opened 2020-02-20
Another attempt to make bananium tiles work

Raptorizer:
# Pull Request Template

### Purpose
as titles says but without map conflics

### Notes:


CodeKalias:
Currently this will not work in a build, an editor launch from StartUp or Lobby or another map as the new sounds are added only in the OutpostStation file.
In the case of the build/startup/lobby this is due to how managers travel between scenes, removing the ones in a newly loaded scene.

The additions to SoundManager should be applied to the prefab, not the scene files
image

Make sure there aren't duplicate entries in OutpostStation's version of SoundManager afterwards, this will likely require another scene save/commit.

Issue #2358, opened 2019-11-26
Fake Walls/Hidden Doors B: 65


fomalsd:
# Description
I think these are rather important for gameplay:
![False_Wall](https://user-images.githubusercontent.com/10403536/69636257-17bd5900-104e-11ea-9314-632dfd7c7f7e.gif)![False_Wall_r](https://user-images.githubusercontent.com/10403536/69636305-2f94dd00-104e-11ea-8672-1ced75e98b6e.gif)

https://tgstation13.org/wiki/Guide_to_construction#Normal.2FReinforced_Hidden_Door

Implementation thoughts:
When built, these should probably still be pure wall tiles in closed state (but with a flag in metadata to indicate that it's a fake wall).
Then check for such flag when clicking wall and replace wall tile with an airlock-like object while it's open (and turn into wall tile again when it's closed).
Since everyone can see all the walls at the moment (and gets all the tile updates and object spawn/despawn messages) it won't be 100% secret (nearby people will see wall tile disappear), but we can't do much in that regard until we introduce network culling anyway. At least it will remain a secret while it's closed.

chairbender:
Adding a bounty of 65. Foma mentioned these are important for gameplay and they're cool. I think it'll be a little complicated due to it needing to act like a wall tile until it's bumped, then turning into an airlock when opened. But foma outlined a reasonable approach to it which helps.

CykaGuy:
I'm working on it

Robustin:
I'm testing a couple of approaches here. I'm leaning toward adding a "dumb door" prefab that would function much like the basic stone & wood doors in /tg/code and the false-wall would just be a variety of that prefab with a different animation and a couple unique interactions with tools so that it completely imitates the behavior of the original.

fomalsd:
fixed in #3370

PR #3246, opened 2020-02-20
Adds False Walls

Robustin:
### Description:
False walls! The cornerstone of any stealthy antagonist. The behavior is identical to how false walls functioned on /tg/code, including the massive headache of getting it to smooth with nearby walls.

This also resolves a bug where GeneralDoorAnimator.cs would break any door that used less than 6 frames for animation and it adds "Non-Automatic" door behavior that can be used for mineral doors and other unpowered doors that don't simply open from being bumped.

This is a bounty (#2358). 

### Remaining Work:

1) Add deconstruction, I need to update my branch with the recent "Weldable Doors" PR and reconcile some interactions to ensure that false walls can be properly deconstructed and can't be welded.

2) Assess how the community feels about highlighting. Highlighting objects by mousing over them is a feature unique to Unity Station, that means you can also highlight false walls - which has some gameplay implications for how easy or hard they are to discover. Should false walls be an exception to highlighting? 

3) Multiplayer testing.


Evan-R-Rogers:
Be sure to add it to the 'doors' tile palette please!

Bod9001:
Only issue is, security, idk what we're doing for that Since client knows there's a wall door there

UnityProject/Assets/Scripts/Doors/GeneralDoorAnimator.cs
```diff
@@ -11,26 +15,40 @@ public class GeneralDoorAnimator : DoorAnimator
...
+	[Tooltip("Only check if the door changes appearance (smooths) based on nearby walls")]
+	public bool Smoothed;
+	private Tilemap tilemap;
+	private MetaTileMap metaTileMap;
+	private TileChangeManager tileChangeManager;
 	private SpriteRenderer doorbase;
 	private Sprite[] sprites;
+	private static readonly int[] map =
+	{
+			0, 2, 4, 8, 1, 255, 3, 6, 12, 9, 10, 5, 7, 14, 13, 11, 15, 19, 38, 76, 137, 23, 39, 46, 78, 77, 141, 139, 27, 31, 47, 79, 143, 63, 111, 207, 159,
```
fomalsd, line 37: woah, what's this?

Robustin:
@Bod9001

There doesn't seem to be many solutions for that atm. I'm not too concerned since cheating in SS13 has always been as easy as having a second account observing the round - when I was an admin on /tg/ it wasn't too difficult to notice when a player had a pattern of finding secret rooms and hidden stuff. Other observers and Ahelps did a good job of weeding out people who cheated.

That is relevant to highlighting though, harder to distinguish between a cheater and a lucky guess when the wall glows on mouseover.

Bod9001:
yeah that's true

UnityProject/Assets/Scripts/Doors/GeneralDoorAnimator.cs
```diff
@@ -74,6 +92,52 @@ private IEnumerator Delay()
...
 	}
 
+	private bool HasWall(Vector3Int position, Vector3Int direction, Quaternion rotation, Tilemap tilemap)
+	{
+		TileBase tile = tilemap.GetTile(position + (rotation * direction).RoundToInt());
+		ConnectedTile t = tile as ConnectedTile;
+		return t != null && t.connectCategory == ConnectCategory.Walls;
+	}
+
+	private int SmoothFrame()
```
fomalsd, line 102: what is this for?

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/TileChangeManager.cs
```diff
@@ -142,6 +142,12 @@ public LayerTile GetLayerTile(Vector3Int cellPosition, LayerType layerType)
 		return metaTileMap.GetTile(cellPosition, layerType);
 	}
 
+	[Server]
+	public MetaTileMap GetMetaTileMap()
+	{
+		return metaTileMap;
+	}
+
```
fomalsd, line 152: it's not java, you can use
`public MetaTileMap MetaTileMap => metaTileMap;`
instead

fomalsd (changes requested):
Something must be wrong, couldn't make the fake wall open. It's also pretty obvious:
img

And yeah, Bod is right about secutiry: no client should know that certain wall is fake until it's interacted with and therefore is in opened state. I explained it in discord some time ago:
```
false wall would be an absolutely normal tile based wall (with a flag in metadatanode) unless it's touched
then wall tile will be replaced with an opening door-like object that remains until it's closed. 
then it's transformed into a tile again
```
Girder code looks good though!

Also here's a list of exceptions I got:
```
IndexOutOfRangeException: Index was outside the bounds of the array.
GeneralDoorAnimator.Awake () (at Assets/Scripts/Doors/GeneralDoorAnimator.cs:50)
UnityEngine.Object:Instantiate(GameObject, Vector3, Quaternion, Transform)
Spawn:_PoolInstantiate(GameObject, SpawnDestination, Boolean&) (at Assets/Scripts/Lifecycle/Spawn.cs:432)
SpawnablePrefab:SpawnAt(SpawnDestination) (at Assets/Scripts/Lifecycle/Spawnable/SpawnablePrefab.cs:57)
Spawn:Server(SpawnInfo) (at Assets/Scripts/Lifecycle/Spawn.cs:262)
Spawn:ServerPrefab(GameObject, SpawnDestination, Int32, Nullable`1) (at Assets/Scripts/Lifecycle/Spawn.cs:138)
Girder:ConstructFalseWall(HandApply) (at Assets/Scripts/Objects/Construction/Girder.cs:206)
<>c__DisplayClass14_0:<ServerPerformInteraction>b__0() (at Assets/Scripts/Objects/Construction/Girder.cs:84)
<>c__DisplayClass1_0:<ServerUseToolWithActionMessages>g__ProgressComplete|0() (at Assets/Scripts/Tool/ToolUtils.cs:34)
StandardProgressAction:OnServerEndProgress(EndProgressInfo) (at Assets/Scripts/UI/ProgressBar/StandardProgressAction.cs:241)
ProgressBar:Update() (at Assets/Scripts/UI/ProgressBar/ProgressBar.cs:233)
```
```
IndexOutOfRangeException: Index was outside the bounds of the array.
GeneralDoorAnimator+<PlayOpenAnim>d__23.MoveNext () (at Assets/Scripts/Doors/GeneralDoorAnimator.cs:186)
UnityEngine.SetupCoroutine.InvokeMoveNext (System.Collections.IEnumerator enumerator, System.IntPtr returnValueAddress) (at C:/buildslave/unity/build/Runtime/Export/Scripting/Coroutines.cs:17)
UnityEngine.MonoBehaviour:StartCoroutine(IEnumerator)
GeneralDoorAnimator:OpenDoor(Boolean) (at Assets/Scripts/Doors/GeneralDoorAnimator.cs:63)
DoorAnimator:PlayAnimation(DoorUpdateType, Boolean) (at Assets/Scripts/Doors/DoorAnimator.cs:21)
<Process>d__4:MoveNext() (at Assets/Scripts/Messages/Server/DoorUpdateMessage.cs:18)
UnityEngine.SetupCoroutine:InvokeMoveNext(IEnumerator, IntPtr) (at C:/buildslave/unity/build/Runtime/Export/Scripting/Coroutines.cs:17)
```
```
IndexOutOfRangeException: Index was outside the bounds of the array.
GeneralDoorAnimator+<PlayCloseAnim>d__22.MoveNext () (at Assets/Scripts/Doors/GeneralDoorAnimator.cs:152)
UnityEngine.SetupCoroutine.InvokeMoveNext (System.Collections.IEnumerator enumerator, System.IntPtr returnValueAddress) (at C:/buildslave/unity/build/Runtime/Export/Scripting/Coroutines.cs:17)
UnityEngine.MonoBehaviour:StartCoroutine(IEnumerator)
GeneralDoorAnimator:CloseDoor(Boolean) (at Assets/Scripts/Doors/GeneralDoorAnimator.cs:74)
DoorAnimator:PlayAnimation(DoorUpdateType, Boolean) (at Assets/Scripts/Doors/DoorAnimator.cs:24)
<Process>d__4:MoveNext() (at Assets/Scripts/Messages/Server/DoorUpdateMessage.cs:18)
UnityEngine.SetupCoroutine:InvokeMoveNext(IEnumerator, IntPtr) (at C:/buildslave/unity/build/Runtime/Export/Scripting/Coroutines.cs:17)
```

Robustin:
@Evan-R-Rogers 

I'm looking at Assets/Tilemaps/Resources/Tiles right now. I'm not exactly sure what these assets even do. I don't see any reference to them in the scene and I don't recall seeing them in any script.

UnityProject/Assets/Scripts/Doors/GeneralDoorAnimator.cs
Robustin: It's the bitmap necessary to select the correct sprite frames for smoothing, most of this code is adapted from ConnectedTiles.cs

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/TileChangeManager.cs
Robustin: Thanks, I meant to ask about this, this was my bandaid solution.

UnityProject/Assets/Scripts/Doors/GeneralDoorAnimator.cs
Robustin: This was the week-long headache for this PR: getting the false wall to "smooth" with nearby walls.

The appearance of the closed false-wall depends on other nearby walls, including NE/NW/SW/SE, so this function works like the one in ConnectedTiles.cs, it looks for nearby tiles that connect to walls and comes up with a frame # for the spritesheet for the correct appearance.

Robustin:
@fomalsd 

I've seen that error in earlier stages of testing, can't seem to reproduce it now. I'll try a clean version of this branch and see why it's not working. Edit: Pulled onto a clean branch, can't reproduce your issues.

Regarding the highlighting, that was something I wanted to clarify, I should be able to disable it.

I considered starting with a modified wall but went for an object since that was how /tg/ handled it. I fear there would be a lot of bugs introduced by treating them as walls first and foremost when they are essentially stealth doors, especially as we introduce new features to walls or doors every week. 

**This is further complicated by layer logic. Walls exist "below" objects/effects/decals/etc., if the False Wall is truly a wall, then it would have the self-defeating effect of showing stuff that you would expect to be hidden by the door. Currently if you close the false wall over a blood stain, weapon, corpse, etc. then it would be "hidden" as expected, but if the false wall turns into a "real" wall when closed when it would slap all the other stuff on top. I could try to change that behavior too but at that point I think it would have to tell the client that the wall isn't the same as other walls.** 

So, ultimately, I'm not sure if its currently possible to create a false wall that can 100% trick clients into thinking it's real while retaining all the functionality we expect from them. And I'm not convinced it would even matter since the best "hack" this game has is observing as a ghost and it's trivial to execute, the only real security that SS13 has ever had is diligent admins/players who take note when someone keeps finding hidden bodies/false walls and the excuse of "just lucky" only works for so long.

fomalsd:
Note: I tested this PR after pulling from develop to this branch first (around 400 commits). will try on another machine later and see if I still get the exceptions.

I *think* tg had it this way: all station walls could be clicked, and it usually responded with “you push the wall, but nothing happens” if it wasn’t a fake one. And if it was indeed a fake wall, it transformed into a door and opened up. So I think it’s pretty clear code wise when to remove a real wall tile with an `IsFakeWall` flag in MetaDataNode and spawn a fake wall door object on its place. 
People will eventually make more elaborate hacks than simple FoV bypass and I’d much rather not tell them that certain wall tile is fake before they ever see someone use it.

What do you think?

Robustin:
@fomalsd 

I haven't resolved conflicts with the main branch yet, while the conflicts aren't explicit it's very possible that Door Welding changed some of the Door Controller/Door Animator logic that could interfere with my changes. 

You are correct about the regular wall behavior on /tg/, but the false walls were structural objects that didn't inherit any wall code - no transformation involved (see: https://github.com/tgstation/tgstation/blob/2591ce17a6609002109fc8eafa48ae0479fc0b8e/code/game/objects/structures/false_walls.dm). 

My chief issue is that it would require some pretty hacky solutions for stuff related to atmos, wall damage, wall interactions, and most importantly the layer issue I bolded above. Basically is false walls are actual walls when closed, their layer would be below the object layer and you'd have blood stains and objects sitting on top of the wall. On /tg/, because they are on the object layer, the doors "cover" anything on that tile, which helps keep them stealthy since objects/blood/etc. would normally never be sitting on a regular station wall. 

I could move the IsFakeWall to a different layer but then that gets us back to the original issue, a wall sitting on a different layer would be an immediate tell to the client that it's not real, right?

fomalsd:
[MetaDataNode](https://github.com/unitystation/unitystation/blob/4395c3d131be651d8180418363f6f4bd90e38bc6/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/MetaDataNode.cs) is layer-agnostic and stores a bunch of metadata for given tile position. My proposal is simply marking normal wall tile as IsFakeWall when building one and checking for that flag when one clicks the wall. Then swap normal wall with fake wall object you have implemented for the duration when it’s opened. Doesn’t sound too complicated, does it?

Robustin:
No, that implementation doesn't seem too complex on its face. However, layer agnostic still means that the wall would need to be on a different layer or else the issues with objects/blood/etc. appearing on the wall would crop up. If I put the wall on a different layer, then the client would know it's not a real wall. How would you propose handling this issue? 

There's plenty of other potential concerns I'd have to sort out as well, I saw a lot of Tile code that assumes that all walls will be on the Wall layer, whereas this would have to be an exception.

fomalsd:
Exceptions confirmed on another machine, updated to latest develop.

I think there might be a misunderstanding. MetaDataNode totally does NOT imply that you'd have to place wall tiles for fake walls on a different layer. MetaDataNodes are created in runtime for each position and store stuff like current gas mixes. There is just one MetaDataNode per local position. You don't have to modify wall ConnectedTile asset for that in any way.

Robustin:
Ok I'll have to look to read about them more, although I still don't understand what is supposed to exist on the closed false door tile. What will exist above the object layer to conceal stuff on the tile?

fomalsd:
when false door closes its object gets despawned and replaced with a usual wall tile. no need to conceal anything or do tricks with layers

Robustin:
That's precisely my concern. The typical "use case" for false walls is using it as a secret hideout for traitors/cultists/revolutionaries where they stash corpses, weapons, evidence, etc.

At the very minimum that involves a likely blood trail, but it also frequently has dropped items, ejected bullet casings, or even a corpse or two or five. In /tg/ code when you close the false wall, it exists on the object layer and almost everything under it - and that is what players like me expect from it. Doors behave the same way, people generally expect doors to cover what's "under" them. 

Your version of the false wall, where the object is deleted when closed and only a "flagged" wall remains, wouldn't be able to perform like that. The wall layer is below objects, blood, mobs, etc. and so the false wall would still reveal all of those things when closed. Not only would it obviously fail to serve one of its core purposes, but it would also create visual inconsistencies where the door would "cover" objects as it closes, then promptly spit them back out on top as the object is deleted and replaced with a flagged wall. 

The solution to that would be to have the "flagged" wall also exist on a different layer, but at that point I don't see the point in making all these changes because surely then the client would know that it's not a true wall because of the layer discrepancy. Of course I'm not entirely sure if that is avoidable but I am 100% sure that using a "usual wall tile" would create a significant deviation from /tg/code's false wall behavior.

fomalsd:
well, that can be worked around by simply removing overlays like blood/wet floor and blowing items away to an unoccupied direction, I don't think it's that much of a problem... unless someone seriously uses it as a one tile safe to keep stuff at?

Robustin:
Yea that's exactly the sort of oddball snowflakey solutions I considered but felt to be too much of a deviation from /tg/ behavior - none of them are ideal. Having a door that deletes atmos/blood and vomits out objects is already a significant deviation from the /tg/ behavior but there's also dozens of other issues that I haven't explored like how to handle bullet holes, damage, anchored objects, atmos generation, etc.

The "real wall" approach is going to make this feature harder to maintain, introduce new bugs, and deviate from /tg/ behavior all in the name of addressing a concern about hackers when there's already an easier and more effective way to "hack" the game simply by opening another client or metacommunicating with an observer. SS13 has always been able to address that cheating the same way they would address "wall hacks", through vigilant admins. I'm just making sure everyone understands this tradeoff before I have to dig into the guts of this PR and rewrite most of it.

fomalsd:
okay, do it how you see fit. though I'm wondering if it's possible to make fake wall object mimic the "shape" (texture) of real wall and avoid seams on sides? otherwise it's pretty easy to detect with a naked eye

Robustin:
That's what this PR does, was the hardest part. Once I address conflicts with the upstream I'll figure out how to remove the highlighting and and make the smoothing happen when it's first created (right now it doesn't remove seams until you close it for the first time).

Robustin:
I'm just going to close this since I clearly screwed up pulling the Upstream, I've got a clean branch ready.

PR #3247, opened 2020-02-20
Cargo Extension Fix

Ranzarth:
Extended Cargo to allow free passage to the cargo shuttle.
Also added two more cargo consoles in HoP's office and in cargo area.
Fixes Arrivals shuttle issues


waffielz:
ogga

PR #3250, opened 2020-02-21
Allow ammo to be stored in secbelts

AutumnNova:
Allows any type of ammo to be stored in secbelts


Evan-R-Rogers:
Any? Including LMG boxes? I think it should be limited to carrying small-arms magazines/clips.

AutumnNova:
what about grenades/shells?

Ranzarth:
Grenades and shotgun shells should fit inside belts

fomalsd (approved):

PR #3251, opened 2020-02-21
Makes hit icons obey palletting

Ruzihm:
### Purpose
Makes sprite that indicate hits with an item use the palette of the item being hit with

### Notes
Hit icons get a Property Block from the weapon being used, which applies the palette (or lack of palette) appropriately.


Jibbajabbafic (approved):
Looks good, tested in game with the captains jumpsuit and it seemed to work!

PR #3252, opened 2020-02-21
Food is now containers

Fishfish458:
food

### Purpose
an AMAZING feature of ss13 is it's ability to have reagents in freaking everything. Cooking is probably one of the first places that really benefits from it.

I really feel like now is an important time to be joining these chemistry systems all together for cooked/grown foods to be working with reagents.

### Notes:
I will take any help or feedback I can get because I am definitely not a c# developer and this container feature needs to be implemented on all food, players, and in a smaller form, materials.

I made this PR because I hope to get some feedback or help in making sure this code actually makes sense before attempting to implement it on every single prefab

### Todo:
- [ ] add nutriment from player "stomach" to their nutrition over time

- [ ] process toxins in user "stomach" to poison them

- [ ] Add container component to literally all current food and drink components

- [ ] Change cooking/grow size behaviour to add nutriment to a food / divide upon slicing


UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
```diff
@@ -458,6 +458,47 @@ public void RpcBeforeBodyTransfer()
 		GetComponent<MouseInputController>().enabled = false;
 	}
 
+	//FOOD
+	[Command]
+	public void CmdEatFood(GameObject food, NamedSlot fromSlot, bool isDrink)
```
fomalsd, line 463: This whole Cmd was previously removed for a reason. Please use TryEat method in Edible instead

UnityProject/Assets/Scripts/Chemistry/ReagentContainer.cs
```diff
@@ -64,6 +64,12 @@ private set
 	[FormerlySerializedAs(nameof(TransferAmount))]
 	private float InitialTransferAmount = 20;
 
+	public bool isFood;
+
+	public bool isDrink;
+	public bool isPlayer;
+	public GameObject leavings;
```
fomalsd, line 71: Nope, these fields don't belong to ReagentContainer. 
Instead, make Edible hold a ReagentContainer reference, and possibly add a `[RequireComponent(typeof(ReagentContainer))]` to the top of Edible as well.
then refer to ReagentContainer and set TransferMode accordingly (InputOnly).
Also you might want to add Syringe trait to Reagent Whitelist so that you could only inject food with a syringe, for example.

fomalsd (changes requested):
Thanks for the PR!
Unfortunately it won't work well with this approach.
Please see notes below.

PR #3302, opened 2020-02-23
Aliases for chat commands and new sing command

corp-0:
### Purpose
Talking about PR's nobody asked for, this one will implement "*emote" as an alias for "/me emote", also "/w text" for whispering and a new chat command; "/s text" for singing.

GIF 23-02-2020 1-30-59

### Notes:
- I honestly don't remember how singing worked on TG. Right now each character has a chance to be capitalized.
- As you can see in the GIF, chat bubble is not displaying ♫ character. No idea how to fix it.

### DIFF
--- a/UnityProject/Assets/Scripts/Chat/Chat.Process.cs
+++ b/UnityProject/Assets/Scripts/Chat/Chat.Process.cs
@@ -370,15 +370,15 @@ public partial class Chat
 
 		foreach (char c in m)
 		{
-			var current = c;
+			char current = c;
 			if(Random.Range(1,6) == 1)
 			{
 				current = char.ToUpper(c);
 			}
-			song = song + current;
+			song += current;
 		}
 
-		song = song + " ♫";
+		song += " ♫";
 
 		return song;
 	}
--- a/UnityProject/Assets/Scripts/Chat/Chat.Process.cs
+++ b/UnityProject/Assets/Scripts/Chat/Chat.Process.cs
@@ -102,7 +102,7 @@ public partial class Chat
 			chatModifiers |= ChatModifier.Sing;
 		}
 		// Sing alias
-		else if (message.StartsWith("/s"))
+		else if (message.StartsWith("/s" ))
 		{
 			message = message.Substring(3);
 			chatModifiers |= ChatModifier.Sing;
--- a/UnityProject/Assets/Scripts/Chat/Chat.Process.cs
+++ b/UnityProject/Assets/Scripts/Chat/Chat.Process.cs
@@ -96,18 +96,14 @@ public partial class Chat
 			chatModifiers |= ChatModifier.Whisper;
 		}
 		// Sing
-		else if (sentByPlayer.Script.mind != null &&
-				sentByPlayer.Script.mind.occupation != null &&
-				message.StartsWith("%"))
+		else if (message.StartsWith("%"))
 		{
 			message = message.Substring(1);
 			message = Sing(message);
 			chatModifiers |= ChatModifier.Sing;
 		}
 		// Sing alias
-		else if (sentByPlayer.Script.mind != null &&
-				sentByPlayer.Script.mind.occupation != null &&
-				message.StartsWith("/s "))
+		else if (message.StartsWith("/s "))
 		{
 			message = message.Substring(3);
 			message = Sing(message);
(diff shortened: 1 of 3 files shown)


Evan-R-Rogers:
I love it

AutumnNova:
If i had to guess the ♫ character doesn't display because the font used for the chat bubbles because that font does not have any ♫ character to display, you could fix it by adding the character to the font manually or by using a different font

corp-0:
> If i had to guess the ♫ character doesn't display because the font used for the chat bubbles because that font does not have any ♫ character to display, you could fix it by adding the character to the font manually or by using a different font

Yes. The thing is when I searched for what font the chat bubble uses I found a .mat file, not the classic ttf so no frigging idea how to add characters to that.

vlandemart (approved):

UnityProject/Assets/Scripts/Chat/Chat.Process.cs
```diff
@@ -83,6 +89,24 @@ private static (string, ChatModifier) ProcessMessage(ConnectedPlayer sentByPlaye
...
+			chatModifiers |= ChatModifier.Whisper;
+		}
+		// Sing
+		else if (message.StartsWith("%"))
+		{
+			message = message.Substring(1);
+			chatModifiers |= ChatModifier.Sing;
+		}
+		// Sing alias
+		else if (message.StartsWith("/s" ))
```
daalta: I think there's a space missing here. Should this be
`else if (message.StartsWith(" /s" ))`
instead?

corp-0: it should be "/s ". I was sure I corrected that, let me see.

daalta:
Looks really good. My requests (as discussed on Discord):

* Chat bubbles and the chat log will contain different capitalization because the message is processed twice. The message should be consistent.
* Add the 🎵 suffix to the chat log only. The chat bubbles will probably have different visuals in the future, which will have the effect you desire! :) 💭🗯💬

UnityProject/Assets/Scripts/Chat/Chat.Process.cs
```diff
@@ -96,15 +96,21 @@ private static (string, ChatModifier) ProcessMessage(ConnectedPlayer sentByPlaye
...
 		{
 			message = message.Substring(1);
+			message = Sing(message);
 			chatModifiers |= ChatModifier.Sing;
 		}
 		// Sing alias
-		else if (message.StartsWith("/s " ))
+		else if (sentByPlayer.Script.mind != null &&
+				sentByPlayer.Script.mind.occupation != null &&
+				message.StartsWith("/s "))
```
daalta: These checks were needed for the clown because we look up the player's occupation - but we don't do that for singing, so there's no need to check the player's mind.

daalta (changes requested):
Cool, thanks for fixing those things! Just once more thing: The clown checks if mind is null because that's where we find their job. Singing doesn't need the mind check 😅

UnityProject/Assets/Scripts/Chat/Chat.Process.cs
corp-0: done

fomalsd:
🎵 is not displayed in chat bubbles because neither default chat bubble font (chatbubble SDF rendered from one of Ubuntu ttf fonts) nor fallback one (Liberation SDF) have emojis. Text chat however doesn't rely on TextMeshPro and uses FontAwesome font which has a ton of emojis in

corp-0:
Suspicion confirmed, then. 

I already got rid of that character in the bubble, as requested by Fax.

daalta (approved):
Alright! This looks clean and good. Let's merge it. 🙂

Here some stray thoughts:

- I would have documented `Sing` - though I suppose it's fairly obvious what it does.
- We might want to come up with a better way to check all the different aliases. We could iterate through a list of aliases for each modifier instead of having a long list of if-else blocks.
- I didn't extensively test how different modifiers interact with the new aliases and sing modifier. Some of our modifiers are mutually exclusive - chat bubbles cannot whisper and sing at the same time.
- Chat rework is on the way, so we'll see what happens 😗

corp-0:
Since it is an else if list, it takes priority on those commands who are first in the list. You can’t sing and whisper, or sing and yell, code stops looking for modifiers as soon as you started singing.

PR #3307, opened 2020-02-23
Allow Atmospheric analyzers in toolbelt

AutumnNova:
Allows storage of atmos analyzers in toolbelts

I added an itemtrait for the atmos analyzer to allow it to be whitelisted in toolbeltcapacity.asset, the second commit in this pr seems to have reordered the atmos analyzer prefab for whatever reason

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Traits/CommonTraits.cs
+++ b/UnityProject/Assets/Scripts/Items/Traits/CommonTraits.cs
@@ -34,5 +34,4 @@ public class CommonTraits : SingletonScriptableObject<CommonTraits>
 	public ItemTrait Welder;
 	public ItemTrait Shovel;
 	public ItemTrait Knife;
-	public ItemTrait AtmosAnalyser;
 }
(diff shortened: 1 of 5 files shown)


UnityProject/Assets/Scripts/Items/Traits/CommonTraits.cs
```diff
@@ -34,4 +34,5 @@ public class CommonTraits : SingletonScriptableObject<CommonTraits>
 	public ItemTrait Welder;
 	public ItemTrait Shovel;
 	public ItemTrait Knife;
+	public ItemTrait AtmosAnalyser;
```
vlandemart: Should it really be a common trait? I mean, I can imagine 10 types of knives and why this is a trait but can't think of the same for atmos analyzer

AutumnNova: well how else do you want me to do it?
the whitelist system for the belt uses itemtraits

vlandemart: Maybe we should instead have "Tool" trait which will allow items to be put in toolbelts/toolboxes and possibly something else

AutumnNova: the other reason i did it this way because the health analyzer was done the same way

PR #3309, opened 2020-02-23
ok, interaction system again, Please review this in good time xD

### LINKED ISSUE #2298: Action System B: 100

Bod9001:
Fixes #2298
look at https://github.com/unitystation/unitystation/pull/3092


Bod9001:
for teh dfuckcjkvhgdkg gvfkj.

Issue #3317, opened 2020-02-24
Staging 4039 issues


DooblyNoobly:
List them here

AutumnNova:
- banana peel is not dropped from banana, banana plant stopped dropping bananas after a few harvests (had a go at fixing it, wasnt successful however)

- shoes are not rendered on your character on client and the captains shoes icon is green

- removing a grill with wirecutters no longer drops rods, worked previously in 4036 (fix to be merged)

the below were also in versions before 4039
eating food sometimes displays "you unwillingly eat <food item>" (apparently not a bug, TIL)

- cappuccino in an edible cup leaves a cup after consuming it (fix to be merged)

- putting a cutlet in a microwave will always give 3 cooked cutlets, no mater the amount of cutlets put into the microwave

corp-0:
“You unwillingly eat” is not an issue. It displays that text when you eat without hunger.

ReiderJack:
![NoRods](https://user-images.githubusercontent.com/42844726/75145979-1a8e1f80-570b-11ea-85aa-b6b940f3f9a9.gif)
Rods are not spawned when grille is wirecutted.

ghost:
> ![NoRods](https://user-images.githubusercontent.com/42844726/75145979-1a8e1f80-570b-11ea-85aa-b6b940f3f9a9.gif)
> Rods are not spawned when grille is wirecutted.

Should also somehow get rid of the possibility to affect things diagonally when you can't go there

Mazhavil:
Can't craft metal rods

ReiderJack:
![LockerClosed2](https://user-images.githubusercontent.com/42844726/75160802-c182b380-572b-11ea-9a5c-9b7b942c5ad8.gif)

Can't open lockers

DooblyNoobly:
Server NRE spamming that is causing all of the rubber banding: 
```
NullReferenceException: Object reference not set to an instance of an object
  at SpriteFunctions.CompleteSpriteSetup (SpriteSheetAndData textureAndData) [0x0002d] in /home/unitystation/UnityProject/Assets/Scripts/Sprite Handler/SpriteFunctions.cs:28 
  at ClothingV2.SetUpSheetForClothingData (ClothingData clothingData) [0x0002a] in /home/unitystation/UnityProject/Assets/Scripts/Clothing/ClothingV2.cs:151 
[...]

PR #3289, opened 2020-02-22
Paletteable shoes

Ruzihm:
Makes paletteable shoes where the front and back colors are paletteable. re-implements black and brown shoes with them.

example

### Purpose
Progress towards #2110

### Notes
recommended test: spawn as staff assistant, search for "BlackShoes", change palette elements 0 and 1 in ItemAttributesV2, right click on ItemAttributesV2 and select "Propagate Palette Changes"

Sprites created using [this tool](https://codepen.io/ruzihm/full/ZEGWWGM)

PR #3345, opened 2020-02-25
Fixes paletteable shoes

Ruzihm:
### Purpose 
Fixes bugs from #3289 reported in #3317 (shoe & NRE bugs)

### Notes 
bug was caused by not saving necessary changes to white/brown shoe assets to disk or committing them


fomalsd:
seems to work fine, only nitpick is that spawner previews don't seem to support your system yet
img

PR #3349, opened 2020-02-25
x2 food buff because we are all starving... help

AutumnNova:
just a quick and dirty buff to food, increases all nutrition values by x2 because goddamn are we all fuckin starving


Scorr:
let them eat cake

fomalsd:
how about changing one line of hunger rate instead?
https://github.com/unitystation/unitystation/blob/193d58af492629939293de477a567cada5cd5353/UnityProject/Assets/Scripts/Health/BloodHealth/MetabolismSystem.cs#L103
set it to 1 and call it a day

Issue #3282, opened 2020-02-22
struggling out of cuffs  doesn't have a progress bar


Bod9001:
struggle out of cuffs  observe lack of progress bar

vladb99:
I would create the progress bar in `RestraintOverlay.ServerBeginUnCuffAttempt()` with
`StandardProgressAction.Create(new StandardProgressActionConfig(StandardProgressActionType.Uncuff), null)
					.ServerStartProgress(thisPlayerScript.gameObject.RegisterTile(), 30f, thisPlayerScript.gameObject);`
Now I'm not sure if this is the correct way to do it. I saw that when a player gets uncuffed by someone else, the progress bar gets created through `RequestUncuffMessage`. 
What would be the right way to do this?

Akronymus-git:
I'll take a look on friday, if you aren't working on it anymore.

PR #3350, opened 2020-02-25
Adds progress bar to struggling player

vladb99:
Fix for #3282 
Changes resist time from Handcuffs.prefab to 30.


fomalsd (changes requested):
Thanks, it works when you're standing still!
The only issue is that progress bar disappears if you change facing direction while otherwise standing still, but uncuff sequence isn't really interrupted and in 30s you get uncuffed without the progress bar.

vladb99:
Thanks for the correction. For some reason I assumed the uncuffing process would be cancelled on direction change and didn't think of proving the assumption. It should work now.

fomalsd (changes requested):
Relying on clothing icon doesn't seem like a good solution.
Existing uncuffing implementation is unorthodox, but it doesn't have to stay like that. 
I think you can safely convert that sketchy coroutine with a progress action of yours – since it was most likely implemented before we introduced progress bar system

vladb99:
I am closing the PR till I find a better solution.

PR #3371, opened 2020-02-27
bugfix and a janitorbelt change

AutumnNova:
Meat drops raw cutlets instead of cooked cutlets when chopped
janitor belt can now hold flashlight


JesterX666:
Review in progress

JesterX666 (approved):
Looks good

PR #3373, opened 2020-02-27
Adds the medical belt

Dingo-Dongler:
# The Medical Belt

### Purpose
Adds the medical belt, which can store any items with the new "medbeltable" trait.

### Notes:
Currently to only way to obtain this is to spawn as CMO, as I do not know how to map stuff in.
In future it might be useful to implement other belts with a trait system, e.g. "janibeltable"


Dingo-Dongler:
whoops, someone already did this and me aren't smart

PR #3388, opened 2020-02-27
Wrenchable basic tables

### LINKED ISSUE #3010: There should be a way of deconstructing tables
There's literally no reason for the player not be able to deconstruct table, every other code I know has it, the player can become stuck if he manage to trap itself between a wall and a table, without any way of deconstructing the wall

ReiderJack:
fix #3010

Wrenchabletables


fomalsd:
oh! that's pretty useful!
do you think you can add wood planks prefab so that wood tables aren't turned into metal?
https://tgstation13.org/wiki/Guide_to_construction#Planks

ReiderJack:
Working on it, also wrenchable chairs are coming in HOT

Ranzarth:
Are the reinforced tables harder to dissemble as well?

fomalsd:
now that wooden planks are merged, could you make wooden tables drop wood, glass tables drop glass etc according to [this list](https://tgstation13.org/wiki/Guide_to_construction#Tables.2FReinforced_Table)? and a construction recipe would be nice as well.
also if you have time, could you look at wooden items not having any names?
img

ReiderJack:
TablesDeconstrucrable

UnityProject/Assets/Scripts/Furniture/TableBuilding.cs
```diff
@@ -0,0 +1,114 @@
...
+	}
+	public void ServerPerformInteraction(HandApply interaction)
+	{
+		if (interaction.TargetObject != gameObject) return;
+		if (Validations.HasItemTrait(interaction.HandObject, CommonTraits.Instance.Wrench))
+		{
+			ToolUtils.ServerUseToolWithActionMessages(interaction, 0.5f,
+						"You start deconstructing a table frame...",
+						$"{interaction.Performer.ExpensiveName()} starts deconstructing a table frame...",
+						"You are deconstructing a glass table.",
```
fomalsd:
```suggestion
						"You are deconstructing a table frame.",
```

fomalsd:
constructing table from table frame doesn't consume any metal/glass/wood sheets (but requires clicking with one)

fomalsd:
also constructing, say, glass table requires 2x rods and 1x glass but gives 2x glass on deconstruction.. shouldn't build/deconstruct drops match?

ReiderJack:
wow, good point, it does not on TG as I remember
I will make it do

UnityProject/Assets/Scripts/Furniture/TableBuilding.cs
```diff
@@ -0,0 +1,115 @@
...
+	{
+		Spawn.ServerPrefab("Rods", gameObject.WorldPosServer(), count: 2);
+		Despawn.ServerSingle(gameObject);
+	}
+	[Server]
+	private void SpawnTable(HandApply interaction, LayerTile tableToSpawn)
+	{
+		var interactableTiles = InteractableTiles.GetAt(interaction.TargetObject.TileWorldPosition(), true);
+		Vector3Int cellPos = interactableTiles.WorldToCell(interaction.TargetObject.TileWorldPosition());
+		interaction.HandObject.GetComponent<Stackable>().ServerConsume(2);
```
fomalsd: I suspect this will give an error if you have just one sheet, but will still build a table.
I reckon you should check if stackable.Amount is sufficient before doing any build logic

ReiderJack: `public void ServerConsume(int consumed)
	{
		if (consumed > amount)
		{
			Logger.LogErrorFormat("Consumed amount {0} is greater than amount in this stack {1}, will not consume.",
				 Category.Inventory, consumed, amount);
			return;
		}
		SyncAmount(amount, amount - consumed);
		if (amount <= 0)
		{
			Despawn.ServerSingle(gameObject);
		}
	}`

ReiderJack: Is another check neccesary to put?

fomalsd (approved):
yess

PR #3406, opened 2020-02-28
Change to lowercase.

RedlineTriad:
### Purpose
Fixes casing of executable built by GitHub Actions so launcher works without any intervention.

### DIFF
--- a/Docker/docker-compose.yml
+++ b/Docker/docker-compose.yml
@@ -10,4 +10,4 @@ services:
       HUB_PASSWORD: $HUB_PASSWORD
       SERVER_NAME: Unitystation - EU02 Staging
     volumes:
-      - /etc/UnityStation/admin:/UnityStation_Data/StreamingAssets/admin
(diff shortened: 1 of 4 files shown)


Docker/docker-compose.yml
```diff
@@ -10,4 +10,4 @@ services:
       HUB_PASSWORD: $HUB_PASSWORD
       SERVER_NAME: Unitystation - EU02 Staging
     volumes:
-      - /etc/UnityStation/admin:/UnityStation_Data/StreamingAssets/admin
\ No newline at end of file
+      - /etc/UnityStation/admin:/Unitystation_Data/StreamingAssets/admin
```
fomalsd, line 13: `/etc/UnityStation/` – is this fine?

RedlineTriad: This is just the file it uses on the host system, so i would have to rename that on the GER02 server, which i am too lazy to do.

PR #3430, opened 2020-02-29
Adds clownsteps, bootsteps and barefoot sounds

corp-0:
# Pull Request Template

### Purpose
Adds checks for footwear so we can play the proper sound when walking. Takes in consideration footwear, then surface.

### Notes:
- In this implementation, footwear sounds (clown and boots) sound **instead** of the normal surface sound. I don't remember if they played together, I think they didn't.
- When barefoot, takes in consideration the surface and plays the proper barefoot sound (Currently implemented normal barefoot when walking on any WalkingSoundCategory "floor", carpet barefoot when walking over carpet and wood barefoot when walking over wood).
- Adds a new field to BasicTile so other barefoot sounds can be declared for any tile.
- Adds 2 new traits to tag items as Squeaky (like clown shows) and Boots (for anyhting you want to sound like *thump thump*.

- [x] The PR has been tested on moving / rotating / rotated-before-joining matrices (if applicable)

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Items/Clothing/Feetwear/ClownShoes.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Items/Clothing/Feetwear/ClownShoes.prefab
@@ -7,6 +7,11 @@ PrefabInstance:
   m_Modification:
     m_TransformParent: {fileID: 0}
     m_Modifications:
+    - target: {fileID: 1762207313667118248, guid: 502f496beb9034b9b9c97868e845ed7e,
+        type: 3}
+      propertyPath: m_AssetId
+      value: 
+      objectReference: {fileID: 0}
     - target: {fileID: 1864679013969520924, guid: 502f496beb9034b9b9c97868e845ed7e,
         type: 3}
       propertyPath: m_Name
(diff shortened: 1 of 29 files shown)


fomalsd:
how about not letting normal boots override floor sounds?

corp-0:
Only boots? Can do that, but I’d confirm first if that’s how it was on byond

fomalsd:
yeah, I'd check with tg first

corp-0:
Boots sound (heavy) plays instead of the normal surface sound in byond. Only exception to that rule is water and lava. That means this implementation would work for now, but it is not a good representation of how Byond did their thing. I will make some changes and use the same categories TG uses.

corp-0:
Done. With this new implementation I recreated the data structure used in TG, but it also let us fully customize sounds for each tile against type of step (maybe we want a bootOnGrass sound for reasons?). 

Each tile has the fields:
- WalkingSoundCategory
- BarefootWalkingSoundCategory
- ClawWalkingSoundCategory
- HeavyWalkingSoundCategory
- ClownWalkingSoundCategory

Fields have a default sound so you only change the one you need to be different. They all follow the classic TG behavior.

Notes:
- I didn't add claw sounds to SoundManager neither did the method to reproduce claw steps, because we don't have creatures with claws yet.

PR #3431, opened 2020-02-29
Fix for invisible clothing and removal of redundant files.

PetMudstone:
### Purpose
This PR redoes the assets for the following:
- All Bio Suits and Hoods _except_ for the generic and security variations.
- White Bomb Suit
- Security Bomb Hood

Previously, these items would not properly appear on players when they were equipped, resulting in wearers becoming naked and bald.

This PR also removes a lot of duplicate .json (and a png and two folders) files that were creating .meta files everytime Unity was opened.

### Notes:
I did not change the assets for the generic and security bomb suits and hoods because they actually appeared on players when equipped, at least on my end.

I had to delete the asset files and create new ones, since the old ones wouldn't properly save after exiting out of Unity. I could modify the old asset files, save, test them in play mode, and see that all clothing was apparently working fine. However, the issues would return after exiting and reopening Unity.

I also had to change .prefab files so that they would point to the new .asset files, since I had to delete the old ones. 


RedlineTriad:
Your previous PR/s seem to have been missing some `.meta` files causing everyone to generate them when opening the project, could you include those in your PR as well?

PetMudstone:
I'll see what I can do.

PetMudstone:
well, instead of adding .meta files, i decided to delete all the duplicate .json files that were causing the .meta files to be created.

PR #3443, opened 2020-03-01
Closet open close

Macoron:
### Purpose
Now player unlock closet just by click. Works only if player wearing right ID.
To lock closet you still need to close locker door and use your ID.

Also add cosmetic change to locker right click menu. Now it changes name depending on closet state (open/close).


fomalsd (approved):
thanks, works alright

PR #3444, opened 2020-03-01
Fix WinDoor hitbox rotation

AutumnNova:
# Pull Request Template

### Purpose
Correct the rotation of windoors to fix issues in maps like box and pogstation

### Notes:
was also going to make the windoor drop glass shards instead of metal sheets, but i couldnt figure it out


Evan-R-Rogers:
Pog doesn't use windoors but this is still great! This will fix the windoors not being able to function correctly sideways, yes?

AutumnNova:
it should do, yes

ThatDan123:
If you want the test to succeed you just need to merge develop/upstream

But it should be fine to merge as it looks like the issue failing it has been fixed and there arent any other failures.

fomalsd:
I suppose you meant Outpost instead of Pog, since outpost has obvious one at captains place

ThatDan123:
Looks like windoors might be broken again.

PR #3451, opened 2020-03-01
Fix reinforced tables

AutumnNova:
### Purpose
change reinforced tables so it no longer drops 3 plasteel and a reinforced girder


JesterX666:
Review in progress

JesterX666:
Can't seem to be able to unwrench it anymore.  At least, not on my branch.  It only puts the wrench on the table (even if i'm in harm intent)

AutumnNova:
you gotta use the welder then crowbar on it

JesterX666 (approved):

PR #3467, opened 2020-03-02
Cleaned up and added more lockable lockers

Robustin:
Attempt 2: Fixes up a lot of the locker prefabs and adds more lockable locker prefabs.

Several locker prefabs hadn't been given formal "lockable" treatment. Unfortunately @Ranzarth added lockability directly to the OutpostStation scene in #3355 instead of changing the prefabs and they weren't set up properly. I've added lockability to these prefabs, but there are probably some manually-added locker lights on OutpostStation that will need to be hunted down and removed.


Robustin:
Alright, looks good. Thanks to @Bod9001 for figuring out why the test was failing.

PR #3469, opened 2020-03-03
Usable Sprite Animation System - Simple World Icons

### LINKED ISSUE #2590: Usable Sprite Animation System - Simple World Icons B: 50
Sub-task of #2311
Refer to that ticket for guidance / overview of the plan for the sprite animation system.

Implement a usable system for defining, changing, and playing world icon animations of a particular object, which includes non-animated / static sprites, and syncing them over the network.  A world icon is the appearance of the object in the world.  

This only needs to support simple world icons, where there's only one sprite renderer (for example, air vents). Composite icons, such as plant trays, will be done in #2591. Inventory icons will be done in #2592.

Following criteria need to be met:

- [ ] Objects should be capable of defining possible animations (which might just be a single / static sprite) and switching between them dynamically by calling server-side methods.
- [ ] Each animation should have a pre-defined animation rate (how fast to switch through the sprites in the animation).
- [ ] The server should only need to tell the client which animation to play, it shouldn't tell the client about every frame of the animation.
- [ ] The design should be flexible so it can support future use cases.
- [ ] A few objects should be migrated to use this new system to prove it out.
- [ ] Dev team agrees that it meets our needs as far as usability is concerned. Expect some feedback and iteration.

You don't need to completely migrate everything to this new system, it can be set up to coexist alongside the old system so it can be migrated in a piecemeal fashion.

Refer to the suggestions in #2311 for guidance on the design / implementation.

dapocalypse:
### Purpose
Fixes #2590

### Notes:


dapocalypse:
Do Not Merge, I'm having this here so that I can get feedback on what I'm doing, its WIP.

dapocalypse:
No longer Do Not Merge

dapocalypse:
I think this fixes #2311 now that im looking at them

RedlineTriad (changes requested):
I don't think you can have constructor parameters for `MonoBehaviour`s and if you don't use `Start()` and `Update()` you should probably remove it.

UnityProject/Assets/Scripts/Sprite Handler/AnimationHandler.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class AnimationHandler : MonoBehaviour
+{
+	#region Server
```
Jibbajabbafic, line 7: This logic will probably be running on the client not the server, so I don't think this is appropriate

UnityProject/Assets/Scripts/Sprite Handler/AnimationHandler.cs
```diff
@@ -0,0 +1,48 @@
...
+	private IEnumerator coroutine;
+
+	public void Animator(AnimationInfo anim, SpriteRenderer spriteRenderer)
```
Jibbajabbafic, line 10: Change the name of this to match the class name so it's a proper constructor: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/constructors

UnityProject/Assets/Scripts/Sprite Handler/AnimationHandler.cs
```diff
@@ -0,0 +1,48 @@
...
+
+		if (len == 0)
+		{
+			spriteRenderer.sprite = anim.sprites[0];
+		}
+		else if (len > 0)
+		{
+			for (int i = 0; i > len; i++)
+			{
+				coroutine = WaitAndShow(1 / FPS, spriteRenderer, anim, i);
```
Jibbajabbafic, line 23: Coroutines are expensive, so it would be better if you created a single one which loops through all of the frames instead of creating a new coroutine per frame.

UnityProject/Assets/Scripts/Sprite Handler/AnimationHandler.cs
```diff
@@ -0,0 +1,48 @@
...
+	void Start()
+    {
+        
+    }
+
+    // Update is called once per frame
+    void Update()
+    {
+        
+    }
```
Jibbajabbafic, line 47: Don't include empty functions if they don't do anything

UnityProject/Assets/Scripts/Sprite Handler/AnimationInfo.cs
```diff
@@ -0,0 +1,97 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class AnimationInfo : MonoBehaviour
```
Jibbajabbafic, line 6: This class would be better off as a ScriptableObject. Have a look at the guidance in #2311 and try to follow it's advice.

UnityProject/Assets/Scripts/Sprite Handler/AnimationInfo.cs
```diff
@@ -0,0 +1,97 @@
...
+			handanim.sprites[insertedSprite] = sprite;
+		}
+		return handanim;
+	}
+	// Sets the ammount of frames played per second.
+	public AnimationInfo setFPS(AnimationInfo animhand, int FPS)
+	{
+		animhand.FramesPerSecond = FPS;
+		return animhand;
+	}
```
Jibbajabbafic, line 82: I'm not sure if we would need to create custom animations at runtime. They should all be setup in the editor (similar to something like ItemAttributes) and then loaded/changed out as needed. 
For example, you would make an 'on' animation and an 'off' animation in the editor, and then at runtime in the code you can switch the animation when the welder is activated or deactivated.

Jibbajabbafic (changes requested):
This doesn't fix #2311 since that's an epic which is broken down into multiple smaller issues which still need implementing. This is a good start but it still has a long way to go before it's a fully usable system. 

This isn't gonna be an easy task and it's gonna take a lot of work. I think you should start with the subtasks in #2311 if you're serious about completing the whole thing.

Edit: looks like this was supposed to fix #2590, but it still needs to handle networking and be implemented on some items

UnityProject/Assets/Scripts/Sprite Handler/AnimationHandler.cs
Jibbajabbafic: Actually nevermind this, MBs can't have constructors.

Issue #3312, opened 2020-02-23
Shotgun treating Slug as 2-shot magazine


DeeJayhX:
# Issue Template

## Description
Currently, you can only load one slug into the shotgun, fire the shotgun twice (during which, the shotgun ejects regular bullet casings instead of spent slugs), and then eject the slug like an empty magazine.

Each slug should be one shot, the slug should be ejected like a casing when fired (and not eject a bullet casing), the slug should look blown open at the plastic end when ejected, and you should be able to load at least 4 slugs into the shotgun (this is the BARE minimum, for a super short combat shotgun with just a cap in place of a magazine tube...)

The ammo capacity should probably be as follows for the current shotguns:
Sawn_Shotgun_Ballistic = 4 (1 in the chamber, 3 in the magazine tube)
Shotgun_Ballistic = 4 (1 in the chamber, 3 in the magazine tube)
Cshotgun_Ballistic = 6 (1 in the chamber, 5 in the magazine tube)

This would probably be a good easy/starter issue.

### Steps to Reproduce
Please enter the steps to reproduce the bug or behavior:  

1.  Grab a shotgun
2.  Fire it twice
3.  Notice the bullet casings hitting the floor
4.  Eject the unspent-looking, magical 2 shot slug
5.  Praise the honkmother

drneuris:
just spitballing here, but could we have the slugs be stackable, to maybe like 5-8? I feel it takes TONS of room in inventory to have a reasonable amount of bullets if you decide to go shotty

DeeJayhX:
> just spitballing here, but could we have the slugs be stackable, to maybe like 5-8? I feel it takes TONS of room in inventory to have a reasonable amount of bullets if you decide to go shotty

There's no way for that to be feasible. You couldn't realistically cram more than three 10 or 12 gauge shotgun shells into your pocket and still use them to reload reliably. At that point, what's even the point of making them stackable. Plus, this is exactly what the shotgun bandolier is for.

Dropgunner:
A normal combat shotgun holds 7+1 rounds in his chamber. 2 rounds are definitely not enough

DeeJayhX:
> A normal combat shotgun holds 7+1 rounds in his chamber. 2 rounds are definitely not enough

[...]

PR #3474, opened 2020-03-03
Internal magazines and clips(partial fix of #3312)

Tjarr737:
### Content
- Magazines now may be classified as "clips" by setting a toggle in the editor. Clips can be used to reload other magazines. The name is modeled after things like stripper clips, but also includes singular bullets and ammo boxes.
- Guns may now have internal magazines by setting a toggle in the editor. These can not be ejected and only be reloaded with clips applied to the gun.
- All shotguns are set to use internal magazines. The combat shotgun holds 7 shells, the others hold 5. (see #3312)
- Shotgun slugs and the 9mm ammo box have been made clips.

### Notes:
- MagazineBehaviour was rewritten in large parts, as the syncing previously only served to decrease client-side ammo if it was too high, so increases in content would not be properly synced. (On a sidenote, the RNG was moved into a table instead of starting it again and advancing to the wanted position.)
- Clips are currently just deleted when emptied. While this makes sense for shotgun slugs (if the gun casings are replaced at least) for stripper clips we might want some garbage and we might just want to use regular magazines for reloading other magazines.
- Weapons with internal magazines spawn their own magazine on initialization and adjust the size and content. The spawned magazine is a new prefab which is a variant of the generic magazine.
- There is no pump action in this. Shotguns are still semi-automatic.

### Self-Checks:

* [x] Code is sufficiently commented
* [x] Code is indented with tabs and not spaces
* [ ] The PR does not include any unnecessary .meta, .prefab or <b>.unity (scene) changes</b> [I think using a new internal magazine instead of a generic one and making internal magazines a new type of ammunition for mag slot purposes is reasonable]
* [x] The PR does not bring up any new compile errors
* [x] The PR has been tested in editor
* [x] Any new files are named using PascalCase (to avoid issues on case sensitive file systems)
* [x] Any new / changed components follow the [Component Development Checklist](https://github.com/unitystation/unitystation/wiki/Component-Development-Checklist)
* [ ] Any new objects / items follow the [Creating Items and Objects Guide](https://github.com/unitystation/unitystation/wiki/Creating-Items-and-Objects%3A-Now-With-Prefab-Variants) (especially concerning the use of prefab variants) [The generic internal magazine is placed right next to the generic magazine instead of in some folder like all other variants, though I think this is reasonable]
* [x] The PR has been tested in multiplayer (with 2 clients and late joiners, if applicable) [local]
* [x] The PR has been tested with round restarts.
* [x] The PR has been tested on moving / rotating / rotated-before-joining matrices (if applicable)

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Items/Magazines/Generic Internal Magazine.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Items/Magazines/Generic Internal Magazine.prefab
@@ -1,0 +1,92 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!1001 &820886777150758510
+PrefabInstance:
+  m_ObjectHideFlags: 0
+  serializedVersion: 2
+  m_Modification:
+    m_TransformParent: {fileID: 0}
+    m_Modifications:
+    - target: {fileID: 3091929310079895539, guid: 2c406df2572ceba47af8a9de4f9a50ba,
+        type: 3}
+      propertyPath: m_LocalPosition.x
+      value: 0
+      objectReference: {fileID: 0}
+    - target: {fileID: 3091929310079895539, guid: 2c406df2572ceba47af8a9de4f9a50ba,
+        type: 3}
+      propertyPath: m_LocalPosition.y
+      value: 0
+      objectReference: {fileID: 0}
+    - target: {fileID: 3091929310079895539, guid: 2c406df2572ceba47af8a9de4f9a50ba,
+        type: 3}
+      propertyPath: m_LocalPosition.z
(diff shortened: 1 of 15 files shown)


DooblyNoobly:
Might throw this one to staging for test

DooblyNoobly:
This accidentally made it into production and players can no longer reload their weapons. Can you look into it please. Also are you on discord?

AutumnNova:
reloading works fine

Tjarr737:
For me reloading works on build 4052 with local host, 30ms ping (EU) and 300ms ping (AUS). Both normal guns and shotguns. It might be specific guns or some other situation I did not encounter.
Edit: Can't find any faults with the prefabs either.

PR #3483, opened 2020-03-04
Fix all of the holes in the floor below windows in outpost station

AutumnNova:
### Purpose
Currently most if not all windows on outpost station have holes that lead out into space below them, this fixes that issue

### Notes:
Ive only fixed the holes on the main station

### Please make sure you have followed the self checks below before submitting a PR:

- Code is sufficiently commented
- Code is indented with tabs and not spaces
- The PR does not include any unnecessary .meta, .prefab or <b>.unity (scene) changes</b>
- The PR does not bring up any new compile errors
- The PR has been tested in editor
- Any new files are named using PascalCase (to avoid issues on case sensitive file systems)
- Any new / changed components follow the [Component Development Checklist](https://github.com/unitystation/unitystation/wiki/Component-Development-Checklist)
- Any new objects / items follow the [Creating Items and Objects Guide](https://github.com/unitystation/unitystation/wiki/Creating-Items-and-Objects%3A-Now-With-Prefab-Variants) (especially concerning the use of prefab variants)
- The PR has been tested in multiplayer (with 2 clients and late joiners, if applicable)
- The PR has been tested with round restarts.
- The PR has been tested on moving / rotating / rotated-before-joining matrices (if applicable)

PR #3491, opened 2020-03-05
Fixes Missing Floors on Outpost

Ranzarth:
### Purpose

- Fixes missing floors on Syndi and escape shuttle


DooblyNoobly:
Already fixed in #3483 ?

Ranzarth:
This is for the syndi and escape shuttle, that PR only fixed the main station

PR #3492, opened 2020-03-05
Adds Cargonia game mode

DooblyNoobly:
Adds a new game mode in rotation called Cargonia.

Implied goal:
- Cargo is to secede from the rest of the station
Antag Objectives:
- 50% of all cargonians need to be alive by round end
- Only cargonians are allowed to escape alive on the shuttle

image


Jibbajabbafic:
Looks like you deleted the trigger nuke objective

DooblyNoobly:
> 
> 
> Looks like you deleted the trigger nuke objective

fixed, also set detonated to true on the server so hopefully green text works again

Jibbajabbafic:
Nice

PR #3511, opened 2020-03-07
Pog-Station spawnpoint update

Evan-R-Rogers:
Research and RD spawnpoints are now in the research sector, Chemists no longer spawn in the wall.


AutumnNova:
scuse me, you what now

PR #3539, opened 2020-03-08
Added chat typing icon

Macoron:
### Purpose
Adds chat typing icon that indicates when other player is typing. This will make IC conversation much more convenient. OOC or non verbal chats (like binary) doesn't show typing icon.
Re-exported all chat sprite animations from .DMI. They were missing animation frames.

### Notes:
Chat typing status sends by messages only to nearby players.
Classic chat icons (whit bubbles with expression) are not function prior this PR. Chat toggle doesn't do anything. Don't know is that intentional or bug.

Added new field UprightSprites to ignore "extra" rotation. Sprites added there will be always up no matter what. Tested it with stunned player and on rotation matrices.


daalta:
> Classic chat icons (whit bubbles with expression) are not function prior this PR.

I think those might have been intentionally disabled recently.

This is a fun addition! Won't be able to review it myself, but after taking a cursory glance at your code I'd say it's good. Hope performance is OK, and I personally wouldn't if you had more documentation :)

PR #3544, opened 2020-03-09
Improve fire performance (and various fire-related atmos fixes)

### LINKED ISSUE #2917: Plasma fires make server lag hard when there are many players connected
Players experience rubberbanding and can't do much when there's a large fire spreading throughout the station. Server lag is quite possibly caused by vast amount of update messages sent to all players for each fire tile

chairbender:
### Purpose

Should fix #2917 to an extent. There is still room for improvement but this should be pretty significant, reducing frequency and amount of GC coming from ReactionManager.

Went through and did whatever I could to reduce GC caused by fires, also fixed some weird atmos bugs related to fires, and issues where fires / plasma overlays were conflicting with each other.

There is a bit of jank specifically with how tile "overlays" and the stuff in the effects layer (plasma and fire) is working right now. It should mostly work properly, but should be refactored and improved so it's easier to work on that stuff. I have called out some of this logic, search "OVERLAYS" in todo comments.

1. Ashes are now tiles, so minimal GC associated with spawning them. These are using the new "UpdateOverlay" and "RemoveOverlay" methods in tile change manager which simply puts them at z level -1. This is not a great long term solution because we will most likely have more than one overlay on tiles.
2. Adds some more logic for tile overlays to make the ash stuff possible.
3. Fires and Plasma now go in the effects layer at z level -3 and -2. This prevents them from having to keep checking for each other and replacing each other. The way it is implemented is not that easy to make sense of so this should be refactored in the future.
4. Overall, less object instantiation is done when destroying things during fires. My justification is "it's just melting completely", but the underlying reason is that object instantiation creates significant GC and fixing this would require pre-allocating pools for objects that spawn a lot during fires which was a bit much for this PR.
5. Adds more profiling tags to make it easier to see where GC is coming from related to fires.
6. Fixed an issue where fires would actually ADD to the gasmix rather than subtracting from it (consuming) when burning.


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/TileChangeManager.cs
```diff
@@ -101,15 +125,24 @@ public void RemoveTile( Vector3Int cellPosition )
 		}
 	}
 
+	/// <summary>
+	/// Removes the tile at the indicated cell position. By default all tiles in that layer
+	/// at all z levels will be removed. If removeAll is true, then only the tile at cellPosition.z will
```
fomalsd, line 130: should be
>If removeAll is **false**, then only the tile at cellPosition.z will be removed.

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Matrix.cs
```diff
@@ -187,6 +192,22 @@ public bool IsFloatingAt(GameObject[] context, Vector3Int position, bool isServe
...
+	/// GetComponent so there's no GC. The OTHER benefit is that normally iterating through these
+	/// would throw an exception if the RegisterTiles at this position were modified, such as
+	/// being destroyed are created. This method uses a locking mechanism to avoid
+	/// such issues.
+	/// </summary>
+	/// <param name="localPosition"></param>
+	/// <returns></returns>
+	public void ForEachRegisterTileSafe(IRegisterTileAction action, Vector3Int localPosition, bool isServer)
+	{
+		(isServer ? ServerObjects : ClientObjects).ForEachSafe(action, localPosition);
```
fomalsd, line 208: very cool! we should use it more

PR #3561, opened 2020-03-10
Show the reason why a person cannot be cloned

joelhed:
# Show the reason why a person cannot be cloned

### Purpose
At the time, when you try to clone someone and it fails, there's no way to see why it failed. The cloning console simply displays the message `Initialisation failure.` This can lead to some confusion, since it is not obvious what initialisation failure means.

This PR sets the status display to something readable when cloning fails, such as `Spirit cannot be found` for when the player has disconnected, or `Person is still alive` when, you guessed it, the person is still alive.

### Notes:
For cloning to work as it does today, you kind of have to admit the existence of spirits, or something similar, in character. There might be some other more appropriate name, like Mind, as the class is named, or soul, or whatever. It's arbitrary really. But you can't avoid the fact that sometimes, the cloning machine mysteriously fails to work. If you were to build a cloner today that could perfectly recreate a scanned dead body and make it alive again, then it would have to be possible to create many copies of a single person. Since UnityStation will only allow one alive clone per physical player, you have to come up with something that can adequately explain why you can't clone the clown 50 times. Spirits fill that purpose at least somewhat.

I also simplified some logic in the code I touched. I couldn't help it!

  *No new items or objects!*

  *Tested with host-client and client*

### DIFF
--- a/UnityProject/Assets/Scripts/Medical/ClonableStatus.cs
+++ b/UnityProject/Assets/Scripts/Medical/ClonableStatus.cs
@@ -1,0 +1,13 @@
+/// <summary>
+/// This is used to represent the status of a player's Mind object when it comes to
+/// cloning. The only item here that allows for cloning is Cloneable, the others are
+/// just there to inform why the player can't be cloned.
+/// </summary>
+public enum CloneableStatus
+{
+	Cloneable,
+	OldRecord,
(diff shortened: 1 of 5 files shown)


fomalsd (approved):
looks good, thanks!

PR #3573, opened 2020-03-11
Make Emote, Whisper and Sing commands ignorecase.

craftxbox:
makes text commands ignore case so that `/ME` and such will still work.


Evan-R-Rogers:
Hurrah!

PR #3580, opened 2020-03-11
Fix substring for emote commands

craftxbox:
adds a check to set the start length of the substring if its using a /command, because /command is longer than the single character way

### DIFF
--- a/UnityProject/Assets/Scripts/Chat/Chat.Process.cs
+++ b/UnityProject/Assets/Scripts/Chat/Chat.Process.cs
@@ -87,21 +87,21 @@ public partial class Chat
 		// Emote
 		if (message.StartsWith("*") || message.StartsWith("/me ",true,CultureInfo.CurrentCulture))
 		{
-			message = message.Replace("/me","") // note that there is no space here as compared to the above if
-			message = message.Substring(1);     // so that this substring can properly cut off both * and the space
+			message = message.Replace("/me",""); // note that there is no space here as compared to the above if
+			message = message.Substring(1);      // so that this substring can properly cut off both * and the space
 			chatModifiers |= ChatModifier.Emote;
 		}
 		// Whisper
 		else if (message.StartsWith("#") || message.StartsWith("/w ",true,CultureInfo.CurrentCulture))
 		{
-			message = message.Replace("/w","")
+			message = message.Replace("/w","");
 			message = message.Substring(1); 
 			chatModifiers |= ChatModifier.Whisper;
 		}
 		// Sing
 		else if (message.StartsWith("%") || message.StartsWith("/s ",true,CultureInfo.CurrentCulture))
 		{
-			message = message.Replace("/s","")
+			message = message.Replace("/s","");
 			message = message.Substring(1); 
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Chat/Chat.Process.cs
```diff
@@ -87,19 +87,19 @@ private static (string, ChatModifier) ProcessMessage(ConnectedPlayer sentByPlaye
 		// Emote
 		if (message.StartsWith("*") || message.StartsWith("/me ",true,CultureInfo.CurrentCulture))
 		{
-			message = message.Substring(1);
+			message = message.Substring(message.StartsWith("/me ") ? 4 : 1);
```
RedlineTriad: Maybe use `message.Replace("/me", "")` or `message.Remove(3)` instead, so the function is more obvious.

UnityProject/Assets/Scripts/Chat/Chat.Process.cs
```diff
@@ -87,19 +87,19 @@ private static (string, ChatModifier) ProcessMessage(ConnectedPlayer sentByPlaye
...
 			chatModifiers |= ChatModifier.Emote;
 		}
 		// Whisper
 		else if (message.StartsWith("#") || message.StartsWith("/w ",true,CultureInfo.CurrentCulture))
 		{
-			message = message.Substring(1);
+			message = message.Substring(message.StartsWith("/w ") ? 3 : 1);
```
RedlineTriad: Same here.

UnityProject/Assets/Scripts/Chat/Chat.Process.cs
```diff
@@ -87,19 +87,19 @@ private static (string, ChatModifier) ProcessMessage(ConnectedPlayer sentByPlaye
...
 			chatModifiers |= ChatModifier.Whisper;
 		}
 		// Sing
 		else if (message.StartsWith("%") || message.StartsWith("/s ",true,CultureInfo.CurrentCulture))
 		{
-			message = message.Substring(1);
+			message = message.Substring(message.StartsWith("/s ") ? 3 : 1);
```
RedlineTriad: Same here.

UnityProject/Assets/Scripts/Chat/Chat.Process.cs
```diff
@@ -87,19 +87,19 @@ private static (string, ChatModifier) ProcessMessage(ConnectedPlayer sentByPlaye
 		// Emote
 		if (message.StartsWith("*") || message.StartsWith("/me ",true,CultureInfo.CurrentCulture))
```
fomalsd, line 88: imo these checks should be against `message.ToLower()` since someone could write `/Me` and these would fail

craftxbox: No? The second argument of startswith is an ignorecase flag

fomalsd: oh right.. that's not obvious unless IDE gives you a hint

Issue #3026, opened 2020-02-10
Add a basic ahelp system B: 200


DooblyNoobly:
At the moment its really hard for players to contact admins. What we need is a basic way for players to contact admins. 

I propose the /ahelp tag in chat input. If they type /ahelp <message> then it gets sent to the screen of all admins (a pop up like the restart vote pop up) and as soon as an admin clicks on it to reply it disappears off the screen of all the other admins. 

Criteria:
- Add the /ahelp functionality to chat
- Create a new ahelp network message that is sent to the server and then a message is sent to all of the admins
- When the message arrives on the admins client they get a pop up somewhere on screen that they can reply too. 
- Replies will just use the Admin bwoink system (could even auto open it to the admin bwoink page with the players name already selected) 
- If an admin claims the ahelp then send a message that makes it disappear from the screen of the other admins
- in addition to this when an admin bwoinks someone, a dedicated chat dialogue is opened up on their screen. A lot like what admins see in the bwoink menu, but limited to just speaking with that specific admin that initiated the conversation.
- Also add a way for players to send prayers to admins

Contributor will receive a bounty reward of 200 usd for a successful merge of a solution that satisfies the criteria above

The-Scout:
also reliability of ahelps needs to be improved i have noticed sometimes they are not sent or replies are not sent.

hecksadecimal:
Chat is still going to be chaotic in highpop servers. Even after proximity is fixed. People will start using comms more frequently, and admin bwoinks may end up getting drowned out in our current chat implementation. 

So in addition to this I would also propose that when an admin bwoinks someone, a dedicated chat dialogue is opened up on their screen. A lot like what admins see in the bwoink menu, but limited to just speaking with that specific admin that initiated the conversation.

DooblyNoobly:
not a bad idea actually

DooblyNoobly:
Added it to the criteria and increased bounty to 150

fomalsd:
what do you think about ahelp discord integration? ss14 has OOC routed to their discord and it's pretty convenient

hecksadecimal:
That sounds really nice.

Jibbajabbafic:
I think a dedicated menu button, or hotkey like F1 would be good to have too.

DooblyNoobly:
changing from b150 to b200. Bigger job then I thought

DooblyNoobly:
> 
> 
[...]

PR #3592, opened 2020-03-14
Adds new ChatScroll prefab and the first working version of A help chat

### LINKED ISSUE #3582: Create a reusable Chat Scroll Rect GUI element B: 150
## Description:

It has come to my attention that we will need to build the MVVM based chat system now as it will need to be used for 4 other chat windows that are being implemented for admins and players. This chat system will be based on the ViewModel controlling the scroll rect and the child elements in an easy to use ChatScrollRect prefab that can be instantiated in the scene view the tool bar. 

The view prefabs will be spawned from a local pool and shown only when they are inside the scroll rect. Once they go outside of the view area they will reset themselves and return to the pool as another one will be taken from the pool and given the data to be displayed in the view for the new entry that should be showing at the bottom. 

This should allow the chat windows to be performant while also storing as many chat logs as we want.

## Criteria:

- Create a ChatScrollRect element that inherits from ScrollRect and treat it as the VM in this MVVM scenario
- Create a base data object that can be expanded to hold all types of data
- Create the ChatEntryView that will inherit Monobehaviour and control the UI elements
- Create easy to use methods to load data from a model (different types of models?) into a ChatScrollRect
- Create a default prefab for the ChatScrollRect and add it to the Editor UI menu so it can be spawned via a button click
- Test it on the admin chat windows. We will implement it in the main chat system when we do ChatSystem V2
- Test in MP with 2 clients

DooblyNoobly:
fixes #3582 
- You can now create a new chat scroll complete with input via the create menu and toolbar
- Added an example scene for chat scroll
- Added Subscriber events to chat scroll for easy use

Partial of #3026. This adds the new chat scroll to Admin Player Chat. Also spawns a new tool bar for player chat. Merging this so we can test on staging as it is in a basic functional state. Still need to finish of Admin chat, notifications and Prayer stream


DooblyNoobly:
merging now so I can test on staging. Don't have much time to wait for 2 tests

PR #3595, opened 2020-03-15
Armory Renovations to Outpost and General Fixes

Ranzarth:
### Purpose

- Gets rid of armory crates and puts in racks instead
- Adds bomb suits, bullet proof armor, and riot armor to Armory

DNM, will be including a rewiring of some parts of the station, and staging fixes


Ranzarth:
Ready to Merge

PR #3599, opened 2020-03-15
Editor and runtime performance fixes

Macoron:
### Purpose
Fixes some performance problems in editor, especially when working with enabled gismos (from 10 fps to 50 fps). 

### Notes:

- Removed blue icon gismo from NetworkedIdentity
- Electrical cables particle systems (sparks, smoke) are now using lazy initialization on runtime and doesn't clutter scene. Consider using pooling manager for fx in the future.
- All doors are now using sound manager for open/close sound effect. They don't contain audio source in prefab anymore.
- Fixed InvalidArgument exception for UprightSprites. Usually was thrown on scene opening.
- Fixed big performance bottleneck for MetaDataView. Before it was checking if Matrix Check enabled for each cell. Now it first get enabled Checks and then iterating all cells.


fomalsd:
the biggest performance improvement for me was commenting out Update loop in https://github.com/unitystation/unitystation/blob/bcb97e6286e45aaefbe928dbbfdcc5918d8603f8/UnityProject/Assets/Scripts/Objects/UprightSprites.cs#L124-L135
before vs after:
img

could you include this one in your changes as well?

Macoron:
I think it's made on purpose to show correct direction  of the sprite when you editing it in scene. Maybe change it to [OnValidate](https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnValidate.html)? 
OnValidate can miss changes sometimes, but it's usually stable.

fomalsd:
yeah, that might work

DooblyNoobly:
merge conflicts. Macoron can you resolve it and ping us for a speedy merge

Macoron:
Should be fixed now

PR #3633, opened 2020-03-18
Add Basic Seed Extractor and Plant Bag

DVanwart:
# Add Basic Seed Extractor and Plant Bag
### Purpose
- Adds Seed Extractor with GUI that shows seed packets that it has processed as well as their stats

- Adds very basic version the the botanists satchel/plant bag
    - Added to botanist locker populator
    - Currently only works on the floor for picking up and dumping

### Notes:
The seed extractor needs power for it to begin processing it's inventory of produce so an APC reference needs to be added to it in the scene, same as consoles

- [x] The PR has been tested on moving / rotating / rotated-before-joining matrices (if applicable)


Bod9001:
looks good, have a play around on staging just make sure

DVanwart:
@Bod9001 Seems good enough to keep, the UI in the seed extractor freaks out a little as the item count gets higher but no knock-on effects outside of the seed extractor or NREs or anything. 

I'll see If I can figure out the problem but the vendor UI also has the same glitchyness, just less noticeable with the more static item list so I'm guessing they share a root cause

PR #3647, opened 2020-03-19
fix(meta): god damn leftover meta files

Jibbajabbafic:
See title.


Jibbajabbafic:
Will self-merge if it passes checks to stop everyone having to put up with them

PR #3649, opened 2020-03-20
7 new desk items.

Evan-R-Rogers:
Sadly, all items are only cosmetic as of now. Placeholder if you would.

Objects can be found under assets/resources/prefabs/objects/misc

 Pictured below:
desk_items


Evan-R-Rogers:
Lamps do not function, a new script will need to be added it seems. This will be my next project.

PR #3660, opened 2020-03-21
Late Joiners Spawn At Arrivals After 2 Min

### LINKED ISSUE #3259: Make all late joiners spawn on arrivals shuttle
As per title.
If you didn't join within 1-2 mins after round start you should spawn in arrivals shuttle instead of your usual spawn.
Make sure it doesn't interfere with nuke ops spawn point!

DVanwart:
# Late Joiners Spawn At Arrivals

### Purpose
All players, excluding special jobs, will use the assistant spawn location after 2 minutes station time.
Resolves #3259

### Notes:
- Gets current round time from GameManager.stationTime
- In PlayerSpawn.NEVER_SPAWN_ARRIVALS_JOBS can be updated with any new jobs that should ignore this rule, current values are below
```
{
	JobType.AI, 
	JobType.SYNDICATE
});
```
- ~~OK BUT WHAT IF YOU HAD TO WHEEL THE AI INTO THE STATION ON LATE JOIN~~
- PlayerSpawn.ARRIVALS_SPAWN_TIME holds the time for when this rule starts applying
- Also small NRE protection for seed extractor if power state change call happens before awake is called.

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/SceneConstruction/SpawnPoints - V2.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/SceneConstruction/SpawnPoints - V2.prefab
@@ -44,6 +44,50 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   Department: 24
+--- !u!1 &2483236536562526403
+GameObject:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  serializedVersion: 6
+  m_Component:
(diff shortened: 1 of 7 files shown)


DVanwart:
Just noticed POG doesn't have assistant's spawn on the shuttle so I'm going to rework this to check for a shuttle spawn and just use assistant as a fallback

PetMudstone:
I would like to note that SeedExtractor.cs is included in this PR and has a single character ("?" to be exact) added to a line. Is this intentional?

DVanwart:
@PetMudstone yep! Was getting a NRE when clients joined and the power loaded before the seed extractor did
> Also small NRE protection for seed extractor if power state change call happens before awake is called.

DVanwart:
If tests pass this should be good to merge

PR #3666, opened 2020-03-22
Whitelist

### LINKED ISSUE #2773: automatic admin if running and hosting in editor
# Issue Template
I've never gotten the admin to work on the editor, it would be very convenient for it just to automatically start with admin

### LINKED ISSUE #3614: Feature: Whitelisting
## Description

A config option and whitelist.txt similar in usage to admins.txt to allow a 'panic bunker'-like setup, where people must first be whitelisted before they're able to connect to the server. Being able to change this config through RCON while the a round is ongoing would also be nice. 

Bonus points if it were to allow for different modes. 
Examples:
Admins only - Only IDs defined in admins.txt can connect. 
Whitelist - IDs in admins.txt and whitelist.txt can connect. 
Off - No ID checking. 

By allowing only administrators on the server without having to completely change the current whitelist.txt, it would allow admins to set up events prior to other players joining for the round.

ThatDan123:
Adds the ability to have a whitelist.

Admins are automatically whitelisted

Server now gets admin.

Ready to test on staging.

known issue: The box telling you why you cant join has decided not to show in my testing anymore, might work on proper server though.

fixes #3614 

fixes #2773 (Just putting this here to close the issue, as im pretty sure editor already has admin)


UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
```diff
@@ -211,7 +246,23 @@ private async Task<bool> CheckUserState(string userid, string token, ConnectedPl
 			return false;
 		}
 
+		//whitelist checking:
+		//Checks whether the userid is in either the Admins or whitelist or is server AND that the whitelist file has something in it.
+		//Whitelist only activates if whitelist is populated.
+
+		var lines = File.ReadAllLines(whiteListPath);
+
+		if ((!whiteListUsers.Contains(userid) | !adminUsers.Contains(userid)) && CustomNetworkManager.Instance._isServer == false && lines.Length > 0)
```
DooblyNoobly: `if ((!whiteListUsers.Contains(userid) | !adminUsers.Contains(userid)) && CustomNetworkManager.Instance._isServer == false && lines.Length > 0)
`
So no need to worry about this check `CustomNetworkManager.Instance._isServer ` as its already the server you are running the code on. 

change this `(!whiteListUsers.Contains(userid) | !adminUsers.Contains(userid))` too

`(!whiteListUsers.Contains(userid) && !adminUsers.Contains(userid)` So if you are on the whitelist or if you are on the admin list it won't trigger the kick

DooblyNoobly (changes requested):

ThatDan123:
Ready to test on staging.

PR #3688, opened 2020-03-23
Table climb fixes

aleios:
Adds essential checks to table climbing:

- Checks the distance between the interacting player and the target. This is so you can't run 9000 tiles away and slingshot someone onto a table.

- Performs a collision check to make sure that the tile you're climbing onto is actually passable. Necessary so that you can't climb onto a table that has been blocked by shutters.

IsPassableAt was modified to take optional parameters for exclusion of specific layers and/or tile types from the collision checks. In most cases these two parameters are going to be null and thus skipped over. 

Assuming the C# runtime adequately flags the branches as likely false there shouldn't be any visible change in the performance of these functions.

**Notes:** 
Will still need some extensive tests done. There's only so many cases that can be tested locally.
Notably need to figure out if there are any cases where the distance between source and target are not enough to prevent magically teleporting people.


ThatDan123:
Does this work with windoors?

PR #3691, opened 2020-03-23
27 New Hostile Mobs

ThatDan123:
27 new hostile mobs

Also reorganised NPC folder.

Todo in the future: 
Make their AI more specific.
         ~-Eg statue AI, where it shouldnt be able to move if seen unless in darkness.~
          Not possible :(, But alternative is only move if player is facing away.


ThatDan123:
Pls merge

waffielz:
nice

PR #3697, opened 2020-03-23
Chair fixes.

Evan-R-Rogers:
Due to the 'Unshootable Machines' layer causing issue with physics 2D, the shuttle chairs for now have been changed to the 'Machines' layers. Changed the object types for other new seats to better fit what they are.


DooblyNoobly (changes requested):
The actual layer should be fixed via the physics2d matrix instead. Untick all layers that it does not need to interact with

Evan-R-Rogers:
Changes Psychics2D so that the unshootable layer doesn't interact with fucking everything. All chairs/seats are now under the 'Unshootable Machines' layer.

Tested on headless on station and on emergency shuttle. Seems to work just fine.

PR #3701, opened 2020-03-24
Fix security records console entry editing

joelhed:
### Purpose

Currently, when you try to edit an entry in the security records console, the popup window that appears simply contains what was previously written in that same popup, regardless of what field you're editing. With this PR, the popup will instead contain the current value of the field being edited.

### Notes

I noticed some wonkiness to the records console, namely that I sometimes couldn't move after clicking a field for editing. I couldn't reproduce it though, and it's unlikely that this PR was the one to introduce it, so I'm not gonna do anything about it right now.

I also noticed that when two people are looking at the same record at different consoles, and one of them edits a field, it does not trigger an update on the other person's console. Since I'm a networking newb, I won't try to change anything about that.

-  The PR has been tested with round restarts.
-  The PR has been tested on moving / rotating / rotated-before-joining matrices (if applicable)


DooblyNoobly:
Testing on staging. Thanks

PR #3716, opened 2020-03-25
Traitor Populator

Ranzarth:
### Purpose

Adds a traitor populator based on server population.

### Notes:
Probably not the best way implement this, but see how it goes in the review.


DooblyNoobly:
see how it goes in deployment instead

PR #3733, opened 2020-03-26
Update SpawnPoints

ThatDan123:
Adds missing spawnpoints to the prefab

Current Maps Will Need To **Remove** And **Move** SpawnPoints Around.

DONT build servers with this merged until the live maps have been updated

Renamed Spawnpoints so they dont have the same numbers.

Also ordered them by department in the prefab.


ThatDan123:
DO NOT MERGE unless all maps are ready to be updated

PR #3744, opened 2020-03-27
Change the notes ui so it isnt entirely white and color the teleport scrollbar because i forgot

AutumnNova:
h


DooblyNoobly:
I'm making changes to ghost hud as well to fix teleport

PR #3750, opened 2020-03-27
Mob Animator

ThatDan123:
Mob Animator

-Allows as many different animations as you want.

Simple Animator, has only on animation direction.

Complex Animator, automatically changes animation if mob direction is changed

And some mobs which use the new system.


DooblyNoobly:
Not bad, lots of room for improvement in the code tho. You could remove the coroutines and put the logic in an UpdateMe loop that is registered to UpdateManager to reduce cpu load. You could also remove the Animation Loop and just have a loop flag on the MobAnimation component.

PR #3761, opened 2020-03-28
Traitor Values adjustment

Ranzarth:
Adjust the amount of traitors that spawn for balancing
Also readds a check for the second return to stop duplications

### DIFF
--- a/UnityProject/Assets/Scripts/GameModes/Traitor.cs
+++ b/UnityProject/Assets/Scripts/GameModes/Traitor.cs
@@ -24,6 +24,10 @@ public class Traitor : GameMode
 		{
 			TraitorAmount = 4;
 		}
+		else if (PlayerList.Instance.InGamePlayers.Count >= 24)
+		{
+			TraitorAmount = 4;
+		}
 		else if (PlayerList.Instance.InGamePlayers.Count >= 15)
 		{
 			TraitorAmount = 2;
--- a/UnityProject/Assets/Scripts/GameModes/Traitor.cs
+++ b/UnityProject/Assets/Scripts/GameModes/Traitor.cs
@@ -1,12 +1,17 @@
 using UnityEngine;
 using Antagonists;
 using System.Collections.Generic;
+using System;
 
 [CreateAssetMenu(menuName="ScriptableObjects/GameModes/Traitor")]
 public class Traitor : GameMode
 {
-	
-	private float TraitorAmount = 0;
+
+	[Tooltip("Ratio of traitors to player count. A value of 0.2 means there would be " +
+			 "2 traitors when there are 10 players.")]
+	[Range(0, 1)]
+	[SerializeField]
+	private float TraitorRatio;
 
 	/// <summary>
 	/// Set up the station for the game mode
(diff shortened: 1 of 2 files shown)


daalta:
Looks OK to merge! Having 4 traitors at ten players seems silly, your numbers are much better.

50+ players: 5 traitors
30+ players: 4 traitors
15+ players: 2 traitor
00+ players: 1 traitor

What do you think of adding a player count at which 3 traitors spawn? At, let's say, 23 players?

daalta:
I've gone ahead and simplified the check. https://github.com/unitystation/unitystation/pull/3761/commits/6e66e4c933ddc3ac6df0d88515761cf49a13f05b

corp-0:
make it a switch! We don't want our code to look like Yandere Simulator

daalta:
@corp-0 A switch would be better... but eh, I think it's not worth the effort. Code won't be any shorter, and the variable is only referenced thrice. Good enough! :)

daalta:
Added 3 traitors when there are at least 24 players, discussed it with @Ranzarth .

daalta:
This doesn't really make sense to me - why is this a for loop in the first place? https://github.com/unitystation/unitystation/pull/3761/commits/c3f5575d843ac551bb959c5fc57e087d739d602d

~~Either way, I think this can be merged.~~ Uh, nevermind

`AntagManager.Instance.AntagCount == TraitorAmount`

This part eludes me. Should is be less / more or equal than?

Maybe someone else can pick up the review o.o

UnityProject/Assets/Scripts/GameModes/Traitor.cs
```diff
@@ -68,11 +76,12 @@ public override void StartRound()
...
 	{
 
-		for (int i = 0; i < TraitorAmount - 1; i++) {
+		for (int i = 1; i < TraitorAmount; i++) {
 			return !LoyalImplanted.Contains(spawnRequest.RequestedOccupation.JobType)
 					&& PlayerList.Instance.InGamePlayers.Count > 0;
 		}
 		return !LoyalImplanted.Contains(spawnRequest.RequestedOccupation.JobType)
+					&& AntagManager.Instance.AntagCount == 0
 					&& PlayerList.Instance.InGamePlayers.Count > 0;
```
daalta: Why the for loop? Why not replace
```
for (int i = 1; i < TraitorAmount; i++) {
	return !LoyalImplanted.Contains(spawnRequest.RequestedOccupation.JobType)
			&& PlayerList.Instance.InGamePlayers.Count > 0;
}
return !LoyalImplanted.Contains(spawnRequest.RequestedOccupation.JobType)
			&& AntagManager.Instance.AntagCount == 0
			&& PlayerList.Instance.InGamePlayers.Count > 0;
```
with
```
return !LoyalImplanted.Contains(spawnRequest.RequestedOccupation.JobType)
			&& AntagManager.Instance.AntagCount <= TraitorAmount
			&& PlayerList.Instance.InGamePlayers.Count > 0;
```

Ranzarth: Hmm I didn't really take that into account

Jibbajabbafic (changes requested):
You can remove TraitorAmount all together and replace it with a ratio instead. See `NukeOps.cs` for an example.

Jibbajabbafic:
Nice, the ratio might need adjusting to find a good balance but we can figure that out later on

Jibbajabbafic (approved):

Ranzarth:
Indeed, I just set it to the same value as nukeops for now

Jibbajabbafic:
I'll let @daalta decide if he wants to merge this as is

PR #3767, opened 2020-03-29
Fix Input Fields and Add a limit to cap announcements

### LINKED ISSUE #3710: Exploits on the captains console
No cooldown on threat level changes.
You can make blank messages using blank unicode characters (And a few more it seems ,like ᚛                 ᚜ )
https://www.compart.com/en/unicode/U+2800
(and Bod9001 i want the streetcred don't steal my space roubles)

AutumnNova:
Change Input fields for Ghost Teleport Menu, Admin Sound Player and Notes Menu
Adds a 200 character limit to caps announcements (Fixes #3710)
Added an observer message for digging
Changed the name of a grill in the construction menu from grilleobject to grill
Fixed the construction of a wall displaying the message for a false wall being constructed
Reworded several construction/deconstruction messages


DooblyNoobly:
legend thank you!!!

PR #3768, opened 2020-03-29
CentComm and shuttle sounds and adds nuke to Outpost

Ranzarth:
### Purpose

Adds CentComm and the Escape Shuttle sounds. The Escape Shuttle will now travel to a far away CentComm and dock, it will then end the round after 1 minute.
Adds nuke and nukedisk for the Syndicate Gamemode rework to Outpost Scene.

### Notes:
Pogstation needs to be updated with this before being put on the main servers.
~~CentComm is largely barebones and without lighting atm~~


Ranzarth:
Should be good for review

PR #3779, opened 2020-03-29
Antagonists | Proper Nuclear Operatives

### LINKED ISSUE #3173: Proper nuclear operatives rewrite B: 80
Right now nuke ops is a very basic game-mode and some things need major changes.

These changes would allow nuke operatives rounds to be a lot more dynamic and IMO more fun. this changes the dynamic from nukies rush sec and put in code to nukies have to get the disk, put the nuke in place, arm it, and get out of dodge before it blows.  while the station crew defend the captain and in turn prevent and keep the ops from getting said disk.

[Here is a tgstation wiki link for nuclear operatives](https://tgstation13.org/wiki/Nuclear_Operative)

1.  The captains quarters should have a nuclear authentication disk which is required to be placed inside of the nuke before it can detonate. Once armed the disk is able to be taken out of the nuke with the countdown still going, the nuke also cannot be disarmed without the disk inside.

2. the nuke should be in the nukies shuttle behind a glass door, that stays anchored until the nukies insert the disk and de-anchor it.

3. The nuke should be bolted and unmovable until the nuke disk is inserted and should have a button to anchor it again in the GUI.

4. The nuke should have a timer with a minimum of 3 minutes to make sure people have enough time to potentially counterattack the nuke ops and get the disk back, and for the nuke ops to be able to get a safe distance. with the ops able to choose to make it longer by setting a time after the code is set

5. If the nuke ops die when the nuke goes off it counts as a partial failure.

6. A nuke disk tracker that spawns on all nukies and in the HOS locker, that constantly updates its sprite to a arrow pointing to the disk, when you are close it turns to a green arrow and finally a dot when you are right on it.

ReiderJack:
fixes #3173

Pointer

- [ ] Need to put nukes on syndie shuttles.
- [ ] Probably put disk on tables on all stations not in the locker.
- [ ] Station nukes shoud be replaced with destruction terminal.
- [ ] Delta alert toggle should be removed from Communication console.


DooblyNoobly:
okey dokey, merging to staging

PR #3791, opened 2020-03-31
More stringent checks on account handling

Bod9001:
So  I am unable to investigate the token Authentication stuff 
so idk if it actually has anything wrong with it, Since I can't view the code for it,
I cleaned the other bits up and added naming to Highlight what Variables are Unchecked soShouldn't be trusted from client

Using Userid more dependable than client ID,
for logout bodies, hoping that that HTTP request authenticates the
Userid and token properly,
added more stringent checks on null return on the HTTP request 

### automatic kick for multiple accounts on one IP

A brief check in editor,
Not tested relogging

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
+++ b/UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
@@ -164,20 +164,20 @@ public partial class PlayerList
 		return adminUsers.Contains(userID);
 	}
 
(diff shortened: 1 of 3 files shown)


DooblyNoobly:
`Using Userid more dependable than client ID` problem with this is that you can't use one account to test with 2 or 3 clients in development. I think we have a IsRelease flag in some kind of build options class that will return true if it was built via the build server so maybe we can use that to flick between the two conditions

Bod9001:
Done requested changes merging

PR #3816, opened 2020-04-03
Mag boots

### LINKED ISSUE #3011: Spacewind is too deadly, we need magboots. B: 30
Description 
Currently there is essentially no way to deal with space wind as magboots are not in.

Proper magboots need 5 things

- Apply a speed reduction to wearer when toggled on unless outside of gravity IE space.
- Make the user unable to be slipped.
- Make the user unable to be force-ably moved by space winds.
- Ability to be toggled on and off via clicking the boots some way IE control or alt clicking.
- User cannot be pushed nor pulled unless via grab intent, speed when grabbed should be very slow.

Contributor will receive a bounty of 30usd on a successful merge of a solution that meets the criteria above.

ReiderJack:
fixes #3011

AdvBootToggle

UnPushable

UnSlippable


UnityProject/Assets/ItemMagBoots.cs
```diff
@@ -0,0 +1,185 @@
...
+		if (!isOn)
+		{
+
+
+			playerMove.InitialRunSpeed = newSpeed;
+			playerMove.RunSpeed = newSpeed;
+
+			//Do not change current speed if player is walking
+			//but change speed when he toggles run
+			if (playerMove.GetComponent<PlayerSync>().SpeedServer < newSpeed)
```
fomalsd: client doesn't know SpeedServer, this will have issues on headless

fomalsd: also you can avoid tons of GetComponent<> here by doing `var playerSync = playerMove.PlayerScript.PlayerSync`

UnityProject/Assets/ItemMagBoots.cs
```diff
@@ -0,0 +1,185 @@
...
+
+	// Client only method
+	public void OnInventoryMoveClient(ClientInventoryMove info)
+	{
+		if (CustomNetworkManager.Instance._isServer && GameData.IsHeadlessServer)
+		{
+			return;
+		}
+
+		var hand = PlayerManager.LocalPlayerScript.playerNetworkActions;
```
fomalsd, line 108: > hand

wat

UnityProject/Assets/ItemMagBoots.cs
```diff
@@ -0,0 +1,185 @@
...
+		if (isOn && info.FromSlot.NamedSlot == NamedSlot.feet)
+		{
+			ServerChangeState(this.player);
+		}
+	}
+
+	// Client only method
+	public void OnInventoryMoveClient(ClientInventoryMove info)
+	{
+		if (CustomNetworkManager.Instance._isServer && GameData.IsHeadlessServer)
```
fomalsd, line 103: are you sure this check is even needed here?

UnityProject/Assets/ItemMagBoots.cs
```diff
@@ -0,0 +1,185 @@
...
+			if (playerMove.GetComponent<PlayerSync>().SpeedServer < newSpeed)
+			{
+				return;
+			}
+
+			playerMove.GetComponent<PlayerSync>().SpeedClient = newSpeed;
+		}
+		else
+		{
+			playerMove.InitialRunSpeed = 6;
```
fomalsd: I still don't get why you have to change `InitialRunSpeed`, it's not supposed to be modified and should be used as a reference to what's the normal running speed of this player

UnityProject/Assets/ItemMagBoots.cs
```diff
@@ -0,0 +1,185 @@
...
+			playerMove.RunSpeed = 6;
```
fomalsd: Avoid magic numbers like `6` everywhere, refer to InitialRunSpeed instead

UnityProject/Assets/ItemMagBoots.cs
ReiderJack: > client doesn't know SpeedServer, this will have issues on headless

Didn't cause any issues as far as I saw in loggs and console.

UnityProject/Assets/ItemMagBoots.cs
ReiderJack: I am using as an example code in InteractableStorage.cs It's handled there this way

UnityProject/Assets/ItemMagBoots.cs
ReiderJack: I tried doing it as other systems which are changing speed of the player handle it and it did not work. Without this line, player won't be slowed.

Bod9001:
Make sure to mark them as a Resolved once you've resolved them

PR #3874, opened 2020-04-10
Tiny Fan

ThatDan123:
image

Is not atmos passable. 


fomalsd:
be aware that non-atmos passable objects are buggy:
https://github.com/unitystation/unitystation/issues/1902#issuecomment-575571921
so if someone destroys such fan the tile will most likely stay non-atmos passable

ThatDan123:
Ah right.... humm tricky

ThatDan123:
Do you still have the special command?

corp-0:
from what I understood, the problem is with objects. If you make it a tile when active, it should work... Maybe.

fomalsd:
yes, forcefully updating tile can be done with alt+click when hosting

ThatDan123:
Can that be done in code?

ThatDan123:
Just gonna get this in, i accidentally deleted the other stuff like the holograms which i was working on and cant find the files anywhere rip.

PR #3886, opened 2020-04-12
Fix for hostile AI going braindead when they spot a player, bots being non-functional

### LINKED ISSUE #3884: Xenos are brain dead
# Xenos are brain dead

## Description
Xenos don't attack players anymore. They don't even move. Looks like something happen with mobs AI.

CodeKalias:
### Purpose
Fixes #3884
Also bots will once again move around and perform their tasks

### Notes:

### DIFF
--- a/UnityProject/Assets/Scripts/NPC/AI/MobAgent.cs
+++ b/UnityProject/Assets/Scripts/NPC/AI/MobAgent.cs
@@ -130,13 +130,26 @@ public class MobAgent : Agent
 
 	void MonitorDecisionMaking()
 	{
-		if (health == null || integrity == null) return;
+		if (health == null && integrity == null) return;
 		// Only living mobs have health.  Some like the bots have integrity instead.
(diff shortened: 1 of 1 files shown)


CodeKalias:
Should be good to go

UnityProject/Assets/Scripts/NPC/AI/MobAgent.cs
```diff
@@ -130,13 +130,26 @@ protected virtual void AgentServerStart()
...
-			return;
+			if ((health.IsDead || health.IsCrit || health.IsCardiacArrest))
+			{
+				//can't do anything this NPC is not capable of movement
+				return;
+			}
+		}
+		else //Bot
+		{
+			if (integrity.integrity <= 0)
```
fomalsd, line 144: integrity can be null which will give an NRE

CodeKalias: Line 133 should mean it can't get there if integrity is null
edit- since health would have to be null at 138, so integrity can't be as well

PR #3909, opened 2020-04-14
add Survive Objective for nuke ops

corp-0:
### Purpose
The idea is to have an incentive for Nuke Ops to leave the nuke and flee. This is also how it is played in original SS13.

### Notes:
I just added the scriptable object to the Nuke Ops antags.

### DIFF
--- a/UnityProject/Assets/Scripts/Antagonists/Antags/NuclearOperative.asset
+++ b/UnityProject/Assets/Scripts/Antagonists/Antags/NuclearOperative.asset
@@ -18,5 +18,6 @@ MonoBehaviour:
   needsEscapeObjective: 0
   coreObjectives:
   - {fileID: 11400000, guid: 1a15330eccadc7044b35b5a49a4154b4, type: 2}
+  - {fileID: 11400000, guid: b6107e1c868bada4da25dc2fc840e411, type: 2}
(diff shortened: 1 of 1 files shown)


Jibbajabbafic:
Does this work? The intention was for escape objectives to be generated by ticking 'NeedsEscapeObjective' but that will randomly pick the escape or survive objective. There's no logic to *only* choose specific escape objectives yet which I could add since I've been modifying some of the antags in my PR.

corp-0:
as far as I know, "escape" means escaping in the evac shuttle which is not what we want... I think. Survive objective only checks if you're alive at the end of the round.

The idea here is to make them flee after planting the nuke as the explosion has a radius in which it is deadly (see ``nuke.cs``).

Jibbajabbafic:
Hmm the naming makes this a bit confusing. When I said 'escape objectives' I meant objectives like Escape or Survive, so if 'NeedsEscapeObjective' is true it will randomly pick one of those.

I'll rename 'NeedsEscapeObjective' to 'NeedsFinalObjective' or something like that to make it clearer.

Anyways, I think you will need to increase NumberOfObjectives to 2 for this to work like you intended.

corp-0:
closing as the changes will be included in a bigger PR by @Jibbajabbafic

PR #3920, opened 2020-04-15
Fixed Bikehorn sound not playing and dealing damage #3519

### LINKED ISSUE #3519: On Help intent, bike horns should not deal damage.
It should just be annoying, not actual assault.

Adalino:
Fixes #3519 by modifying the Horn.cs class. 
The horn now only does damage when harm intent is selected and makes noise again.

Notes:
I have written a method that returns the location where the honk-sound should be played. If possible, this behaviour should not be controlled by each item individually but on a higher level.
Also first pull request ever, so a little bit of a test.

Please make sure you have followed the self checks below before submitting a PR:

- [X] Code is sufficiently commented
- [X] Code is indented with tabs and not spaces
- [X] The PR does not include any unnecessary .meta, .prefab or **.unity (scene) changes**
- [X] The PR does not bring up any new compile errors
- [X] The PR has been tested in editor
- [X] Any new / changed components follow the [Component Development Checklist](https://github.com/unitystation/unitystation/wiki/Component-Development-Checklist)
- [X] The PR has been tested in multiplayer (only 1 client)


corp-0:
maybe this 
```c
ItemSlot itemslot = gameObject.GetComponent<Pickupable>().ItemSlot;
return itemslot != null ? itemslot.ItemStorage.gameObject : gameObject;
```
could go directly in the method that plays sound, so every item that needs it can benefit from it.

Adalino:
Yes, this is what I meant by "If possible, this behaviour should not be controlled by each item individually but on a higher level". I'm going to test if there is a problem when i put the inventory check into the soundmanager.

corp-0:
make it return null if it doesn't have the Pickupable component and it should be alright

```c
var itemSlot = gameObject.GetComponent<Pickupable>()?.ItemSlot;
```

PR #3922, opened 2020-04-15
Need to test this, temporary fix for the bans?

Bod9001:
since client ID was recently been implemented,
the old bands had serialised to empty IDs,
but that means if anything connects with an empty ID it would be counted as banned


DooblyNoobly:
I got one coming

PR #3951, opened 2020-04-18
Conveyor Fix

ThatDan123:
Hacky fix for built in game conveyors not animating.


Bod9001:
What happens when you convert it into a Synchronise var hook?

Bod9001:
I was told to use Syncvar hooks Since they were more Performit apparently, vs Syncvar  
but That was on unet so idk if it's changed

PR #3960, opened 2020-04-19
Change to french accent and unique icon for clothing button

AutumnNova:
I'm is now je suis instead of j
a is now un or une
the is now les, la or le
i is now je istead of j'
friends is changed to amis
very is changed to tres
bread is changed to baugette

as they say in france, c'est un tres bon baugette


Bod9001:
ok looks good

PR #3966, opened 2020-04-20
Make mobs great again, Part one: The Xenos

corp-0:
## Purpose

This PR will clean up Mob scripts to actually make use of OOP fundaments, reducing code duplication and making them easier to maintain.

Also will implement AI Xenomorph life cycle.
- Xeno queens have a chance to lay eggs
- Eggs will start to grow. They are very fragile when small and can be easily stomped.
- Once they are big, you can't destroy them that easily and if you let them unatended, a facehugger will be born.
- Facehuggers jump into people and insert a larvae in their esophagus. You can prevent it using some face protection, but beware, they may break if you receive enough damage.
- People can save you in the frame of time the facehugger takes to "impregnate" you, but once the deed is done, there is not much you can do (until we have surgery).
- After some time, a larvae will burst out of your chest, killing you in the act.
- Larvaes are small and represent almost no threat to anyone... That's it until they grow into a Xenomorph we all know and love.

## Notes
Despite the initial intention being reduce code duplication, some sins had to be done for the facehugger, otherwise I would have had to refactor a deeper part of the mob AI core.

I tested this PR in MP and all, no game breaking problems found. Only issue we noticed is that some times spawned AI would be unresponsive for a second, but then it would be back at it.


corp-0:
review me, sempai

PR #3973, opened 2020-04-20
Added some common Integrity Templates

### LINKED ISSUE #3946: Create templates for object integrity / resistances
Unity 2019.3 has a new "Templates" feature which allow us to create different preset values for any component.

If we had templates for small objects / crates / doors / indestructible objects / etc. , then setting an object's health would be quicker and more consistent. 🎉

... Templates would, of course, be useful for other components, too. But integrity seems like a good place to start.

Fortune117:
Added common elemental metals, some fabrics, glass and wood.

# Basic Integrity Templates

### Purpose
Fixes #3946
Adds a bunch of common metals, some fabrics, wood and glass templates for object integrity.
There's a lot of room to expand here, and plenty more could be added, but it's a start.

###Notes
This is my first PR, so please let me know if anything is wrong! I wasn't sure exactly what folder these would go in so I made a presets folder under assets.

### DIFF
--- a/UnityProject/Assets/Presets.meta
+++ b/UnityProject/Assets/Presets.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: 3a9203130e1e0e744a265256ec0585b7
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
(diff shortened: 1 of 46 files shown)


daalta:
Good stuff!

PR #4008, opened 2020-04-23
Small cable fixes, new placeholder high power wire sprites

CodeKalias:
### Purpose
Fix some cable assets which didn't have their directions and deconstruct set up correctly
Updated placeholder high power wire sprites, including adding diagonals, they're orange now to match the inhand/high voltage standard
Changed wires only being placed in North/East/South/West directions to an option instead of hard coded to high power wires
High power wires are now allowed to be placed diagonally

### Notes:
If you want to revert high power wires to 4 directional only, tick "Only NESW" in High Cable Coil.prefab Cable Coil script.
(Or just revert commit 6fb8dee below)

Diagonal cables connected to the generator
image
**RIBBON CABLE HELL**
image


Bod9001:
yeah, Part of the idea of high-voltage cables was they were so unflexible, They could only do right angles and Straight on, Personally I think they should still have limitation

PR #4016, opened 2020-04-23
Improve NetUIDynamicList computational complexity.

### LINKED ISSUE #3002: Creating a large amount of new records in the Security Records Console
## Description
Large amount of new (added by the player) records in the Security Records Console may lead the server to very serious lags and even crash the whole server. But if server stopped lagging player who put his ID to the Console is unable to use the Console again and return his ID from this every Security Records Console.

### Steps to Reproduce
1.  Create 100-100000 new records in the Security Records Console.

### LINKED ISSUE #3169: NetUIDynamicList needs a rewrite
#### NetUIDynamicList is unrobust! 

- Current implementation wasn't designed for a large amount of entries.
- Every new entry increases amount of sent messages exponentially, also network message size can end up getting too high. You can really feel the lag when there are 15 entries or more and a new one is added. See #3002 
- It can occasionally break, resulting in "wonky value import" warnings. It signifies a desync between client and server, and that client no longer sees everything he's supposed to.

To see how bad it gets, set logging level for NetUI category to Trace, open Security Records tab and start adding new records. 

#### Good implementation should have following qualities:

- Handle 200 entries without noticeable performance impact for server or client
- Never go beyond network message size limit
- Shouldn't send more than one update message when a new entry is added to the list, regardless of its current size
- Make sure that all tabs using dynamic lists still work properly (Shuttle radar, Cloning console, ) and no "wonky value import" warnings occur

RedlineTriad:
### Purpose
Fix #3002

### Notes:

- [ ] Code is sufficiently commented
- [ ] The PR has been tested in multiplayer (with 2 clients and late joiners, if applicable)

### DIFF
--- a/UnityProject/Assets/Scripts/Messages/Server/TabUpdateMessage.cs
+++ b/UnityProject/Assets/Scripts/Messages/Server/TabUpdateMessage.cs
@@ -88,7 +88,7 @@ public class TabUpdateMessage : ServerMessage
 		}
 
 		msg.SendTo(recipient);
-		Logger.LogTrace(msg.ToString(), Category.NetUI);
+		Logger.LogTraceFormat("{0}", Category.NetUI, msg);
 		return msg;
 	}
 }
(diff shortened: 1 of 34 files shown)


fomalsd:
are you sure it fully fixes #3169? because this one is most likely not caused by performance but some bug in the logic:
> It can occasionally break, resulting in "wonky value import" warnings. It signifies a desync between client and server, and that client no longer sees everything he's supposed to.

RedlineTriad:
@fomalsd can't really test that unless i have a way to reproduce it.

PR #4022, opened 2020-04-24
NRE fixes

ReiderJack:
image

### DIFF
--- a/UnityProject/Assets/Scripts/Lighting/LightSource.cs
+++ b/UnityProject/Assets/Scripts/Lighting/LightSource.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Runtime.Remoting.Messaging;
 using Light2D;
 using Lighting;
 using Mirror;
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Lighting/LightSource.cs
```diff
@@ -133,8 +134,9 @@ void OnDrawGizmosSelected()
 		var sprite = GetComponentInChildren<SpriteRenderer>();
 		if (sprite == null)
 			return;
-		if (relatedLightSwitch == null && !isWithoutSwitch)
+		if (relatedLightSwitch == null)
```
fomalsd, line 137: this doesn't fix the exceptions, according to the screenshot they are on line 144 – most likely due to `relatedLightSwitch.transform.position` if` relatedLightSwitch` is null

ReiderJack: I think it does. If it's null it won't pass to that section of code, cuz I fixed the check.

fomalsd: ah, I think you're right... an if-else would probably more readable tho

PR #4026, opened 2020-04-24
add basic random loot system

### LINKED ISSUE #2334: Add loot tables and spawners for map makers
# Description
For areas in maint, lavaland, or otherwise, map makers in DM can define loot tables to spawn random items. The loot tables could probably be SOs, with an option to make an overridden loot table for a specific spot. 

These should be able to define how many items can spawn at that location, if the same item can be spawned more than once, and if the items should be locked into the center of the tile or can be randomly scattered.

It would also be worth having tables that can be reused in multiple places. For example, a maintenance table that spawns items like trash, costumes, or medical items.

Loot tables can also define weights of the items. All items have a default weight of 1, but can be defined to have a bigger chance of spawning. For example, to give a 75% chance of an iron coin spawning, with a 25% chance of a gold coin, those two items in the list would have weights of 75 and 25. This does not mean that the weights are required to add up to 100.

https://github.com/tgstation/tgstation/blob/422ed7df0b72de0c04bc49b36f8038b4454dc58f/code/game/objects/effects/spawners/lootdrop.dm

corp-0:
### Purpose
Adds the ability for mappers to set random loot spawners. Fixes #2334

### How does it work:
From this:
image

To this:
image

The actual Spawner is a game object that accepts a list of item pools (scriptable objects). At round start it will pick a random pool and roll the pool probability of spawning an item from this pool, when met it will choose a random item from the pool and roll probability for that item, if met it will spawn it. Simple.

image

image

The system is designed to be flexible and allow devs to combine pools and set their probabilities how ever they want.

I recreated some pools from TG's code, although incomplete because we lack some items.

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Items/RandomItemSpot.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Items/RandomItemSpot.prefab
@@ -1,0 +1,89 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!114 &1158969463323420989
+MonoBehaviour:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 2839352587806555520}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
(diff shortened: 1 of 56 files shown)


corp-0:
found some problems with the object destruction when it shouldn't spawn anything. I'm fixing them

corp-0:
Fixed the problem + made the spawner even better. I'm ready for review!

Issue #3027, opened 2020-02-10
Fire Doors B: 120


Jibbajabbafic:
###  Description
They should fulfil the following criteria:
- [x] Animated opening and closing with sounds
- [x] Trigger automatically when they detect these things near them: 
  - [x] Fires or a high temperature
  - [x] Low and high pressure
- [x] Be fire resistant
- [x] Allow crowbar to open and close them manually

Other nice to have improvements:
- [x] Add a way to easily connect fire alarms to fire doors (copy the way the APCs handle connecting devices, see APCConnectingEditor.cs)
- [x] Ensure overlapping doors can still be bumped into regardless of the order fire doors were placed

Once this is done #4118 can be completed.

The-Scout:
you should add that they should lower whenever a room gets depressurized. very important with our current atmos implimentation

StrawberryMoses:
Depending on how hard the implementer wants to go on this- iirc on TG fire doors work with a sort of "zone" concept. The station is broken up into rooms / halls and are used for two purposes (that I can remember): electricity and fire doors. The apc powers everything in that room, and similarly, when a fire is detected or the rooms fire thingy is pulled, all of the fire doors in the room close.
Similarly, this is used when someone tries to steal the captains antique gun- when the case is broken, all fire doors in the captains office close.

Jibbajabbafic:
Do the 'zones' have dedicated sensor objects for fires? I think just having the doors do it for now would be good enough.

StrawberryMoses:
If I was going to guess, I'd say it would be the fire alarm control thingy, the air alarm, or the vents.
I agree, the doors doing that would probably be best / easiest for now- just a thought that might be worth keeping in mind

Jibbajabbafic:
Ye, we can make an issue for those too

Misterixmax:
There is air alarm on TG. It has a sensor and some other stuff, for example you can configure vents or scrubbers - it's needed to clear room from toxins, co2 or so. Also air alarm controls filters and air composition with configured % of each gas type.  And it sends data about atmos inside the room to atmos technician's room where all the station's atmos can be monitored. Also one monitor is on the bridge.

h3killa:
Ill snag this one if no one else has

RedlineTriad:
Neither superscout or @h3killa are responding, so I am marking this Issue as free again.

smugmann:
can i get assigned to this so i would feel responsible?

RedlineTriad:
[...]

PR #4041, opened 2020-04-25
Firelocks

Aranclanos:
### Purpose
Adds firelocks!
They close on detecting low pressure on their tile or being damaged by a hotspot

Related to #3027 

- [X] Code is sufficiently commented
- [X] Code is indented with tabs and not spaces
- [X] The PR does not include any unnecessary .meta, .prefab or **.unity (scene) changes**
- [X] The PR does not bring up any new compile errors
- [X] The PR has been tested in editor
- [X] Any new files are named using PascalCase (to avoid issues on case sensitive file systems)
- [ ] Any new / changed components follow the [Component Development Checklist](https://github.com/unitystation/unitystation/wiki/Component-Development-Checklist)
- [ ] Any new objects / items follow the [Creating Items and Objects Guide](https://github.com/unitystation/unitystation/wiki/Creating-Items-and-Objects%3A-Now-With-Prefab-Variants) (especially concerning the use of prefab variants)
- [X] The PR has been tested in multiplayer (with 2 clients and late joiners, if applicable)


ThatDan123:
Missing reference found in scriptable object FireDoor field Object

  Expected: <empty>
  But was:  < (FireDoor, Object) >at Tests.MissingAssetReferences.CheckMissingRefenceFieldsScritpableObjects () [0x0016a] in /github/workspace/UnityProject/Assets/Tests/Asset/MissingAssetReferences.cs:172[Tests] Missing reference found in scriptable object FireDoor field Object

Aranclanos:
ok fixed!!! go team!!!

Jibbajabbafic:
One problem I've found is that the fire doors stop the bump interaction with doors underneath them:

jtnlRbkdix

edit: This only happens when the fire lock is placed first and then the door on top. If the fire lock is placed on top of the door then it still works. I think it should be made more robust to stop mappers accidentally placing them in the wrong order

UnityProject/Assets/Scripts/Doors/FireLock.cs
```diff
@@ -0,0 +1,47 @@
...
+	public override void TryOpen(GameObject performer)
+	{
+
+	}
+
+	public void TickUpdate()
+	{
+		if (!Controller.IsClosed)
+		{
+			if (metaNode.GasMix.Pressure < AtmosConstants.MINIMUM_OXYGEN_PRESSURE)
```
Jibbajabbafic: The doors need to be more sensitive to pressure changes.

Uj52N3CunD


Try using this constant instead:
```suggestion
			if (metaNode.GasMix.Pressure < AtmosConstants.WARNING_LOW_PRESSURE)
```

Jibbajabbafic: I think it would be good if they also checked if it's greater than WARNING_HIGH_PRESSURE/HAZARD_HIGH_PRESSURE too.

Jibbajabbafic (changes requested):
A few things:
- Make doors more sensitive to low pressure
- Add checks for high pressure
- Allow the doors on the same tile to be bumped regardless of the order they are place (e.g. if the fire lock is on top of the door it seems to work fine, but the other way round it stops working).
- Make the firelock prefab default to the open state to make it easier for mappers
- Check the tiles around the firelocks since they don't seem to help stop fires spreading much https://imgur.com/a/zjIWARV

Aranclanos:
Right so to avoid fires from rushing through faster than what the firelocks can close, I enabled the fire alarms so they have a hub to "report" a fire so all the firelocks linked to said alarm get closed, acting like a room. But it's under the mappers hands to add the firelocks to the alarm list.

The pressure check is also now under the alarm instead of all firelocks to avoid affecting performance hits and they're more sensible, easier to trigger.

UnityProject/Assets/Scripts/Objects/FireAlarm.cs
```diff
@@ -0,0 +1,205 @@
...
+		}
+	}
+
+	private IEnumerator SwitchCoolDown()
+	{
+		isInCoolDown = true;
+		yield return WaitFor.Seconds(coolDownTime);
+		isInCoolDown = false;
+	}
+	*/
```
Jibbajabbafic, line 203: Delete this or add a comment explaining why it's commented out if you want to use it in future

Jibbajabbafic (changes requested):
The fire locks are way better at stopping fires now!

image

One problem with the fire alarms is that they need to check the tile next to them in the direction that they are facing, otherwise they only check the atmos inside the wall.
image

The tests need to be fixed before I can merge this though, so delete all the existing fire alarms in outpost and pog and I can merge this before Doobly's prefab lockdown.

PR #4070, opened 2020-04-28
Holdable Chairs

ReiderJack:
### Purpose
Make bar fights with stools real.

### Notes:
Added validation for Interactable Furniture. Added bar stool to build list. Put ItemMagBoots script in proper folder.

DraggableChairs


VernaSaurus:
this fixes two small things that were bugging me lmao, i always wondered why magboots weren't in the right folder (and also the barstool thing)

PR #4088, opened 2020-04-30
fix missing reference

AutumnNova:
fixes missing reference to lawyers headset


chairbender:
lgtm, thanks for fixing it!

Issue #2526, opened 2020-01-03
Too many warnings


Jibbajabbafic:
# Description
There are currently about ~500 warnings when opening the Lobby scene which clutters up the console significantly and can give new developers a bit of a scare the first time they open the project.

Most of them can be easily fixed so it would be nice to clean them up at some point. 

![image](https://user-images.githubusercontent.com/4493127/71724835-abdac700-2e29-11ea-996d-f710fe90fd09.png)

## Steps to Reproduce
Please enter the steps to reproduce the bug or behaviour:
1. Open a Scene
2. See all the warnings in the console

voidel:
For the variables that are unused but might be indicative of future functionality, should they be commented out or just removed?

Jibbajabbafic:
I say remove them, they can always be added back in easily if they're needed and we'll double check them in your PR.

If you could do a separate commit for removing the variables it'll make the reviewing easier too.

daalta:
More progress on the issue: https://github.com/unitystation/unitystation/pull/3878

PerfectTangent:
I think once we get the number of warnings in the console down to less than 40, this issue should be considered satisfied. How many warnings are spammed on a fresh branch of develop right now?

HuntorWorks:
129 as of today

Issue #3362, opened 2020-02-26
Finish & polish all job descriptions


daalta:
We have job descriptions for most of our jobs. But these still contain some errors and some jobs still lack a description.

See our progress here: https://docs.google.com/document/d/1ZYBtoh4tAWKykmbViitqhFMdmdLVpPHIN_y2MgsYzI4/edit

Last related PR: https://github.com/unitystation/unitystation/pull/3234

corp-0:
To anyone who contributes to the descriptions. Try to don't modify the green ones unless it is truly necesarry. 

Check the state of the description and see if you can make one that hasn't been done yet (Mime and Chaplain at the moment of this comment).

daalta:
Looks like most descriptions are OK. We'll continue updating them as required, it'll be an ongoing task.

PR #4099, opened 2020-04-30
Fixed some compilation warnings

daalta:
More progress on https://github.com/unitystation/unitystation/issues/2526 .

* Fixed some of the usual default value warnings
* Camera recoil was accidently using a local variable, which I fixed. Could someone please test if the camera recoil feels different? (Because it... should? Maybe?)
* Some obsolete functions were replaced - Behaviour *should* be identical in those cases.

I did a quick editor test, and things seem to be in order. Still, the prospect of having broken something is *terrifying*.


corp-0:
from what I saw in the changes, everything needs a default value when declared? Could we make a test for it so peopel don't forget it?

daalta:
@corp-0 I'm still not sure what the best way to handle that warning is.

Unity, by default, expects variables to get assigned somewhere. It doesn't know if that happens through the inspector, so it warns us. Just in case.

Most IDEs, however, see assigning variables to their default value as redundant. Rider suggests *removing* default values.

Seems like we can't really win..?

daalta:
@corp-0 I'm still not sure what the best way to handle that warning is.

Unity, by default, expects variables to get assigned somewhere. It doesn't know if that happens through the inspector, so it warns us. Just in case.

Most IDEs, on the other hand, think that assigning variables to their default value is redundant. Rider suggests *removing* default values.

Seems like we can't really win..?

Bod9001:
looks good, and the only way to find out is to test it : P

daalta:
https://github.com/unitystation/unitystation/issues/3362

PR #4083, opened 2020-04-29
PlayerSpawn Queue system

ThatDan123:
Adds a queue for the playerspawn, to hopefully fix multiple people getting same job.

- [x] Code is sufficiently commented
- [x] Code is indented with tabs and not spaces
- [x] The PR does not include any unnecessary .meta, .prefab or **.unity (scene) changes**
- [x] The PR does not bring up any new compile errors
- [x] The PR has been tested in editor
- [x] The PR has been tested in multiplayer (with 2 clients and late joiners, if applicable)
- [x] The PR has been tested with round restarts.

PR #4104, opened 2020-05-01
Fixes a bunch of server killing bugs

DooblyNoobly:
- reverted #4083 as it caused logging in issues for players and would crash servers
- Made it impossible to create duplication of wire tiles
- Fixed wire dupes on outpost
- Fixed egg code (not tested but separated server and client code preventing server methods being called on client)


DooblyNoobly:
test runner passes locally

PR #4105, opened 2020-05-01
Dirty draft PR. Fixing player access violation, lighting, and more.

Ninjaboi8175:
Just a quick and dirty draft PR. Fixes lighting errors, client accessing server-only code, player attempting to edit object data they have no authority over, and more.

### DIFF
--- a/UnityProject/Assets/Scripts/Clothing/FacehuggerImpregnation.cs
+++ b/UnityProject/Assets/Scripts/Clothing/FacehuggerImpregnation.cs
@@ -37,6 +37,14 @@ namespace Clothing
 
 		public override void OnStartServer()
 		{
+			// Testing to see if one of the several references to this method get called on the client. Can be removed
+			// or add a [Server] line above it fire warning, but this is more specific at a glance.
+			if (!isServer) 
+			{
+				Logger.Log("Client tried to run OnStartServer() in FacehuggerImpregnation.cs. Forcing return.", Category.Server);
+				return;
+			}
+
 			base.OnStartServer();
 			if (!isToy) return;
 
--- a/UnityProject/Assets/Scripts/Lighting/LightMountStates.cs
+++ b/UnityProject/Assets/Scripts/Lighting/LightMountStates.cs
@@ -87,6 +87,12 @@ public class LightMountStates : NetworkBehaviour, ICheckedInteractable<HandApply
 
 	public void ServerPerformInteraction(HandApply interaction)
 	{
+		if (!isServer) 
+		{
+			Logger.Log("Client tried to run ServerPerformInteraction() in LightMountStates.cs. Tsk Tsk");
+			return;
+		}
+
 		if (isInCoolDown) return;
 		StartCoroutine(CoolDown());
 		var handObject = interaction.HandObject;
@@ -150,7 +156,8 @@ public class LightMountStates : NetworkBehaviour, ICheckedInteractable<HandApply
 
 		if (newState == LightMountState.On)
 		{
-			lightSource.ServerChangeLightState(LightState.On);
+			// Most likely shouldn't be running client-side.
+			//lightSource.ServerChangeLightState(LightState.On);
 			spriteRenderer.sprite = GetSprite(spriteListFull);
 			spriteRendererLightOn.sprite = GetSprite(spriteListLightOn);
 			integrity.soundOnHit = "GlassHit";
@@ -173,7 +180,8 @@ public class LightMountStates : NetworkBehaviour, ICheckedInteractable<HandApply
 			spriteRenderer.sprite = GetSprite(spriteListFull);
 			integrity.soundOnHit = "GlassHit";
 		}
-		lightSource.ServerChangeLightState(LightState.Off);
+		// Most likely shouldn't be running client-side.
+		//lightSource.ServerChangeLightState(LightState.Off);
 		spriteRendererLightOn.sprite = null;
 	}
 
@@ -241,19 +249,25 @@ public class LightMountStates : NetworkBehaviour, ICheckedInteractable<HandApply
 
 	private void SyncLightState(LightMountState oldState, LightMountState newState)
 	{
-		state = newState;
+		//state = newState;
 		ChangeLightState(newState);
 	}
 
 	public override void OnStartClient()
 	{
-		SyncLightState(state, state);
+		// Commenting this line out for now as you shouldn't be running a syncvar hook outside of the syncvar getting updated data from server.
+		//SyncLightState(state, state);
 		base.OnStartClient();
 	}
 
 	[Server]
 	private void ServerChangeLightState(LightMountState newState)
 	{
+		if (!isServer) 
+		{
+			Logger.Log("Client tried to run ServerChangeLightState() in LightMountState.cs. Tsk Tsk");
+			return;
+		}
 		state = newState;
 	}
 
--- a/UnityProject/Assets/Scripts/Lighting/LightSource.cs
+++ b/UnityProject/Assets/Scripts/Lighting/LightSource.cs
@@ -50,7 +50,8 @@ public class LightSource : ObjectTrigger, IAPCPowered, IServerDespawn
 		wallMount = GetComponent<LightMountStates>();
 		emergencyLightAnimator = GetComponent<EmergencyLightAnimator>();
 		SwitchState = true;
-		mState = InitialState;
+		if(isServer)
+			mState = InitialState;
 		lightSprite.Color = lightStateColorOn;
 	}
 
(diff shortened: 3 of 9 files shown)


UnityProject/Assets/Scripts/Clothing/FacehuggerImpregnation.cs
```diff
@@ -37,6 +37,14 @@ private void Awake()
 
 		public override void OnStartServer()
 		{
+			// Testing to see if one of the several references to this method get called on the client. Can be removed
+			// or add a [Server] line above it fire warning, but this is more specific at a glance.
+			if (!isServer)
```
DooblyNoobly, line 42: OnStartServer is only ever called on the server or a host so `!isServer` is not needed here

UnityProject/Assets/Scripts/Lighting/LightMountStates.cs
```diff
@@ -87,6 +87,12 @@ public bool WillInteract(HandApply interaction, NetworkSide side)
 
 	public void ServerPerformInteraction(HandApply interaction)
 	{
+		if (!isServer)
```
DooblyNoobly, line 90: ServerPerformInteraction is only ever called on server via Interface. `!isServer` is not needed here

UnityProject/Assets/Scripts/Lighting/LightMountStates.cs
```diff
@@ -241,19 +249,25 @@ private void OnDamageReceived(DamageInfo arg0)
 
 	private void SyncLightState(LightMountState oldState, LightMountState newState)
 	{
-		state = newState;
+		//state = newState;
```
DooblyNoobly, line 252: you need to set the SyncVar again with the new state in SyncVar hooks. Its silly I know

UnityProject/Assets/Scripts/Lighting/LightMountStates.cs
```diff
@@ -241,19 +249,25 @@ private void OnDamageReceived(DamageInfo arg0)
...
 		ChangeLightState(newState);
 	}
 
 	public override void OnStartClient()
 	{
-		SyncLightState(state, state);
```
DooblyNoobly, line 250: Calling the syncvar hook in OnStartClient is actually the correct way to do it. SyncVars are loaded with the correct value from server in the OnStartClient hook but the SyncVar hook is not called. You need to do it manually here.

UnityProject/Assets/Scripts/Lighting/LightMountStates.cs
```diff
@@ -241,19 +249,25 @@ private void OnDamageReceived(DamageInfo arg0)
...
+		// Commenting this line out for now as you shouldn't be running a syncvar hook outside of the syncvar getting updated data from server.
+		//SyncLightState(state, state);
 		base.OnStartClient();
 	}
 
 	[Server]
 	private void ServerChangeLightState(LightMountState newState)
 	{
+		if (!isServer)
```
DooblyNoobly, line 266: The [Server] Attribute will stop this method being called on Client so !isServer condition is not needed. If it is called on the client though a warning is logged in console.

UnityProject/Assets/Scripts/Lighting/LightSource.cs
```diff
@@ -50,7 +50,8 @@ private void Awake()
 		wallMount = GetComponent<LightMountStates>();
 		emergencyLightAnimator = GetComponent<EmergencyLightAnimator>();
 		SwitchState = true;
-		mState = InitialState;
+		if(isServer)
```
DooblyNoobly, line 53: IsServer won't be set in Awake. Override the OnStartServer NetworkBehaviour lifecycle method and set the mState value there

UnityProject/Assets/Scripts/Lighting/LightSource.cs
```diff
@@ -107,7 +108,8 @@ public override void Trigger(bool newState)
 	public override void OnStartClient()
 	{
 		base.OnStartClient();
-		SyncLightState(mState, mState);
```
DooblyNoobly, line 110: This is required for late joiners. See previous comments about initializing SyncVars with hooks

UnityProject/Assets/Scripts/Lighting/LightSource.cs
```diff
@@ -118,7 +120,7 @@ public void ServerChangeLightState(LightState newState)
 
 	private void SyncLightState(LightState oldState, LightState newState)
 	{
-		mState = newState;
+		//mState = newState;
```
DooblyNoobly, line 123: as per other comments this is actually needed

UnityProject/Assets/Scripts/Switches/LightSwitchV2.cs
```diff
@@ -50,6 +50,11 @@ public bool WillInteract(HandApply interaction, NetworkSide side)
 
 		public void ServerPerformInteraction(HandApply interaction)
 		{
+			if (!isServer)
```
DooblyNoobly, line 53: ServerPerformInteraction is only ever called on server

DooblyNoobly (changes requested):
These were the most obvious problems I could see.

UnityProject/Assets/Scripts/Clothing/FacehuggerImpregnation.cs
Ninjaboi8175: I agree. Early on I started with a shotgun approach and simply tossed isServer checks around in various places with a log to look out for once I started the client. I planned to clean up the ones that never showed up and then incrementally remove the ones that did get run by the client somehow after I figured out how/why they were being called.

UnityProject/Assets/Scripts/Lighting/LightMountStates.cs
Ninjaboi8175: Also completely agree with this.

UnityProject/Assets/Scripts/Lighting/LightMountStates.cs
Ninjaboi8175: I'm a bit confused by this, would you be able to explain why? In the Mirror docs for hooks it even states that the hook is always called AFTER the property has been set, so wouldn't that mean that it would have already been received and changed by the client before proceeding with the hook?

> The Hook is always called _after_ the property value is set. _You don't need to set it yourself_.

Especially that last part implied that setting it yourself is unnecessary in either case, plus from what I recalled from UNet you should never write to a syncvar variable as a client and instead treat it as read-only due to serialization issues that arise.

UnityProject/Assets/Scripts/Lighting/LightMountStates.cs
Ninjaboi8175: That actually makes sense. The docs state exactly the same thing with the value being the latest from the server prior to OnStartClient but never actually runs it. Makes sense that hooks won't start running until the client has started.

UnityProject/Assets/Scripts/Lighting/LightMountStates.cs
Ninjaboi8175: Sorry about that one. Again shotgunned the approach and put one there. To be honest I thought all the [Server] attribute did was give a warning when the client attempted access but apparently it also prevents it.

UnityProject/Assets/Scripts/Lighting/LightSource.cs
Ninjaboi8175: You're right as the object itself wouldn't be ready on Awake. Did what you suggested.

UnityProject/Assets/Scripts/Lighting/LightSource.cs
Ninjaboi8175: Agreed that would be necessary.

UnityProject/Assets/Scripts/Lighting/LightSource.cs
Ninjaboi8175: My previous comment on what you said earlier still stands. I'll add them back in just to be sure, but if you're able to explain why I'd much appreciate it.

UnityProject/Assets/Scripts/Switches/LightSwitchV2.cs
Ninjaboi8175: You're right. Another shotgun moment. Never saw that log ( or those other unnecessary logs ) ever pop up in my tests so these isServer checks were only the chopping block either way.

UnityProject/Assets/Scripts/Lighting/LightMountStates.cs
DooblyNoobly: yep I've read that too, and also tried it in a test project. Client wouldn't take the new value on a hook without setting it. Would be worth another retest now that we have the latest mirror

DooblyNoobly:
I see the reider has taken over. I'll close this pr

PR #4116, opened 2020-05-02
Cremators

NoooneyDude:
### Purpose
Adds cremator prefab for use in crematoriums! 

### Notes
Activated via right click method until wallmounted buttons can be assigned a generic component and not just DoorController.
> No map changes.


waffielz:
someone add it to a map smh

PR #4119, opened 2020-05-02
Electrical refactors and performance optimizations

CodeKalias:
### Purpose
Got rid of some string use in WireDirections, ElectrictricityFunctions should run a bit faster and be a little easier to follow, some minor optimizations by replacing .ContainsKey usage and early returns from loops.

### Notes:
There are so many pre-existing comments and functions with no references in ElectricityFunctions, would be nice if they could be trimmed.
See commit message c9fc7ea for an overview of the ElectricityFunctions changes

No obvious issues in testing, though I'd welcome an expert electricsman to verify the results

### DIFF
--- a/UnityProject/Assets/Scripts/Electricity/ConnectionMap.cs
+++ b/UnityProject/Assets/Scripts/Electricity/ConnectionMap.cs
@@ -5,191 +5,98 @@ public static class ConnectionMap
 	/// Given the output position of a wire and an adjacent tile to check
 	/// this method returns true if the adjacent tile has a connection 
 	/// input point that is intersecting the origin tiles output point
-	/// 
-	/// originP = the two connection points of the WireTile that is doing the checking
-	/// adjP = the two connection points of the WireTile that is being checked
-	/// Adjtile = the direction to the adjacent tile
 	/// </summary>
 	public static bool IsConnectedToTile(Connection ConnectionDirection, ConnPoint AdjacentConnections)
 	{
+		Connection validDirection;
 		switch (ConnectionDirection)
 		{
 			case Connection.North:
-				{
-					if (AdjacentConnections.pointA == Connection.South || AdjacentConnections.pointB == Connection.South)
-					{
-						return (true);
-					}
-				}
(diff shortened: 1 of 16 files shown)


Bod9001:
looks good so far

Bod9001:
I'm happy to answer any questions you have about electrical

Bod9001:
oh didn't see c9fc7ea

Bod9001:
only comment is missing the log check, since surprisingly the mappers tend to map cables on top of each other a lot

CodeKalias:
The log section moved to line 36, I'm thinking about looking into moving that into unit tests, though probably in a different PR.
Added one additional commit cb64001 , mostly reducing how deep indentation goes and clearing up some duplicated code, no functional changes. Done for now.

UnityProject/Assets/Scripts/Electricity/ConnectionMap.cs
```diff
@@ -5,191 +5,98 @@ public static class ConnectionMap
...
+				return false;
+			}
 		}
-		return (false);
-	}
 
+		return (AdjacentConnections.pointA == validDirection || AdjacentConnections.pointB == validDirection);
+	}
 
 	public static bool IsConnectedToTileOverlap(Connection ConnectionDirection, ConnPoint AdjacentConnections)
```
Bod9001, line 73: This is meant for overlapping connections  on a tile
like 

image


Should technically be still connected

Even though to a bug I introduced accidentallyThis doesn't work anymore,  since it expects connections to be a new you Hashset every time, so You have to do record if it got any connections the usual way, if not then check if there is overlap

Bod9001: It seems to be empty so therefore these wouldn't connect as in it only triggers with machine overlap for some reason? is this an error?

CodeKalias: First comment should be addressed in 1e62f26
Second in 85ab8fd

PR #4134, opened 2020-05-03
IF2 - allow clients to defer to the server when determining which interact to trigger

### LINKED ISSUE #2579: IF2 Should Support Client Triggering Wrong Component
We added RequestInteractMessage.ComponentType so you could manually trigger specific interactions, such as for the right click menu.

However, this also means that IF2 requests from the client specify the exact component they want to trigger. This is sometimes a problem, because the client doesn't always have all the correct info to know what component should be triggered (for example, for butchering, they don't know if a given object is a corpse). This would result in the interaction never happening because client keeps telling the server to perform the wrong interaction, and the server keeps going "lol no".

Change it so that regular IF2 workflow doesn't specify ComponentType and it goes back to the default behavior of the server checking WillInteract on each component, but manually triggering the interaction is also still supported.

EDIT: Make sure docs are updated with this

chairbender:
### Purpose
Fixes #2579 

Now when triggering a normal IF2 interaction, the client defers to the server to decide which component is triggered. Previously the client would tell the server the exact component it wanted to trigger, and if the server said no, nothing happened. This allows clients to trigger interactions without having the full knowledge that the server has w.r.t. which component should actually trigger.

Clients are still able to manually trigger specific interactions (such as for right click menu).

Also applies to Tile interactions.

The main entry point for the changes is in RequestInteractMessage (particularly ProcessInteraction) and InteractableTiles.ServerProcessInteraction

### DIFF
--- a/UnityProject/Assets/Scripts/Input System/InteractionV2/InteractionUtils.cs
+++ b/UnityProject/Assets/Scripts/Input System/InteractionV2/InteractionUtils.cs
@@ -38,7 +38,6 @@ public static class InteractionUtils
 	/// </summary>
 	/// <param name="interactable"></param>
 	/// <param name="interaction"></param>
-	/// <param name="side"></param>
 	/// <typeparam name="T"></typeparam>
 	/// <returns>true if an interaction was triggered (even if it was clientside-only)</returns>
 	public static bool ClientCheckAndTrigger<T>(this IBaseInteractable<T> interactable, T interaction) where T: Interaction
--- a/UnityProject/Assets/Scripts/Input System/InteractionV2/InteractionUtils.cs
+++ b/UnityProject/Assets/Scripts/Input System/InteractionV2/InteractionUtils.cs
@@ -23,7 +23,7 @@ public static class InteractionUtils
 	/// <param name="interaction">interaction to perform</param>
 	/// <param name="interactableComponent">component to handle the interaction, null
 	/// if the server should determine which component will trigger from the interaction</param>
-	/// <typeparam name="T"></typeparam>
+	/// <typeparam name="T">type of interaction</typeparam>
 	public static void RequestInteract<T>(T interaction, IBaseInteractable<T> interactableComponent = null)
 		where T : Interaction
 	{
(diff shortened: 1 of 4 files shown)


chairbender:
keeping this as a draft until I can confirm that this feature is working correctly in nooney's branch

chairbender:
Actually this is good to review. I've tested a lot of different interactions and it seems they are working fine, but there might be some subtle thing or edge case that doesn't work the same anymore that we'll only find out after a bit of actual playtime or staging testing.

chairbender:
updated docs: https://unitystation.github.io/unitystation/development/Implementing-Interactions/#how-does-willinteract-work-my-client-does-not-have-enough-info-to-implement-it

UnityProject/Assets/Scripts/Input System/InteractionV2/InteractionUtils.cs
```diff
@@ -14,31 +14,60 @@ public static class InteractionUtils
...
 		if (!Cooldowns.TryStartClient(interaction, CommonCooldowns.Instance.Interaction)) return;
 		RequestInteractMessage.Send(interaction, interactableComponent);
 	}
 
+
+	/// <summary>
+	/// Checks if this component would trigger and sends the interaction request to the server if it does. Doesn't
+	/// send a message if it only triggered an IClientInteractable (client side only) interaction.
+	/// </summary>
+	/// <param name="interactable"></param>
```
daalta: The parameter documentation is empty, I think you should remove it or complete it

chairbender: @daalta fixed

chairbender:
@daalta ready for another review, thanks for keeping me honest, bad habit of leaving those empty

daalta:
Docs look great now, thanks for taking the time to improve them.

I have to delegate the review to someone else though. Though GitHub does insist on calling me a 'reviewer' for making a comment :)

fomalsd (approved):
looking good!

Bod9001:
Shall I merge seems to be okay?

Bod9001:
merging

PR #4140, opened 2020-05-04
More logging

ThatDan123:
Adds more admin logging.

Adds two txt files, one for prev round and one for current which logs all admin actions.


ThatDan123:
Ready to merge

UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
```diff
@@ -65,6 +67,84 @@ void InitAdminController()
...
+
+			//Todo add to file round number here
+			file.WriteLine("false");
+
+			file.Close();
+		}
+
+		var first = File.ReadLines(adminLogPath).First() == "true";
+
+		var second = File.ReadLines(adminLog2Path).First() == "true";
```
chairbender: why is it writing "true" / "false" to the files?

UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
```diff
@@ -65,6 +67,84 @@ void InitAdminController()
...
+
+		if (first && !second)
```
chairbender: comments are great, but if you're going to add comments it would be more helpful to explain "why" rather than "what"...which leads me to my next question, why is it doing this check?

UnityProject/Assets/Scripts/Messages/Client/Admin/RequestRespawnPlayer.cs
```diff
@@ -37,8 +37,12 @@ void VerifyAdminStatus()
 
 	void TryRespawn(ConnectedPlayer deadPlayer)
 	{
-		UIManager.Instance.adminChatWindows.adminToAdminChat.ServerAddChatRecord(
-			$"{PlayerList.Instance.GetByUserID(Userid).Name} respawned dead player {deadPlayer.Name}", Userid);
+		string msg = $"{PlayerList.Instance.GetByUserID(Userid).Name} respawned dead player {deadPlayer.Name}";
```
chairbender: this logic is duplicated in quite a few places, refactor it to a method

UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
```diff
@@ -400,6 +480,16 @@ public void ProcessAdminEnableRequest(string admin, string userToPromote)
 		}
 	}
 
+	public void AddToAdminLog(string log)
```
chairbender: Why is a logging / chat function and log file management in PlayerList? I think logic related to admin logging should be somewhere else. PlayerList is for dealing with the list of players on the server.

chairbender (changes requested):
not quite clear what the file loading is doing, doesn't seem like it should be necessary to store info on the first line of the files but maybe I'm mistaken.

there's lots of duplicated logic that you should refactor to a shared method, maybe on PlayerList.Instance. Also I don't think PlayerList is the right place for managing log files / admin logs.

A more general question about this is...if we put this into a file then the only way to view it is to have physical access to the server. How will server admins who don't have access to the server view these logs?

UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
ThatDan123: Because for some reason even though this script is on a manager the bool way of changing which file is written to, as i wanted to keep a prev round log file resets on map change so theres no way to tell it to change file, so i added the true / false to the files instead.

UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
ThatDan123: Thats all part of it deciding which file to write to next.

Heres a pic of what i had to do:

image
The number in the circle is which file it will use, the true/false is whats read from the file.

UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
ThatDan123: Humm might be able to fix it now player list wasnt on the manager i thought it was on.

ThatDan123: Think i'll make a logging manager, also be good for player logging when thats added.

UnityProject/Assets/Scripts/Messages/Client/Admin/RequestRespawnPlayer.cs
ThatDan123: Yeah ive moved it to its own method.

UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
ThatDan123: Yeah ive moved it all to its own script, will need to be placed on a manager though

ThatDan123:
Atm admin actions are logged into the admin chat window, but this allows to record the last rounds logs as well, also theyre more detailed, they have the Admin ID and timestamp.

Also adds logging to i think all the admin actions now as only some were done before.

UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
ThatDan123: Cant seem to get the variable to persistent through round restarts so im gonna have to stick with this method for now.

chairbender: @ThatDan123 I don't think a variable or writing these values should be necessary to implement rolling logs. Just maintain 2 log files for example - currentRound.log, prevRound.log. At the start of each round, if currentRound.log exists, copy it and overwrite prevRound.log if it exists.

RedlineTriad:
I don't think file logging is the right approach, it's easy to hook up something to standard output and record it, files are more cumbersome.
And it would make more sense to use a logging library anyways, something like serilog would be good.

ThatDan123:
So ive changed it to be using the unity logger instead.

Just need to put it on a manager which i'll do after lockdown ends.

chairbender:
@RedlineTriad does the new approach look good to you?

chairbender:
@ThatDan123 looks like just 2 conflicts to resolve

ThatDan123:
No idea why test failed.

fomalsd (commented):
You made that whole new AdminLogging class – why not integrate that functionality in our existing Logger instead? 
So instead of `AdminLogging.Instance.AddToAdminChatAndLog(Message, Userid);`
It'd be `Logger.LogAdmin(Message, Userid);` – much more concise

Issue #4141, opened 2020-05-04
Pogstation needs renovations


Ranzarth:
## Description

Currently Pogstation is a mess in its current state, large parts need to be remade to make it playable
I've made a list of the problems and also personal opinions on what can be done to improve the map.

### Medical

- Needs to be redesigned with a more understandable layout.
- Chemistry needs to have access to hallway with secure windoors for two chemists.
- Virology needs to be more isolated from medical, possibly making it a small satellite from medical.
- Cloning needs to be separated from geneticists in its own separate room which connects to the genetics room.
- Surgery needs to be downsized and an observation room added.
- Chief Medical Officer's office needs to be moved to a better location.
- Security office needs to be added.

### Science

- Science is basically non existent, only a room in medical and some weird xeno room above evac.
- Security office needs to be added.

### Security

- Is missing permanent residence.
- Lacks an execution room.
- Needs a secure suit storage separate from armory.
- Needs an evidence room with evidence lockers.

### Command

- Current Command is too small, only containing a small captain's room, bridge, and hop's room.
- Missing a general hallway connecting everything.
- Is missing a meeting room.
- Hop needs to have a waiting line and not a waiting room.
- Missing Telecomms room.

### Engineering

- Weirdly designed and with too many windows.
- Suit storage should be located in the secure room.
- Missing secure storage.
- Technical storage needs to be moved as its own separate room from engineering and with a secure technical storage at the end for high risk boards.
- Room needs to be added for a future possible nuclear reactor (bod).
- CE's office needs shutters and shutter over the front entrance to engineering, with the button in CE's office.
- Security office needs to be added.

### Atmospherics

- Too small, will make fitting piping difficult once added.
- Too unsecure, easy to break in, should have a middle area between front and the atmospherics piping.
- Should have a two way windoor office with the atmos secure storage there, and a shutter to move canisters out to the hallway.

### Service

- Kitchen is way too big, needs be downsized, with a table with shutters connecting it to main hallway, and a secure windoor access to botany, and possibly bar as well.
- Bar needs to be put in its own separate area with a backroom for the bartender.

### Cargo

- Cargo shuttle needs to be adjusted so it doesn't crash into the docking bay
- QM's office needs to be able to clearly see cargo bay, and no door to main hallway
- Needs a security Office.

### AI

- Needs a separate AI upload somewhere in the station
- Needs proper AI core room, glass surrounding the AI won't work.

### Other 

[...]

PR #4142, opened 2020-05-04
Removes Pogstation from rotation

Ranzarth:
### Purpose
It's not good and needs some work

See #4141 for details.

Also lowers boxstation pop requirement to 30 players.


Evan-R-Rogers:
I'd appreciate being told of this before hand, have you discussed this with Doobly or anyone else at all yet?

corp-0:
Most of the issues presented to justify this action are personal opinion. If anything, Pog needs tweaking and maybe set the required population to choose the map to something like less than twenty.

Ranzarth:
Doobly will do it himself

PR #4148, opened 2020-05-05
Refactor Electrocution, Add Electrocution Overlay

NoooneyDude:
### Purpose
Refactors the old Electrocution class into an Electrocution object that can be passed through an IElectrocution interface. This interface is currently implemented by LivingHealthBehaviour as a basic mob-damaging function, but this logic is extended in PlayerHealth, specific to humanoid players. This adds things like better resistance calculation, stun, SFX etc that are specific to players.

Also adds a maximum damage constant (arbitrarily set to 100) so that cutting 760 kV isn't instant death (some players didn't like that). However, at maximum damage of 100, the character still becomes unconscious but is able to move around (assuming started with full health) for a short while. A small compromise.

**Update**
- Adds mouse cable-chewing behaviour.
- Converts BurningDirectionalOverlay to a generic PlayerDirectionalOverlay
- Adds electrocution overlay

Hopefully there are no atrocious programming calamities that I introduce... no partial classes today...


UnityProject/Assets/Scripts/Health/IElectrocutable.cs
```diff
@@ -0,0 +1,12 @@
...
+/// Implement this on a component to cause that component to do something when electrocuted.
+/// It's up to the component what it does in response to electrocution.
+/// </summary>
+public interface IElectrocutable
+{
+	/// <summary>
+	/// Expose this component to electrocution.
+	/// </summary>
+	/// <param name="exposure">information on the exposure</param>
+	ElectrocutionSeverity Electrocute(Electrocution exposure);
```
chairbender: If the contract is "do whatever it wants", then why have a return value? What is the contract for the return value? IMO it should just be void. The logic relying on the return value should be moved into the implementation or the class that's consuming the instance of IElectrocutable.

For example of a possible issue, I notice that ElectrocutionSeverity is a return value, but it's also a field on the exposure. My implementation could return None but could also modify exposure to have a severity of Mild (which is also why I suggest making Electrocution immutable and removing severity from Electrocution).

UnityProject/Assets/Scripts/Health/Electrocution.cs
```diff
@@ -1,162 +1,50 @@
...
+	/// <summary> Exposure to voltage like that of a small battery (no consequential effects) </summary>
+	Mild,
+	/// <summary> Exposure to high voltage - akin to an electric fence (drop actively held item, minor burns) </summary>
+	Painful,
+	/// <summary> Exposure to very high voltage (stun, severe burns, death possible) </summary>
+	Lethal
+}
 
-	private const int NON_INSULATED_ITEM_RESITANCE = 20000; // 20 kilo Ohms
+public class Electrocution
```
chairbender, line 22: If an electrocution is supposed to indicate an external exposure to electricity, then this object should be immutable and all information on the exposure should be present upon construction. There should be no ability or need to modify / set fields. 

Adding the ability to modify the details of the exposure by any implementation of IElectrocutable will add complexity / confusion to this process, and I don't see why it should be needed anyway...if you're exposing a bunch of things to electricity they shouldn't need to communicate between each other, other than knowing how strong the exposure is in terms of some quantitative value.

UnityProject/Assets/Scripts/Health/IElectrocutable.cs
```diff
@@ -0,0 +1,12 @@
+/// <summary>
+/// Implement this on a component to cause that component to do something when electrocuted.
+/// It's up to the component what it does in response to electrocution.
+/// </summary>
+public interface IElectrocutable
```
chairbender: What's the purpose of this interface if only one component is implementing it? Can anything other than players be electrocuted? Even if so, that system isn't currently implemented (current logic only electrocutes LivingHealthBehavior) so we don't have a good idea if the classes involved are designed according to those needs. 

I would suggest adding logic which tests out this / implements IElectrocutable on another component so we can see if there's any issues with the way these classes are designed.

UnityProject/Assets/Scripts/Health/Electrocution.cs
```diff
@@ -1,162 +1,50 @@
...
-	/// <para>Lethal: high voltage line (stun, severe burns, death possible)</para>
-	/// </summary>
-	public enum Severity
-	{
-		None,
-		Mild,
-		Painful,
-		Lethal
-	}
+	/// <summary> No potential difference (no side effects) </summary>
```
chairbender, line 8: Severity seems very specific to humans, why is it an enum and not just part of the internal implementation in PlayerHealth / LivingHealthBehavior? Then those components can decide how they want to respond based on the strength of the electrocution. I don't think this should be part of Electrocution,, each implementation should be able to set the range of their behavior.

UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
```diff
@@ -694,6 +694,87 @@ public void UpdateClientBodyPartStats(BodyPartType bodyPartType, float bruteDama
...
+
+	/// <summary>
+	/// Electrocutes a mob, applying damage to the victim depending on the electrocution power.
+	/// </summary>
+	/// <param name="electrocution">The object containing all information for this electrocution</param>
+	/// <returns>Returns an ElectrocutionSeverity for when the following logic depends on the elctrocution severity.</returns>
+	public virtual ElectrocutionSeverity Electrocute(Electrocution electrocution)
+	{
+		float resistance = ApproximateElectricalResistance(electrocution.Voltage);
+		electrocution.shockPower = Electrocution.CalculateShockPower(electrocution.Voltage, resistance);
```
chairbender: I think this just shows how severity can just be part of the internal implementation in LHB / PlayerHealth. Just check the range and call a different virtual method depending on the severity, IDK if you even need an enum.

UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
```diff
@@ -694,6 +694,87 @@ public void UpdateClientBodyPartStats(BodyPartType bodyPartType, float bruteDama
 		}
 	}
 
+	/// ---------------------------
+	/// Electrocution Methods
+	/// ---------------------------
+	/// Note: Electrocution for players is extended in PlayerHealth deriviative.
```
chairbender, line 701: So PH and LHB have different logic for electrocution...how can I test what happens when something that's just an LHB gets electrocuted? Seems like all I can do is electrocute players so I'd never be able to know if this implementation works.

UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
```diff
@@ -694,6 +694,87 @@ public void UpdateClientBodyPartStats(BodyPartType bodyPartType, float bruteDama
...
+		return electrocution.severity;
+	}
+
+	/// <summary>
+	/// Finds the severity of the electrocution.
+	/// In the future, this would depend on the victim's size. For now, assume humanoid size.
+	/// </summary>
+	/// <param name="electrocution">The object containing all information for this electrocution</param>
+	/// <returns>Severity enumerable - None, Mild, Painful, Lethal</returns>
+	protected virtual ElectrocutionSeverity SetElectrocutionSeverity(Electrocution electrocution)
```
chairbender: If anything this should just be a "GetElectrocutionSeverity", it should not modify Electrocution as a side effect. Avoiding side effects and state mutation where possible helps keep complexity down.

UnityProject/Assets/Scripts/Input System/InteractionV2/TileInteraction/ElectricalCableDeconstruction.cs
```diff
@@ -53,9 +53,10 @@ public override void ServerPerformInteraction(TileApply interaction)
 			// Electrocute the performer. If shock is painful enough, cancel the interaction.
 			ElectricityFunctions.WorkOutActualNumbers(ElectricalData.InData);
 			float voltage = ElectricalData.InData.Data.ActualVoltage;
-			var electrocutionSeverity = new Electrocution().ElectrocutePlayer(
-				interaction.Performer, interaction.WorldPositionTarget, "cable", voltage);
-			if (electrocutionSeverity > Electrocution.Severity.Mild) return;
+			var electrocution = new Electrocution(voltage, interaction.WorldPositionTarget, "cable");
+			var performerLHB = interaction.Performer.GetComponent<LivingHealthBehaviour>();
+			var severity = performerLHB.Electrocute(electrocution);
+			if (severity > ElectrocutionSeverity.Mild) return;
```
chairbender: I'm not clear what effect the severity has on the electrocution logic. If there needs to be different behavior based on that, why not put that into LHB / PlayerHealth?

chairbender (changes requested):
Great documentation.

I think you should consider some changes to the architecture of how electrocution works...If this is going to be something implemented by multiple components (as the interface implies), then we should be careful about the design of that interface and the system as a whole. That usually means implementing more than one case that demonstrates the system in action. 

Right now the only case implemented is shocking a player trying to deconstruct a cable, which doesn't even necessitate this interface. You should prove the need for the interface while also proving that it supports some various use cases. It's the ultimate test case for building new systems.

Give my comments a read and lmk if you get stuck.

NoooneyDude:
Hi @chairbender , thank you for the review. I found returning an ElectrocutionSeverity value was useful. For example, some interactions may or may not proceed based on the severity. Take the wirecut example. If the electrocution was only mild, there's no reason to stop the interaction, but if it was painful, then yes, it does make sense to cancel it - the performer has probably dropped the tool in surprise. This value is also useful for the upcoming electrified grille interaction, with similar reasoning - if the electrified grille merely gave some tingling in the fingers, there's no reason to stop the unsecuring or deconstruction interaction.

You suggest that any logic that depends on the strength of the electrocution should exist inside LHB - for the most part, this is the current behaviour (stun, SFX, etc.) - except for interaction logic. I'm not quite sure how, or why, interaction logic should reside there. Could you perhaps explain this a bit?

As for the IElectrocutable interface itself, I agree - electrocution is only really applicable to LHB and its deriviatives, meaning there is only one implementation of the interface. I was planning for non-player mobs to be electrocuted too, such as mice chewing cables, but I've seen that they use the SimpleAnimal component which extends LivingHealthBehaviour anyway. I'll remove the interface.

The ElectrocutionSeverity enum. You're right in saying that at the moment, its only use is for players.  I was leaving the door open for non-player mobs to do something with the electrocution severity (similar to my reasoning in the first paragraph) too, where the severity at a given voltage depends on the mob's size (or ultimately resistance). I think where the enum resides - or if it should even exist - depends on whether interaction logic or other external code still takes into account the severity.

I was somewhat averse to adding variables to LHB - I'm not even sure myself what my reasoning was at the time... it gave rise to weird side effects like mutable variables inside what should be an immutable Electrocution class. I'll fix that.

Thanks!

chairbender:
Thanks for clarifying @NoooneyDude , then if that enum is only going to be used for communicating the effect on a living thing, it makes sense how you set it up / used it. I was thinking electrocution might be applied to objects (like a piece of machinery blowing up), so in that sense it would make sense to me for it to just be a quantifiable value rather than a living-thing-specific enum. It's fine as an enum then for this.

Interaction logic can live outside LHB like how you had it, that makes sense.

NoooneyDude:
Awesome, sounds good. I too was thinking electrocution could be applied to other things too, like fire is. But I won't be working on that, and it isn't any effort to re-add the interface if needed.

Bod9001:
ready to review?

NoooneyDude:
Yep, ready to review.

Bod9001:
it's been through a review so should be good now, and if there's any problems can always revert

PR #4154, opened 2020-05-06
Changes to tile damage calculations by corp-0, and tweaks to armor and health values.

### LINKED ISSUE #4137: Tile Armor and Resistances aren't applied correctly
## Description
In TilemapDamage, the DoDamageInternal method applies damage to tiles. At the top, it uses the Armor field of the BasicTile (if the tile is a BasicTile). But lower in the method, it calls various Add(Blah)Damage methods which also apply a hardcoded armor value (like WALL_ARMOR). Also, only the "Indestructable" resistance flag on the tile is checked, other resistance flags are ignored.

For this ticket:
1. Remove all hardcoded Armor settings (WALL_ARMOR and the like). Armor should only come from the BasicTile.Armor field.
2. Add the logic to properly check all the possible Resistances from BasicTile.Resistances.
3. Ensure appropriate tiles have armor values filled in (you can reference TG). Because we were previously relying on hardcoded values, some of the tile assets may have 0 armor, which may make them fragile. As a start, you could set them to use the same hardcoded values that were deleted in step 1 if they don't have any filled in.

PetMudstone:
### Purpose
This PR contains tweaks to TilemapDamage.cs and BasicTile.cs **by corp-0** so that there are no longer hardcoded values for armor and resistances for all kinds of tiles. Changes to the aforementioned scripts also make fire deal fire damage to windows and grilles instead of melee damage, for some reason.

This PR also tweaks the armor, health, and resistance values of **many** tiles in their assets now that said values are no longer hardcoded. I used /tg/ values where applicable, and used the old hardcoded values if /tg/ values were not available.

May or may not resolve #4137.

### Notes:
**Again, the changes to scripts were done by corp-0 (Gilles on Discord). I just inserted values into the assets for all the tiles.**

Also, fires still break windows within seconds. Most windows should actually break even faster since the health (integrity) values I added are lower than the hardcoded ones.

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/TilemapDamage.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/TilemapDamage.cs
@@ -36,12 +36,6 @@ public class TilemapDamage : MonoBehaviour, IFireExposable
 	//TODO: this needs a refactor. BaseTile has useful fields that should be used instead of this.
 	//also this implementation isn't designed for tile variations, essentially being limited to one tile kind per layer
 
-	private const int MAX_TABLE_DAMAGE = 50;
-	private const int MAX_WALL_DAMAGE = 100;
-	private const int MAX_FLOOR_DAMAGE = 70;
-	private const int MAX_PLATING_DAMAGE = 100;
-	private const int MAX_WINDOW_DAMAGE = 100;
-	private const int MAX_GRILL_DAMAGE = 60;
 	private static readonly float TILE_MIN_SCORCH_TEMPERATURE = 100f;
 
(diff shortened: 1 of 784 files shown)


corp-0:
A couple things to mention:
My changes to TilemapDamage are just contributing to the spaghetti and that class should be properly refactored to take even more data from BasicTile scriptable objects. I added more fields to BasicTile which are unused at the moment.

Our problems with fire are quite interesting, because even when we’re effectively reducing damage with the armor values, for tiles objects and creatures, it seems to do way more damage than what it should. Need someone to take a look and see what we’re doing different from tg so we can balance the game using tg numbers.

PR #4174, opened 2020-05-08
Removes Maint Access from non essential roles.

ThatDan123:
Now that hacking is on staging, rebalancing access is needed.

As having maint access means players would have to get through only one door in order to break into higher security areas through maints.

Removes maint access from most jobs.


ReiderJack:
Not only assistants don't have access to maint tho. Only security, engineering, cargo techs/qm, janitor, heads have access to maint, as i remember.

PR #4181, opened 2020-05-09
adds special respawns for admins

corp-0:
### Purpose
This PR will allow admins to respawn any dead player but now changing their occupation and gear to any occupation declared in a list of jobs for admins.

### Notes:
- Adds Centcomm Commander, Centcomm Officer, Centcomm Intern and Death Squads occupations. Only way to select them is by this new button.
- Only dead players can be respawned and their body will keep existing in the world (just like regular respawn).
- Once the job is changed, you can't go back to your normal job.
- Players will respawn at their designed spawnpoints (arrivals for the jobs that have no spawn point). I have no idea how this will work with subscenes, but eventually Death Squads and Centcomm should spawn at Centcomm.
- Tested in editor. I can't test in MP because builds are broken (Hologram shader).

Some pictures:
image
image
image
image
image
Adding new jobs is just as easy as adding Occupations to this list.


UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
```diff
@@ -474,8 +474,13 @@ public void CmdRespawnPlayer()
 	}
 
 	[Server]
-	public void ServerRespawnPlayer()
+	public void ServerRespawnPlayer(string occupation = null)
 	{
+		if (occupation != null)
+		{
+			playerScript.mind.occupation = Resources.Load<Occupation>($"ScriptableObjects/Occupations/"+occupation);
```
Bod9001: could you use the SingletonScriptableObject to get the references from, you will have to add it to the SingletonScriptableObject  loader script

Bod9001: ScriptableObjectSingletons in there

corp-0: Yes, I can do that. Would it work if I add a scriptable object that contains a list of occupations? So we don’t have to add all jobs to the Singleton list

corp-0: done

corp-0:
I'm taking the Occupation reference from a singleton list now. Also fixed a problem I didn't notice before, when trying to respawn if you have no spawn point mapped you would just bug out of existence. It defaults to arrivals now but still can return null if arrivals hasn't been mapped and logs the error.

Bod9001:
all good

PR #4182, opened 2020-05-09
Revert "Camera Effects"

DooblyNoobly:
This PR breaks builds. Please be more careful

Reverts unitystation/unitystation#4164

PR #4183, opened 2020-05-09
Fixes Camera Effects

ThatDan123:
Reverts unitystation/unitystation#4182

And removes the bad shaders.

Tested in editor and has been built.


daalta:
Nice!

The glitch effect has a license, did you write the other two shaders yourself?

ThatDan123:
See text files for references.

Bod9001:
looks good, If it becomes an issue we can sort it out

PR #4185, opened 2020-05-09
Boxstation Updates

ThatDan123:
Added fire locks

Added some more blastdoors 

Added more items in different rooms.

Readds Caverna and Palace to the awaysites SO and build list.


ThatDan123:
Ah just remembered ive got to put random loot in.

PR #4186, opened 2020-05-09
Tests build settings match the new scenelist scriptableobjects

CodeKalias:
### Purpose
Test that scenes listed in the new MainStationListSO, AwayWorldListSO and AsteroidListSO are correctly included and enabled in the build settings

### Notes:
Will need updating if we change how the scriptableobjects work


ThatDan123:
well at least we know your tests works :p 

fixed in #4185

Bod9001:
what is the check failing on?

Issue #4184, opened 2020-05-09
Staging Issues


ThatDan123:
![image](https://user-images.githubusercontent.com/56727168/81470670-dc6dab80-91e3-11ea-8e18-2e719e1d9360.png)

![image](https://media.discordapp.net/attachments/312454684021620736/708608341483454464/aaaaaaa.png?width=1203&height=143)

-Error messages when joining a server, might only happen when the round hasnt started(or is restarting), late joining seems to not have these errors.

![image](https://user-images.githubusercontent.com/56727168/81471072-83534700-91e6-11ea-85a7-5480bd73c183.png)
-Doing a round restart locks you to this screen, server still works and if you join after it works, but players already in game cant get past this screen.
-Same errors as the pictures above.

![image](https://media.discordapp.net/attachments/312454684021620736/708613115574485022/unknown.png?width=661&height=562)
-Late joining as a ghost people have no clothes (visual bug).
-rejoining doesnt fix the issue and causes any people who joined after you to also have no clothes.

https://imgur.com/a/FJe3uA3
-going into a different matrix seems to keep pushing you back
Seems like exploration sites are ok though.
Ghost only, players seem fine, however:

![image](https://user-images.githubusercontent.com/56727168/81470699-18a10c00-91e4-11ea-98a2-b84fbdf33ce1.png)
-Teleporting to a player that's on a shuttle sends you somewhere else.

![image](https://user-images.githubusercontent.com/56727168/81470723-41c19c80-91e4-11ea-846b-46095e252b3d.png)
-Logged off players seem to go to weird coords. (The two middle ones)
-And when they rejoin they lose all their equipment (not a visual bug).

![image](https://user-images.githubusercontent.com/56727168/81470760-746b9500-91e4-11ea-8a7c-c8ebe54a2cbc.png)
-Not sure whats caused this, but a player has had their sprite stuck to the forwards version.

CodeKalias:
Leaving and rejoining, appear at proper location, teleport to 0,0 shortly after
Errors due to no gear showing up on client
![image](https://user-images.githubusercontent.com/7896673/81471182-46d41b00-91e7-11ea-9cfe-e66461afb069.png)
![image](https://user-images.githubusercontent.com/7896673/81471189-4f2c5600-91e7-11ea-85ef-b5417b5cc024.png)
I put down a mask in the shuttle before the rejoin, it thinks it is here as well
[...]

PR #4191, opened 2020-05-10
Fixes rejoining issues to do with subscenes

DooblyNoobly:
- Changes the way things are synced so the new client receives the data after it becomes an observer of an object

- Created a Rejoining UI popup when you are rejoining a round

- Fixed an issue where server list was not reporting the current mainstation that was loaded

^ all from #4184


DooblyNoobly:
I ran testrunner locally

PR #4194, opened 2020-05-10
Fixes subscene restart round issues

DooblyNoobly:
- Creates a dedicated dictionary on NetworkServer that lists what scenes the NetworkConnection is allowed to observe. This is then used by OnRebuildObservers and OnCheckObserver methods on NetworkBehaviour

- Ensures the scene list is properly synced

- Sets all music to streaming to save mem usage

Fixes more stuff in #4184


DooblyNoobly:
completed testrunner locally

PR #4200, opened 2020-05-11
Prevents random loot from doing an occasional infinite loop

corp-0:
Hot fix for a very stupid error that could end with an infinite loop on every round start.

This implementation will be dropped eventually for a real weighted random selection, once I find a good approach.


chairbender:
needs a null check at the top of SpawnItems, poolData may be null, so you need to avoid the possible NRE
`if (poolData == null) return;`

chairbender:
👍

Issue #4138, opened 2020-05-03
Object Lifecycle V2


chairbender:
## Description
There are a number of issues with the object lifecycle system.
- All these custom interfaces which don't even seem to work reliably (OnSpawnServer, OnSpawnCLient, OnDespawnClient, etc...)
- Bloating network components with all those spawn hooks
- Not using standard mirror / unity hooks.
- Pooling everything without regard for what actually needs it
- Pooled objects ending up interactable (hiddenpos hijinks players recently discovered are possible)
- Getting confused about what hooks are called in what order (mirror can't guarantee that unity events are called before it, like Awake, but mirror CAN guarantee its own hooks are called in a particular order)
- Pool grows unbounded! (memory leak possibility)
- On clientside, objects in pool are kept active! They should be deactivated on despawn, activated on spawn.

Let's fix these so the lifecycle is more reliable and easier to integrate with components. I want to minimize confusion / bloat caused by the inconsistent ordering of event hooks, make it more clear the order in which things happen and provide ways to guarantee order (for example if one networked component depends on another component having already been initialized).

Make sure to update the lifecycle docs

chairbender:
I think for step 1, I'll do a PR where I 1) shrink / limit the pools to a constrained size and only pool specific objects that actually benefit from it 2) ensure objects in the clientside / serverside pools are deactivated (but still reactivate them properly so they don't miss any messages when they are respawned from the pool).

According to mirror guys, using Unspawn is a good idea for it - you pool the object on the server side and use Unspawn, which keeps it around server side but destroys it client side, then when you go to spawn it, you reuse the object, the server doesn't need to instantiate it, and the client does, so it decreases the load on the server when creating a new networked object.

PR #4202, opened 2020-05-11
Memory Leak Fix and Better Object Pooling

### LINKED ISSUE #4167: Memory leak
There's a nasty memory leak that seems to happen during round restarts

### Steps to Reproduce

1. Launch client and join a populated server, join as observer
2. Open process explorer and look at ram usage
3. See memory usage increasing after every round start

chairbender:
### Purpose
Fixes #4167 
Partial implementation of #4138 

### Notes:
Object pool now actually has configured limits / which objects are pooled (in PoolConfig singleton SO). At the moment only projectiles are in the pool as they are often created / destroyed, but I have tested it with networked objects (bullets are non networked). The pool is configured to be very small right now, probably should be expanded / limits increased based on observations. Object pool now also correctly clears out everything in it and destroys it if necessary.

Clientside / server side pools work a bit differently - networked objects can be pooled on server side but not on the clientside - we use Mirror's Unspawn and Spawn to handle this. But as I said earlier no network objects are configured to be pool-able at the moment because I couldn't think of any obvious candidates.

No more clientside spawn hooks also. This is part of what will eventually be a reworking of how lifecycle hooks / networking init stuff works to make them more usable. I refactored those to just use OnEnabled at the moment, which works because unpooled and newly instantiated objects are always deactivated / activated. Documentation will need updating, I'll get around to it. 

This should be tested on staging before rolling it out since it's kinda big.

Main entry point for the new Pooling logic is Spawn.Server, Spawn.Client, Despawn.Server, Despawn.Client


DooblyNoobly:
very nice, doesn't fully stamp out the mem leak but significantly reduces it. Merging now and I'll take a look at the remainder of the leak today.

PR #4218, opened 2020-05-11
minor changes to the scotsman accent

Fiftyllama:
### Purpose
tweaks the Scotsman accent, adds a few words, replaces some that are not appropriate for simple word substitution, and removes some entirely for accuracy


duskwalker05:
was downricht supposed to be spelled like that or is it downright?

Fiftyllama:
It's downright, I made a Booboo. I'll fix it next time, thanks

PR #4221, opened 2020-05-12
Fixes a bunch of bugs

DooblyNoobly:
- Adjusts produce nutrition rates
- Fix UpdateClientValue NRE
- Added EnsureInit to PlayerDirectionalOverlay
- Fixes Alien Egg, Mob Attack and OveryLayCrits bugs
- Joining and Alien egg fixes
- Adds RTT checks to mob attack so mobs don't attack too quickly on slower connections


DooblyNoobly:
completed the testrunner locally with all green

PR #4224, opened 2020-05-12
AILAWS

Buldinn:
A funny minute long skit by Cuboos on Soundcloud.

If timed correctly with an admin nuclear detonation would be a good round end sound clip. (59 seconds before nuke goes off play the soundfile)


ThatDan123:
https://soundcloud.com/cuboos/ailaws link for future reference.

daalta:
I think we should be careful when adding large (audio) files. Too many and our updates will take a long time.

Maybe this could be compressed more heavily in the import settings? I did that a few months back for most large audio files and no one seems to have noticed 😘

fomalsd:
it's a 7 megabyte .wav – can you compress it to ogg with low bitrate so it's like 2mb?
currently on limited connection and it really hurts to download extra 60mb of music bundled with each and every new build. (see #4236)

fomalsd:
stale PR, closing

PR #4226, opened 2020-05-12
Finishing with firelocks maybe hopefully perhaps i hope yeah

Aranclanos:
### Purpose
Fixes firelocks not closing if there's a nonpassable on the same tile (ex: tables, other doors)
Adds an animation for fire alarms when they activate.
Fixes fire alarms performing the pressure check on the wall instead of the floor tile in front of them.
Adds a sound for fire alarm activation.


UnityProject/Assets/Scripts/Objects/FireAlarm.cs
```diff
@@ -12,11 +12,17 @@ public class FireAlarm : NetworkBehaviour, IServerLifecycle, ICheckedInteractabl
 	public float coolDownTime = 1.0f;
 	public bool isInCooldown = false;
 
+	public SpriteHandler spriteHandler;
+	public Sprite topLightSpriteNormal;
+	public SpriteSheetAndData topLightSpriteAlert;
```
daalta, line 17: Do these variables cause a Unity compilation warning? 

https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0649

PR #4233, opened 2020-05-13
Big bug fix pr

DooblyNoobly:
- Robust init for SpriteHandler, may it never NRE again
- fixed clothing item NRE
- fixes main station server list reporting
- GUI Hacking NRE check
- Null check examine message
- Another attempt at fixing Overlay NRE
- Attempt at staggering NetworkObject awake messages so not to cause ph…
- Fix huge syncing flaw that resulted in naked station on rejoin


DooblyNoobly:
Ran tests locally with all green ticks

PR #4234, opened 2020-05-13
Adds Pixel Perfect cam zoom and reverts spritehandler changes

DooblyNoobly:
- Adds pixel perfect zoom
- Reverts spritehandler because my earlier changes broke a lot of shit
- Cleaned up SubManager code so its easier to read


DooblyNoobly:
ok made a fix and now all tests pass locally

PR #4237, opened 2020-05-13
Enable compression for linux builds again

AutumnNova:


Bod9001:
The reason why it is like that is so you can get Stack traces on the server

Bod9001:
https://discordapp.com/channels/273774715741667329/312454684021620736/710124120821465138

PR #4245, opened 2020-05-14
Added Pixel Perfect Camera from Unity.2D.pixel-perfect package

DooblyNoobly:
Adds the pixel perfect camera and adjusts the PPU for zoom in / out. Moved to update method as it was a lot smoother and worked with Matrix Move.


DooblyNoobly:
50 green ticks on the test runner locally

PR #4260, opened 2020-05-16
xeno food and crafting setup (and truly fixes the whole... plants not giving you food thing)

VernaSaurus:
mmmmm tasty aliens


VernaSaurus:
It looks like, for whatever reason, the hydroponics sprites don't update on client. This is probably the cause of the invisible wires on client glitch too.

PR #4275, opened 2020-05-17
Buckle UI button fix for Despawn

### LINKED ISSUE #3694: Disappear with chair
So, with any chair you can buckle yourself to, you can disassemble it while buckled to become invisible on the spot. Also, invulnerable. However, you can be hit by things flying past you, or bullets. You can also get flung to the escape shuttle (top right of the map, I believe), which has nothing to collide with, so you sort of just drift down and to the left.

https://www.youtube.com/watch?v=-Xz3giO3S0Y
### Steps to Reproduce

1.  Buckle to a chair
2.  disassemble the chair
3.  unbuckle to reappear
4. works with other people, dead bodies etc
5. if the chair becomes sideways and a dead body is "disappeared" on it, it will never be found again because you cannot build sideways chairs and the dead body cannot unbuckle itself

additional note: you can "summon" the person using the bug by building the same chair, given that they hadn't sat on a turned chair.

ReiderJack:
fixes #3694
Could not reproduce this issue, but this event should prevent it.

### DIFF
--- a/UnityProject/Assets/Scripts/Furniture/BuckleInteract.cs
+++ b/UnityProject/Assets/Scripts/Furniture/BuckleInteract.cs
@@ -58,6 +58,8 @@ public class BuckleInteract : MonoBehaviour, ICheckedInteractable<MouseDrop>, IC
 		var playerMove = drop.UsedObject.GetComponent<PlayerMove>();
 		playerMove.ServerBuckle(gameObject, OnUnbuckle);
 
+		gameObject.GetComponent<Integrity>().OnServerDespawnEvent += playerMove.Unbuckle;
+
 		//if this is a directional sprite, we render it in front of the player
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/Furniture/BuckleInteract.cs
```diff
@@ -58,6 +58,9 @@ public void ServerPerformInteraction(MouseDrop drop)
 		var playerMove = drop.UsedObject.GetComponent<PlayerMove>();
 		playerMove.ServerBuckle(gameObject, OnUnbuckle);
 
+		var objectIntegrity = gameObject.GetComponent<Integrity>();
```
Bod9001: do you think it's worth having an reference for it so don't have to get component all the time?

PR #4279, opened 2020-05-18
undo bods change as it breaks fire extingushers

AutumnNova:
gaddamnit bod


Bod9001:
Did you test If it works with this change?

PR #4280, opened 2020-05-18
traitor objectives, firelocks on outpost

AutumnNova:
also adds the vendor refill to cargo
traitor objectives are to steal the hos' gun, cap's gun, adv magboots or abl trenchcoat


PetMudstone:
What happens if an objective item isn't available on the map? The ablative trenchcoat doesn't exist on Outpost or Box at the moment.

AutumnNova:
i added it to outpost
but you just cant complete it if the item doesnt exist

PR #4283, opened 2020-05-18
Fire extinguisher cool down

ReiderJack:
Checking in Will Interact does not allow clients to interact at all.


UnityProject/Assets/Scripts/Items/Others/FireExtinguisher.cs
```diff
@@ -71,15 +71,12 @@ public bool WillInteract(AimApply interaction, NetworkSide side)
 	{
 		if (!DefaultWillInteract.Default(interaction, side)) return false;
 
-		if (reagentContainer.ReagentMixTotal < reagentsPerUse || safety) return false;
-
-		if (isCoolDown) return false;
```
fomalsd, line 76:
```cs
if (side == NetworkSide.Server) {
  if (reagentContainer.ReagentMixTotal < reagentsPerUse || safety) return false;		
  if (isCoolDown) return false;
}
```
would also work

PR #4284, opened 2020-05-18
Fix ability to pass through certain NPCs

CodeKalias:
### Purpose
NPCs were ignoring the Passable flag set in their prefab/scene instance as it was being overwritten
Now they remain properly passable.

### Notes:
- [x] The PR has been tested in multiplayer (with 2 clients and late joiners)


fomalsd (approved):
looking good, thanks!

PR #3862, opened 2020-04-08
AutoDrobe Part 1.5: Werewolf costumes and Populator Tweaks. Also Magboots in closets.

PetMudstone:
### Purpose
This PR adds werewolf costumes **ported over from [Goonstation](https://github.com/goonstation/goonstation)** at the request of 0ozymandias on Discord. 
This PR also adds a few items to the AutoDrobe such as:
- The new werewolf costumes (and masks).
- Normal (non-skirted) versions of the clown and mime uniforms and masks.
- A maid apron.
- A black skirtsuit.

This PR also makes magboots appear in the inventories of relevant closets/lockers. Standard magboots appear in engineering lockers, advanced ones in CE lockers, and syndie ones in syndie closets.

### Please make sure you have followed the self checks below before submitting a PR:
- Code is sufficiently commented
- Code is indented with tabs and not spaces
- [x] The PR does not include any unnecessary .meta, .prefab or **.unity (scene) changes**
- [x] The PR does not bring up any new compile errors
- [x] The PR has been tested in editor
- [x] Any new files are named using PascalCase (to avoid issues on case sensitive file systems)
- Any new / changed components follow the [Component Development Checklist](https://github.com/unitystation/unitystation/wiki/Component-Development-Checklist)
- [x] Any new objects / items follow the [Creating Items and Objects Guide](https://github.com/unitystation/unitystation/wiki/Creating-Items-and-Objects%3A-Now-With-Prefab-Variants) (especially concerning the use of prefab variants)
- [ ] The PR has been tested in multiplayer (with 2 clients and late joiners, if applicable)
- [ ] The PR has been tested with round restarts.
- The PR has been tested on moving / rotating / rotated-before-joining matrices (if applicable)

PR #4287, opened 2020-05-18
Add ForeignAssets folder and put Goon assets inside.

PetMudstone:
### Purpose
This PR adds a ForeignAssets folder in the Assets folder. Inside ForeignAssets is a GoonAssets folder in which assets taken from Goon exist. These assets include the werewolf costume's sprites from #3862, FireAlarm.ogg from #4226, and cannabis sprites from #3587

Included in the GoonAssets folder are unmodified versions of LICENSE.MD and README.MD from /tg/'s "goon" folder.

In relevant folders where .prefabs and .assets point to Goon sprites, I've added .txt files named "zForeignAssets.txt" that state which files use Goon sprites, give the path to said sprites, and what license they are under (in this case, CC BY-NC-SA 3.0 US).

image

### Notes
FireAlarm.ogg and the cannabis sprites were added to GoonAssets after a look in /tg/'s goon folder showed they were imported from Goonstation's 2016 release.

"ForeignAssets" might need a better name.


Bod9001:
do we have the weed sprites for goon?

Jibbajabbafic:
I'd keep the prefabs in their original place, we just need the sprites and sounds separate with the license since we can always switch them out later on.

PetMudstone:
Well, those are all the Goon assets I could find, for now.

Bod9001:
cool beans, if we find anything else that needs to be moved over should be no problem to do it in another PR

PR #4289, opened 2020-05-19
Fixes Server NRE's, Fire Extinguisher and Vendor list empty bug

DooblyNoobly:
- Fixes a bunch of NRE's on the server for updating new players. This occurred because I had not accounted for ghosts. This is fixed now
- Fixed up the cooldown on Fire Extinguisher so it works now. Only 1 burst per second is allowed.
- Fixed the issue where you had to double click a vendor to see its contents


DooblyNoobly:
testrunner is good locally

PR #4293, opened 2020-05-19
finalize cooking stuff

VernaSaurus:
mixing bowls, working grinders, and flour bags, oh my!

there are a few minor changes to outpost's kitchen here, like adding a high capacity water tank for water and modifying the fridge, and also adding a mixing bowl to the table


VernaSaurus:
is someone going to look this over or

DooblyNoobly:
Code looks good. Will throw it on staging

PR #4304, opened 2020-05-21
adds the ability to range limit a bullet type

AutumnNova:
tested with pka, doesnt break it.
limits the flamethrower and tazers range

### DIFF
--- a/UnityProject/Assets/Scripts/Weapons/Projectiles/BulletBehaviour.cs
+++ b/UnityProject/Assets/Scripts/Weapons/Projectiles/BulletBehaviour.cs
@@ -82,9 +82,8 @@ public class BulletBehaviour : MonoBehaviour
 		
 	}
 
-	private IEnumerator countTiles()
+	public IEnumerator countTiles()
 	{
-		Vector2 startPos = gameObject.AssumedWorldPosServer();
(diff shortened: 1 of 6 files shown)


UnityProject/Assets/Scripts/Weapons/Projectiles/BulletBehaviour.cs
```diff
@@ -71,6 +75,20 @@ public virtual void Shoot(Vector2 dir, GameObject controlledByPlayer, Gun fromWe
...
+		if (isRangeLimit);
+		{
+			StartCoroutine(countTiles());
+		}
+		
+	}
+
+	private IEnumerator countTiles()
+	{
+		Vector2 startPos = gameObject.AssumedWorldPosServer();
```
fomalsd: so why do you need `startPos`? it's not used

AutumnNova: this is reusing code from the pka's bullet, dont think its needed for this however so ill remove it

fomalsd: then it's worth removing both from bulletbehaviour and kineticbullet

PR #4311, opened 2020-05-22
Let mappers place items in exact positions

CodeKalias:
### Purpose
Mappers want to do this
image

But currently everything snaps to the middle of a tile at runtime.
So I've added this (it's turned ON for everything by default)
image

Now they can uncheck that on stuff with a CustomNetTransform to make sure it stays where they put it.
image

### Notes
Maintaining object sorting order (see the gloves/multitool/vest) is a different issue that will also need fixing elsewhere.

- [x] The PR has been tested in multiplayer (with 2 clients and late joiners)
- [x] The PR has been tested on a headless server.


fomalsd (approved):
nice, we could also eventually add support for that in dev spawner

PR #4327, opened 2020-05-25
Fix MetaTileMap bounds calculation

CodeKalias:
### Purpose
MetaTileMap GetBounds() would treat an empty layer as a single tile at the origin, causing bounds problems with maps drawn off-center.
Changes the calculation to skip them.

Also fixes an int rounding error in GetWorldBounds() which could result in bounds dimensions being off by 1.
image

### Notes:
Allows proper compression of bounds on stations like Pog, or those trailing bounds behind shuttles.
image


fomalsd (approved):
very nice, thanks!

PR #4328, opened 2020-05-25
makes the easter egg its own matrix

VernaSaurus:
whoops


VernaSaurus:
also contains some files that are currently unused and shouldn't impact anything as far as i know

PR #4334, opened 2020-05-27
Mime spells, Spell system, Action tooltips

### LINKED ISSUE #4098: SerializableDictionary Naming Conflict
`SerializableDictionary.cs` has a naming conflict which causes dozens of Unity compilation warnings.

`Assets\Scripts\Tilemaps\Utils\SerializableDictionary.cs(304,35): warning CS0436: The type 'SerializableDictionary<TKey, TValue>' in 'D:\Repos\UnitystationRepo\UnityProject\Assets\Scripts\Tilemaps\Utils\SerializableDictionary.cs' conflicts with the imported type 'SerializableDictionary<TKey, TValue>' in 'SerializableDictionary, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'. Using the type defined in 'D:\Repos\UnitystationRepo\UnityProject\Assets\Scripts\Tilemaps\Utils\SerializableDictionary.cs'.
`

Related: https://github.com/unitystation/unitystation/issues/2526 (Too many editor warnings)

fomalsd:
Adds a first iteration of spell system - only tested for mime spells, to be expanded
Scr 2020-05-28 в 1 10 47

Adds fully functional mime spells:
- Speech (make or break vow of silence): you can't talk as a mime if you made a vow
- Invisible Chair
- Invisible Box
- Invisible Wall

Adds action tooltips
Scr 2020-05-28 в 1 12 20

Fixes #4098

Bumped Mime progress to 80%


Bod9001:
I had a quick look at the UIaction changes, and they look good

Bod9001:
one issue I found, run to clients let the non-hosted one joint first with mime as high priority, getting started, wasn't ready on the main one then notice that the action buttons appear on the server hosted one and not on the client

UnityProject/Assets/Scripts/Chat/Chat.cs
```diff
@@ -60,7 +60,17 @@ public static void AddChatMsgToChat(ConnectedPlayer sentByPlayer, string message
 		// This step is skipped when speaking in the OOC channel.
 		(string message, ChatModifier chatModifiers) processedMessage = (string.Empty, ChatModifier.None); // Placeholder values
 		bool isOOC = channels.HasFlag(ChatChannel.OOC);
-		if (!isOOC) processedMessage = ProcessMessage(sentByPlayer, message);
+		if (!isOOC)
+		{
+			processedMessage = ProcessMessage(sentByPlayer, message);
+
+			if (player.mind.occupation.JobType == JobType.MIME && player.mind.IsMiming
```
Ostaf, line 69: Just an idea - rather than checking based on JobType and IsMiming specifically, could we have a combination of them set an effect on the player instead (muted?) then check for that here? That would then allow us to reuse the functionality for other things - a gag, a "silence" spell etc.

I'm not really familiar with all the code so may have missed something

fomalsd: being gagged/muted would probably require a different warning message. I'd revisit that when we introduce more reasons for being unable to speak and/or a decent player effects system

UnityProject/Assets/Scripts/Player/Mind.cs
```diff
@@ -52,9 +94,24 @@ public static void Create(GameObject player)
...
+		if (occupation != null)
+		{
+			int i = 0;
+			foreach (var spellData in occupation.Spells)
+			{
+				var spellScript = spellData.AddToPlayer(playerScript);
+				Spells.Add(spellScript);
+			}
+
+			if (occupation.JobType == JobType.MIME)
```
Ostaf: I think it'd be better if this was set by the job somehow rather than the mind so conceptually they're separate which will be more maintainable in the future. Job specifics shouldn't be in the domain of the Mind, otherwise you could end up with a long list in here in the future.

You could have a list of properties defined in a job which they set on the body/mind when initialised (similar to the inventory populators for example).

UnityProject/Assets/Scripts/Chat/Chat.cs
Ostaf: it would, but the message could be coupled with the effect - that's fine, just from experience having implementation specific details outside of their area will make it harder to maintain in the future. Refactoring when you need too is valid though

UnityProject/Assets/Scripts/Player/Mind.cs
fomalsd: yeah, I'll try integrating that into Occupation ScriptableObject

fomalsd:
It's fixed now - had to add support for remote spell action instantiation

Bod9001:
I did a quick test and everything seems okay, I'm good for merging

PR #4339, opened 2020-05-30
Discord Webhook

ThatDan123:
Sets up a one way messaging system, which takes OOC messages in a server and shows them in a discord channel.

Also has admin ahelp integration (one way still).

Also has an announcement system to post a message when a round is starting and when one has ended.

All configured in the config.json

-When setting up multiple unity game servers using the same discord channel use a separate webhook for each server, due to discord rate limiting.

### DIFF
--- a/UnityProject/Assets/Scripts/Chat/Chat.cs
+++ b/UnityProject/Assets/Scripts/Chat/Chat.cs
@@ -86,7 +86,7 @@ public partial class Chat : MonoBehaviour
 			Instance.addChatLogServer.Invoke(chatEvent);
 
 			//Sends OOC message to a discord webhook
-			DiscordWebhookMessage.SendWebHookMessage(Urls.DiscordWebhookOOCURL, message, chatEvent.speaker, ServerData.ServerConfig.DiscordWebhookOOCMentionsID);
+			DiscordWebhookMessage.SendWebHookMessage(DiscordWebhookURLs.DiscordWebhookOOCURL, message, chatEvent.speaker, ServerData.ServerConfig.DiscordWebhookOOCMentionsID);
 
 			return;
 		}
--- a/UnityProject/Assets/Scripts/Chat/Chat.cs
+++ b/UnityProject/Assets/Scripts/Chat/Chat.cs
@@ -2,6 +2,8 @@
 using System;
 using AdminTools;
 using Tilemaps.Behaviours.Meta;
+using DiscordWebhook;
+using DatabaseAPI;
 
 /// <summary>
 /// The Chat API
@@ -82,6 +84,13 @@ public partial class Chat : MonoBehaviour
 			}
 
 			Instance.addChatLogServer.Invoke(chatEvent);
+
+			//Sends OOC message to a discord webhook
+			if (!string.IsNullOrEmpty(ServerData.ServerConfig.DiscordWebhookOOCURL))
+			{
+				DiscordWebhookMessage.SendWebHookMessage(ServerData.ServerConfig.DiscordWebhookOOCURL, message, chatEvent.speaker, ServerData.ServerConfig.DiscordWebhookOOCMentionsID);
+			}
+
 			return;
 		}
 
--- a/UnityProject/Assets/Scripts/Chat/Chat.cs
+++ b/UnityProject/Assets/Scripts/Chat/Chat.cs
@@ -86,10 +86,7 @@ public partial class Chat : MonoBehaviour
 			Instance.addChatLogServer.Invoke(chatEvent);
 
 			//Sends OOC message to a discord webhook
-			if (!string.IsNullOrEmpty(ServerData.ServerConfig.DiscordWebhookOOCURL))
-			{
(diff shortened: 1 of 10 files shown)


ThatDan123:
DNM, still gotta do some testing.

UnityProject/Assets/Scripts/Chat/DiscordWebhook.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+	public static class DiscordWebhookMessage
+	{
+		public static byte[] Post(string url, NameValueCollection pairs)
+		{
+			using (WebClient webClient = new WebClient())
+				return webClient.UploadValues(url, pairs);
+		}
+
+		public static void SendWebHookMessage(string url, string msg, string username, string mentionID = null)
```
Ostaf: I'd suggest refactoring this slightly.

Perhaps have an enum (or similar) of channels (OOC, Admin, etc.) to pass in rather than a url, and then have this method work out the appropriate url. You could then also pull the three repeated string.IsNullOrEmpty calls into here - that'd also abstract away the implementation detail of the URL from what the callers of the method really care about which is the channel, and allow you to make one message go to multiple channels in the future.

UnityProject/Assets/Scripts/Chat/DiscordWebhook.cs
```diff
@@ -0,0 +1,47 @@
...
+
+
+namespace DiscordWebhook
+{
+	public static class DiscordWebhookMessage
+	{
+		public static byte[] Post(string url, NameValueCollection pairs)
+		{
+			using (WebClient webClient = new WebClient())
+				return webClient.UploadValues(url, pairs);
```
Ostaf: The return from this isn't used anywhere as far as I can see.

UnityProject/Assets/Scripts/Chat/Chat.cs
```diff
@@ -82,6 +84,13 @@ public static void AddChatMsgToChat(ConnectedPlayer sentByPlayer, string message
 			}
 
 			Instance.addChatLogServer.Invoke(chatEvent);
+
+			//Sends OOC message to a discord webhook
+			if (!string.IsNullOrEmpty(ServerData.ServerConfig.DiscordWebhookOOCURL))
```
Ostaf: Potentially pull this null check into the SendWebHookMessage - see the commet on DiscordWebhook.cs

UnityProject/Assets/Scripts/Chat/DiscordWebhook.cs
```diff
@@ -0,0 +1,47 @@
...
+					msg
+				}
+			});
+		}
+
+		public static string MsgMentionProcess(string msg, string mentionID = null)
+		{
+			var newmsg = msg;
+
+			newmsg = Regex.Replace(newmsg, "<@", " ");
```
Ostaf, line 52: I think the code guidelines say all Regex have to be commented.

Ostaf (changes requested):

UnityProject/Assets/Scripts/Chat/DiscordWebhook.cs
ThatDan123: Cool, comment added

UnityProject/Assets/Scripts/Chat/Chat.cs
ThatDan123: Added.

UnityProject/Assets/Scripts/Chat/DiscordWebhook.cs
ThatDan123: Think ive done what youve suggested.

UnityProject/Assets/Scripts/Chat/DiscordWebhook.cs
ThatDan123: removed

ThatDan123:
Good to merge, unless you want to review again.

Bod9001:
all good to merge?

Ostaf:
> all good to merge?

Only minor thing is that the enum could be renamed to be more descriptive, but I don't know how much you guys care about stuff like that.

UnityProject/Assets/Scripts/Chat/DiscordWebhook.cs
```diff
@@ -0,0 +1,85 @@
...
+					return ServerData.ServerConfig.DiscordWebhookAdminURL;
+				case Urls.DiscordWebhookAnnouncementURL:
+					return ServerData.ServerConfig.DiscordWebhookAnnouncementURL;
+				default:
+					return null;
+			}
+		}
+	}
+
+	public enum Urls
```
Ostaf: This could potentially be renamed to be more descriptive, and if the enum is called "Urls" the entries don't need URL in them.

I'd call it "DiscordChannels", and just have OOC, Admin, and Announcement.

ThatDan123: Ah cool, will fix

ThatDan123:
Good to merge, thx for the reviews Ostaf.

PR #4350, opened 2020-06-02
Fix microwaves losing stackable meals, cutlet cooking works again

### LINKED ISSUE #3999: Microwaved cutlets disappear if you don't pick up the previous cooked batch first.
## Description
Microwaved cutlets disappear if you don't pick up the previous cooked batch first.

### Detailed Bug Report
Tested on USA servers on build 4091. It is true for both stations. 
Cutlets should stack on one tile, at least like Meat Steak does.

### Steps to Reproduce
1. Cook 3 cutlets 
2. Do not pick up cooked ones from the table 
3.  Cook 3 more cutlets
4. Wait for them to finish cooking, pick up cutlets from the table.
5. Notice you only have 3 cutlets, not 6. The other 3 disappeared.

Laggy low res video:
https://youtu.be/8Pq_tz4P3uA

### LINKED ISSUE #4341: Cutlets can't be microwaved anymore.
## Description
As the title says, cutlets can't be microwaved into cooked cutlets anymore. Not in stacks, and not individually either. You can still cook (individual) cutlets by attacking them with lit welders however.

### Steps to Reproduce
1.  Obtain cutlets.
2. Try to place them in a microwave.  
3.  Despair as the cutlet doesn't go in and a message in the chat log tells you that you can't microwave cutlets.

CodeKalias:
### Purpose
Fixes #3999 , Fixes #4341
Stackable objects stack with other items on spawn, microwave was doing calculations to adjust the output stacks without realizing this happens

### Notes:
Microwave spawning is more robust and will better handle non-Stackable outputs, differences in maximum stack amounts between input and output

Current microwave behavior gives you a number of meals equal to the stacks of ingredient in your hand, mostly ignoring the recipe.
This works currently but will likely need work in the future, ideally adding code to CraftingDatabase.cs/Recipe.cs
Spawn.cs could also benefit from being aware of Stackable and handling stack counts instead

However these are out of scope for this fix

### DIFF
--- a/UnityProject/Assets/Scripts/Machines/Microwave.cs
+++ b/UnityProject/Assets/Scripts/Machines/Microwave.cs
@@ -135,33 +135,23 @@ public class Microwave : NetworkBehaviour
 
 			if (stck != null)   // If the meal has a stackable component, set the correct number of meals
 			{
-				int spawnedSoFar = stck.InitialAmount;
+				int mealDeficit = mealCount - stck.InitialAmount;
 
-				while (spawnedSoFar < mealCount)
+				while (mealDeficit > 0)
 				{
-					int spaceInStack = stck.MaxAmount - stck.Amount;
+					mealDeficit = stck.ServerIncrease(mealDeficit);
 
-					if (spaceInStack > 0)
-					{
-						// Add all the missing meals to the stack, or as many as the stack can contain
-						int mealDeficit = mealCount - spawnedSoFar;
-						int stacksToAdd = Mathf.Min(spaceInStack, mealDeficit);
-						stck.ServerIncrease(stacksToAdd);
-						spawnedSoFar += stacksToAdd;
-					}
-
-					if (spawnedSoFar < mealCount) // Still not enough meals
+					if (mealDeficit > 0)
 					{
 						result = Spawn.ServerPrefab(mealPrefab, spawnPosition, transform.parent);
 						stck = result.GameObject.GetComponent<Stackable>();
-						spawnedSoFar += stck.InitialAmount;
+						mealDeficit -= stck.InitialAmount;
 					}
 				}
 
-				if (spawnedSoFar > mealCount) // Reduce the stack if our last spawned stackable results in too many meals
+				if (mealDeficit < 0) // Reduce the stack if our last spawned stackable results in too many meals
 				{
-					int mealSurplus = spawnedSoFar - mealCount;
-					stck.ServerConsume(mealSurplus);
+					stck.ServerConsume(-mealDeficit);
 				}
 			}
 			else if (mealCount > 1) // Spawn non-stackable meals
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Machines/Microwave.cs
```diff
@@ -124,30 +124,50 @@ public void RpcStartCooking()
...
-					stck.ServerIncrease(stckChanger);
-				} else if (stckChanger < 0)
+					int spaceInStack = stck.MaxAmount - stck.Amount;
+
+					if (spaceInStack > 0)
+					{
+						// Add all the missing meals to the stack, or as many as the stack can contain
+						int mealDeficit = mealCount - spawnedSoFar;
+						int stacksToAdd = Mathf.Min(spaceInStack, mealDeficit);
+						stck.ServerIncrease(stacksToAdd);
```
Ostaf: Maybe out of scope of this but it seems like this must be a pretty common problem.

A nice solution might be to have this ServerIncrease return the extra amount it couldn't spawn due to exceeding the stack - you could then just loop calling the method and spawning a new stack each time until it returned 0.

UnityProject/Assets/Scripts/Machines/Microwave.cs
```diff
@@ -124,30 +124,50 @@ public void RpcStartCooking()
...
+				while (spawnedSoFar < mealCount)
 				{
-					stck.ServerIncrease(stckChanger);
-				} else if (stckChanger < 0)
+					int spaceInStack = stck.MaxAmount - stck.Amount;
+
+					if (spaceInStack > 0)
+					{
+						// Add all the missing meals to the stack, or as many as the stack can contain
+						int mealDeficit = mealCount - spawnedSoFar;
```
Ostaf: I think Stackable has a "CanAccomodate" method which might be able to take care of some of this logic for you?

CodeKalias: CanAccomodate only deals with whether two items can exchange stacks, doesn't work for a single object or deal with remaining space.

UnityProject/Assets/Scripts/Machines/Microwave.cs
CodeKalias: I'll have a look at adjusting Stackable to make it a bit easier to work with

UnityProject/Assets/Scripts/Machines/Microwave.cs
Bod9001: could you add it to can accommodate?
idk if it would cause funny things though

CodeKalias: It's not something I'd add there, that method deals with unrelated things
I've adjusted Stackable further to address this though

Bod9001:
have you done the changes?

CodeKalias:
Yeah, the latest commit has them

PR #4354, opened 2020-06-02
Round will restart after 60 seconds now.

ThatDan123:
Round will restart after 60 seconds and a message in chat will state this.


ThatDan123:
Good to merge.

PR #4362, opened 2020-06-04
Window stuff: More steps, sheets, deconstruction, variants, and more.

PetMudstone:
### Purpose
- Adds several more steps for deconstructing reinforced windows, similar to /tg/.
   - Allows all windows to be unsecured and deconstructed. Additional .prefabs.
   - ChangeTileWhenItemUsed.cs can now check for harm intent.
- Adjust window health values once more.
- Add titanium glass, plastitanium glass, plasma glass, and reinforced titanium glass sheets.
  - Add plasma and reinforced plasma windows.
  - Make most windows constructible, can't obtain new materials though...
- ReinforcedWindowObject.cs and WindowObject.cs are gone, merged into WindowFullTile.cs.
  - Can set variables for layer tiles and materials in the inspector for all full tile window .prefabs now.
  - Bare-bones directional window object script, only used for materials dropped on destruction.
- Added StructureBase.prefab.
  - Made new directional windows children of it. 
    - **Kept old directional windows around as not to disturb maps**.
  - Full tile window prefabs are variants of relevant directional windows, it's a bit of a mess.

### Notes:
- I made window prefabs drop the right materials (even rods) thanks to script changes, but I did not do the same for tiles because I don't know how the hell they work.
- How WindowFullTileObject.cs and WindowDirectionalObject.cs spawn both rods and glass shards upon window destruction may not be ideal. Would like criticism/advice on it.
- Directional windows have half health (and materials) of full tile windows. Window .prefabs have proper stats from /tg/ [while tiles have boosted fire armor](https://github.com/unitystation/unitystation/pull/4246) (except for plasma windows, which have 99 fire armor even in /tg/).
- Directional windows drop materials as defined in the Window Directional Object component and also two sheets of metal due to the Register Door component. Might need to alter the latter later.
- Screws don't cool down automatically after about 30 seconds after completing step 1 below, unlike /tg/.
- Plastitanium and shuttle window .prefabs drop plastitanium glass and titanium glass sheets instead of shards upon being destroyed, for now at least.

**How to deconstruct reinforced/plastitanium/shuttle windows:**
1. Weld *on harm intent* to heat screws (all steps after can be done on any intent). [15 sec.]
2. Unscrew screws with a screwdriver. [5 sec.]
3. Crowbar to pop out panel. [4 sec.]
4. Wirecutters to cut connecting bars. [2 sec.]
5. Wrench to unsecure the frame. [6 sec.].
6. Wrench again to deconstruct the window, as per usual.

**How to (re)secure reinforced/plastitanium/shuttle windows:**
- After step 2, you can quickly put the screws back in using a screwdriver.
- After step 4, or after constructing a new reinforced/plastitanium/shuttle window and screwing it into place, you can use a crowbar to completely secure the window. Beware, it takes about 12 seconds to do so.


Bod9001:
we got a conflict, Will merge once you resolve conflict

PetMudstone:
Should be good now. I think the conflict was because of my crates PR....

PR #4363, opened 2020-06-04
Job Selection Queue

ThatDan123:
Readds #4083 and #4095 and should work properly this time.

-Adds queue to after round job select to stop people spawning more of the same job than allowed.

~Still testing.~

Tested on headless, late joining and readying up, no errors works good.

Should have a robust staging test, just to make sure it all works ok.

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/GameManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/GameManager.cs
@@ -79,7 +79,13 @@ public partial class GameManager : MonoBehaviour
 	private bool loadedDirectlyToStation;
 	public bool LoadedDirectlyToStation => loadedDirectlyToStation;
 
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Managers/GameManager.cs
```diff
@@ -287,6 +293,13 @@ private void Update()
 			stationTime = stationTime.AddSeconds(Time.deltaTime);
 			roundTimer.text = stationTime.ToString("HH:mm");
 		}
+
+		timeElapsedQueueCheckServer += Time.deltaTime;
```
Bod9001, line 297: this seems like it should be in update,  currently it is in private void Awake()

ThatDan123: It is in the update bod :p Just github view collapsed it.

ThatDan123: image

Bod9001:
tested it it looks good

PR #4365, opened 2020-06-05
Making spawn aware of stackables, recipe crafting rework

CodeKalias:
### Purpose
When spawning something which is stackable you wouldn't get the amount you asked for, you'd get that amount times by the Stackable initialAmount.
This is probably part of why we have things like two versions of each wire prefab with 1 or 30 in the stack.
Spawn now looks at their Stackable and correctly sets their amounts.

Recipe crafting has been updated to match the new functionality
- Crafting attempts now properly use the recipe, making sure to use and produce the correct amounts
- Reduced a bunch of code copy+paste

### TODO?

Currently considering a better place to put the new CraftingManager code and the Recipe.Consume, particularly because they can't be marked as [Server] where they are.

The grinder has its own Recipe and CraftingDatabase classes, wondering if that can be made unnecessary.

Should be able to add a spawn amount to the admin spawner/cloner with this.

Need to do a comment pass

## Tests
Find any places stackables might be spawned and make sure they are working correctly
See where changes can be made to use the new spawning
- [x] Crafting recipes
- [x] Construction / Deconstruction / Destruction
- [ ] Admin spawner
- [ ] Vending machines
- [x] Map placed items
- [ ] Bags/Belts + Lockers + Cargo orders
- [ ] More I haven't thought of

### DIFF
--- a/UnityProject/Assets/Hatchet.cs
+++ b/UnityProject/Assets/Hatchet.cs
@@ -30,34 +30,9 @@ public class Hatchet : MonoBehaviour, ICheckedInteractable<InventoryApply>
 	}
 	public void ServerPerformInteraction(InventoryApply interaction)
 	{
-
-		//is the target item chopable?
-		ItemAttributesV2 attr = interaction.TargetObject.GetComponent<ItemAttributesV2>();
-		Ingredient ingredient = new Ingredient(attr.ArticleName);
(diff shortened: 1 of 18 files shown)


CodeKalias:
Closing this until I figure out how to handle this with populators

PR #4368, opened 2020-06-05
A quick fix to get visual wiring overlays to a workable state

DooblyNoobly:
There is still a few bugs to work out with the interaction validations but this PR is just to get the current version into a workable state.

See changes below:

wiring01


DooblyNoobly:
btw test runner has failed to start a few times today on github actions. I ran the test runner locally on this PR and it was all passes

PR #4372, opened 2020-06-07
Adds "Use Exact Amounts" option to Reactions

VernaSaurus:
allows you to make reactions require exact amounts rather than ratios, as shown with the dough and cake batter recipes

also i'm back haha gottem


UnityProject/Assets/Scripts/Chemistry/Reaction.cs
```diff
@@ -33,28 +34,62 @@ public bool Apply(MonoBehaviour sender, ReagentMix reagentMix)
...
+					var reactionResult = reactionAmount * result.Value;
+					reagentMix.Add(result.Key, reactionResult);
+				}
+
+				foreach (var effect in effects)
+				{
+					effect.Apply(sender, reactionAmount);
+				}
+			}
+			else if (useExactAmounts == true)
```
vlandemart: 90% of this else is just copypaste of if statement
Can't you just insert
`if (useExactAmounts && !ingredients.All(ingredient => reagentMix[ingredient.Key] >= ingredient.Value))
    return;`
after if statement at line 41 not to make unneeded nesting?

vlandemart (changes requested):

VernaSaurus:
I noticed that only after I already closed everything out, I'll fix it

VernaSaurus:
alright, i did it

ThatDan123:
image

I would suggest making the bool (useExactAmounts) the first condition of the If statement just for performance reasons,  unless a code reviewer doesnt think that'll be necessary.

UnityProject/Assets/Scripts/Chemistry/Reaction.cs
```diff
@@ -33,6 +34,12 @@ public bool Apply(MonoBehaviour sender, ReagentMix reagentMix)
 			{
 				return false;
 			}
+
+
+			if (!ingredients.All(ingredient => reagentMix[ingredient.Key] >= ingredient.Value) && useExactAmounts == true)
+			{
+				return false;
+			}
```
RedlineTriad: This will use inexact amounts for when there is more of an ingredient than is strictly needed.

Bod9001:
all good?

VernaSaurus:
yep

PR #4373, opened 2020-06-07
Pog cargo shuttle fix

ThatDan123:
Fixes pog cargo shuttle from crashing when turning to go back to centcom.


ThatDan123:
Good to merge.

PR #4374, opened 2020-06-07
Vendor fixes and improvements

### LINKED ISSUE #2838: Vendor GUI glitches out on client side
## Description

On client side, sometimes when you scroll down and buy something on a vendor, the UI glitches out and all the entries get shifted to the left. Reopening the GUI fixes it.

Macoron:
### Purpose
General vendor improvements and fixes:
- Refactoring of Vendor and GUI_Vendor scripts. Now vendors use cooldown system to avoid player spamming (used to be custom one). Also now View and Logic part are isolated.
- Fixes #2838 - vendors list layout is no longer broken after player bough something
- Most of the vending machines now have access restrictions. No more bar looting 😒
- Now vending show correct items name in UI selection list, not the GameObject IDs
- Remove "can of worm" text for non-initialized vendor entry (looked weird). Now it is just "...". Non-initialized entries appear if there is too much list entities.
- Vendor UI doesn't reload entity list from the server, but only last changed items. This reduced bandwidth a bit and also fixed bug with jumping to center viewport of the list,

### Notes:
Added NetButtonAuth to pass player that pressed button. Maybe there was some build-in way to do that, but I didn't find it.


Bod9001:
not sure the player base is ready for ID restrictions but everyone in design discussion seems okay with it so Should be okay

UnityProject/Assets/Scripts/UI/Net/Elements/Dynamic/NetUIDynamicList.cs
```diff
@@ -115,10 +116,15 @@ public virtual void Clear()
...
-		UpdatePeepers();
+		if (MasterTab.IsServer)
+		{
+			NetworkTabManager.Instance.Rescan(MasterTab.NetTabDescriptor);
+			RefreshPositions();
+			UpdatePeepers();
+		}
+
+		// rebuild layout to fix bug with moved UI elements position
+		LayoutRebuilder.MarkLayoutForRebuild(transform as RectTransform);
```
fomalsd, line 127: oh cool, does that fix the need to open tabs twice?

UnityProject/Assets/Scripts/UI/GUI_Vendor.cs
```diff
@@ -1,179 +1,113 @@
...
 		for (int i = 0; i < vendorContent.Count; i++)
 		{
 			VendorItemEntry item = itemList.Entries[i] as VendorItemEntry;
 			item.SetItem(vendorContent[i], this);
 		}
 	}
 
-	public void VendItem(VendorItem item)
+	/// <summary>
+	/// Update only single item and send new state to clients UI
```
fomalsd, line 85: is this actually required to update client properly? I'd like to avoid having to write stuff like that in each GUI class and instead have that encapsulated in net dynamic list code

UnityProject/Assets/Scripts/UI/Net/Elements/Dynamic/NetUIDynamicList.cs
Macoron: If you are about bug when vendors UI moves to the wrong corner after first opening - it is.

UnityProject/Assets/Scripts/UI/GUI_Vendor.cs
Macoron: This one fixes annoying bug when you selected some item to buy and lists sliders starts moving like crazy. It's because UI dynamically download and rebuild same entities again client side (can be hundred in one vendor). This function doesn't remove any elements, but only update single one values.

Not sure how to encapsulate this function to dynamic list. It updates values using SetItem function which will be different for each list application. Though I'm not very familiar with NetUI yet, so it may be made generic somehow.

fomalsd:
alright, going to merge this but left a note regarding 'moving like crazy' fix in #3169

PR #4381, opened 2020-06-08
infinite fire stack issue + fire state tracking 

### LINKED ISSUE #4343: Fire stacks build up forever if you're wearing fire-resistant clothing.
## Description
If you're on fire and wearing clothing with 100 fire armor, then fire stacks will keep building up (maybe). Unsure if there's an upper-limit, but most I've seen in the variable viewer was about 4555 stacks.

### Steps to Reproduce
1. Have admin permissions, so you can use stuff like the variable viewer.
2.  Hold control and right click on yourself, select the "View Script Variables" bubble.
3. Put on a set of clothing with 100 fire armor, such as an emergency firesuit and firefighter's helmet.
4. Set yourself on fire, either using a spawned-in flamethrower or by starting a plasma fire. Being in a plasma fire should make fire stacks on you increase rather quickly.
5. Open the variable viewer in the admin tab in the top right.
6. Click on the PlayerHealth button on the window that shows up, another window that looks like a book on the right will appear.
7. Click on the button on the book window that shows an ASCII arrow pointing right, it will "turn a page".
8. On the right page of the book window, you will see "FireStacks" and "NetworkfireStacks" (sic). They tell you have many fire stacks they have on you.

stvnksslr:
### Purpose
Fixes #4343  
two main issues are 
* That there was no fire stack cap so that's been added, stacks will continue to tick up forever and  probably cause issues down the line.
* When someone is on fire they are themselves a source of fire and will just add new stacks to themselves faster than the fire stack decay rate.

I am working on a better way to implement fire state tracking to prevent someone in fire proof clothing from being an infinite source of new fire stacks even if they are now capped.

### Notes:
Any questions or comments msg Zasze on discord

### DIFF
--- a/UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
+++ b/UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
@@ -105,6 +105,8 @@ public abstract class LivingHealthBehaviour : NetworkBehaviour, IHealth, IFireEx
 	//can go up or down based on possible sources of being on fire. Max seems to be 20 in tg.
 	[SyncVar(hook=nameof(SyncFireStacks))]
 	private float fireStacks;
+	private float maxFireStacks = 20f;
+	private bool isOnFire = false;
 
 	/// <summary>
 	/// How on fire we are. Exists client side - synced with server.
@@ -379,7 +381,16 @@ public abstract class LivingHealthBehaviour : NetworkBehaviour, IHealth, IFireEx
 
 		if (attackType == AttackType.Fire)
 		{
-			SyncFireStacks(fireStacks, fireStacks+1);
+			if (fireStacks <= maxFireStacks && !isOnFire)
+			{
+				SyncFireStacks(fireStacks, fireStacks+1);
+				isOnFire = true;
+			}
+
+			if (fireStacks <= 0f)
+			{
+				isOnFire = false;
+			}
 		}
 
 		//For special effects spawning like blood:
--- a/UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
+++ b/UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
@@ -381,17 +381,19 @@ public abstract class LivingHealthBehaviour : NetworkBehaviour, IHealth, IFireEx
 
 		if (attackType == AttackType.Fire)
 		{
-			// TODO: issue here is that fire stacks need to apply freely when inside a fire source but not forever from self sources
+			// fire stacks should not exceed 20, and not apply if already at the cap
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
```diff
@@ -379,7 +381,23 @@ private void SyncFireStacks(float oldValue, float newValue)
 
 		if (attackType == AttackType.Fire)
 		{
-			SyncFireStacks(fireStacks, fireStacks+1);
+			// fire stacks should not exceed 20, and not apply if already at the cap
+			if (fireStacks <= maxFireStacks && !maxFireStacksReached)
```
fomalsd, line 385: I think `maxFireStacksReached` bool isn't even needed here, a `if (fireStacks < maxFireStacks)` check should be enough

UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
```diff
@@ -105,6 +105,8 @@ protected set
 	//can go up or down based on possible sources of being on fire. Max seems to be 20 in tg.
 	[SyncVar(hook=nameof(SyncFireStacks))]
 	private float fireStacks;
+	private float maxFireStacks = 20f;
```
fomalsd: 20 seems like too much in my opinion, I don't think it should be more than 5. Correct me if I'm wrong though (if /tg/ has 20 too then that's fine)

UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
stvnksslr: You will forever set yourself on fire and generate new stacks without some flag to prevent yourself from also being a fire source. I puzzled with this one alot, The best long term solution might be to add a trait to fire to determine if its coming from the environment or the character. 

or accept that being fire proof should not allow you to be set on fire / generate fire stacks given how fire currently works.

Though im still not sure of the best way to go about that.

UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
stvnksslr: When I asked in discord 20 was the number thrown around and referenced in our code comments, but I have no real opinion on this as anything over 5 seems to be fatal even with stop drop and rolls.

fomalsd: sorry for late reply. make it 5

stvnksslr: dont sweat it, changed to 5

PR #4383, opened 2020-06-09
Fix firelocks

### LINKED ISSUE #4382: Firelocks should be only be manually openable using crowbars.
## Description
Right now, players can just open closed firelocks using anything. Empty hands, welders, guns, they all open firelocks if you use on them with said items. In /tg/, you can only open firelocks using crowbars (and other prying tools I believe, like fire axes). Since anyone can just open firelocks whenever, any fool can open one to let a fire spread.

Curiously, you can already manually close firelocks but only using crowbars and at any time,

On a side note in /tg/, if you click on a firelock with an empty hand you bang on it according to chat but no sound plays.

image

### Steps to Reproduce
1.  Go to Outpost or Box so you have firelocks available, Pog has none.
2.  Get a firelock closed, either by pulling a nearby alarm or by using a crowbar.
3.  Use anything on the firelock, it should open.

AutumnNova:
fixes #4382  and fixes a mistake in the french accent

### DIFF
--- a/UnityProject/Assets/Resources/ScriptableObjects/Speech/French.asset
+++ b/UnityProject/Assets/Resources/ScriptableObjects/Speech/French.asset
@@ -95,7 +95,7 @@ MonoBehaviour:
     - je suis
   - original: maint
     replaceWith:
-    - les banlieues
+    - banlieues
   - original: my
     replaceWith:
     - mon
--- a/UnityProject/Assets/Scripts/Doors/FireLock.cs
+++ b/UnityProject/Assets/Scripts/Doors/FireLock.cs
@@ -9,30 +9,6 @@ public class FireLock : InteractableDoor
 	public override void ServerPerformInteraction(HandApply interaction)
 	{
 		this.interaction = interaction;
-
-		if (Validations.HasItemTrait(interaction.HandObject, CommonTraits.Instance.Crowbar))
-		{
-			TryCrowbar();
-		}
-		else if (!Controller.IsClosed)
-		{
-			TryClose(); // Close the door if it's open
-		}
-		else if (Validations.HasUsedActiveWelder(interaction))
-		{
-			TryWelder(); // Repair or un/weld door, or deconstruct false wall
-		}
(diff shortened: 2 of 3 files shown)


Aranclanos:
this interaction already exists, reimplementing it is a bad idea

take a look at TryCrowbar() in InteractableDoor.cs, it worked originally but I believe the door hacking broke it, but I can't say how without taking a closer look
just make sure that before fixing it, you understand what's causing the error

AutumnNova:
nothing in TryCrowbar() will fix the issue as it is caused by the ServerTryOpen at the end of ServerPerformInteraction(). Replacing it with a return did fix the issue.

Aranclanos:
Again, do not attempt to fix an issue without knowing the source of the error
if I'm not wrong, the functions TryOpen() and TryClose() from InteractableDoor dont get called because of a bad merge conflicts
replace line 129 and 133 with those calls

Aranclanos:
probably remove the changes in firelock.cs I think that override will make all interactions do nothing

AutumnNova:
I cant imagine changing things in TryCrowbar() will fix opening it with just your hand, but alright

Aranclanos:
oh yeah right, change the ServerTryOpen() calls from bump and hand interaction to the TryOpen()
looking at the commit that broke all this would be the best, just to see if theres something more

AutumnNova:
> Again, do not attempt to fix an issue without knowing the source of the error
> if I'm not wrong, the functions TryOpen() and TryClose() from InteractableDoor dont get called because of a bad merge conflicts
> replace line 129 and 133 with those calls

replacing those calls would prevent opening the firedoors with a crowbar, no?
as firelock.cs overwrites them to nothing?

Aranclanos:
sounds about right, the ones inside trycrowbar() were fine

ThatDan123:
You have compile errors.

Bod9001:
good to merge?

AutumnNova:
yep

PR #4394, opened 2020-06-11
Ambient sound

ReiderJack:
### Purpose
Manager for ambient sounds.

### Notes:
- Made a scriptable object with all ambient sounds.
- Made a manager for ambient sounds which also is an audio sources pool:
  - It handles initialization for audio sources pool automatically so developers just need to add new sounds to scriptable object which manager references.
- Removed not ambient sounds from folder Sounds/Ambient .
- Moved ambience sound area to client side and added that prefab to oupost.
- The volume for ambience was in range 0 - 0.25f. Made it 0 - 1.0f


Bod9001:
thought you going to use SO references instead of string name?

Bod9001:
Talked with ReiderJack in DMs , it's part of the changing over to the new system

PR #4404, opened 2020-06-12
fix issue with pog and a few more changes

AutumnNova:
move clown room, add in ian, add engie maint and a few other small things


ThatDan123:
PogStation: "APC - Atmospherics" has a null value in the list.
PogStation: "APC - Atmospherics" has a null value in the list.
PogStation: "APC - Atmospherics" has a null value in the list.

PR #4409, opened 2020-06-13
Admin Teleport

ThatDan123:
Adds: 

-Admin to player teleport

-Admin to player teleport but as Aghost

-Player to Admin teleport

-Teleport ALL players to any other player, (can only teleport to ghosts IF it an admin ghost who does the action)

-Fixes NRE with wallmounts

-kick message fix.

-rebalanced glass and metal crates buy cost to 850 from 700, as you could sell for 800.

Code problem discovered: ghosts position cannot be found using usual methods on headless, it just produces the same coord each time not matter what position, which is why only admins can do the all players to them when ghost as i had to send their client coord to the server.

### DIFF
--- a/UnityProject/Assets/Scripts/Messages/Client/Admin/RequestAdminTeleport.cs
+++ b/UnityProject/Assets/Scripts/Messages/Client/Admin/RequestAdminTeleport.cs
@@ -1,0 +1,102 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+using Mirror;
+using UnityEngine;
+
+public class RequestAdminTeleport : ClientMessage
+{
+	public string Userid;
+	public string AdminToken;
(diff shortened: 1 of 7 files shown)


UnityProject/Assets/Scripts/Messages/Client/Admin/RequestAdminTeleport.cs
```diff
@@ -0,0 +1,194 @@
...
+	{
+		if (OpperationNumber == 1)
+		{
+			DoAdminToPlayerTeleport();
+		}
+		else if (OpperationNumber == 2)
+		{
+			DoPlayerToAdminTeleport();
+		}
+		else if (OpperationNumber == 3)
```
Bod9001: you could probably use enums instead?

Bod9001:
apart from that everything else looks good

PR #4410, opened 2020-06-13
Custom editor inspector for Audio Clips Array

ReiderJack:
### Purpose
More friendly inspector for audio clips array.

Before:
InspectorBefore

After:
After

### Notes:
- Button "P" to play the clip.
- Button "S" to stop. (Stops any clip)
- Button "X" to delete from array.
- Button "Add clip" opens PickerWindow with all AudioClips in the project. Double click or select and close the window to add the clip.


Bod9001:
looks good

PR #4411, opened 2020-06-14
Admin vote veto

ThatDan123:
Allows admins to veto a vote.

Allows users to change their vote.

-If an admin is online the vote cant auto pass until after 15 seconds even if they have the numbers, to allow time for veto

-Fixes tooltip bug


Bod9001:
looks good

Issue #4355, opened 2020-06-02
Assistant Maintenance Access


Humonitarian:
### Detailed Bug Report
Assistants do not start with maintenance access, even though it is one of the most important parts of being an assistant.

ThatDan123:
It was removed as hacking was added, we use tg as a guideline and assistants dont have maint access on there.

PetMudstone:
Last I checked on Fulpstation (a downstream of /tg/), assistants didn't have maint access on there either. Strangely enough, the [/tg/ wiki page on assistants](https://tgstation13.org/wiki/Assistant) states that they have maint access in the header and in the traitor guide section at the bottom.

ThatDan123:
humm, strange couldve sworn they didnt.

ThatDan123:
Guess we could make a community poll.

PetMudstone:
It would appear that whether or not assistants in /tg/ start with maint access is changeable in /tg/ configuration.

![image](https://user-images.githubusercontent.com/38266309/83572688-dadf8c80-a4de-11ea-9330-1a230eda64e5.png)
![image](https://user-images.githubusercontent.com/38266309/83572717-e7fc7b80-a4de-11ea-9fc2-4ec55958f1b8.png)

On my local /tg/ repo, I spawned as assistant and could open maint doors by default without changing config. That may have been because of additional access given from the station having a skeleton crew (a.k.a. just me).

It appears it can also be changed if security gets maint access or if _everyone_ gets maint access.
![image](https://user-images.githubusercontent.com/38266309/83573226-e2536580-a4df-11ea-8644-2fbd488e2de9.png)

NoooneyDude:
This issue should be closed. #4413 gives them maint. access.

corp-0:
closing as it was fixed by #4413

PR #4413, opened 2020-06-15
Assistants gets Maint access

C4BRA:
### Purpose
This PR gives assistants Maint access ( #4355 )
Majority of the community agrees that it should be implemented


C4BRA:
*clicks tongue* noice

PR #4420, opened 2020-06-16
few changes to booze and soda dispensers

AutumnNova:
adds reagent list back
adds ethanol to booze and watermelon juice to soda dispensers


Bod9001:
I've seen screenshots they loook good

PR #4450, opened 2020-06-22
Machines and computers now drop stuff when destroyed

### LINKED ISSUE #4352: Computers and machines should drop components/frames when destroyed.
## Description
Right now, if you destroy computers and any machine (like a plasma generator) they don't drop anything. In /tg/, if you destroy a computer or machine, it drops all the materials and components used in its construction. 

Computers and machines don't get destroyed immediately when damaged enough in /tg/ however. First, the display on a computer breaks and a machine turns back into a frame and drops all the components and the board except for the wiring added before said board. If additional damage is done then the computer or machine frame gets destroyed and drops the rest of the materials and components used within. Wires, metal, and board for a computer without a display; just wires and metal for a machine frame.

It should be noted that the materials dropped when a computer or machine is destroyed should be based on their state of construction. A computer and machine frame shouldn't drop wiring if no wires were put in them. 

Additionally, computers and machines in /tg/ drop **all** materials used in them and can easily be rebuilt should they be destroyed. It might be better if there was a chance for some materials to be destroyed instead of dropping so that someone can't immediately put whatever computer or machine that got destroyed back together.

ThatDan123:
Machines and computers now drop stuff when destroyed.

fixes #4352  Only change i made to this was that it'll spawn a random amount of metal, glass and cable, as the machine was destroyed so i felt some of the material would be lost. You'll always get the circuit board and stock parts though.

### DIFF
--- a/UnityProject/Assets/Scripts/Construction/Computer.cs
+++ b/UnityProject/Assets/Scripts/Construction/Computer.cs
@@ -43,11 +43,18 @@ public class Computer : MonoBehaviour, ICheckedInteractable<HandApply>
 			$"{interaction.Performer.ExpensiveName()} disconnects the monitor.",
 			() =>
 			{
-				WhenDestroyed();
+				WhenDestroyed(null);
 			});
 	}
 
-	public void WhenDestroyed()
+	private void Awake()
+	{
+		if (!CustomNetworkManager.IsServer) return;
+
+		GetComponent<Integrity>().OnWillDestroyServer.AddListener(WhenDestroyed);
+	}
+
+	public void WhenDestroyed(DestructionInfo info)
 	{
 		//drop all our contents
 		ItemStorage itemStorage = null;
--- a/UnityProject/Assets/Scripts/Construction/Computer.cs
+++ b/UnityProject/Assets/Scripts/Construction/Computer.cs
@@ -43,21 +43,26 @@ public class Computer : MonoBehaviour, ICheckedInteractable<HandApply>
 			$"{interaction.Performer.ExpensiveName()} disconnects the monitor.",
 			() =>
 			{
-				//drop all our contents
-				ItemStorage itemStorage = null;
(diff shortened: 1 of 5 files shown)


PetMudstone:
Hey king you should link to the relevant issue: #4352

ThatDan123:
Tested on headless, no errors server side or client.

Good to merge.

UnityProject/Assets/Scripts/Construction/Computer.cs
```diff
@@ -43,21 +45,37 @@ public void ServerPerformInteraction(HandApply interaction)
...
+		if (gameObject != null)
+		{
+			itemStorage = GetComponent<ItemStorage>();
+		}
+
+		if (itemStorage != null)
+		{
+			itemStorage.ServerDropAll();
+		}
+		var frame = Spawn.ServerPrefab(framePrefab, SpawnDestination.At(gameObject)).GameObject;
```
vlandemart, line 74: We can enter this part of code with `gameObject == null`
I know that you just moved old code, but I think it might be a good time to fix this

UnityProject/Assets/Scripts/Construction/Machine.cs
```diff
@@ -59,26 +61,42 @@ public void ServerPerformInteraction(HandApply interaction)
...
 
-					frame.GetComponent<MachineFrame>().ServerInitFromComputer(this);
+			if (itemStorage != null)
+			{
+				itemStorage.ServerDropAll();
+			}
 
-					Despawn.ServerSingle(gameObject);
-				});
+			var frame = Spawn.ServerPrefab(framePrefab, SpawnDestination.At(gameObject)).GameObject;
```
vlandemart, line 92: Same here - I think there will be errors if `gameObject == null`

vlandemart (changes requested):
Just one minor issue

vlandemart (approved):

PR #4465, opened 2020-06-25
Light bulbs and tubes construction 

Aranclanos:
### Purpose
Enables the construction of light bulbs and tubes fixtures.
Wallmount construction overlays and object spawn will now support wall protrusion wallmount types.


PetMudstone:
Damn, it's like you can read my mind or something. Was planning to make an issue about light fixtures.

Do these fixtures come with lights when you build them? Are they also automatically powered upon being built? I noticed that both of those things happened while testing in public servers on the Hub when using light fixtures that an admin spawned in.

Aranclanos:
They spawn without the tube/bulb, they're just an empty frame. But they ARE magical gaining power from nothing, I'll plan to follow up this with another PR about buttons, air alarms and other wallmounts where the player would like to connect two objects

Aranclanos:
Alright I borked this

PR #4481, opened 2020-06-26
reducing editor code

ReiderJack:
### Purpose
Made an abstract class to reduce editor code for subscribing triggers on the scene(Like lights, firelocks etc).


UnityProject/Assets/Scripts/Editor/Inspector/SubscriptionControllerEditor.cs
```diff
@@ -0,0 +1,97 @@
...
+		{
+			isSelecting = false;
+			return;
+		}
+
+		if (HasPressedLeftClick(e))
+		{
+			Ray ray = HandleUtility.GUIPointToWorldRay(e.mousePosition);
+			RaycastHit2D[] hits = Physics2D.RaycastAll(ray.origin, ray.direction);
+			List<GameObject> hitObjects = new List<GameObject>();
```
Bod9001: hey, I did some changes to this because it was having trouble picking up objects, sometimes doubling them so I cast them to game objects then put them an hashset to remove duplicates

UnityProject/Assets/Scripts/Editor/Inspector/SubscriptionControllerEditor.cs
```diff
@@ -0,0 +1,97 @@
...
+
+			foreach (var hit in hits)
+			{
+				hitObjects.Add(hit.transform.gameObject);
```
ReiderJack: `var hashits = new HashSet<GameObject>(hits.Select(x => x.transform.gameObject));` isn't it the same?

Bod9001: List vs has

Bod9001: it was mainly done just to get rid of duplicates, they shouldn't be any now since I technically removed all the duplicate box Colliders

Bod9001:
thanks, all good to merge

Bod9001:
oh. there is a bug that you can't select stuff on the syndicate outpost on outpost? you can look into that later

PR #4490, opened 2020-06-28
Turning space suits and hardsuits into proper variants and more kinds of them.

PetMudstone:
### Purpose
- Turns space suits and hardsuits into proper variants.
   - **Old non-variant clothing still exists and has been moved to its own folder as not to cause errors on maps.**
- Adds "several" new space suits and hardsuits, such as syndicate space suits in many colors, MK. I and MK II SWAT suits, and elite syndicate hardsuits.
- Pirate clothing, including that of the space-proof variety. Yarr.
- Removed a redundant component from Outerwear.prefab.

### Notes
As stated before, the old .prefabs for hardsuits and space suits have been preserved as not to cause map errors. I (and hopefully other mappers) will see to it that they are replaced on maps eventually.


Bod9001:
can you make an issue for Updating the map with the new spacesuits and deleting the old files

PetMudstone:
Can do.

PR #4491, opened 2020-06-28
PogStation Overhaul Part 2

C4BRA:
### Purpose

this PR focuses on issue #4141 

"Virology needs to be more isolated from medical, possibly making it a small satellite from medical"
virosat

"Chief Medical Officer's office needs to be moved to a better location"
CMOoffice

"...which connects to the genetics room."
GeneticsCloning

"Medical needs to be redesigned with a more understandable layout."
medicallayout

Also managed and added more APCs


ThatDan123:
DNM, has files not relevant to PR.

PR #4502, opened 2020-06-30
fixes the mixing bowl (in inventory)

VernaSaurus:
no more deleting dough with bowl in inventory, and also makes buns made with dough slices instead of dough balls

### DIFF
--- a/UnityProject/Assets/Scripts/Chemistry/Effects/Item.cs
+++ b/UnityProject/Assets/Scripts/Chemistry/Effects/Item.cs
@@ -11,6 +11,7 @@ namespace Chemistry.Effects
 		public GameObject spawnItem;
 		public override void Apply(MonoBehaviour sender, float amount)
 		{
+			amount = (int)Math.Floor(amount);
 			if (sender.gameObject.GetComponent<MixingBowl>() != null)
 			{
 				playerHolding = sender.gameObject.GetComponent<MixingBowl>().playerHolding;
--- a/UnityProject/Assets/Scripts/Chemistry/Effects/Item.cs
+++ b/UnityProject/Assets/Scripts/Chemistry/Effects/Item.cs
@@ -7,11 +7,22 @@ namespace Chemistry.Effects
 	[Serializable]
 	public class Item : Chemistry.Effect
 	{
+		public RegisterPlayer playerHolding;
 		public GameObject spawnItem;
 		public override void Apply(MonoBehaviour sender, float amount)
 		{
-			amount = (int)Math.Floor(amount);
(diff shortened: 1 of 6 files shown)


VernaSaurus:
ok it's good now

UnityProject/Assets/Scripts/Chemistry/Effects/Item.cs
```diff
@@ -7,11 +7,23 @@ namespace Chemistry.Effects
 	[Serializable]
 	public class Item : Chemistry.Effect
 	{
+		public RegisterPlayer playerHolding;
 		public GameObject spawnItem;
 		public override void Apply(MonoBehaviour sender, float amount)
 		{
 			amount = (int)Math.Floor(amount);
-			Spawn.ServerPrefab(spawnItem, sender.gameObject.RegisterTile().WorldPositionServer, null, null, (int)amount);
+			if (sender.gameObject.GetComponent<MixingBowl>() != null)
```
ReiderJack, line 15: cache `sender.gameObject` before `if`.

ReiderJack: then check if the `sender.gameObject` is null.
then cache `GetComponent<MixingBowl>()` and check if it's null.

UnityProject/Assets/Scripts/Items/Others/MixingBowl.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Mirror;
+public class MixingBowl : NetworkBehaviour, IServerInventoryMove
```
ThatDan123, line 2: Does this need to be a network behaviour? Try switching to mono and then if that works remove the using mirror.

UnityProject/Assets/Scripts/Chemistry/Effects/Item.cs
VernaSaurus: the sender's gameobject will never be null, since the sender is always attached to a gameobject

UnityProject/Assets/Scripts/Items/Others/MixingBowl.cs
VernaSaurus: it does, i tried monobehavior and it doesn't work

Bod9001:
all good to Merge?

UnityProject/Assets/Scripts/Items/Others/MixingBowl.cs
ThatDan123: Ah cool.

VernaSaurus:
all good to merge

PR #4513, opened 2020-07-01
Converting lockers and crates and related objects to variants.

### LINKED ISSUE #2247: Restructure lockers and crates to be prefab variants
# Description
We have a lot of lockers and crates, all of which are separate prefabs, which creates maintenance issues / tedium. Restructure them to have one base Locker and Crate prefab and each different locker / crate can be a variant.

Will require editing the tile palette asset and remapping.

PetMudstone:
### Purpose
- Converts lockers and crates into proper variants of each other and the structure base .prefab.
  - Body bags, coffins, fridges, included.
  - Add big cardboard boxes and some other stuff.
  - Body bag items included.
- Allows closets/crates to drop material other than metal when they are destroyed.
  - Add useless cloth, cardboard, and alien alloy.
- Change references to make use of new variants.
- Move all the non-variant orphans to separate folders, as before. Their files start with BVOld_.

Scenes are all unchanged. Will need to do scene changes in different PRs. Should resolve #2247, maybe.


PetMudstone:
Should be ready now.
Should I make an issue for replacing closets and lockers on maps?

PR #4515, opened 2020-07-01
Fuel Tank Explosion

ThatDan123:
Adds explosions when fuel tank is used with active welder.

or if destroyed.

-Bug: isnt damaged by guns for some reason, melee works fine.


ThatDan123:
Good to merge after tests, been tested on headless all good.

PR #4521, opened 2020-07-01
Drinking Alcohol Now Makes You Drunk

ThatDan123:
Adds the visual effect for being drunk.

Currently set to add how much you drink to the time, eg 5u = 5 seconds. 

However this happens for each type of alcohol, so the more types of alcohol in a drink the more drunk you get from each sip.


ThatDan123:
Ready for a review, tested on headless, two clients and was fine.

UnityProject/Assets/Scripts/Food/DrinkableContainer.cs
```diff
@@ -71,11 +72,31 @@ private void Drink(PlayerScript eater, PlayerScript feeder)
...
+		if ((int) drinkAmount == 0) return;
+
+		foreach (var reagent in container.CurrentReagentMixGet)
+		{
+			//if its not alcoholic skip
+			if (!AlcoholicDrinksSOScript.Instance.AlcoholicReagents.Contains(reagent.Key)) continue;
+
+			//The more different types of alcohol in a drink the longer you get drunk for each sip.
+			playerEatDrinkEffects.ServerSendMessageToClient(eater.gameObject, (int)drinkAmount);
+			playerEatDrinkEffects.ServerSendMessageToClient(eater.gameObject, 0);
```
Bod9001: why does it send it then send 0?

UnityProject/Assets/Scripts/Camera/CameraEffectControlScript.cs
```diff
@@ -27,5 +32,24 @@ public void ToggleNightVisionEffectState()
 		{
 			nightVisionCamera.enabled = !nightVisionCamera.enabled;
 		}
+
+		private void Update()
```
Bod9001: maybe better served to use the update manager?

UnityProject/Assets/Scripts/Food/DrinkableContainer.cs
ThatDan123: Ah, good catch, thats was when i was testing with sync vars.

UnityProject/Assets/Scripts/Camera/CameraEffectControlScript.cs
ThatDan123: changed.

PR #4544, opened 2020-07-05
implements burstfire! (also fixes soda dispenser)

AutumnNova:
implements the burstfire mode to weapons
fixes the soda dispenser
update some prefabs on fallstation

notes about burstfire implementation:
uses an IEnumerator, was only used due to a lack of a better solution (editing the firecooldown resulted in the weapon being broken)
contains a considerable amount of if statements, i cant say i like the implementation, but it works


AutumnNova:
tests are failing due to the soda dispenser now requiring connection, not something i can fix without scene changes
  Expected: <empty>
  But was:  < (BoxStationV1, SodaDispenser (1)), (OutpostStation, SodaDispenser), (PogStation, SodaDispenser (1)), (CentCom, SodaDispenser (1))

the-other-autumn:
only map to error should be outpost as i cant change that due to my branch being out of date

AutumnNova:
should be good to merge now

PR #4548, opened 2020-07-06
Geiger Counter Sound Fix

ThatDan123:
Makes Sound local to geiger Counter, not global.

### DIFF
--- a/UnityProject/Assets/Scripts/Tool/GeigerCounter.cs
+++ b/UnityProject/Assets/Scripts/Tool/GeigerCounter.cs
@@ -58,19 +58,19 @@ public class GeigerCounter : MonoBehaviour, IInteractable<HandActivate>, IServer
 		if (node  == null) return;
 		if (node.RadiationNode.RadiationLevel > 1000)
 		{
-			SoundManager.PlayNetworkedAtPos(Noise[Level.Extreme][RNG.Next(0,3)], registerItem.WorldPositionServer);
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Tool/GeigerCounter.cs
```diff
@@ -57,19 +58,19 @@ public void CycleUpdate()
 		if (node  == null) return;
 		if (node.RadiationNode.RadiationLevel > 1000)
 		{
-			SoundManager.PlayNetworked(Noise[Level.Extreme][RNG.Next(0,3)]);
+			SoundManager.PlayNetworkedAtPos(Noise[Level.Extreme][RNG.Next(0,3)], registerItem.WorldPositionServer, sourceObj: gameObject);
```
Bod9001, line 61: forgot to say if this is in someone's inventory will return hidden POS

vlandemart, line 61: "forgot to say if this is in someone's inventory will return hidden POS" - Needs to be fixed

vlandemart (changes requested):

UnityProject/Assets/Scripts/Tool/GeigerCounter.cs
ThatDan123: Yeah sorry, was just finishing up the lava land stuff to a reasonable place before coming back to fix this.

ThatDan123:
Just did a two client test and this seems to work even if its in a pocket

Bod9001:
interesting,  I will merge

PR #4550, opened 2020-07-07
Enables button and light switch construction.

Aranclanos:
### Purpose
Enables button and light switch construction.
Adds multitool linking for light switches with wallmount lights and buttons with doors.

### DIFF
--- a/UnityProject/Assets/Scripts/Lighting/LightSource.cs
+++ b/UnityProject/Assets/Scripts/Lighting/LightSource.cs
@@ -5,7 +5,7 @@ using Mirror;
 using UnityEngine;
 using Random = UnityEngine.Random;
 
-public class LightSource : ObjectTrigger, ICheckedInteractable<HandApply>, IAPCPowered, IServerLifecycle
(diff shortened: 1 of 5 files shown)


AutumnNova:
would it not then be possible to build a button and link a door to it allowing you to bypass the id restriction on the door?

UnityProject/Assets/Scripts/Lighting/LightSource.cs
```diff
@@ -45,6 +45,19 @@ public class LightSource : ObjectTrigger, ICheckedInteractable<HandApply>, IAPCP
 	private GameObject itemInMount;
 	private float integrityThreshBar;
 
+	[SerializeField]
+	private MultitoolConnectionType conType = MultitoolConnectionType.LightSwitch;
+	public MultitoolConnectionType ConType  => conType;
+
+	public void SetMaster(ISetMultitoolMaster Imaster)
+	{
+		var lightSwitch = (Imaster as Component)?.gameObject.GetComponent<LightSwitchV2>();
```
Bod9001, line 54: you should be able to use 
(Imaster  as LightSwitchV2)

PR #4553, opened 2020-07-08
Connected Soda Dispensers to relevant APCs

Akbadain29:
### Purpose
Fixes Soda Dispensers on BoxStation V1, Fallstation and OutpostStation, as they are added to their APC's Connected Devices List so they are powered and work properly.

### Notes:
I think there should be a UI element that should pop up to alert the player that the device is not connected to an APC or that it does not know what APC it should connect to. That way it would be easier to spot when such a device isn't properly connected.


Bod9001:
that are Tests on the top right for detecting if any components are missing an APC reference 
image

PR #4564, opened 2020-07-10
[DNM] the mess that is ComplexMealBase

VernaSaurus:
complexmealbase was supposed to be how to make things that needed more than two ingredients but i never finished it and it's a huge mistake

only pr'ing this to see if anyone wants to fix this horrid mess


VernaSaurus:
ass we can, seems like i fucked up some files

let me delete those

VernaSaurus:
that should do it, still don't merge it though

kruezbergbier:
Fucked up

VernaSaurus:
i'm just going to close this, it's never going to be finished

PR #4575, opened 2020-07-11
Mass Driver

ThatDan123:
Adds mass driver and general switch to connect it.

-Also moves centcom to different routine.


Bod9001:
What happens if you put a player into it?

ThatDan123:
Pushes them two tiles, same with objects. Items it uses the throw mechanic.

Its a very basic mass driver atm. Would be nice to have a velocity setting and such the in the future. But for now its good for flinging stuff into space.

Bod9001:
cool

PR #4592, opened 2020-07-13
PDA-Update minor fixes

Nukeuler123:
### Purpose
This fix is like two lines, all it does is add the PDA tab to NetTab(Somehow it went missing) and makes the IDcard variable go null when the PDA's ID slot changes, meaning that it will properly update when someone takes the ID out via the slot and not through the UI.


ThatDan123:
Missing reference found in prefab PDA component Sprite field Sprite

Nukeuler123:
> Missing reference found in prefab PDA component Sprite field Sprite

Strange, ill look into it

Nukeuler123:
Should be fixed (?)

PR #4602, opened 2020-07-15
When using cutters on tile, show popup of cables to cut TB #4525

### LINKED ISSUE #4525: When using cutters on tile, show popup of cables to cut TB
## Description

Currently, cutting wires is very buggy as there is no easy way to determine which wire was clicked on within a tile. We need to create a small pop up window with all of the wires listed that are on that tile so the player can easily choose which one to cut.

## Criteria:

- Change the current behavior of wire cutters so that a pop up is shown instead of trying to cut a wire directly
- If no wires are on that tile then do not show the pop up
- Run through all of the Z levels (the range is about 50) on that tile and cache all of the wire tiles (for an example of searching through the Z levels on TileMap click here: [Example](https://github.com/unitystation/unitystation/blob/096d81af0544af72fa1cac6b054bac5a78518960/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/UnderFloorLayer.cs#L25) )
- Dynamically load a list of buttons with an image of the wire into the pop up window
- Make sure the pop up window automatically closes if the player moves away from the tile or if they click somewhere else
- Make a close button (Small X top right) to manually close the window
- Process the original cutting logic using the players choice of wire tile when they click on a wire cut button
- Thoroughly test to make sure there is no bugs

michalczemierowski:
### Purpose
Fixes #4525 

### Notes:
cutting
cutting_hurt


Bod9001:
only thing I would change make the UI dark blue, since it's quite strong currently with the pure white

PR #4629, opened 2020-07-18
Job bans

### LINKED ISSUE #4500: Job bans.
## Description
The ability to ban people from playing certain jobs (or even in entire departments) would be helpful to admins. Some people just aren't good fits for certain jobs, most often those in the security or command departments. Too often there are people who "main" jobs such as captain or officer yet are completely inept in those positions _at best_. Often, they use their job to create negative experiences for everyone else in the game through abuses of power. 

If admins were able to hand out job bans, we might have less bad security (also known as "shitsec"), captains, HoPs, etc. around.

ThatDan123:
Adds a job ban system for admins to use.

Fixes #4500

-Cant do department bans though.

### DIFF
--- a/UnityProject/Assets/Scripts/Player/JoinedViewer.cs
+++ b/UnityProject/Assets/Scripts/Player/JoinedViewer.cs
@@ -6,6 +6,7 @@ using Mirror;
 using Newtonsoft.Json;
 using UnityEngine;
 using UnityEngine.SceneManagement;
+using System.Globalization;
 
 /// <summary>
 /// This is the Viewer object for a joined player.
@@ -100,6 +101,8 @@ public class JoinedViewer : NetworkBehaviour
 		}
 
 		PlayerList.Instance.CheckAdminState(unverifiedConnPlayer, unverifiedUserid);
+
+		PlayerList.ClientJobBanDataMessage.Send(unverifiedUserid);
 	}
 
 	/// <summary>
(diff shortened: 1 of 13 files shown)


UnityProject/Assets/Scripts/Player/JoinedViewer.cs
```diff
@@ -174,6 +180,12 @@ private void CmdRequestJob(JobType jobType, string jsonCharSettings)
 			return;
 		}
 
+		if (PlayerList.Instance.FindPlayerJobBanEntryServer(playerID, jobType, true) != null)
```
Bod9001, line 183: I feel like this would be a bit insecure since you could send someone else's ID?

ThatDan123: I did originally use the SentByPlayer.Connection from the client message but it was returning some values null. It is insecure in the sense they could grab any players ban entry but it is limited to that ban info, which is just a list of this:

[Serializable]
public class JobBanEntry
{
	public JobType job;
	public bool isPerma;
	public double minutes;
	public string dateTimeOfBan;
	public string reason;
}

You cant grab Ip, or Ids

Bod9001: but if you're banned from captain what stop you from sending someone else's ID to avoid the ban?

ThatDan123: The server does its own check, this is just the client asking for the data to do a client side check which while can be hacked around, the server side check would still apply.

ThatDan123: Its just a way to stop the client from asking the server to keep doing checks unless it knows that the check will pass.

PR #4645, opened 2020-07-20
Fixes gateway teleport.

ThatDan123:
For some reason RegisterTile WorldPositionServer was returning the wrong coord, even though the matrix was correct.

So I had to switch over to transform.position


vlandemart:
I don't think that it is right way to address this issue

ThatDan123:
I think in this case its fine as the awaysite is never going to move, so as its not on a movable matrix there shouldnt be any issues using the transform position.

ThatDan123:
But there is might be some underlying issue as the registertile should get the right coordinate but isnt.

fomalsd:
is there an issue # to understand what you are trying to fix?

ThatDan123:
Just been tinkering around with it and its probably an initialise problem.

fomalsd (approved):

Issue #4654, opened 2020-07-21
Disposals Issue and features B:150


NoooneyDude:
Known issues from #4524.
### Known issues

- [x] fix #6568.
- [x] 22. Items ejected from a disposal pipe terminal sometimes permanently spin, as described in issue #5551.
- [x] 3. Constructed pipes with non-standard grey colour return to default grey when welded to form a tile.
- [x] 4. Sometimes items can be ejected at a strange angle.
- [x] 13. Players are exposed to vacuum when traversing disposal pipes that are in space.
- [x] 14. Disposal machines do not use power from APCs.
- [x] 15. Disposal virtual containers can be seen by players if they bring up the context menu for the same tile the container is on. This is issue #4619.
- [x] ~~16. Comply with requested changes in #4524.~~ _**Partially addressed by PR #5786.**_
- [x] 7. ~~Dropping items while in a disposal instance does not add them to the disposal virtual container, and so they are lost to the void.~~ Items dropped while in a disposal instance are invisible once ejected.
- [x] 8. Coloured machine does not apply to overlays (bin overlay spritehandler).
- [ ] 9. Tiles aren't actually broken when items are ejected from pipe end - the update is merely visual.
- [x] 19. DisposalVirtualContainer can be spawned via the dev spawner menu.
#
### New features
- [x] G. Dump the contents of the trash bag into the disposal bin without dumping the bag itself, as described in #4938. This should probably be a property of the bag and not the bin.
- [x] H. Disposal bins should actually take in ambient air, store it, and send it with the disposal traversal. Lungs inside the instance should then breathe from this gas mix. Should solve issue part 13. in the process.
- [x] C. A tooltip for the pipe dispenser's dispensables would be nice, too.

 #
- [x] **Disposals crawling** - When a player enters a disposal pipe system, they should be able to stop flowing along the pipes by pressing the resist key. After struggling for a period of time, they will stop moving with the flow and instead be able to crawl through the pipes in any direction they want to. 

They should also be able to enter disposal pipe openings in a similar manner to disposal bins to initiate crawling at any time they would like, allowing them to navigate pipes in a similar way to small creatures vent crawling. They should also be able to crawl out of disposal pipe openings and disposal bins to get free of the system. 

Disposals crawling will greatly expand the possible applications of the disposals system and add an effective counter to being trapped in disposal pipe loops set up by griefers.

PerfectTangent:
[...]

PR #4660, opened 2020-07-21
Fixes teledisposals bug

### LINKED ISSUE #4646: There's no position validation when drag-and-dropping players into disposals
## Description
According to oksts2904#4959 on Discord, you can drag-and-drop anyone into disposals regardless of their position.
The only limitation is that the other player should stand still during the drag operation.
gif

Both current player and dragged player positions should be validated to be adjacent to the disposals bin before starting progress action.

### Steps to Reproduce

1.  Join server with your friend or foe
2. Click and drag his character and release when hovering nearest disposals bin
3. Make sure they don't move until progress completes

NoooneyDude:
### Purpose
Fixes #4646, teledisposals bug. Part 11. of #4654.

### DIFF
--- a/UnityProject/Assets/Scripts/Disposals/DisposalBin.cs
+++ b/UnityProject/Assets/Scripts/Disposals/DisposalBin.cs
@@ -187,6 +187,10 @@ namespace Disposals
 		public bool WillInteract(MouseDrop interaction, NetworkSide side)
 		{
 			if (!DefaultWillInteract.Default(interaction, side)) return false;
+			if (!Validations.IsInReach(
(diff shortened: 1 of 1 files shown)


ThatDan123:
I would say to add the check in the server perform as well, as its currently only a client side check.

At least i think WillInteract is only called on client side

NoooneyDude:
> I would say to add the check in the server perform as well, as its currently only a client side check.
> 
> At least i think WillInteract is only called on client side

I'm somewhat certain IF2 serverside will run `WillInteract` before `ServerPerformInteraction`, seeing the `NetworkSide side` parameter.

Issue #4285, opened 2020-05-18
Canisters consume tanks and dont spit them out


braincrab:
## Description
When you shove a tank into a canister with intent to fill it, the canister wont let you take it out, saying that no tank is inserted. attempting to put in another tank says that the canister already has a tank inside

### Steps to Reproduce
Please enter the steps to reproduce the bug or behaviour:  

1.  take a canister and a tank
2.  shove tank in canister
3.  the tank is now both in and not in the canister

Nukeuler123:
Claiming this, ill look into it.

Nukeuler123:
Uhh, this is a bit more complicated that I expected

NoooneyDude:
This issue should be closed as of #4662.

Nukeuler123:
Can this get closed?

PR #4662, opened 2020-07-21
Canister Bugfix

Nukeuler123:
### Purpose
Fixed the bug with the canisters eating tanks  #4285 

### Notes:
Iv also modified the HasNetTab component to allow people to grab a reference to a player's gameobject, may be useful in the future.


ThatDan123:
Why dont you check for the active hand slot, and then check if thats empty? If not empty put on floor.

Currently if you have an item in both hands it wouldnt work.

ThatDan123:
if you cant check for active hand slot, then add the check for left hand is null and then an else if right and left arent not null to drop.

Nukeuler123:
Updated the code, should try to put something in the active hand, if it cant it will try and put it in the other , then if it still cant it will do nothing, making it look like the item was dropped at the canister.

PR #4667, opened 2020-07-21
Rice and some other things

VernaSaurus:
i have no idea if this works but I can't test it so can someone else please do so


PetMudstone:
The rice bowl and wish soup reactions do indeed delete the bowl and spawn the appropriate items... as long as the bowl is outside the player's inventory. If the reaction occurs inside the player's inventory, the held bowl is emptied of reagents but continues to exist. No rice bowl or wish soup is dropped either.

I'm not sure how the wish soup script for randomly giving it a different nutrition value works either.
image
Is the above supposed to be zero? Not sure. Also, it takes two strokes of the Z key to eat the wish soup for some reason.

Edit: I don't see any errors in the console, by the way.

VernaSaurus:
ah shite, seems like I did forget something

PetMudstone:
Uh, I was gonna say that the in-hand reactions still don't work. The bowl appears to actually turn into the item it's supposed to... but you can't eat it and it just ceases to exist once you drop it. You can't examine or right click on it either. Everything else appears to be fine though.

VernaSaurus:
well i should prooooobably have fixed that before it got merged-

PR #4674, opened 2020-07-22
Decreases Editor play start times

Bod9001:
for some reason, unity re-packs atlases every single time you start the game if they're set to build, I know it's dumb but this turns it off so it only is active in builds, this should decrease the in play mode FPS That doesn't seem too bad for a Quicker loading time


fomalsd:
did you mean "Decreases Editor play start times"?

Bod9001:
yeah xD

Issue #4351, opened 2020-06-02
"Running" while malnourished shouldn't make you slip on wet floors.


PetMudstone:
## Description
If you're malnourished and "running", you are still able to slip on wet floors. I don't think this should be possible as if you're malnourished and "running" you are moving at roughly the same speed as you would be while walking and well-fed. 

It is also frustrating to be malnourished and having to walk due to wet floors since you move at an absurdly slow speed, especially if you happen to also be wearing clothing that slows you down.

stvnksslr:
I would argue its the intent of the speed that should determine whether you trip or not vs the actual speed. 

When you run malnourished you are still rushing and therefore should be susceptible to slipping even if your moving slower because of a negative status.

Disagree on this being a bug either way.

fomalsd:
that's a fair point actually. maybe make malnourished walking speed even lower than 3 and this one can be closed

yoysam:
should we increase the staving penalties to balance it out

yoysam:
the pull request fixed it

Mute-Man:
PR #4678

PR #4678, opened 2020-07-22
Makes you not slip while malnourished 

yoysam:
### Purpose
fixes issue #4351 

### Notes:
added a simple check


PetMudstone:
I don't think "HungerState.Malnourished" is the state that applies the speed debuff, despite the chat messages. Only "HungerState.Starving" is the state that applies the speed debuff.

image

yoysam:
i changed it to only have straving in it

PR #4724, opened 2020-07-27
Jukebox new

JesterX666:
New jukebox!

Contains: 
The jukebox will play all song found in the lobby screen
A UI panel to skip songs, play, stop playing sound.
When a song finishes, the next song starts playing.
The jukebox has a playing animation.
The jukebox has a different animation when damaged (50% health).  It also stops being able to play.

The jukebox will be added to Outpost station on a different PR after this one.

But now you can spawn it or build on any other station maps.

Recipe for the jukebox machine : 
Machine Frame (built from iron plates)
Jukebox Circuit Board
2 Medium Cable
Micro Laser
Standard component bin
2 pane of glass

Sound volume decreases with distance and is muffled (for walls).  (See below for next iteration).

A lot of adjustements were made to the SoundPlayMessage and SoundManager prefab to be able to specify sound mixer, spatial blend (for attenuation), min distance, max distance and such.

Since the message and many functions had WAY TOO MANY parameters, 2 new objects were made to pass multiple parameters together :

AudioSourceParameters that allows to personalize how the sound will output.
ShakeParameters that regroups all things related to ground shake effect.

This was tested within Unity editor and also with headless server + 2 clients.

Current limitations and stuff for future iterations:

The current SoundManager keep tracks of every sound by it's name.  If two sounds have the same name, you can't stop one without the other stoping at the same time.  Someday we should keep track of sounds by it's emitter or something similar.

If a client late arrives to the station while the jukebox is playing, he will start listening only the next song.  We should incorporate a client join message to all Jukebox for them to broadcast which song is playing and at what time.

Most machines are still working even without APC power connection.  We should refactor and verify this globally.

Synchronisation is "near perfect" : Messages are sent to all client when a song start and when a song stops.  If we want millisecond "PERFECT" synchronisation, we will have to incorporate a SoundSynchronize event.


corp-0:
I'd say you should get rid of scene changes and just let the mappers incorporate the jukebox

JesterX666:
> I'd say you should get rid of scene changes and just let the mappers incorporate the jukebox

Yup, will prevent the conflicts.

JesterX666:
Do not merge.  I should remove unwanted changes to the Lobby and OutpostStation.

Also, teleporter sounds don't seem to work on Box Station map.  And I hear a mouse in the lobby?!?!  I should fix those.

UPDATE: FIXED

Bod9001:
still DNM?

JesterX666:
I removed the DNM, I've fixed the bug that I found.  You can do the code review now.

UnityProject/Assets/Scripts/Machines/Jukebox.cs
```diff
@@ -0,0 +1,263 @@
...
+
+	/// <summary>
+	/// The current state of the jukebox powered/overpowered/underpowered/no power
+	/// </summary>
+	[HideInInspector] public PowerStates CurrentState;
+
+	/// <summary>
+	/// The current state of the jukebox powered/overpowered/underpowered/no power
+	/// </summary>
+	[HideInInspector] private APCPoweredDevice APCConnectionHandler;
```
ReiderJack: Don't think you need this `[HideInInspector]` here, it's already private.

UnityProject/Assets/Scripts/UI/Jukebox/GUI_Jukebox.cs
```diff
@@ -0,0 +1,60 @@
...
+	{
+		labelTrack.Value = jukeboxController.TrackPosition;
+		labelSong.Value = jukeboxController.SongName;
+		labelArtist.Value = jukeboxController.Artist;
+		prefabImagePlayStop.Value = jukeboxController.PlayStopButtonPrefabImage;
+	}
+
+	public void PlayOrStop()
+    {
+		if (jukeboxController.IsPlaying)
```
ReiderJack, line 40: `jukeboxController.IsPlaying ? jukeboxController.Stop() : jukeboxController.Play();`

UnityProject/Assets/Scripts/Machines/Jukebox.cs
JesterX666: Ah... I think I copied this from somewhere else... :)  Gonna fix it.  Thanks

UnityProject/Assets/Scripts/UI/Jukebox/GUI_Jukebox.cs
JesterX666: That won't be legal C#.  I was surprised of it too.  Apparently, you just can't use the conditional operator on it's own without assigning to something in return or piping the result to a function.

int a = 5;
string b = (a == 5) ? a.ToString() : 666.ToString();  // This is legal
Console.WriteLine((a == 5) ? a.ToString() : 666.ToString(); // This is legal
a == 5 ? a.ToString() : 666.ToString(); // This isn't legal (And it doesn't even make sense...)

PR #4734, opened 2020-07-29
Let''s see if this works, Build fix 

Bod9001:
Build fix 
apparently only building atlases in builds apparently caused unity to freeze , so it's just unity being dumb dumb


fomalsd:
looks like it worked

PR #4747, opened 2020-07-30
Mirror Update

ThatDan123:
Updates mirror to latest version.

Got these weaver errors: 

image

Clients crash from this:

image

The weaver errors might be related, or theres some custom code i havnt copied or mirror has changed something.


ThatDan123:
~Feels like this is worth the hassle upgrading as we cut network update time in halve.~

corp-0:
I think Doobly did some stuff regarding networking with the sub scenes update. All prefabs have a new component that may or may not be connected to the errors.

ThatDan123:
Might have been comparing a lobby to in game where theres more objects anyway, so doesnt seem to increase performance by that much. Its like 0.5 ms change, so closing this as its too much work for not that much gain.

Issue #4690, opened 2020-07-23
Server NRE when locker is opened after grenade exploded in it


NoooneyDude:
## Description
Server NRE when locker is opened after grenade exploded in it.

![unitystation_kwij5Gqqmf](https://user-images.githubusercontent.com/6926225/88246978-48928280-ccf0-11ea-9b02-514b6ff1c8a1.png)

### Steps to Reproduce
1. As the host on a development build, place an activated grenade in a locker and close it.
2. Step back and wait for the explosion.
3. Attempt to open the locker and observe the NRE produced.

JesterX666:
Another Exception caused by the same way:

MissingReferenceException: The object of type 'ObjectBehaviour' has been destroyed but you are still trying to access it.
Your script should either check if it is null or you should not destroy the object.
UnityEngine.Component.GetComponent[T] () (at <1245d947a91b4a14b7c7e2c7335512b2>:0)
ClosetControl.OpenItemHandling () (at Assets/Scripts/Closets/ClosetControl.cs:465)
ClosetControl.ServerHandleContentsOnStatusChange (System.Boolean willClose) (at Assets/Scripts/Closets/ClosetControl.cs:453)
ClosetControl.OnWillDestroyServer (DestructionInfo arg0) (at Assets/Scripts/Closets/ClosetControl.cs:148)
UnityEngine.Events.InvokableCall`1[T1].Invoke (T1 args0) (at <1245d947a91b4a14b7c7e2c7335512b2>:0)
UnityEngine.Events.UnityEvent`1[T0].Invoke (T0 arg0) (at <1245d947a91b4a14b7c7e2c7335512b2>:0)
Integrity.CheckDestruction () (at Assets/Scripts/Health/Objects/Integrity.cs:264)
Integrity.ApplyDamage (System.Single damage, AttackType attackType, DamageType damageType) (at Assets/Scripts/Health/Objects/Integrity.cs:221)
Explosion.DamageObjects (UnityEngine.Vector3Int worldPosition, System.Int32 damage) (at Assets/Scripts/Explosion/Explosion.cs:149)
Explosion+<ExplosionRoutine>d__11.MoveNext () (at Assets/Scripts/Explosion/Explosion.cs:84)
UnityEngine.SetupCoroutine.InvokeMoveNext (System.Collections.IEnumerator enumerator, System.IntPtr returnValueAddress) (at <1245d947a91b4a14b7c7e2c7335512b2>:0)
[...]

PR #4751, opened 2020-07-31
NRE Correction Issue #4690

JesterX666:
NRE Correction Issue #4690

Correction of a server side NRE when a grenade explodes in a locker.
The grenade was despawned but still in the locker list of item.

Suggestion for a more intelligent software : the locker should not handle it's own list of item.  We should one day create a component for lockers, boxes, crates, and other containers.


NoooneyDude:
Yes, I was considering your suggestion when I added disposals. The DisposalVirtualContainer gameobject could probably be made more generic, so that it can be used for any container.

PR #4752, opened 2020-07-31
radio broadcast to all matrixes

VernaSaurus:
until someone implements a better solution, i'm just tired of not being able to hear what's going on on-station just because i stepped onto a stupid shuttle

i didn't delete anything, just commented it out for now

i need someone to test this in multiplayer


Bod9001:
isn't the problem with this at the nuclear operatives broadcast the nuke codes?

ThatDan123:
Should be fixed now that you need the disk to do it and the station nuke has different codes.

Bod9001:
ok cool, @MajorMoonjump  make sure to make an issue for the radio system

Bod9001:
NOte: need to test on staging

PR #4759, opened 2020-08-01
Jukebox Volume Control, Jukebox volume attenuation tweak, Change Audio Source Parameters live

JesterX666:
New message to be able to change Audio Source Parameters on a playing sound

Volume slider in the Jukebox GUI

Adjusted the attenuation of the jukebox to hear it a little bit farer (100% up to 4 spaces, 50% at 7 spaces, 0% at 10 spaces.)


corp-0:
Question. If we're using the name of the sound as the identifier to change parameters on the fly, what would happen if you're walking next to the bar and "song x" is playing then you start playing "song x" in your personal player?

JesterX666:
> Question. If we're using the name of the sound as the identifier to change parameters on the fly, what would happen if you're walking next to the bar and "song x" is playing then you start playing "song x" in your personal player?

Both will stop and that would suck.  It's one of the current limitation of the SoundManager.  The name of the sound is the "key" in the playing sound collection, which is terribly a bad idea.  One refactor we will need in a near future, is to make the key something else, for exemple, the NetId of what generates the sound or something like that.

corp-0:
That's what I thought, thanks!

I'm testing the build right now, I do like the distance changes now! Maybe and only maaaybe one more tile of tail for max distance and it is perfect imo. What do others here think? :eyes:

One more thing, I tried sliding the slider but it moved the entire pop up instead. I was able to slide after I clicked in a position of the slider thing, but direcctly trying to slide it didn't work.

JesterX666:
> One more thing, I tried sliding the slider but it moved the entire pop up instead. I was able to slide after I clicked in a position of the slider thing, but direcctly trying to slide it didn't work.

It did that to me also, once.  I don't know what causes that and how to fix it.  Who is really good in Unity UI?  Maybe I'll need counseling on that.

corp-0:
I think it is fine as is. I won't review code because I'm not well versed on messages and stuff but I tested it live with another person and it seems to work.

PR #4762, opened 2020-08-02
Turning vending machines into proper .prefab variants. Some other things.

PetMudstone:
### Purpose
- Turns vending machines into proper .prefab variants (of the machine base). Smartfridges and mining thing not included.
  - Most vending machines in, not all have inventories though.
  - Construction lists for all, but no actual way to build vending machines without deconstructing them and getting the board inside.
- Converting a few more sprites to have SOs.
  - Correcting spray bottle and hypospray variants.
- Rename a lot of chemical bottles.

### Notes:
As stated before, you can't build vending machines without deconstructing one and retrieving the board inside (which can be used to make another vending machine). This is because there is only one vending machine board, and it doesn't have the functionality to be switched to build different vendors as in /tg/. See #4763 for more details.


NoooneyDude:
Merging as builds failing is just due to docker memory issues (again).

PR #4764, opened 2020-08-02
Smite!

JesterX666:
Really basic SMITE command for Admins.
Allows to instantly gib a player.
Only works on players, not on all living beings.


Bod9001:
oh no, the bug from foma's PR Is causing issues

ThatDan123:
I would move this all over to use the AdminCommandsManager.cs scripting system, just to keep them all together.

Has all the net message stuff for you to use there, just remember to put the parameters of your method in the right order.

ThatDan123:
Also add in the admin action message stuff, see the other methods in the admin commands script.

JesterX666:
> I would move this all over to use the AdminCommandsManager.cs scripting system, just to keep them all together.
> 
> Has all the net message stuff for you to use there, just remember to put the parameters of your method in the right order.

Why do we have two places for admin commands?  I've decided to put smite at the same place as all the other commands that targets a player (ban, kick, job ban, respawn, boink).

So, the "official" way to go is AdminCommandManager now?

ThatDan123:
The bans and stuff are in the playerlist as theyre loaded in there. But yeah stick to the manager now.

JesterX666:
Personally, I'm not a fan of he "versiontwo", "versionthree", "versionfour" kinda stuff.  We are still "alpha" and already have 5 ways to send the messages.  In the future, we might arrive to a point that we will have so many different type of calls, that it will become a mess.  

Reflection is known to be slow
Magic strings (i.e.: "string Action") is prone to error.  
Untyped parameters all in the same parameter variable is also prone to errors.

I can see what you tried to do there, it's a very noble idea to standardize stuff.  It's not "bad" code, as it works, but it's not the type of code I prefer.  But, for now, it works perfectly fine.  So, I'll use it.

JesterX666:
@ThatDan123 changes made.  You can proceed for a second review.

UnityProject/Assets/Scripts/Managers/AdminCommandsManager.cs
```diff
@@ -298,6 +294,36 @@ public void CmdSendBlockShuttleRecall(string adminId, string adminToken, bool to
...
+				return;
+
+			var players = PlayerList.Instance.GetAllByUserID(userToSmite);
+			if (players.Count != 0)
+			{
+				foreach (ConnectedPlayer player in players)
+				{
+					var message = $"{player.Name} has been smiten by {PlayerList.Instance.GetByUserID(adminId).Username}: Username: {player.Username}";
+					Logger.Log(message);
+
```
ThatDan123: Add this here too

`UIManager.Instance.adminChatWindows.adminToAdminChat.ServerAddChatRecord(msg, null);
DiscordWebhookMessage.Instance.AddWebHookMessageToQueue(DiscordWebhookURLs.DiscordWebhookAdminLogURL, msg, "");`

ThatDan123: And could you reformat the message to like this one:

$"{PlayerList.Instance.GetByUserID(adminId).Username}: {state} shuttle recalling.";

Admin username, then action

ThatDan123: `var message = $"{PlayerList.Instance.GetByUserID(adminId).Username}: Smitted Username: {player.Username}, {player.Name}";`

Something like that maybe, though whether its smited or smitted, stupid words :p

JesterX666:
Done and tested

image

PR #4783, opened 2020-08-04
Wooden Cleanbot Sprites

BootlegBow:
Adds wooden cleanbot sprites, and separates the wooden cleanbot sprites and blue cleanbot sprites within the cleanbot folder.

The sprites are not functional yet, this just adds the .png files needed.


Bod9001:
do you have access to the unity editor? 
 Since we need the .asset files for referencing them
is ok, if you don't we just make an issue for it

BootlegBow:
I do, I didn't make the .asset files because I'm not entirely sure how to be honest

Bod9001:
image
then the object it makes make sure to name it the exact same thing as your sprite png (might took a second or two to go through)
then make the variance list one
then make the frame list one
then drag in your Sprite then press
Control + s
to seve
And that should be done

BootlegBow:
I’ve tried 3 times, all of which resulted in Unity crashing. Someone else will need to do it

PR #4797, opened 2020-08-06
Fix the antag box bug

VernaSaurus:
Have you ever had a perfect antag round lost by your items being in a box? Well no more! With the new stuff I'm committing here, you can rest easy knowing your stolen goods will be counted regardless of where they are!

Note: Has not actually been tested.

### DIFF
--- a/UnityProject/Assets/Scripts/Antagonists/Objectives/Steal.cs
+++ b/UnityProject/Assets/Scripts/Antagonists/Objectives/Steal.cs
@@ -78,13 +78,23 @@ namespace Antagonists
 			description = $"Steal {Amount} {ItemName}";
 		}
 
+		/// <summary>
+		/// yes it looks stupid as hell and very repetitive but it SHOULD work to fix the box bug
+		/// </summary>
 		protected override bool CheckCompletion()
 		{
 			int count = 0;
-			foreach (var slot in Owner.body.ItemStorage.GetItemSlotTree())
+			List<ItemStorage> storageList1 = new List<ItemStorage>();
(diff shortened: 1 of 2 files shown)


corp-0:
what are the two lists used for?

VernaSaurus:
the lists are used to write/read each itemstorage found to/from

VernaSaurus:
basically it stores every itemstorage it finds in the first check, and then checks if it's found the item, if it has it ends, if it hasn't it checks to see if there are any itemstorages, if there are it repeats the process until either A: it's gone four layers deep, or B: it finds the item

corp-0:
I see. It's a little bit hard to read, don't be afraid to divide your logic and create new functions to abstract the tasks. Also, remember the DRY rule, don't repeat yourself. If you need to do something n times, loop it n times, don't repeat the code n times.

Try doing a little refactor, pretty sure you can!

VernaSaurus:
i can't loop it because even though it LOOKS like it's performing the same check, it's actually performing a slightly different check every time

VernaSaurus:
This is actually required, otherwise I wouldn't do it

ThatDan123:
The first two loops look the same. And try to reduce nesting.

corp-0:
would it work if we iterated all item storages and appended content to a list, then from that list iterate again and append until there are no more itemStorages and then check this single list with all items the character has if it contains the requested item?

VernaSaurus:
i have no idea what any of this means i'm sorry

VernaSaurus:
because i'm a lazy idiot this'll probably never be merged so i might as well close it huh

PR #4804, opened 2020-08-08
Fancy item storage

Macoron:
### Purpose
Added new component FancyItemStorage. The 'fancy' component defines objects like donuts boxes, egg boxes or cigs packets that show how many items are in the ItemStorage on the sprite itself. [Check original DM code](https://github.com/tgstation/tgstation/blob/1aa293ea337283a0191140a878eeba319221e5df/code/game/objects/items/storage/fancy.dm) for more details.

### Notes:
Not implemented open/close sprite change yet. Storages are always "open" by default.
Cigarettes and cigars use much more advanced mechanic and not yet implemented (several slots for different type of items inside).


Bod9001:
looks good, once I finish with the Sprite handler stuff shouldn't need to network with sprites at all

PR #4811, opened 2020-08-08
Storage check

ThatDan123:
Cleaned up storage check.


corp-0:
Looks good and checks passed. I can't think of a solution without looping as @ReiderJack suggested.

PR #4817, opened 2020-08-09
Laser bouncing

ReiderJack:
Bouncing

Bounce1

## Notes
Moved Gun.cs to namespace.
Moved all scriptable objects scripts to namespaces.
Also added OnDisable to all behaviours for clearing references.
Added HitProcessorBehaviout for behaviours which need InteractableTile component to reduce GetComponent calls.
Added Condition SOs for injecting validation to behaviours(So you are able in write your own validation for bouncing and just assign it in the unity editor.)
Reduced default ambient sound volume cuz it was too loud.


Bod9001:
looks good

PR #4818, opened 2020-08-09
NetUI Packet size fix

ThatDan123:
Adds a system which splits and sends multi messages for the arrays if they are too big.

Has a check for if the messages dont arrive in time, and will not continue.

### DIFF
--- a/UnityProject/Assets/Scripts/Messages/Server/TabUpdateMessage.cs
+++ b/UnityProject/Assets/Scripts/Messages/Server/TabUpdateMessage.cs
@@ -18,12 +18,13 @@ public class TabUpdateMessage : ServerMessage
 
 	private static readonly ElementValue[] NoValues = new ElementValue[0];
 
-	private static Dictionary<int, ElementValue[]> ElementValuesCache = new Dictionary<int, ElementValue[]>();
+	private static Dictionary<int, Tuple<ElementValue[], int>> ElementValuesCache = new Dictionary<int, Tuple<ElementValue[], int>>();
 
 	private static HashSet<int> ServerUniqueIdCache = new HashSet<int>();
 
 	public int ID;
 	public int UniqueID;
+	public int NumOfMessage;
 
 	public override void Process()
 	{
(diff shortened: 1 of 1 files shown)


ThatDan123:
Ive added some comments.

Has been tested on headless opening a vendor with 200 items and works fine.

UnityProject/Assets/Scripts/Messages/Server/TabUpdateMessage.cs
```diff
@@ -55,68 +108,178 @@ public static void SendToPeepers(GameObject provider, NetTabType type, TabAction
...
+					Send(recipient, provider, type, TabAction.Close);
+					return null;
+				}
+				break;
+		}
+
+		// SingleMessage: 0, start/middle: 1, end: 2
+		var id = 0;
+
+		var uniqueID = Random.Range(0, 2000);
```
Bod9001: does it need to be random? I feel like it would be safe if it was just a counter that went up?

Bod9001: then rolled over at some point

UnityProject/Assets/Scripts/Messages/Server/TabUpdateMessage.cs
```diff
@@ -15,10 +18,60 @@ public class TabUpdateMessage : ServerMessage
 
 	private static readonly ElementValue[] NoValues = new ElementValue[0];
 
+	private static Dictionary<int, Tuple<ElementValue[], int>> ElementValuesCache = new Dictionary<int, Tuple<ElementValue[], int>>();
+
+	private static HashSet<int> ServerUniqueIdCache = new HashSet<int>();
+
+	public int ID;
```
fomalsd: create a new `enum TabMessageType {MoreIncoming, EndOfMessage}` for that instead of calling it "ID" if it's just going to be "1" or "2"

UnityProject/Assets/Scripts/Messages/Server/TabUpdateMessage.cs
fomalsd: that's certainly not unique lol
expect collisions

ThatDan123: It only has to be unique for the sending, then as soon as its sent can be reused. The chance it fails both checks is minuscule, so i dont really expect collisions.

ThatDan123: But yeah i'll see about changing it to a counter.

ThatDan123: Yeah i'll see about changing it to a counter.

UnityProject/Assets/Scripts/Messages/Server/TabUpdateMessage.cs
```diff
@@ -55,68 +108,178 @@ public static void SendToPeepers(GameObject provider, NetTabType type, TabAction
...
+		var elementValuesLists = new Dictionary<List<ElementValue>, int>();
+
+		if (values != null && tabAction != TabAction.Close)
 		{
 			// get max possible packet size from current transform
 			int maxPacketSize = Transport.activeTransport.GetMaxPacketSize(0);
+
 			// set currentSize start value to max TCP header size (60b)
-			int currentSize = 60;
+			float currentSize = 60f;
```
RedlineTriad: I haven't read this much but that it's a float makes me very suspicious.

UnityProject/Assets/Scripts/Messages/Server/TabUpdateMessage.cs
ThatDan123: Done

UnityProject/Assets/Scripts/Messages/Server/TabUpdateMessage.cs
ThatDan123: I changed it up so it casts to float in the Math.Ceiling

UnityProject/Assets/Scripts/Messages/Server/TabUpdateMessage.cs
ThatDan123: Goes to 10000, then resets

Bod9001:
so, is this covered underneath the bounty?

DooblyNoobly:
sure is

DooblyNoobly:
if it is actually fixed tho

ReiderJack:
> Adds a system which splits and sends multi messages for the arrays if they are too big.

So it splits one entry to multiple? Didn't read code tbh.
foma wrote that

> Shouldn't send more than one update message when a new entry is added to the list, regardless of its current size

ThatDan123:
I wouldnt personally say the bounty is done, this is just a side fix that should fix the errors with the current UI screens. And i wouldnt take the money as it was just like 3 hours of my time, up to doobly and foma for whether they think the current situation would close the Issue.

Yeah does need testing, though if it doesnt work then only the screens which were erroring before should be broken.

Bod9001:
might as well merge

PR #4820, opened 2020-08-09
Smoking and functional lighters

Macoron:
### Purpose
Implemented cigarettes (smoking tube, cigars) and lighters. Finally you can light cigarette by fancy zippo, open plasma canister and enjoy the show.

### Notes:
Added new class for fire sources like lighters, welders and burning cigarettes. They create hotspot and may lead to plasma fire.
Clothing animation seems to not working yet, so burning cigarettes doesn't play burning animation in hands or mouth.

Smoking doesn't affect players health for now.
I didn't find any smoking vending machines on maps, so it's good time to start adding them.

### DIFF
--- a/UnityProject/Assets/Scripts/Fire/FireSource.cs
+++ b/UnityProject/Assets/Scripts/Fire/FireSource.cs
@@ -64,6 +64,18 @@ public class FireSource : MonoBehaviour, IServerSpawn
 		pushPull = GetComponent<PushPull>();
 	}
 
+	private void OnDisable()
+	{
+		// unsubscribe hotspot from updates
+		if (pushPull && CustomNetworkManager.IsServer)
+		{
+			if (isBurning)
+			{
+				UpdateManager.Remove(CallbackType.PERIODIC_UPDATE, CreateHotspot);
+			}
+		}
+	}
+
 	private void CreateHotspot()
 	{
 		// send some heat on firesource position
--- a/UnityProject/Assets/Scripts/Fire/FireSource.cs
+++ b/UnityProject/Assets/Scripts/Fire/FireSource.cs
@@ -10,7 +10,7 @@ public class FireSource : MonoBehaviour, IServerSpawn
 {
 	[SerializeField]
 	[Tooltip("Does this fire emit flame from start?")]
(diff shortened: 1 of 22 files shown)


UnityProject/Assets/Scripts/Fire/FireSource.cs
```diff
@@ -0,0 +1,63 @@
...
+
+/// <summary>
+/// Defines fire sources like candles, lighters or activated welder
+/// Fire source can lead to plasma fire or used to ignite flamable objects
+/// </summary>
+public class FireSource : MonoBehaviour, IServerSpawn
+{
+	[SerializeField]
+	[Tooltip("Does this fire emit flame from start?")]
+	private bool isBurningOnSpawn;
```
NoooneyDude: Could you assign a value or `null` or `default` to any non-public `[SerializeField]` fields? Just helps keep the warnings down. Thanks.

UnityProject/Assets/Scripts/Fire/FireSource.cs
```diff
@@ -0,0 +1,63 @@
...
+	/// Flame may lead to ignite flamable objects or plasma fire
+	/// </summary>
+	public bool IsBurning { get; set; }
+
+	private void Awake()
+	{
+		pushPull = GetComponent<PushPull>();
+	}
+
+	private void Update()
```
NoooneyDude: I would suggest using `UpdateManager` to handle updates. That way you can subscribe for updates only when needed - for example when it is the server and `pushPull` is not null, and `IsBurning`.

ReiderJack:
I like how you name booleans! 👍

UnityProject/Assets/Scripts/Fire/FireSource.cs
Macoron: Should be done now

UnityProject/Assets/Scripts/Fire/FireSource.cs
Macoron: Yep, implemented that. Didn't touch Cigarette and Lighters update loops, cause they will refactored soon for new sprite system.

PR #4822, opened 2020-08-10
Cargo shuttle now doesn't sell people

LuisDSG:
'Edited the CargoShuttle script and separated its item scanner from the "UnloadCargo" function before attaching it to a public "SearchForObjectsOnShuttle" function that returns a Transform Object
-Added a "LifeformChecker" function in CargoManager that looks for any players or animals on the shuttle.
-Modified the "CallShuttle" function on CargoManager so that, using "LifeformChecker", it avoids flying off from the station with a player or animal.

### Purpose
Fixes the issue that allowed the cargo shuttle to fly off from the station with characters on board and erase them afterwards by preventing the shuttle from flying as long as any object that could be alive is aboard.
### Notes:
None

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/CargoManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/CargoManager.cs
@@ -66,9 +66,9 @@ public class CargoManager : MonoBehaviour
 	bool CheckLifeforms()
 	{
 		Transform ObjectHolder = CargoShuttle.Instance.SearchForObjectsOnShuttle();
-		for (int i = 0; i < ObjectHolder.childCount; i++)
+		foreach (Transform child in ObjectHolder)
 		{
-			if (ObjectHolder.GetChild(i).transform.gameObject.layer == 12 || ObjectHolder.GetChild(i).transform.gameObject.layer == 8)
+			if (child.transform.gameObject.layer == 12 || child.transform.gameObject.layer == 8)
 			{
 				return true;
 			}
--- a/UnityProject/Assets/Scripts/Managers/CargoManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/CargoManager.cs
@@ -65,13 +65,15 @@ public class CargoManager : MonoBehaviour
 	/// <returns>true if a lifeform is found, false if none is found</returns>
 	bool CheckLifeforms()
 	{
+		LayerMask layersToCheck = LayerMask.GetMask("Player", "NPC");
 		Transform ObjectHolder = CargoShuttle.Instance.SearchForObjectsOnShuttle();
 		foreach (Transform child in ObjectHolder)
 		{
-			if (child.transform.gameObject.layer == 12 || child.transform.gameObject.layer == 8)
+			if(((1<<child.gameObject.layer) & layersToCheck) == 0)
 			{
-				return true;
+				continue;
 			}
+			return true;
 		}
 		return false;
 	}
--- a/UnityProject/Assets/Scripts/Managers/CargoManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/CargoManager.cs
@@ -63,12 +63,12 @@ public class CargoManager : MonoBehaviour
 	/// Checks for any Lifeforms (dead or alive) that might be aboard the shuttle
 	/// </summary>
 	/// <returns>true if a lifeform is found, false if none is found</returns>
-	bool LifeformChecker()
+	bool CheckLifeforms()
 	{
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Managers/CargoManager.cs
```diff
@@ -59,6 +59,23 @@ private void OnDisable()
 		SceneManager.activeSceneChanged -= OnRoundRestart;
 	}
 
+	/// <summary>
+	/// Checks for any Lifeforms (dead or alive) that might be aboard the shuttle
+	/// </summary>
+	/// <returns>true if a lifeform is found, false if none is found</returns>
+	bool LifeformChecker()
```
ReiderJack: Good rule is to call methods with a verb. In this case for example: `CheckLifeforms`

ReiderJack: Also we specify access modifiers. At least doing it lately. So please add `private`

UnityProject/Assets/Scripts/Managers/CargoManager.cs
```diff
@@ -59,6 +59,23 @@ private void OnDisable()
...
+	{
+		Transform ObjectHolder = CargoShuttle.Instance.SearchForObjectsOnShuttle();
+		for (int i = 0; i < ObjectHolder.childCount; i++)
+		{
+			if (ObjectHolder.GetChild(i).GetComponent<LivingHealthBehaviour>() != null)
```
ReiderJack: This is really expensive way to check if an object is a player. Did you try to check by layers?
<https://docs.unity3d.com/Manual/Layers.html> 
There is a Player layer.

ReiderJack: `if(((1<<gameObject.layer) & layers) == 0)` return `true`;
Where layers has Player layer. 
<https://docs.unity3d.com/ScriptReference/LayerMask.html>

ReiderJack: You can do ` LayerMask.GetMask("Player")`, but cache it before iterating in `for` to not call it everytime.

ReiderJack (changes requested):

UnityProject/Assets/Scripts/Managers/CargoManager.cs
ReiderJack: Also you can iterate without using `for`
```
foreach (Transform child in transform)
{
  //child is your child transform
}
```

corp-0:
@ReiderJack it is mimicking ss13 behavior. It doesn’t just check players but anything that has/had life.

corp-0:
That being said, you can check player and NPC layer to get the same result.

Edit: and dead layer too

ReiderJack:
But Ian has value and he can be sold

corp-0:
In OG SS13 you can’t sell Ian not even dead mobs/players

ReiderJack:
I see the point in not selling alive players, but don't see one for mobs or dead bodies.

corp-0:
Hiding bodies would be way too easy if you can just sell them.

AutumnNova:
not to mention it deletes the player, meaning they cannot be revived with admin intervention at all

ReiderJack:
Just use layers to determine if there is a living thing on a matrix, not `GetComponent`

UnityProject/Assets/Scripts/Managers/CargoManager.cs
```diff
@@ -59,6 +59,23 @@ private void OnDisable()
...
+	/// <summary>
+	/// Checks for any Lifeforms (dead or alive) that might be aboard the shuttle
+	/// </summary>
+	/// <returns>true if a lifeform is found, false if none is found</returns>
+	bool CheckLifeforms()
+	{
+		Transform ObjectHolder = CargoShuttle.Instance.SearchForObjectsOnShuttle();
+		foreach (Transform child in ObjectHolder)
+		{
+			if (child.transform.gameObject.layer == 12 || child.transform.gameObject.layer == 8)
```
ReiderJack: Can't you just do `child.gameObject.layer`? Because `child` is already `transform`.

ReiderJack: Cache `LayerMask` before `foreach`
`LayerMask layersToCheck = LayerMask.GetMask("Player", "NPC")`

Then do the check `if((1<<child.gameObject.laer & layersToCheck) == 0) continue;`
<https://answers.unity.com/questions/422472/how-can-i-compare-colliders-layer-to-my-layermask.html>

ReiderJack (approved):

PR #4827, opened 2020-08-10
Improve jukebox GUI and create PressStart2P TextMeshPro asset

### LINKED ISSUE #4761: Jukebox Volume Slider doesn't register click every time.
Jukebox volume slider doesn't registrer click and drag correctly.
Also, we can see that the "handle"  doesn't even register the mouse over color change!
I've tryied solving it but can't find the cause.  (I'm not very good in Unity UI).

It might be something related to Raycast target, bounding boxes, placement of controls and such.

It might even be simple enough to be a good first issue.  No code is involved.

Steps to reproduce:

1.  Click on the jukebox
2.  Try to move the volume handle (doesn't work)
3.  Click somewhere like the middle of the slider
4.  Now the volume handle works.

Korywon:
### Purpose

Fixes #4761. Improves the jukebox GUI by using the PressStart2P font, used by the other tabs. I also added panels and made it to where it looks like an actual console. The sliders are also made larger and looks proportional to the rest of the tab.

### Notes:

I created a TextMeshPro asset of the PressStart2P font which is much sharper than using the normal `.ttf` file. I took the `.ttf` file and threw it into FontDrop to find the character indexes. I then took these character indexes and shoved them into Unity's Font Asset Creator to create the TextMeshPro asset.

`NetLabel` requires the `TMPro` component now, hence why I created the asset. I encourage everyone to start using this asset and convert any tabs still using `Text` components with the normal `.ttf` file. `TMPro` is much sharper and scales better then normal `Text` components.

I also removed "GUI specific" code from `Jukebox.cs`. The labels (i.e. "Arist:", "Song:", etc.) should either be handled in `GUI_Jukebox.cs` or by the prefab (which I did here).


Bod9001:
nice, thanks for Making the font asset!

PR #4845, opened 2020-08-12
Temporary patch for  VV Working not on some books

Bod9001:
Needs to be looked into the actual issue of it since this happens quite a lot, VV  Generally needs a good Clean through, currently working on pipes ,  Then optimisation after this


corp-0:
is this ready to merge?

Bod9001:
yeah

PR #4846, opened 2020-08-13
Added floor guide decal sprites.

nuke-haus:
### Purpose
Floor guide decals make stations more navigable.

Here's what they look like: https://cdn.discordapp.com/attachments/402986688114065409/743258433574142013/Untitled.png

Note that the command decal has been changed to the nanotrasen symbol. Slightly outdated screencap but you get the idea.

### Notes:

Decals are not implemented yet apparently but i figure doesn't hurt to collect the assets for it.


corp-0:
Thank you! I'm merging this since it is just assets.

PR #4856, opened 2020-08-14
All MonoBehaviour singletons now derive from MonoBehaviourSingleton class.

nuke-haus:
### Purpose

Tidy up some code and make all singletons derive from the same singleton class.

### Notes:

I wanted to do this with NetworkBehaviour deriving classes too but it's not possible to use generics with NetworkBehaviours, sadly.

More info on the issue: https://forum.unity.com/threads/networkbehaviour-cannot-have-generic-parameters-why.400728/


Macoron:
Looks like UI manager is now broken if you start game from lobby or restart round
image

PR #4860, opened 2020-08-14
implement the ability to have more then one projectie fired per shot

AutumnNova:


UnityProject/Assets/Scripts/Weapons/Gun.cs
```diff
@@ -652,6 +651,13 @@ private void DequeueAndProcessServerShot()
 			{
 				b.Suicide(shooter, this, damageZone);
 			}
+			else if (ProjectilesFired != 1)
```
ReiderJack: You can just have one `else` statement without `if`. Iterate through one projectile.

UnityProject/Assets/Scripts/Weapons/Gun.cs
```diff
@@ -652,6 +651,13 @@ private void DequeueAndProcessServerShot()
...
+			{
+				for (int n = 0; n > ProjectilesFired; n++)
+				{
+					b.Shoot(ApplyRecoil(finalDirection), shooter, this, damageZone);
```
ReiderJack: So the recoil is applied twice, because that passed direction is off set with a recoil.

ReiderJack: Bullets will be offset a little bit. Better to rotate their direction and make rotation dependent on count of projectiles fired.

UnityProject/Assets/Scripts/Weapons/Gun.cs
```diff
@@ -23,9 +23,9 @@ public class Gun : NetworkBehaviour, IPredictedCheckedInteractable<AimApply>, IC
 	*/
 
 		[SerializeField]
-		private GameObject ammoPrefab = null;
+		protected private GameObject ammoPrefab = null;
```
corp-0:
```suggestion
		protected GameObject ammoPrefab = null;
```
Only one access modifier is needed here.

UnityProject/Assets/Scripts/Weapons/Gun.cs
```diff
@@ -37,7 +37,8 @@ public class Gun : NetworkBehaviour, IPredictedCheckedInteractable<AimApply>, IC
 		/// <summary>
 		///     The type of ammo this weapon will allow, this is a string and not an enum for diversity
 		/// </summary>
-		[FormerlySerializedAs("AmmoType")] public AmmoType ammoType;
+		[FormerlySerializedAs("AmmoType")] 
+		public AmmoType ammoType;
```
corp-0, line 53: Use a private property with ``SerializeField`` decorator and a public getter like this:
```suggestion
        [SerializeField]
		private AmmoType ammoType;
		
		public AmmoType AmmoType => ammoType;
```

Also please include Tooltip for all fields exposed to inspector. Using summary is alright for code you're meant to work with in your IDE, but inspector fields benefit from having tooltips so you just hover and see what the thing is about.

UnityProject/Assets/Scripts/Weapons/Gun.cs
```diff
@@ -122,7 +123,7 @@ public class Gun : NetworkBehaviour, IPredictedCheckedInteractable<AimApply>, IC
 		/// <summary>
 		///     The amount of projectiles expended per shot
 		/// </summary>
-		public int ProjectilesFired;
+		public int ProjectilesFired = 1;
```
corp-0, line 150: if this is meant to be set in inspector, use a private property with ``SerializeField`` decorator and a public getter. If that's not the case, then hide this public from inspector with the ``HideInInspector`` decorator.

PR #4868, opened 2020-08-15
Composite image ui

Macoron:
### Purpose
Previously **UI_ItemSlot** could render only two sprites. Other item sprites were ignored, so some objects with more than two sprite renders looked weird (for instance, welding tool).
Now new class **UI_ItemImage** can render any number of sprites for any item. It manages pool of UI images and create new instances if needed. 
The term "SecondarySprite" was replaced by "Overlay" and used for something like handcufs.

### Note
In future new class should be used for other UI elements, like dev spawn menu or right click item selector.


Bod9001:
is this integrated with the Sprite handler system?
E.G it handles animations

Macoron:
Animation plays same way as in old system. Some custom behaviour script constantly request pickupable.RefreshUISlot() and redraw all sprites updated by SpriteHandler. I just copy-pasted old code to new class.
I'm not a fan of current approach and have some ideas how to move slot animation either in SpriteHandler or UI_ItemImage. But it's for another PR.

PR #4875, opened 2020-08-16
Improved editor compile times

Bod9001:
tested with a build,
you'll have to give it a quick try to see if it improves anything at all or just makes it worse,  first time is going to take about a minute for me then it is about 20 seconds recompiling something like lathe script


corp-0:
Yes, I think I see an improvement. I'm concerned tho with the files marked as changed, are there any you didn't intend to change? There are 982 and some changes look odd.

Bod9001:
thisI had a look through and they seem about right

corp-0:
Alrighty, merging

PR #4877, opened 2020-08-16
Turning projectiles into variants, giving them proper damage, and sprites. Alter Buldinn's CentCom.

### LINKED ISSUE #4864: Bullet damage scriptable objects mostly share the same damage values.
## Description
Title. For some reason most of the SOs in "Assets\Resources\ScriptableObjects\Gun\DamageInfo" share the same damage of 15. At time of writing only Bullet_KineticShot and Bullet_Sniper do different damages.

PetMudstone:
### Purpose
  - Turns all projectiles and decals into variants.
    - Gave some projectiles proper sprites and unique effects.
    - Replace old projectiles in references.
  - Gives all projectiles proper damage values again, fixes #4864.
  - Adds a script that spawns decals upon a projectile despawning instead of on it hitting something. Used for tazers, plasma cutters, and flamethrowers. **Thanks to ReiderJack for providing the code.**
  - Tweak Buldinn's CentCom to replace some old .prefabs and fix references (mainly the latter).
  - Gave disruptor cannon (the pulse rifle) a new firing sound.
  - Start on gun variants. Add a nifty x-ray gun 
    - X-ray bolts pass through anything that isn't a mob, at the cost of reduced range.
  - Tazer stun time increased from 3 seconds to 4 seconds.

### Notes:
- I kept the old .357 bullet and bullet decal around as not to interfere with Autumn's work on Russian revolvers. They have had BVOld added to the start of their names, as is tradition.
- The following errors appear the first time you fire a PKA. Only the first time though. Firing any PKA afterwards creates no further errors and PKAs work perfectly fine.
unknown


ThatDan123:
I would rather see if someone can find out why theres NRE before merging.

corp-0:
I didn't get any NRE shooting PKAs in my pr or before it. What did you change in PKA scripts?

ReiderJack:
Inspector

The kinetic "Bullet" which is actually not a bullet, should not inherit from ProjectileBase.
The errors you get are right. The prefab initialized with all scripts on it and they are removed then.

Kinetic

Just leave it as a separate prefab. It would not have being a bullet in first place if gun.cs was not hardcoded.

So leave the KineticShot as it is now. 

KineticShot

Someday Gun.cs will be refactored and there won't be a need for "Beam" bullets at all.

ReiderJack:
Current implementation of the Plasma Cutter bullet is not what it is supposed to be in terms of hit impacts.

This is on your branch:
PlasmaCutterBroken

This is what it was before you changes:
PlasmaCutterBulletBefore

So the problem is that you removed `IOnHit` decal script and left only `IOnDespawn`.
PlasmaBlast

There are multiple solutions for this:
1) just return `ProjectileDecal` script, but on final hit there will be spawned two decals.

2) Modify your `IOnDespawn` decal script so you can toggle in the editor when you want the decal to appear:
So you add `bool` for controlling it from editor, check the bool first then if the hit collider is not null.
SecondSolution

Also you can remove `bool` from return values in your script.

ReiderJack:
Also you can turn those scripts into one cuz they do the same thing. Just pass `hit.point` to `OnBeamEnd(Vector2 position)` on collision.

TurnIntoOne

The reason i did it with two scripts in `ProjectileAnimateTile` is because it needs to off set `hit.point` to get proper tile it hit.

AnimateTile

PetMudstone:
ALright, I'll try to fix these issues right now.

PetMudstone:
Alright, should be good to go now.

Issue #2593, opened 2020-01-13
Usable Sprite Animation System - Composite Inventory Icons B: 80


chairbender:
Sub-task of #2311
Refer to that ticket for guidance / overview of the plan for the sprite animation system.

Expand on #2591 / #2592  to support composite inventory icons (how an item looks in an inventory slot). Just like simple inventory icons, whatever animations are playing on the object should also play when the object is in inventory.

This will require modifying how the Panel_Hud_Bottom.prefab works, as currently it only supports 2 sprites for inventory items. It needs to support multiple sprites.

Following criteria need to be met:

- [ ] Objects with composite icons in inventory should have the same appearance as if they were in the world, based on what animation is currently playing.
- [ ] The design should be flexible so it can support future use cases.
- [ ] A few objects should be migrated to use this new system to prove it out.
- [ ] Dev team agrees that it meets our needs as far as usability is concerned. Expect some feedback and iteration.

You don't need to completely migrate everything to this new system, it can be set up to coexist alongside the old system so it can be migrated in a piecemeal fashion.

Refer to the suggestions in #2311 for guidance on the design / implementation.

chairbender:
Setting a bounty of 80.

Bod9001:
return to pool

Issue #2592, opened 2020-01-13
Usable Sprite Animation System - Simple Inventory Icons B: 30


chairbender:
Sub-task of #2311
Refer to that ticket for guidance / overview of the plan for the sprite animation system.

Expand on #2590 to support simple inventory icons (how an object looks in an inventory slot). The requirements are simple - the object's inventory icon should reflect whatever the object's current animation is. So, its appearance in inventory should match how it would appear if it was in the world.

This only needs to support simple (single sprite) icons. Composite icons will be done in #2593 .

Following criteria need to be met:

- [ ] Objects in inventory should have the same appearance as if they were in the world, based on what animation is currently playing.
- [ ] The design should be flexible so it can support future use cases.
- [ ] A few objects should be migrated to use this new system to prove it out.
- [ ] Dev team agrees that it meets our needs as far as usability is concerned. Expect some feedback and iteration.

You don't need to completely migrate everything to this new system, it can be set up to coexist alongside the old system so it can be migrated in a piecemeal fashion.

Refer to the suggestions in #2311 for guidance on the design / implementation.

chairbender:
Adding a bounty of 30.

Bod9001:
return to pool

PR #4891, opened 2020-08-16
Automatically animated UI icons 

Macoron:
### Purpose
Previously each animated item should had custom script to update animation frames in UI slots. That lead to a lot stupid, inefficient and repeatable code in different scripts with same line `pickupable.RefreshUISlotImage();`
This PR implements automatic way for UI_ItemImage to update image state based on current SpriteHandler state. Now SpriteHandler fires event when animation frame changed. Subscribed UI_ItemImage change its sprite respectively.

### Notes:
UI image will not be updated if SpriteRender sprite was replaced or animated by another script. It only works if SpriteHandlers have full control over SpriteRender. 
Not all occurrence of `pickupable?.RefreshUISlotImage();` was replaced or deleted. Old approach is still working, but it should be replaced in the future.
Cigarette, lighters and stun button was refactored for new approach. Other items should be refactored in different PR.
Energy sword was broken before this PR. Though now inventory animation works correctly, clothing animation still funny.

### DIFF
--- a/UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
+++ b/UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
@@ -46,6 +46,8 @@ public class SpriteHandler : MonoBehaviour
 
 	private NetworkIdentity NetworkIdentity;
 
+	private Pickupable pickupable;
+
 	private bool isSubCatalogueChanged = false;
 
 	[SerializeField]
@@ -415,6 +417,9 @@ public class SpriteHandler : MonoBehaviour
 			}
 
 			spriteRenderer.SetPropertyBlock(block);
+
+			// update UI icon if avaliable
+			pickupable?.RefreshUISlotImage();
 		}
 		else if (image != null)
 		{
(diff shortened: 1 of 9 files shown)


UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
```diff
@@ -423,8 +482,9 @@ private void SetImageSprite(Sprite value)
 			{
 				image.enabled = false;
 			}
-
 		}
+
+		OnSpriteChanged?.Invoke(value);
```
Bod9001, line 487: how much of a performance impact is this?

Bod9001:
This technically covers
#2593
#2592
you want to take the Bounties?

UnityProject/Assets/Scripts/Sprite Handler/SpriteHandler.cs
Macoron: This event invokes for each frame in object animation (gap between frames usually 0.2-0.4 second). For non-animated items or objects it is just one null check. It's nothing.
For animated items this event just replace sprite in UI Image class. Performance here is even better than before. In old approach custom script used to create coroutine or even Update cycle to manually refresh UI slot.

Keep in mind that this is Client side code only. From my test I can't see any impact on performance.

Macoron:
Tbh, I'm not really interested in bounties and can return it to pool. 
Hope that I didn't break anything in progress, but issues seems to be complete.

Bod9001:
cool, loooks good,merging

PR #4904, opened 2020-08-17
Removes old canisters

ThatDan123:
Replaces the BVoid canisters from outpost and pog.

Deleted the BVoid canister prefabs

-Fall is being updated separately.


ThatDan123:
Tests pass locally

PR #4915, opened 2020-08-19
Fixed MissingRef exception for UI_ItemImage

Macoron:
### Purpose
Quick fix for UI_ItemImage. 
Fixed MissingRef exception when player place animated object in container and close inventory menu.


ThatDan123:
Not sure if related

`
MissingReferenceException: The object of type 'Image' has been destroyed but you are still trying to access it.
Your script should either check if it is null or you should not destroy the object.
UI_ItemImage+ImageAndHandler.OnHandlerSpriteChanged (UnityEngine.Sprite sprite) (at Assets/Scripts/UI/UI Bottom/UI_ItemImage.cs:255)
SpriteHandler.OnDisable () (at Assets/Scripts/Sprite Handler/SpriteHandler.cs:569)
`

Macoron:
Yep, should be fixed in this PR

PR #4928, opened 2020-08-20
Outpost Atmos Part 1

ThatDan123:
-Adds the atmos system to the atmospherics department only, rest of station needs connecting.

-removes syndicate matrixes

-Fixes Gas container Out of Range Exception


ThatDan123:
Tests pass locally

PR #4929, opened 2020-08-20
emag stuff

VernaSaurus:
probably works, i tested it all and it does do things

only adds door, locker, and vendor functionality


VernaSaurus:
oh no not the closets

VernaSaurus:
fixed the closets

PetMudstone:
Do emags remove access restrictions from vendors?

VernaSaurus:
yes, yes they do

PetMudstone:
This doesn't even have merge conflicts anymore. Come on, github-actions!

UnityProject/Assets/Scripts/Items/Emag.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class Emag : MonoBehaviour
```
Bod9001: this seems to be unused?

corp-0:
> This doesn't even have merge conflicts anymore. Come on, github-actions!

The action description says it will get rid of the label when there are no conflicts anymore but that seems to be false. We can either disable that action or just manually unlabel when the conflicts are solved.

UnityProject/Assets/Scripts/Items/Emag.cs
VernaSaurus: Yeah I completely forgot about this, it's 100% unused

VernaSaurus:
removed the unused file, should be good to merge now

PR #4930, opened 2020-08-20
WIP Airlocks revamp

### LINKED ISSUE #4908: A better way to set sprites for airlocks, or: the airlock animator is a nightmare to deal with.
## Description
While trying to work on the goal of #3235 (before I knew it was a thing) I had massive trouble trying to get the airlock animator to work properly (it didn't). In order to actually use the airlock animator, you have to set all the right sprites in the sprite objects in the airlock's hierarchy and then while said airlock's .prefabs is open, modify a random variable before and/or after (I don't know) automatically generating references for the airlock animator so that said modifications to the airlock animator actually save.

This would be inconvenient by itself, but often times the generated references by the airlock animator just don't work like you want them to.

Airlock Atrocity

As can see you above, this opened airlock has floating sparks and lights for some reason. I merely wanted to get windows overlays to work after several tries and it did this for some reason.

It should be noted that the sprites for airlock lights, windows, sparks, and open maint panels (for hacking) all use the same files, which would be why the airlock animator replaces windows with sparks for some reason. Perhaps it would be better to separate groups of overlay sprites (green lights, yellow lights, red lights, windows, sparks, open maint panels, etc.) into their own files? Also, a lot of airlock sprite files could use renaming to make them much less confusing to look through.

You also can't manually edit sprite references in the airlock animator component unless you have multiple .prefabs using it selected. Even then the overlay sprite list is so long that you can't edit it.

corp-0:
Moves airlock's logic to a state pattern design. 
Handling the following door interactions:
- Opening.
- Closing.
- Bolts (open and closed).
- Weld.
- Pry open.
- Fried circuit board.
- Power issues (over voltage, under voltage, unpowered).
- Exposing maintenance panel.
- Disabling AI control.

Moves the updating sprites responsability to SpriteHandler component so everything is networked.

fixes #4908


corp-0:
why would I have missing references in scenes I haven't touched?

corp-0:
closing because I will just nuke and start again with help of Fortune.

PR #4934, opened 2020-08-21
Pointing

Aranclanos:
### Purpose
-Adds player pointing, a playernetwork command to relay a message in the chat and play an arrow effect on all clients.
-Adds a keybind for pointing, left-shift and middlemouse combo.
-Adds a class to add on objects with particle effects if they need to run custom scripts for more structured flexibility.
-Two key-combos that share the same key will now have priority regarding the amount of keys in the combo itself, making the default arrow pointing action not trigger hand swap.
-Pointing will now check for the player status before going off, both client and server.

**Visual update since webm posted in discord:**
-Pointing arrow will now have a fade effect to make the spawn and despawn more subtle.
-Adds changes of speed for the point arrow effect, the bouncing will now have a static speed irregardless of the initial speed.

### DIFF
--- a/UnityProject/Assets/Scripts/Effects/CustomEffectBehaviour.cs
+++ b/UnityProject/Assets/Scripts/Effects/CustomEffectBehaviour.cs
@@ -1,0 +1,10 @@
+﻿using UnityEngine;
+
+public class CustomEffectBehaviour : MonoBehaviour
+{
+	public ParticleSystem particleSystem;
+	public virtual void RunEffect(Vector2 target)
+	{
+
+	}
+}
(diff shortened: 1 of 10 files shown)


Bod9001:
looks good, a lot of work on the key bindings

UnityProject/Assets/Scripts/Effects/CustomEffectBehaviour.cs
```diff
@@ -0,0 +1,10 @@
+﻿using UnityEngine;
+
+public class CustomEffectBehaviour : MonoBehaviour
+{
+	public ParticleSystem particleSystem;
```
corp-0, line 5: is this settable in inspector? If that's the case, do a private and use the ``SerializeField`` decorator with a ``Tooltip``. If you need to get this from another places, do a public getter.

UnityProject/Assets/Scripts/Effects/CustomEffectBehaviour.cs
```diff
@@ -0,0 +1,10 @@
...
+	public virtual void RunEffect(Vector2 target)
```
corp-0, line 6: some docstring would come in handy.

PR #4947, opened 2020-08-22
Atmos Tile Reactions

ThatDan123:
Adds all the atmos reactions from tg repo as of 4 months ago, cus i was too lazy to download the latest.

Only remaining thing to do is the fire effects for tritium/freon, which isnt gonna be in this PR as touching hot spots looks scary.

(plasma fires still use old system will need to be moved into the new system when the other fire effects added)

I also added my own feature, where if water vapour gets too cold ice will spawn, also if ice gets too warn it will turn back into water vapour.


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/ReactionManager.cs
```diff
@@ -221,6 +231,40 @@ private void Update()
 		hotspotsToRemove.Clear();
 		Profiler.EndSample();
 
+		Profiler.BeginSample("GasReactions");
```
Bod9001, line 234: do the reactions need to be on the main thread?

ThatDan123: The reactions are only called on the main thread if they are possible, having it this way might slow down the reactions which could be good. 

I would do a proper deep profile of it but atmos is having issues with the pipe system.

So at the moment i think it's fine to have the actual reaction on the main thread, also allows for unity functions in reaction code.

Bod9001: ok cool

PR #4955, opened 2020-08-24
Add new-er sprites for crates and suit storage units. A few adjustements.

PetMudstone:
### Purpose
- Adds new-er crate sprites **by Cornelius Bogfrollup Fitzgerald**.

Crates
- Adds new suit storage unit sprites **by WatlerJe**.
suit_storage_suit_storage

- Breaks up the images for the crate sprites into individual images for closed and opened crates.
- Makes the cable coil sprite palettable in its SO.
- Gives the jaws of life proper held sprites after the clothing component got deleted.
- Makes riot shotguns fire beanbags, so that they're actually non-lethal as the description claims.

### Notes:
- I modified the science crates in the above sprite sheet to be white instead of grey. Cornelius is aware and didn't object.
crates_science
- Hydroponics crates (the ones with green stripes) above don't look like that in-game. They actually look like this (unedited by me).
crate_hydroponics
- WatlerJe's suit storage unit sprites [can also be found in CEV Eris](https://github.com/discordia-space/CEV-Eris/pull/4431/files). Notably, that PR didn't credit WatlerJe and it also added a few other sprites, which is odd.
  - Speaking of WatlerJe's sprites, he made animated door sprites for the SSU that opens and closes but I added them to the asset library instead.


PetMudstone:
Odd how the actions bot only gave this the sprite work label after my second commit. Is it because that one added new image files? Could have just bugged out too.

PR #4964, opened 2020-08-25
Adds a ContainsKey check into RemoveHandler

Akbadain29:
### Purpose
Stops the RemoveHandler funciton of the EventManager class from checking if the event key is null before checking the dictionary key exists existance.

#### Notes
I was getting an KeyNotFoundException with RemoveHandler, so I thought this might help. Should there be a eventTable.ContainsKey(evnt) for the second if statement as well?


ThatDan123:
Its still going to break on the next check

`if(!eventTable.ContainsKey(evnt)) return;`

Add that to the top of the function

Akbadain29:
I have added that line into the function, sorry about the wait.

PR #4969, opened 2020-08-27
Escaped prefix labels, changed scene labels

daalta:
~~*Requires labels to be updated before merging! Not sure how we can best test this.*~~ *Relevant labels have been renamed*

# Description

Escaped all keys in the label yaml. `js-yaml` *should* be able to handle this, but I only tested it in an online parser.

The escaped keys contain some changes. Map names now have a `Scene: ` prefix to make sure the labels appear next to each other on the list.

### For example:

```
Admin
BoxStation
Engineering
Squarestation
```
would now be:
```
Admin
Engineering
Scene: BoxStation
Scene: SquareStation
```

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,33 +1,30 @@
-Scene Changes:
+"Scene Changes":
   - UnityProject/Assets/**/*.unity
 
-SquareStation:
+"Scene: SquareStation":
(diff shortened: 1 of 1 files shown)


daalta:
The new labels are in. So the old automatic labels are definitely broken!

...Which means that this PR can be merged without breaking anything! And potentially fixing something 🤡

PR #4979, opened 2020-08-28
Fixed Warnings (now below 100 again)

daalta:
Similar to https://github.com/unitystation/unitystation/pull/4180 https://github.com/unitystation/unitystation/pull/4099

Fixed some simple compilation warnings. Such as unassigned variables.

GameManager also had some null checks for floats - which caused warnings, because floats cannot be null.


UnityProject/Assets/Scripts/Alien/AlienEggCycle.cs
```diff
@@ -24,6 +24,7 @@ public class AlienEggCycle : MonoBehaviour, IInteractable<HandApply>, IServerSpa
 		[SerializeField]
 		private EggState initialState = EggState.Growing;
 
+		// TODO This is entirely unused.
```
ThatDan123, line 27: Should be used on caverna away site unless i missed doing that.

daalta: @ThatDan123 It's definitely set inside the editor. But the compiler can't find any references to it being *used* anywhere.

We might want to leave the variable for future use, but are you certain it's implemented?

daalta:
Alrighty, I think I'm done for today. Would be good if we can get this merged today so there's no conflicts.

ThatDan123:
Looks good

PR #4980, opened 2020-08-28
Added in Deleted Tiles To Outpost

Akbadain29:
## Purpose
When I put in some of the atmos pipes, I deleted some tiles to remove some pipes I incorrectly placed. They should be all back now.


corp-0:
thanks, merging.

PR #4984, opened 2020-08-28
fixes Pipes

Bod9001:
fixes a few issues with pipes

fixes all thies
Wrenching/Unwrenching meters doesnt work
Wrenching and unwrenching pipes seem bugged on staging.
Unwrenching should not be instant if pressure is high to stop accidental removal
 Unwrenching pipe seems to not update atmos correctly?
 Need the ability to rotate pipes?

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/AirVent.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/AirVent.cs
@@ -53,6 +53,11 @@ namespace Pipes
 				return;
 			}
 
+			if (metaNode.GasMix.Pressure == 0)
+			{
+				return;
+			}
+
 			float Available =
 				((MaxOutletPressure / metaNode.GasMix.Pressure) * metaNode.GasMix.Moles) - metaNode.GasMix.Moles;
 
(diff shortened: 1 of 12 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/AirVent.cs
```diff
@@ -53,6 +53,11 @@ private void CheckAtmos()
 				return;
 			}
 
+			if (metaNode.GasMix.Pressure == 0)
```
ThatDan123: What does this do? Just if i'm not misunderstanding then If the tile pressure is 0, the vent wont try to output?

Bod9001: yeah, it was dividing by zero so I need to think of a better thing for it, in case it is zero

Bod9001: fixed

corp-0:
the lines with question marks are part of the fixes or problems not fixed in this pr?

Bod9001:
Unwrenching pipe seems to not update atmos correctly?
might need a bit more testing but I fixed one issue That was causing it


Need the ability to rotate pipes?
There is a way for rotating but properly not what is intended

corp-0:
Merging so we can give it a try in staging

PR #4989, opened 2020-08-29
Dividing by zero checks

Bod9001:
see what is causing issues probably won't get everything


ThatDan123:
I dont think just adding a log to it is the most helpful

Bod9001:
I've done some fixes as well

Issue #4952, opened 2020-08-23
Atmos Pipes Performance Issues


ThatDan123:
Calling setup neighbours every frame is decreasing performance. This is a headless server deep profile.

![image](https://user-images.githubusercontent.com/56727168/90978717-c217cd80-e547-11ea-9d3d-282dc60a7ab2.png)
Fall Station, with just vents and scrubbers, not the pipe system

![image](https://user-images.githubusercontent.com/56727168/90978718-c47a2780-e547-11ea-862d-1281218b846f.png)
Square station with the pipe system.

NoooneyDude:
Was this fixed by #5004?

Bod9001:
need some testing, the wind is probably a bit laggy still

corp-0:
now that Aran moved pipes to atmos thread, what's the state of this one?

corp-0:
@Aranclanos can you tell?

ThatDan123:
Dont think this is an issue now

PR #5004, opened 2020-08-30
Atmospheric optimisation

Bod9001:
shouldn't update meta tile neighbours when scrubbers and vents are active


Bod9001:
Should fix #4952

corp-0:
won't be able to test it for a while since building takes time.

PR #5018, opened 2020-08-31
WIP: Lights now fall off more quickly (inverse square law)

daalta:
### Purpose
Lights are currently rather bright. That's because they fall off linearly `brightness ≈ distance`, which makes it easy to light the station evenly. After some distance the brightness of the light reaches zero.

Real lights fall off with the inverse square law `brightness ≈  1 / (distance)^2`. Which is more realistic, but also darker. It makes it harder to light environments evenly.

This PR contains an experimental light texture which falls off with the inverse square law... but there's also 1/3rd of the old light texture mixed in.

### The good

* Lights are darker, changing the station's atmosphere quite a bit... for the better, I think?
* Most things are still visible!
* Lights look... more realistic? Less cartoonish, maybe?

### The bad

* Looks good when it's dark, looks... dark when it's not dark.
  * The size of the light texture was not increased. But it should have been, maybe? We can't really introduce inverse square falloff without increasing the range of lights. Otherwise the station would stay dark.
* The 1/3rd linear part of the texture might have to be tweaked. And there's no easy way to experinent with the textures right now, requiring a file replacement each time.

#### Comparison: Interior 
https://cdn.knightlab.com/libs/juxtapose/latest/embed/index.html?uid=3fd44c4c-eb94-11ea-bf88-a15b6c7adf9a  (Left in inverse square, right is linear. Messed up the labels 🤷‍♀️)
8romcWsiDx

#### Comparison: Dark Hallway
https://cdn.knightlab.com/libs/juxtapose/latest/embed/index.html?uid=b4d74d78-eb94-11ea-bf88-a15b6c7adf9a
iljBCgMyUF

#### More comparisons

The gifs above are compressed, so I'd suggest checking the URLs for a good comparison.

https://cdn.knightlab.com/libs/juxtapose/latest/embed/index.html?uid=45490176-eb95-11ea-bf88-a15b6c7adf9a

https://cdn.knightlab.com/libs/juxtapose/latest/embed/index.html?uid=5d2c9de8-eb95-11ea-bf88-a15b6c7adf9a

#### Texture comparison

JAxZUiFLIZ


ReiderJack:
me want see images, bork

corp-0:
> Real lights fall off with the inverse square law brightness ≈ 1 / (distance)^2. Which is more realistic, but also darker. It makes it harder to light environments evenly.

Sounds like more maps changes will be required

daalta:
> 
> 
> > Real lights fall off with the inverse square law brightness ≈ 1 / (distance)^2. Which is more realistic, but also darker. It makes it harder to light environments evenly.
> 
> Sounds like more maps changes will be required

In its current state the new falloff has big issues, it's just way too dark. But mapping changes might not be needed, depending on the final implementation. I'll post a simple comparison...

daalta:
@ReiderJack comparisons added :)

daalta:
Oh by the way, here's another design consideration: At which distance should lights be absolutely black? 

We could reduce very low light values to 0. The new light sprite goes from a brightness of 4 directly to zero, but it could cut off even sooner.

#### Pros

* Players can't increase their screen brightness to gain an advantage
* Lights would have a clearly defined range
* The area of the light texture would be utilized better

#### Cons

* Overlapping lights at large distances would not "accumulate" their brightness values
* Very intense lights would have a sudden brightness drop at their maximum range

Food for thought 🤷‍♀️

Bod9001:
humm, tell me what you think of the light  Sprite I put up

daalta:
@Bod9001 Do you mean the one you posted on Discord?

Bod9001:
yeah, should be in the game currently

daalta:
Cool. I'll get back to this Merge Request at the end of the week, feel free to close it if it clogs up the list.

daalta:
Closing this due to inactivity (on my end). But this seems to have encouraged some discussion on the topic, I think bod already implemented changes on another branch.

PR #5023, opened 2020-09-01
Fix for freezing vacuum when moving from shuttles

Aranclanos:
### Purpose
Fixes players sometimes being exposed to a freezing vacuum when moving inbetween two matrix (for example, shuttle to station)

### Notes:
this is a crude visual explanation of the bug:
image

now the matrix will update at the start of the movement

### DIFF
--- a/UnityProject/Assets/Scripts/Player/PlayerSync.Server.cs
+++ b/UnityProject/Assets/Scripts/Player/PlayerSync.Server.cs
@@ -472,11 +472,11 @@ public partial class PlayerSync
 		lastDirectionServer = Vector2Int.RoundToInt(newPos - oldPos);
 		ServerState = nextState;
 		//In case positions already match
-		TryNotifyPlayers();
 		if (lastDirectionServer != Vector2.zero)
 		{
 			CheckMovementServer();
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Player/PlayerSync.Server.cs
```diff
@@ -472,11 +472,11 @@ private void TryUpdateServerTarget()
 		lastDirectionServer = Vector2Int.RoundToInt(newPos - oldPos);
 		ServerState = nextState;
 		//In case positions already match
-		TryNotifyPlayers();
```
fomalsd, line 475: why did you remove this one? removing it might make other players receive position updates later

Aranclanos: since this is at the start of the movement, it instantly gets cancelled out because the positions are the same

fomalsd: you probably meant `positions are no longer the same`? since `ServerState` was assigned with `nextState` and `ServerPositionsMatch` is no longer true – yeah, that would make sense
otherwise `SyncMatrix()` call from `TryNotifyPlayers()` would be executed and freezing wouldn't be an issue in the first place

PR #5024, opened 2020-09-01
fixes clothing not refreshing properly

Bod9001:
changes light sprite,
adds colour picker to VV
fixes UI actions
fixes clothing not refreshing properly
addeds Mono behaviour manager Singleton tho tried implementing it but you UI manager had issues for some reason


corp-0:
if the singleton doesn't work, remove the class to avoid anyone of using it thinking it does. Then we merge.

Bod9001:
Singleton does work, it's just some classes do accessing in a strange way/strange times/custom cleaning up

ReiderJack:
If it's not used anywhere, no point keeping it imo

Bod9001:
ok done

NoooneyDude:
Fixed by #5024.

PR #5028, opened 2020-09-02
Gun stuff: foam darts, more calibers, replacing magazines, and better clips.

PetMudstone:
## Purpose
- Adds several new calibers for guns and magazines including: .75, 40mm, 84mm, and foam darts.
- Updated the magazines that most guns use, including internal ones (see below).
   - Revolvers now use internal magazines, instead of ejecting speedloaders, for some reason.
- Adds a script that makes projectiles drop objects on despawn, **courtesy of ReiderJack and Bod9001**. Added so that foam darts fired by toy guns can be reused.
- Disruptor laser shots now do large amounts of damage to walls, enough to usually destroy them in one hit.
- Laser tag guns now actually fire harmless lasers instead of harmful ones.
- Move all the old magazines around.
- Replace references to guns and magazines.
- Changes to Gun.cs and MagazineBehavior.cs, **courtesy of Autumn**.
  - Unitystation has progressed pass the need for a Generic Internal Magazine. Now you can define the unique internal magazine variants to be used in the standard ammo field. Remember to set your gun to use an internal magazines though.
  - Clips can now reload external magazines. Used so that shotgun shells can reload a brand-new drum magazine and that foam darts can reload toy magazines.


ReiderJack:
This PR didn't tackle 
`Attack messages for projectiles are for melee(should be reworked).`
did it?

PetMudstone:
No, no it has not.

ReiderJack:
Alright, ill look into it

PR #5037, opened 2020-09-02
Adds test for Objects that get set at 0,0  And removed some stuff that was causing warnings

Bod9001:
and removed unneeded scenes


Bod9001:
let's see if the test gets picked up on

NoooneyDude:
Yup the test caught them. Nice. Now let's see if my PR fixes the canister ones.

PR #5041, opened 2020-09-03
Add links to action badges

Fogapod:
I'm not sure about `Build Project` workflow, I don't see it in https://github.com/unitystation/unitystation/tree/develop/.github/workflows, last run I could find happened 29 days ago.
Also GitHub does not let allow `Build%20Project` as query for some reason, it only shows 7 months old successful builds.
I used `Build` in query instead.

Rendered version can be seen here: https://github.com/Fogapod/unitystation/blob/badges/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,14 +10,14 @@
 
 _Unitystation has been active since late 2016_
 
-Visit us on Steam:
+Visit us on Steam:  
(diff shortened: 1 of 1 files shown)


Bod9001:
Build Project currently bugging out for some reason, and not finishing (that's why it is disabled/removed), we doing some addressable bundle stuff so once we done that gonna try re-enabling it,

PR #5073, opened 2020-09-07
Fix wood and reinforced table construction

NoooneyDude:
- Wood and reinforced tables are now constructable and do not turn into glass tables.
- Removes most instances of magic prefab strings in favour of inspector fields.


Bod9001:
looks good, some of these Could have Used the common prefabs SO instance

NoooneyDude:
Oh, I had no idea. I'll update them.

Bod9001:
is ok ,quite   obscure it is

NoooneyDude:
Updated. Ready for review.

PR #5101, opened 2020-09-10
fixes  names with numbers on the end

Bod9001:
it checks if the joint viewer is the same account, assuming that you can't have the same account for multiple players


Bod9001:
test feeling for no reason, I suspect it's github's end

PR #5119, opened 2020-09-13
fixes plasma reaction with minus moles

Bod9001:
I'm going to do some testing to see if I can get the vents to bug out


Bod9001:
needs a few more fixes

Bod9001:
ok, fire code should be moved to thread

Bod9001:
good enough for now

PR #5131, opened 2020-09-14
optimises load time

Bod9001:
Removed some find object of type, and split out a Initial Functions of a few frames

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/CentComm.cs
+++ b/UnityProject/Assets/Scripts/Managers/CentComm.cs
@@ -2,13 +2,14 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using Initialisation;
 using UnityEngine;
 using Random = UnityEngine.Random;
 
 ///------------
 /// CENTRAL COMMAND HQ
 ///------------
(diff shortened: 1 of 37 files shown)


UnityProject/Assets/Scripts/Managers/CentComm.cs
```diff
@@ -78,11 +79,14 @@ public void UpdateStatusDisplay(StatusDisplayChannel channel, string text)
 	private string AntagInitialUpdate =
 		"Enemy communication intercepted. Security level elevated.";
 
-	void Start()
+	public InitialisationSystems Subsystem => InitialisationSystems.CentComm;
+
+	void IInitialise.Initialise()
 	{
 		paperPrefab = Resources.Load<GameObject>("Paper");
```
corp-0, line 86: this should probably be a GameObject field rather than magical string loading.

corp-0: I know, not part of this PR. Just commenting for later.

Bod9001: good catch

Bod9001:
build didn't work so well

PR #5141, opened 2020-09-15
Added Thrusters to Shuttles on Box and Outpost Stations

Akbadain29:
### Purpose
Added Thrusters to Shuttles on Box and Outpost Stations, now RCS should work on those maps too.


Bod9001:
looks good, merging

PR #5142, opened 2020-09-16
Fixes #5132 and makes Atmos pressure and temperature functions more readable and intuitive at a glance

### LINKED ISSUE #5132: Absolute zero exhaling and Atmos Poltergeists
## Description
For some reason, the CO2 breathed out by people comes out very, very cold (and possibly spontaneously generated?). normally said co2 expands but is removed by scrubbers before it can cause issues, however, if one wires the waste gas pipe (red) directly to the station atmos pipe (dark blue), possibly because of "hehe traitor turn off life support" antics, said supercooled CO2 is pumped back in, heats up, and massively increases pressure. With ss13 atmos being wonky as fuck, physics gets thrown out the window and this supercooled co2's massive increase in pressure heats up the system because PV = nRT is a poor formula to use in this case and the combined gas law would be better. Regardless, this magic increase in heat (which, surprise, wouldn't occur in real life) causes new cooled co2 coming in to get heated up even more, making more pressure, making more heat, and creating an infinite loop. This creates huge pressure differences and heats up the station considerably, to the point that just being in the air may cause burn damage if it gets far enough. These pressure differences lead to strange other atmos behaviours, such as atmos tornadoes and poltergeists which both have the capacity to gib/kill players instantly.

### Steps to Reproduce
In order to break the laws of physics:

1.  connect the red output into blue input pipe, switch pump so waste goes back into the station air
2.  wait for a very long time (well over an hour)
3.  Observe atmos shit itself and create swirling vortexes of garbage and slam nukies into walls so hard they splatter

In order to fix this, you could either rework atmos completely or make people not breathe out co2 at absolute zero.

coolcat1711:
Modified Accessors for Temperature and Pressure to allow proper setting of Temperature based on Pressure or vice versa without having an infinite loop. Replaced the setter call in other files as needed.

Simplified GasMix functions and procedures for better readability and easier understanding of functions at a glance.

Fixes #5132 - CO2 and O2 now have constant pressure and temperature on exchange.

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/Gas.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/Gas.cs
@@ -12,7 +12,7 @@ namespace Atmospherics
 		public static readonly Gas Plasma = new Gas("Plasma", 200, 40f, true, 0.4f, "PlasmaAir", -4, 0);
 		public static readonly Gas Oxygen = new Gas("Oxygen", 20, 31.9988f, false, 0.4f, "NONE", -3, 0);
 		public static readonly Gas Nitrogen = new Gas("Nitrogen", 20, 28.0134f, false, 0.4f, "NONE", -3, 0);
-		public static readonly Gas CarbonDioxide = new Gas("Carbon Dioxide", 30, 44.01f, false, 0.4f, "NONE", -3, 0);
+		public static readonly Gas CarbonDioxide = new Gas("Carbon Dioxide", 20, 44.01f, false, 0.4f, "NONE", -3, 0);
 
 		public static readonly Gas NitrousOxide = new Gas("Nitrous Oxide", 40, 44.01f, true, 0.4f, "NO2", -5, 10);
 		public static readonly Gas Hydrogen = new Gas("Hydrogen", 30, 44.01f, false, 0.4f, "NONE", -3, 0);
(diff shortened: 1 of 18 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/Gas.cs
```diff
@@ -12,7 +12,7 @@ public struct Gas
 		public static readonly Gas Plasma = new Gas("Plasma", 200, 40f, true, 0.4f, "PlasmaAir", -4, 0);
 		public static readonly Gas Oxygen = new Gas("Oxygen", 20, 31.9988f, false, 0.4f, "NONE", -3, 0);
 		public static readonly Gas Nitrogen = new Gas("Nitrogen", 20, 28.0134f, false, 0.4f, "NONE", -3, 0);
-		public static readonly Gas CarbonDioxide = new Gas("Carbon Dioxide", 30, 44.01f, false, 0.4f, "NONE", -3, 0);
+		public static readonly Gas CarbonDioxide = new Gas("Carbon Dioxide", 20, 44.01f, false, 0.4f, "NONE", -3, 0);
```
NoooneyDude: I'm guessing you made this change while testing issue #5132? Could probably remove this change?

https://i.imgur.com/xLMGVzS.png

corp-0:
```suggestion
		public static readonly Gas CarbonDioxide = new Gas("Carbon Dioxide", 30, 44.01f, false, 0.4f, "NONE", -3, 0);
```

PR #5146, opened 2020-09-16
Some fixes

Bod9001:
some fixes for the staging server


corp-0:
good to merge?

Bod9001:
yeah

UnityProject/Assets/Standard Assets/Mirror/Runtime/NetworkConnection.cs
```diff
@@ -19,7 +19,7 @@ public abstract class NetworkConnection : IDisposable
         // internal so it can be tested
         internal readonly HashSet<NetworkIdentity> visList = new HashSet<NetworkIdentity>();
 
-        Dictionary<int, NetworkMessageDelegate> messageHandlers;
```
Bod9001, line 22: not 100% sure this is a good fix, we'll see if anything breaks

PR #5147, opened 2020-09-16
Fixed UI stackingg

Bod9001:
yea


Bod9001:
oh, need to test

Bod9001:
should be good after tests

PR #5149, opened 2020-09-17
New minor antag: Fugitives

corp-0:
> There's nothing more refreshing than freedom, and these fugitives are willing to kill for it.

## Why?
Fugitives are a special antag that can randomly spawn in a traitor round. Their only objective is to get to the end of the round alive and free (not cuffed). They spawn in a special spawn point called MaintSpawns (or arrivals if not mapped) wearing the prisoner jumpsuit, no ID and a handy toolbox to make their way to freedom.

After a random amount of time (5 to 15 minutes by default, might need tweaking), Centcom sends the station a report and ask them to neutralize the threat.

The intention is to diversify a little the rounds with some mechanically simple antags.

## Notes
- Since the antag system doesn't support weighted antags yet, fugitives are just as likely to spawn as normal traitors. If a player doesn't want to spawn as fugitive, they can simply disable the chance in the antag preferences.
- Some maps are just awful for fugitives, like Fall that has no maints and is very small.

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/SceneConstruction/NestedManagers/GameManager.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/SceneConstruction/NestedManagers/GameManager.prefab
@@ -70,7 +70,7 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   gameManager: {fileID: 5243784332070655206}
-  CommandStatusString: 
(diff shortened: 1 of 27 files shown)


Pockets-byte:
what about bounty hunters?

corp-0:
not in the game yet

wraith-54321:
might want to check if a fugitive always spawns when the event happens

PR #5158, opened 2020-09-18
Update canister and radar entry

NoooneyDude:
- Adds sprites to NetSpriteImage component.
- Fixes compiler warnings.


Bod9001:
you could hook up NetSpriteImage with sprite handler system?
since all you need is an int for the  Sprite location on the manager

NoooneyDude:
I considered it, but the comms console shuttle images (`shuttle_status.psd`) (and the others too, probably) don't have sprite SOs. Should I create the SOs?

Bod9001:
yeah, do it in another PR updating unity

PR #5159, opened 2020-09-18
add check to removegas method preventing removal of gas when no gas exists to be removed

AutumnNova:


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/GasMix.cs
```diff
@@ -328,6 +328,7 @@ public GasMix AddGasReturn(Gas gas, float moles)
 
 		public void RemoveGas(Gas gas, float moles)
 		{
+			moles = Mathf.Clamp(moles, 0);
```
Bod9001: isn't it from minus number to positive number?
with clamp

PR #5162, opened 2020-09-18
Requesting licence file

Bod9001:
to get tests working


Bod9001:
I think it didn't actually update oof

PR #5130, opened 2020-09-14
Adds better player examination

NoooneyDude:
- Adds a system to obscure slot contents.
    - This would be useful for not only examination but also preventing other players from removing items from obscured slots. Not included in this PR.
- Adds better player examination.
    - Slot contents that are not obscured are listed.
    - Player soul state can be found from an examination.
examination

> Only a few clothing prefabs have the obscuring slots defined for testing. All of them should be checked and set accordingly.

PR #5165, opened 2020-09-19
ItemActionButton component

NoooneyDude:
- Adds `ItemActionButton` component.
    - Handles inventory movement, specific slots, death etc.
- Flashlights, PDAs and magboots all now use this component.
- Fixes action button sprites not syncing to client.
- Fixes old action buttons transferring to new character if respawned.


UnityProject/Assets/Scripts/Items/Others/ItemMagBoots.cs
```diff
@@ -3,24 +3,19 @@
 using System.Collections.Generic;
 using UnityEngine;
 using Mirror;
-using Atmospherics;
+using Assets.Scripts.UI.Action;
```
corp-0: can you change this namespace to ``UI.Action`` ? Using the system boundary instead of the folder path.

UnityProject/Assets/Scripts/Inventory/NamedSlot.cs
```diff
@@ -43,11 +43,12 @@ public enum NamedSlot
 [Serializable, Flags]
 public enum NamedSlotFlagged
 {
-None = 1 << 0,
-Outerwear = 1 << 1,
-Belt = 1 << 2,
-Head = 1 << 3,
-Feet = 1 << 4,
+None = 1 << -1,
```
corp-0: why -1 ?

NoooneyDude: Because the equivalent `NamedSlot` is `2048` ~~which would conflict with one of the `NamedSlotFlagged` slots~~, but you cannot bit shift by `2048`.

NoooneyDude: Probably shouldn't bit shift with a negative number. What do you suggest?

NoooneyDude:
Looks like tests are failing due to what is already on develop. Aside from that stuff, tests are fine locally.

UnityProject/Assets/Scripts/Inventory/NamedSlot.cs
Bod9001: Why don't you put the number in exactly 
None = 2048
?

NoooneyDude: Won't that conflict with `NamedSlotFlagged.Hands` (`1 << 11` => `2048`)? I swapped to assigning the number with bitshift instead of a non-shifted number to make it more readable and easier to maintain, as per the suggestion in PR #5130.

Bod9001: I'm confused why you can't just have zero?

NoooneyDude: Ah, of course. Done.

PR #5166, opened 2020-09-19
sunvox preload fix?

Bod9001:
don't know if this fixes the Couldn't find sunvox  Dll  errors
does this work?
should be simple enough to revert


corp-0:
is it safe to just delete sunvox like that?

Bod9001:
I think I'm just setting it to preload it right?

corp-0:
aight, want to give it a try? We can revert if something breaks horribly.

PR #5167, opened 2020-09-20
Small idle performance tweak

NoooneyDude:
- Converts some `UpdateManager` calls to the new periodic type.
- Moves out some unnecessary code from the update methods.

### DIFF
--- a/UnityProject/Assets/Scripts/Construction/Glass/IceShard.cs
+++ b/UnityProject/Assets/Scripts/Construction/Glass/IceShard.cs
@@ -19,18 +19,22 @@ public class IceShard : MonoBehaviour
 
 	private void OnEnable()
 	{
-		UpdateManager.Add(UpdateCycle, 1f);
+		if (CustomNetworkManager.IsServer)
+		{
+			UpdateManager.Add(ServerUpdateCycle, 1f);
+		}
 	}
 
 	private void OnDisable()
 	{
-		UpdateManager.Remove(CallbackType.PERIODIC_UPDATE, UpdateCycle);
+		if (CustomNetworkManager.IsServer)
+		{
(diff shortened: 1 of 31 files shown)


UnityProject/Assets/Scripts/Construction/Glass/IceShard.cs
```diff
@@ -19,18 +19,22 @@ private void Awake()
...
 	}
 
-	private void UpdateCycle()
+	private void ServerUpdateCycle()
 	{
-		if(!CustomNetworkManager.IsServer) return;
-
 		var pos = registerTile.WorldPosition;
 
 		if (pos == posCache && metaDataNode.GasMix.Temperature > AtmosDefines.WATER_VAPOR_FREEZE)
```
ThatDan123: On the `else` (line 46) after this can you change it to `else if (pos != posCache)` seems I forgot that optimisation.

As its just going to change the coord even if its the same if the temp isnt correct.

NoooneyDude: Done!

corp-0 (commented):
Looks pretty good. Did you profile it to see if we got any improvement in performance?

Also, I see you removed a couple IsServer checks, I assume that’s because you’re 100% sure that code ran only on server, right?

NoooneyDude:
Yeah, moved the server checks to `UpdateManager` call instead. And yes, I did profile. See the screenshots. The selected profile in the first image is attributed just to `Integrity` checks on nearly five thousand objects with regards to fire. Managed to move that out of `UpdateManager` and that code will only be run when set on fire.
Note that this is just showing the profile of scripts only, not physics or the like.

> Before - with `UpdateManager => Integrity` selected
before

> After
after

corp-0:
Nice, merging

PR #5030, opened 2020-09-02
Airlock system rework to be more scalable and maintainable

Fortune117:
### Purpose
A major overhaul/refactor to door prefabs and animations. The system is now separated into a central door controller (which handles input, and opening/closing of the door itself) and modules. It is built using the new prefab created by Gilles.

Modules are components that can be drag-n-dropped onto door prefabs to give added functionality to the door. They are independent and completely modular.

Current modules include: Pressure Warning System, Access Restriction, Welding, Door Prying, Door Bolts (not yet implemented, but the base is there) and Hacking (similar position as door bolts).

In addition, the animation system has been overhauled by Gilles, which is pretty great. It's much more usable now, and easier to extend.

I've also added integration with APC Power. When the door is considered 'on' you can open and close it like normal. When the door is 'off', you can pry it open with a crowbar. If it's on, you need the jaws of life. Additionally, if the door is in low voltage, there's a small chance for the access check to return a false positive. The chance for this is paramatised and can be changed or disabled entirely.

Currently only has the base airlock prefab, I haven't created all the door prefabs yet cause it's not really a coding issue, it's just a bit tedious. It'd be a good first issue for someone to look at.

Fixes  #3235.

### Notes:
It's not 100% done and requires some polishing. But it's at a point where mappers are able to begin replacing the different doors on the map with the new prefab.

There's some minor known issues:
The animation for keycard rejection plays behind the doors for some reason when hosted on a server, and not when run in the unity editor. As this is (likely, not 100% sure) due to a the animation player (created by Gilles), I haven't looked into this.
Animation for access denied will sometimes play when prying the door open with a crowbar if you don't have the proper access for the door.

PR #5175, opened 2020-09-21
New airlock variants and airlock assembly tweaks.

PetMudstone:
### Purpose
- Adds a few new airlock variants using the new airlock base introduced in #5030.
- Adds some sprites to the sprite catalogues of the airlock assemblies, for future usage.

### Notes:
- Incapable of properly testing due to current status of my primary desktop computer. Additions and modifications were done on a much less powerful laptop.


corp-0:
Thanks. I can give it a look in editor tomorrow. Did you add the animations? What do you think about the animation workflow now?

PetMudstone:
I did add all the relevant sprites to the sprite catalogues in the .prefabs, yes. I did miss a few since the sprites are absent for some reason in both here and in /tg/. A bit tedious to work with, but at least I can actually work with it unlike the old airlock animator. Doesn't help that all the damn airlock overlay files are named the same so I had to look at file paths.

We could use a generic sprite SO for a blank sprite, come to think of it. I'm not sure how to deal with airlocks that lack filling sprites so I just filled all the catalogue fields with the same open panel "sprite".

corp-0:
Absolutely, a generic empty sprite we all know about is a very good idea. Would you mind doing a copy of the open panel but name it something everyone can understand as an empty sprite?

PetMudstone:
I'd make a blank sprite if my spare laptop wasn't busy importing right now, so I'm not quite capable of doing so at the moment.

PR #5180, opened 2020-09-22
Fix conscious states

NoooneyDude:
- Swaps from using `isServer` to `CustomNetworkManager.IsServer`.
    - Fixes consciousness state, and probably other things.


corp-0:
you're committing your wiz changes here along with the fix, is that ok?

NoooneyDude:
Huh, I don't remember pushing the latest commit. It'd normally be fine, but someone may spawn as wiz if there's enough players. I'll remove it.

NoooneyDude:
Closed in favour of PR #5182.

PR #5200, opened 2020-09-24
apparently no graphics works

Bod9001:
https://github.com/webbertakken/unity-actions/issues/84


Bod9001:
let's see if this works

Bod9001:
for anyone looking over the PR, the test run successfully tho,  looks like we got to fix a few issues on our end thanks for the help

PR #5204, opened 2020-09-25
Fixes players spawning in subscenes exhibiting space exposure symptoms

NoooneyDude:
- Fixes players spawning in additional scenes at roundstart exhibiting space exposure symptoms.
    - This was because additional scenes are loaded at roundstart (as the scenes depend on the gamemode, which is determined at that time) and so `MatrixInfo` was not ready, meaning players were defaulting to the mainstation matrix.

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/GameManager.GameMode.cs
+++ b/UnityProject/Assets/Scripts/Managers/GameManager.GameMode.cs
@@ -135,9 +135,16 @@ public partial class GameManager
 
 		foreach (var job in GameMode.PossibleAntags)
 		{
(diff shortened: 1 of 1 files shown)


corp-0:
Nice. Only errors in tests are the APC related ones that are going to be solved soon. Merging.

ThatDan123:
uh not sure how this fixes the wizard scene as its not the syndicate scene, surely you'd need the frame skip to be outside the loop?

corp-0:
maybe make everyone wait, no matter the job? We might have non antag spawn in other matrixes as well.

PR #5210, opened 2020-09-26
fix player respawn

corp-0:
In my last version of player respawns it was only possible to respawn alive people, this will fix it.

Silly me didn't know ghosts had no health script.


UnityProject/Assets/Scripts/Messages/Client/Admin/RequestRespawnPlayer.cs
```diff
@@ -24,7 +24,9 @@ void VerifyAdminStatus()
 			if (deadPlayer == null || deadPlayer.Script == null) return;
```
ThatDan123, line 24: deadPlayer.Script is already checked here :p

corp-0: one step ahead of you, already fixed it 😎

PR #5214, opened 2020-09-27
2 walls

catcen:
should add two new 3/4 walls to play around with, a base metal one and one for windows. mainly to put it in the hands of people whilst work is done on reinforced etc.


corp-0:
Can you move them to the AssetLibrary folder? So it is only detected by unity when someone moves them back into the Assets folder and tries to implement them

catcen:
I got so confused there, i was working off an older fork where there was no assetlibrary folder lmao. Yeah ill do that

PR #5220, opened 2020-09-29
Convert effects overlays to use SpriteHandler, new DirectionalSpriteV2 component

NoooneyDude:
- Converts `BurningOverlay` and `PlayerDirectionalOverlay` to use `SpriteHandler`.
    - Fixes NRE when items burn in staging.
- Adds new component `DirectionalSpriteV2` - same as `DirectionalSprite` but uses SpriteSOs and their variants as defined in the `SpriteHandler`s instead.
    - Soda and booze dispensers now use this new component.


UnityProject/Assets/Scripts/Health/Objects/Integrity.cs
```diff
@@ -133,8 +134,8 @@ private void EnsureInit()
 		if (registerTile != null) return;
 		if (SMALL_BURNING_PREFAB == null)
 		{
-			SMALL_BURNING_PREFAB = Resources.Load<GameObject>("SmallBurning");
-			LARGE_BURNING_PREFAB = Resources.Load<GameObject>("LargeBurning");
+			SMALL_BURNING_PREFAB = Resources.Load<GameObject>("BurningSmall");
```
corp-0, line 137: these ``Resources.Load`` could be replaced with another way of referencing the prefabs, such as a serialized field.

NoooneyDude: Gave it a go but would have to update every prefab that uses the component. Alternative is to use common SO singleton, but would be rather niche.

corp-0: Then move the reference to a static like the CommonPrefabs

PR #5223, opened 2020-09-29
implements the ability for guns to use different ammotypes

AutumnNova:
demo
requires weapon prefabs to have their projectile var updated


corp-0:
I'm afraid you have compile errors, that's why tests are failing. Try opening in editor and if it doesn't say any error, try building.

AutumnNova:
accidentally removed an extra bracket, 's why every test but that one was successful

PR #5230, opened 2020-09-30
Cleanbots removing slippery factor

### LINKED ISSUE #5222: Cleanbot Mayhem
## Description
Cleanbots make lube invisible instead of actually cleaning it up
### Steps to Reproduce

1.  Create and use lube
2.  Create clearbot near them
3.  Observe results.

Aranclanos:
### Purpose
Cleanbots will now make floors stop being slippery.
Removes a copypasted cleaning method, and now all cleaning will use the Clean() in MetaDataLayer
Fixes half of #5222

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
@@ -918,28 +918,6 @@ public partial class MatrixManager : MonoBehaviour
 		return positionList;
 	}
 
-	public void CleanTile(Vector3 worldPos, bool makeSlippery)
-	{
-		var worldPosInt = worldPos.CutToInt();
-		var matrix = AtPoint(worldPosInt, true);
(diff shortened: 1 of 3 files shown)


corp-0:
I will make it close the issue. If the bust thing is clarified, we can reopen.

PerfectTangent:
I believe it's referring to a case where a cleanbot moved onto a tile that had a hippocrates statue to clean blood under it and it deleted the statue

corp-0:
if the bust is a game object, then it makes no sense.

AutumnNova:
I believe npcs at some point could seek out and consume food, so im assuming somehow the bust is being consumed?

corp-0:
only if someone manually set the bust as food for that particular NPC

Aranclanos:
at the moment of cleaning the cleanbots -used to- run 
`var floorDecal = registerObj.Matrix.Get<FloorDecal>(checkPos, true).FirstOrDefault(p => p.Cleanable);`

But neither prefabs called Bust or BustHippocrates have a FloorDecal

PR #5231, opened 2020-09-30
small electricity related fixes

corp-0:
## Includes:
* Namespace stuff
* Making tedious references auto fill themselves in some components
* Make setting master-slave relation actually persists by making everything dirty with a lot of anger. This fix works for APCs, light switches, Fire alarms, door switches and anything that uses the subscriber system by Reider.


Bod9001:
looks good, tho those extra tabs really do mess with git difference

Issue #5013, opened 2020-08-31
The ability to weld lockers shut.


PetMudstone:
## Description
In /tg/, you can weld lockers shut. This is done to prevent people from just opening lockers from both the outside and from inside. However, you can't weld crates and cardboard boxes, which are children of lockers, shut. Therefore, the ability for a container to be welded should be a variable. There should also be a field for a welded overlay sprite as well.

Related to #4997. You can break out of welded lockers (and other containers, if they are weldable) the exact same way you can break out of locked lockers. It is also important that players be able to break out so that people can't just trap everyone else forever by stuffing them into welded lockers.

It takes roughly 5 seconds to weld or unweld a locker.

jamessimo:
I am working on this issue now and have closets welding shut on my local. 

Is there a reason why you cant weld crates (assuming their metal) in /tg/?

corp-0:
In my opinion, it makes sense to have weldable crates

PetMudstone:
There aren't weld overlay sprites for crates in /tg/. That's it, really. I don't see why you shouldn't be able to weld crates shut.

jamessimo:
#4657 

Some overlap with this issue since closet/crate Emag, weld and crowbar logic will be in its own InteractionsCloset script (same as  how doors work).

corp-0:
Large crates are meant to only be openable with crowbars

jamessimo:
Also touches #5068 since a fix to this would be breaking out of locked/welded lockers

jamessimo:
I think this issue can be closed now, see merges above.

PerfectTangent:
so lockers can be resisted out of, the lock can be broken with an emag, and you can't weld yourself into lockers, correct?

jamessimo:
@PerfectTangent yes, yes and yes!

Issue #4997, opened 2020-08-30
The ability to break out of locked containers such as lockers by using the RESIST button.


PetMudstone:
## Description
In /tg/, when you get locked into into a locker (or any lockable container, really) you can move or click the RESIST button to start breaking out. By default, it takes 2 minutes (or 120 seconds) to break out of a locker, although this number can be adjusted. 
Breaking out of a locker also breaks the lock on said locker, making said locker unable to be locked again.

![image](https://user-images.githubusercontent.com/38266309/91648960-ffd29580-ea22-11ea-80d7-cfcb1bd7cce5.png)
Default breakout time above. Note, it's 1200. Perhaps it's measured in tenths of a second?
![image](https://user-images.githubusercontent.com/38266309/91648953-eaf60200-ea22-11ea-840a-6be003bb6d2e.png)
Message in chat when you start breaking out.
![image](https://user-images.githubusercontent.com/38266309/91649061-66a47e80-ea24-11ea-9a45-c0265bcf68ab.png)
Message in chat when you successfully break out.

jamessimo:
Going to work on this. 

I know in /tg/ its a flat 2 mins to get out but would it be more realistic if it only takes 1min if locked/emaged and 1 min if welded and if both then make it 2 mins? (maybe a lock is easier to break than a weld so we can tweak these).

I know I can't lock the closet anymore, but can I still open and close the door? Technically we could break the door off its hinge (remove the door sprite and drop some metal/planks.

PetMudstone:
The properties of how long it takes to break out of welded/locked closets are up to you.
In regards to opening and closing the door, that is still possible after breaking out of a locker. It's just that the lock becomes non-functional.

Issue #3054, opened 2020-02-11
Further Emag properties


Raptorizer:
### What is wanted

- [x] Emaggable Vendors

- [x] Emaggable Doors

- [ ] Emaggable Wall Mounts

- [x] Emaggable Lockers

- [x] Emaggable Consoles

(Such as ID restricted door switches and APCs)

https://tgstation13.org/wiki/Emag <- Basically all of this

### Vendor Steps
1. Get an emag
2. Find a vendor
3. Emag it
4. Vendor now shows extra items (needs to only happen when emagged)
5. (Optional) emagging vendors also display the extra items that are shown via hacking them.

The script should allow drag and droping of prefabs into a list much like how door switches work.

### Door Steps
1. Get an emag
2. Find a locked door or access restricted door
3. Use emag on door
4. Door sparks and opens
5. (Optional) when hacking is implenented, it should also break the motherboard inside the door. Requiring getting a new one to put inside.

### Wallmount Steps

1. Get an emag
2. Find a wallmounted object
3. Use emag on wallmounted object 
   a. If used on APC,  it allows anyone to have full access to its systems
  b. APC sprite is then changed to have a blue screen
  c. (Optional) When AI is setup and working, it displays a message to the AI that an APC has been messed with but does not give an exact location
  d. If used on an ID restricted door switch, it grants anyone the ability to use it.

### Locker Steps
1. get an emag
2. Find a locker
3. Use emag on locker
4. Locker is unlocked and anyone can open it.

### Console Steps
1. Get an emag
2. Find a console
3. Use emag on console
4. Depending on the console, it should do different things

Emma3DSS:
Emags also need to be able to be used on doors and break said door keeping it over when used, maybe we should make this a general emag features issue

Raptorizer:
Page has been updated to be an emag 
expansion feature

Misterixmax:
Actually first we need make vendors and doors repairable. Also many consoles also has to be possible to emag, for example cargo.

Raptorizer:
> Actually first we need make vendors and doors repairable. 

If we have the functionality set in place for other systems, then what point is there in waiting. Even then, why wait for a function not yet even implemented yet?

>Also many consoles also has to be possible to emag, for example cargo.

This can be added to the list as well

Misterixmax:
[...]

PR #5233, opened 2020-10-01
Resist out of lockers, break lock and no longer weld self in closet

### LINKED ISSUE #5068: Dropping your ID in a Locker Locked by the ID Softlocks You
## Description
Softlock means that the game hasn't crashed, but there's no way to further progress.

### Detailed Bug Report
Dropping your ID in the locked locker makes you unable to pick the ID up, so you're trapped in the locked locker with no way out.

### Steps to Reproduce
1. Get in locker 
2. Lock locker while inside by clicking with ID in hand.  
3. Drop ID while in locked locker.
4. You're stuck in GBJ now, good job.

### Possible Solutions
* Implement resisting out of lockers
* Implement picking up items inside lockers while in lockers
* I've ran out of ideas

### LINKED ISSUE #5163: Can weld self inside closet
## Description
Right now you can weld yourself inside a closet, you can also secure a closet whiles inside it. 

### Detailed Bug Report
Due to old implementation of breaking out of a locker, when ever you move inside a locked or welded locker you will try to use whatever is in your hand. also if your hand is empty you get the following error 
`
ClosetControl.ServerPerformInteraction (HandApply interaction) (at Assets/Scripts/Closets/ClosetControl.cs:506)
`

### Steps to Reproduce

1.  Enter any closet with a weldgun 
2.  Weld the locker with harm
3. Try to move any direction and you will start de-welding 
4. remove welder from hand to see error msg.

jamessimo:
### Purpose

- To allow people to break out of welded/locked lockers using the Resist button or just moving whiles in a locked locker. #4997 
- Added the function to easily break locker locks, used when breaking out of lockers and when you emag lockers
- Added a "IsTrapped" bool to players, when a player enters any closet they are "Trapped", on Closet exit (forcefully or optionally) you are no longer trapped. 
- Added an "Escaping" enum to ProgressActions to be used when escaping anything. 
- Fixes #5163 
- Finishes features listed in #5013
- Can emag closets #3054, made some minor tweaks, now emag doesn't force the door open since we can now break locks on command. 
- Fixes #5068

### Notes:

- Made a UX decision to auto fire the resist out of locker function if you are trapped in a locker and its your only logical choice.
- Made a slight change to EMAG logic, now we dont force the door open but break the lock. The reason for focing open the door and keeping it open was so the lock light wouldnt be visable as we didnt have a realaible way to "Break" it, now we do. EMAGED lockers still cannot be repaired.  
- Added a new progress state on the player to be used when "escaping" anything, a TODO would be to see if climbing out of disposals should leverage this. 
- There is sound code for emagging and resisting out of lockers, however they don't seem to work, I have been told the sound manager is getting a rewrite, closets have had a emag sound for a very long time but from my testing I don't think it ever worked the way it should.

I understand this PR touches alot of places, outside of ClosetControl I have tried to be as sparten as possible with my changes, feel free to raise up any issues!

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Objects/ClosetsAndCrates/ClosetBase.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Objects/ClosetsAndCrates/ClosetBase.prefab
@@ -46,6 +46,7 @@ SpriteRenderer:
   m_LightProbeUsage: 0
   m_ReflectionProbeUsage: 0
   m_RayTracingMode: 0
+  m_RayTraceProcedural: 0
   m_RenderingLayerMask: 4294967295
   m_RendererPriority: 0
   m_Materials:
@@ -127,6 +128,7 @@ SpriteRenderer:
   m_LightProbeUsage: 1
   m_ReflectionProbeUsage: 1
   m_RayTracingMode: 0
+  m_RayTraceProcedural: 0
   m_RenderingLayerMask: 1
   m_RendererPriority: 0
   m_Materials:
@@ -240,6 +242,7 @@ SpriteRenderer:
   m_LightProbeUsage: 0
   m_ReflectionProbeUsage: 0
   m_RayTracingMode: 0
+  m_RayTraceProcedural: 0
   m_RenderingLayerMask: 4294967295
   m_RendererPriority: 0
   m_Materials:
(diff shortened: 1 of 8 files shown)


corp-0:
You seem to have one conflicting file which is odd because I remember you were the last person to modify that script.

jamessimo:
Yes it is odd. I will rebase and see what’s up.

jamessimo:
Seems a rebase on my side fixed it up.

PR #5240, opened 2020-10-01
Turn headsets and stock parts into variants. Add placeholder encryption keys.

PetMudstone:
### Purpose
- Turns headsets into proper variants, with the station-bounced radio as their parent.
- Turns all stock parts into proper variants, adds a few additional stock parts.
  - Adds proper wattages to new power cell variants.
- Updates a large amount of references to use newer items.


Bod9001:
"Turns headsets into proper variants, with the station-bounced radio as their parent."

How did you implement the clothing section of the headsets if you inherit from the station-bounced radio?

PetMudstone:
The headsets are set to occupy the ear slot instead of the belt slot. They also have the Clothing V2 script attached to them so that they render on characters.

Bod9001:
wouldn't it be more maintainable to have the headsets be derived from the clothing prefabs then the handheld radio?

PetMudstone:
If the headsets are derived from the radio, then the headsets will derive all functionality that the handheld radio should eventually have, as in /tg/. I'll try to make an issue about general radio functionality soon.

corp-0:
is all functionality from the radios needed in headsets? If not then it is better to make them a variant of the base clothing object, because of the [interface segregation principle.](https://en.wikipedia.org/wiki/Interface_segregation_principle)

PetMudstone:
Headsets in /tg/ have all functionality that radios do, from toggling receiving and transmitting modes to the ability to switch the frequency.

corp-0:
alrighty then, merging

PR #5241, opened 2020-10-02
Adds the first record in our changelog history

corp-0:
I wrote an action so PRs can declare changes to be mentioned in a CHANGELOG.md file which is generated and maintained automatically. Read more about it [here](https://github.com/corp-0/pr2changelog). 

Not every PR needs to declare changes, only add changes to the changelog if you think they are worth mentioning. The way to do it is quite easy, just do like this:

CL:First record in our changelog history 🎊 

Every line that starts with CL: in the PR's body is a new record. One PR can have more than one record. The changelog file is automatically ordered so the most recent record (among PRs) is first.


corp-0:
imma go ahead and do a gamer move so I can see if the thing worked

PR #5295, opened 2020-10-12
Fixes profile command messages

ThatDan123:
Fixes the command messages for profiling.


Bod9001:
hey, do you know if this turns on the extra logging in the update manager? when you send the logging message?

ThatDan123:
dont think it does, but I can make it

ThatDan123:
humm, but I'd have to remove the #if UNITY_EDITOR, might have a performance impact?

Bod9001:
well, extra functionality so should be fine

Issue #5196, opened 2020-09-24
The ability to set antagonist parameters in a config file.


wraith-54321:
These parameters include: Antagonist spawn rates, meaning ratio to crew and chance of each type of antagonist, such as wizards with a chance of 80 and nuke ops with a chance of 20, meaning wizards spawn 80% of the time and nuke op spawn 20% of the time.(might change this to just be a point system meaning you can have 2 antagonists with 80 and an antagonist with 50 and the 80s are more likely to spawn by 30[I hope that makes sense]) Antagonist amounts, meaning you could have more then one blob or xeno queen and such, you can also change whether or not they are hostile towards each other, you could also in theory have something like a clockwork and blood cult.(these would most likely always be hostile towards each other) adding this might also make testing new antagonists(custom or not) easier.

PR #5342, opened 2020-10-19
Gamemode max antag

ThatDan123:
Adds max antag field to gamemode SO

-Sets max antags for blob to 1

### DIFF
--- a/UnityProject/Assets/Resources/ScriptableObjects/GameModes/Blob.asset
+++ b/UnityProject/Assets/Resources/ScriptableObjects/GameModes/Blob.asset
@@ -16,13 +16,14 @@ MonoBehaviour:
   description: Something unusual is growing inside a crew member. The station must
     survive.
   canRespawn: 0
-  minPlayers: 1
-  antagRatio: 0
+  minPlayers: 20
+  antagRatio: 1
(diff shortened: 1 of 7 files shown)


Bod9001:
could it be adjusted later on to be scalable with population?

ThatDan123:
Already is, thats what the antag ratio variable is, this new one is literally the max it will go.

Bod9001:
2017 TG allowed multiple blobs,  lare you limiting it for it being too overpowered
and
can admins spawn more than the max?

ThatDan123:
Yeah, this doesnt affect admins. Just on round start (or if the gamemode has mid round antags on which blob doesnt atm)

What ratio would you suggest if we want more than one?

Like 1 for every 30 players? (not that we'd get to 60 players atm) 

Or if you wanted to make it hard for the crew i could add so that the blob gains more resources after a certain player limit is hit.

Bod9001:
humm, I think we can tweak it later on

wraith-54321:
[https://github.com/unitystation/unitystation/issues/5196](url)
this is an issue I opened about this

corp-0:
it is related, but this PR doesn't solve your issue just yet.

PR #5344, opened 2020-10-19
Removes unneeded tile Collider and Allows you tto walk across multiple tables

Bod9001:
### Changelog
CL:  you can walk across multiple tables now
CL:  removes unneeded tile map Collider from outpost

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Layer.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Layer.cs
@@ -154,17 +154,18 @@ public class Layer : MonoBehaviour
 		List<TileType> excludeTiles = null)
 	{
 		// There's not tile here so its passable.
-		if (!tilemap.HasTile(to))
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Layer.cs
```diff
@@ -154,17 +154,18 @@ private void OnRotate(MatrixRotationInfo info)
 		List<TileType> excludeTiles = null)
 	{
 		// There's not tile here so its passable.
-		if (!tilemap.HasTile(to))
-			return true;
+		// if (!tilemap.HasTile(to))
```
ThatDan123: This all supposed to be commented out?

Bod9001: yeah, handled in meta-tile map tho still needed for the objects layer since the object layer is special for some reason, actually I can improve it

PR #5432, opened 2020-10-28
Adds WizardAsteriod Scene

Akbadain29:
### Purpose
Adds a Wizard spawn scene on an asteriod, so they'll be an option to spawn in on a ship or on a floating rock.

### Notes:
CL: Adds Wizard Asteroid Spawn Scene.


Bod9001:
WizardSpawnPoint (UnityEngine.Transform) is at 0,0 Please update the prefab/update the map/revert

ThatDan123:
If the scene is at 0,0 move it far away, thx

NoooneyDude:
If something is at 0,0 local to the station and you want it there, you can just move it by 0.2 units to shut the alarm up - but yeah, make sure the root scene is not at 0,0 - something in the ~k range would be ideal - see the other additional scenes.

Akbadain29:
That should be all good now.

PR #5438, opened 2020-10-30
Removed WizardShip from rotation and corrected spelling

Akbadain29:
### Purpose
Takes WizardShip out of rotation until it is completed and corrected the scene's name because it was spelt wrong. Also puts in some missing tables in TestStation that were removed when I added in the tables layer and forgot to put them back in.


Bod9001:
you've replaced it with Wizard asteroid?

Akbadain29:
Yes. I changed the additional scene list and changed asteriod to asteroid in the file name. The WizardShip scene is still there, I duplicated it and modified it into WizardAsteroid,

PR #5439, opened 2020-10-30
Improve tilemap floodfill

RedlineTriad:
### Purpose
This reduces startup time by checking if the tile is already assigned a room number before using flood fill to find room boundaries.

Before:
roomnumbers_before
After:
roomnumbers_after

### Notes:
I tested this and both before and after have correct room numbers and atmos and space wind were both working.

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/MetaDataSystem.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/MetaDataSystem.cs
@@ -105,10 +105,13 @@ public class MetaDataSystem : SubsystemBehaviour
 	{
 		BoundsInt bounds = metaTileMap.GetBounds();
 
+		var watch = new Stopwatch();
+		watch.Start();
 		foreach (Vector3Int position in bounds.allPositionsWithin)
 		{
 			FindRoomAt(position);
 		}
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/MetaDataSystem.cs
```diff
@@ -105,10 +105,13 @@ private void LocateRooms()
 	{
 		BoundsInt bounds = metaTileMap.GetBounds();
 
+		var watch = new Stopwatch();
+		watch.Start();
```
Bod9001, line 109: is it a bad thing that this is never stopped?

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/MetaDataSystem.cs
```diff
@@ -105,10 +105,13 @@ private void LocateRooms()
...
 		foreach (Vector3Int position in bounds.allPositionsWithin)
 		{
 			FindRoomAt(position);
 		}
+		Logger.LogFormat("Created rooms in {0}ms", Category.TileMaps, watch.ElapsedMilliseconds);
```
Bod9001, line 114: and add category if you want to keep it in

PR #5440, opened 2020-10-31
Fix paramedic access

### LINKED ISSUE #5393: paramedic has got engineer access
self-explanatory

craftxbox:
### Purpose
Removes accesses that the paramedic should not have from them.
Closes #5393


craftxbox:
For context, this pr removes the following accesses:
Engineering
Cargo
Mining
Rd (?)
Hydroponics

wraith-54321:
they should have cargo, its that was on TG as far as I know

craftxbox:
should they really have cargo access?
they're in the medical department, i think cargonia being intruded by the paramedic would be quite annoying.

PetMudstone:
Paramedics have general/lobby access to every other departments in /tg/. However, I am fine with this PR. Whether or not paramedics should have lobby access to other departments should be determined by a vote, possibly after access gets reformed.

DooblyNoobly:
sounds good to me. Not hard to change if the community demands it

PR #5442, opened 2020-10-31
Stops hostile dragon behaviour on WizardAsteroid Scene

Akbadain29:
### Purpose
The dragon in the back room of the wizard asteroid spawn can kill you in 6 hits. This has been dealt with and now the dragon will not bother you at all.

### Notes:
I wanted to light the scene with lanterns, but no matter what I do, it just doesn't want to turn the light sources on until I have turned it on in my hand and put it back down again. Am I doing something obviously wrong?


Bod9001:
you can enable the light 2D on the lanterns If it is there

Akbadain29:
Yeah, that was it. Thanks,

PR #5449, opened 2020-11-01
Removes pushing of nonblocking objects and fixes pulling of blocking objects

### LINKED ISSUE #5385: Moving into objects makes you push the items that occupy the same tile as them.
## Description
Title. Moving into a object (including tables) that can't be passed over makes you push the items that are on said object. More specifically, the items that you can pick up and pass over without issue such as drinking glasses. 
Notably, if you move into a movable object (such as a canister) that occupies the same tile as an item (such as a bullet casing), you will push the item first when you try to move into the movable object and not the movable object itself. After the item has been pushed then you can actually move into the object to push it.

push glass

### Steps to Reproduce
1.  Place an item, such a drinking glass, onto a table.
2.  Move into the table.
3.  Watch as the item on the table gets pushed away.
OR
1. Drop an item onto a tile, place a movable object such a canister onto it
2. Move into the object.
3. Watch as you push the item first. Move into the object again and push said object for real.

Ruzihm:
Fixes #5385
Allows pulling of blocking pullable objects (e.g., canisters)
Some code styling

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
@@ -92,7 +92,7 @@ public partial class MatrixManager : MonoBehaviour
 	void OnSceneChange(Scene oldScene, Scene newScene)
 	{
 		ResetMatrixManager();
-		if (!newScene.name.Equals("Lobby"))
+		if (newScene.name.Equals("Lobby") == false)
 		{
 			IsInitialized = false;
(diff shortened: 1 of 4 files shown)


corp-0:
added the label. Why is this DNM?

Ruzihm:
> added the label. Why is this DNM?

@corp-0 Found a regression of not being able to push from inside a tile. fixed it and added a TODO

corp-0:
alright so this is not DNM anymore?

Ruzihm:
That's right!

Bod9001:
looks good

PR #5459, opened 2020-11-01
Improve issue templates

Fogapod:
This commit does multiple things:
- moves templates from docs/ into .github/ folder
- splits issue template into bug report and feature request templates
- adds comments to templates instead of plain text to avoid invalid reports

New issue templates will automatically add labels (Type: Bug and Type: Feature).
There is no need to update repo settings, new templates will be applied automatically.
New issue types can be added easily if needed by using existing ones as example.

Overall these changes should make issue reporting a little bit smoother for both devs and users.


Bod9001:
hey, just saying it puts the Feature tag for the bug and the bug tag for the Feature

Fogapod:
Oh, my bad, sorry

PR #5486, opened 2020-11-04
Adds click move for pulling

### LINKED ISSUE #4659: Click Move For Pulling
In old ss13 while pulling an object you can click on a tile around yourself and move the item to that locations, this is useful to stop getting yourself stuck and to move objects where you want them.

We currently lack this feature.

It would also be helpful for moving objects onto teleporters.

Ruzihm:
Adds click move for pulling
Refactors some interaction logic
Makes true the default for reachability, makes directional windows and magic walls not reachable. <- might require a reimport of prefabs to appear.

Closes #4659

### DIFF
--- a/UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
+++ b/UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
@@ -190,7 +190,8 @@ public partial class PlayerNetworkActions : NetworkBehaviour
 	[Command]
 	public void CmdSlideItem(Vector3Int destination)
 	{
-		if (playerScript.IsInReach(destination, true) == false)
+		if (playerScript.IsInReach(destination, true) == false
+			|| playerScript.pushPull.PulledObjectServer == null)
 		{
 			return;
 		}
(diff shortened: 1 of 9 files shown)


Ruzihm:
Would appreciate a close look at if I handled networking correctly.

UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
```diff
@@ -187,6 +187,19 @@ public void CmdResist()
...
 	}
 
+	[Command]
+	public void CmdSlideItem(Vector3Int destination)
+	{
+		if (playerScript.IsInReach(destination, true) == false)
+		{
+			return;
+		}
+		PushPull pushPull = playerScript.pushPull.PulledObjectServer;
```
ThatDan123, line 198: Think you might need a null check here for this variable as PulledObjectServer could be null.

ThatDan123:
Networking looks fine

Ruzihm:
@Bod9001 Can't check discord at the moment but if there is still trouble
getting the new default touchable=true in build, I was wondering if
touching the .meta files for all the prefabs would help:
https://forum.unity.com/threads/assets-not-being-reimported-in-batchmode.937991/

Ruzihm:
reachable=true I mean

PR #5489, opened 2020-11-04
replace bvold guns on scenes and fix gun logging

AutumnNova:
title


AutumnNova:
thats all of em, box and outpost may still have bvold weaponry, but ive not touched them in case someone is woking on them
note: all awaysites that contained syndicate weaponry have  the unrestricted version of that gun (non trigger lock version)

PR #5490, opened 2020-11-04
.gitignore cleanup

Fogapod:
### Purpose
.gitignore is a little bit messy and needs some love.

### Notes
This commit does not add or remove any entries into .gitignore

This commit includes the following changes:
- removed duplicates
- grouped similar entries together
- removed files that should be ignored, but are still in repo
- removed leading slash for UnityProject entries for consistency
- added python section as there are several python scripts

I removed all `/Assets`  references because I wasn't able to find this top level folder since 2017 in history, the only `Assets` folder is in `UnityProject` folder.
Please check before merging, hence draft PR.

Some entries are probably outdated, but cleaning these up is outside of the scope of this PR.


Fogapod:
image
Failing test tries to delete version file that was deleted in this PR, because it was in .gitignore originally.

Should this step be removed from workflow?
https://github.com/unitystation/unitystation/blob/8020b5d53fb2d7fd947b3313e36d547e807dd9d9/.github/workflows/test.yml#L100-L102

Bod9001:
yeah ok

Fogapod:
Done. Please check the .gitignore on really dirty working directory and this is ready.

.gitignore
```diff
@@ -1,40 +1,44 @@
...
-/UnityProject/[Bb]uilds/
-/UnityProject/[Ll]ogs/
-/UnityProject/Assets/AssetStoreTools*
-/UnityProject/Assets/StreamingAssets/chatLogs*
-/UnityProject/Assets/InitTestScene*
-/UnityProject/Assets/MHLab*
-/UnityProject/PATCH*
-/UnityProject/.idea/
-/UnityProject/.vs/
-.vs
```
Bod9001, line 16: isn't this the visual studio stuff?

Fogapod: These were duplicates and should be covered by "IDE" section

Bod9001: ok cool

Fogapod:
If something breaks, tell me

PR #5493, opened 2020-11-04
More transformations for the null rod.

PetMudstone:
### Purpose
- Adds quite a few transformations for the null rod, greatly expanding the null rod window (seriously it's big someone might want to check it out).
  - The transformations don't have particularly unique properties yet, since I don't know how to code.

image

### Changelog:
CL: Added a couple more transformations for the null rod.


corp-0:
so you spawn the objects in the world and they do nothing or can they do something?

PetMudstone:
You can kill people with the transformations. They mostly share the same properties however.

corp-0:
Alright, yeah we can balance them later.

PR #5494, opened 2020-11-04
fix centcom atmos issues

### LINKED ISSUE #5331: Buldinn's CentCom (used for Outpost and most other maps) has no atmosphere, for some reason.
## Description
Title. For some reason Buldinn's CentCom has no atmos in most of its area despite seeming like it should be sealed. This means that anyone (especially CentCom officials) that go on board experience awful deaths if they don't have proper equipment. There are actually pockets of the station that have atmosphere, but they're completely detached from everything else.

### Steps to Reproduce
1.  Go onto Outpost.
2.  Respawn yourself immediately as a CentCom official.
3.  Spawn, immediately start dying due to a lack of atmosphere.

AutumnNova:
magic

fixes #5331

### DIFF
--- a/UnityProject/Assets/Scenes/AdditionalScenes/CentCom.unity
+++ b/UnityProject/Assets/Scenes/AdditionalScenes/CentCom.unity
@@ -149,7 +149,7 @@ PrefabInstance:
     - target: {fileID: 7373071894442451176, guid: e821287a1c78eafcd801c8a378830506,
         type: 3}
       propertyPath: m_LocalPosition.z
-      value: 1.1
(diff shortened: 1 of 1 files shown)


PetMudstone:
For clarification all the doors and other objects were on different Z axes, which caused problems. Now that everything isn't on separate planes of existence CentCom atmos should be less bad.

PetMudstone:
Oh yeah, this should close #5331.

corp-0:
thanks for the clarification, magic was kind of lacking as an explanation of what happened here

PR #5495, opened 2020-11-04
Some canister code cleanup

Aranclanos:
### Purpose
Replaces the custom gas merge code in GUI_Canister with a gasmix call.
Cleans up a bit canister code.
The connected tanks will now merge and update the pressure level in the UI at the same time than the canister air change. See gif
canister tank sync


corp-0:
what's the difference between ``tankValveOpen`` and ``ValveIsOpen`` ? Does the former need to be public?

Aranclanos:
one is for the connection to the tank (like an emergency oxygen tank) and the other is for the connection to environmental air or pipe connectors

PR #5509, opened 2020-11-05
Wizard Shuttle

PerfectTangent:
### Purpose
Addes a basic, large, mostly empty wizard shuttle in place of the literal 5x5 square box that existed before hand. 

### Notes:
The interior of the wizard's shuttle could use additional work, but it mostly has to wait until new wizard features are added, such as the looking glass, books, or apprenticing.


PerfectTangent:
there's currently a couple problems with the scene itself at the moment. @ThatDan123 is taking a look, but I welcome others to as well.

ThatDan123:
See issue #5514

ThatDan123:
This PR is mostly fine expect it should only have the scene not any other file in the PR, needs cleaning up. (the build settings file especially)

The issues with the populators and lockers is a general bug not this PR scene related

Just one thing is the tile colliders on the matrix are no longer needed on the shuttle, so they need to be removed.

PerfectTangent:
closing in favour of #5525

PR #5518, opened 2020-11-05
Fixes invisible grenades

### LINKED ISSUE #5498: Grenades don't appear when spawned in by admins and minibombs aren't created by the uplink while connected as a client.
### Environment
- Game Version: Staging Build 5491
- OS Version: Windows 10

For some reason, grenades and minibombs spawned in by admins don't seem to exist. Minibombs also don't appear when you buy them with the uplink. Right clicking on tiles where they should be after they get placed done brings up nothing. On the other hand, combustible lemons don't render for clients when spawned in by admins but using right click on tiles where they are shows that they indeed do exist. Lemons can also be picked up and also primed for detonation.

This only happens if you are connected to a server as a client. As the host (including in the editor), you will see everything as you normally would.

### Steps to reproduce.
1. As an admin connected as a client, search up grenades, explosive lemons, and minibombs in dev spawner.
2. Select and spawn them, notice that they don't seem to appear.
3. Right click on their would-be locations where you clicked. Notice that grenades and minibombs just don't exist and that the lemons exist but are invisible.

Aranclanos:
### Purpose
Fixes grenades not existing on clients.
Fixes grown foods having a sprite scale of 0,0 making them invisible if they weren't manually harvested.
Fixes #5498

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Items/Weapons/Explosive/Grenade.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Items/Weapons/Explosive/Grenade.prefab
@@ -80,11 +80,6 @@ PrefabInstance:
       propertyPath: m_AssetId
       value: 
       objectReference: {fileID: 0}
-    - target: {fileID: 114738972662350094, guid: a7abba22164ce49f0bf9b9b219f39298,
-        type: 3}
-      propertyPath: serverOnly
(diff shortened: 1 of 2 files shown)


PetMudstone:
How did you fix grenades? I only see a .prefab change.

Aranclanos:
image

PR #5523, opened 2020-11-05
Disableds collision on particle effects

Bod9001:
a big chunk of CPU was being used by physics 2D particle effects apparently, so disabled physics on the ones I could find


PetMudstone:
You should check "SprayBottle.prefab" as well. It's the parent of various objects that spray stuff.

Bod9001:
ok cool

PR #5524, opened 2020-11-05
Added antags sounds for round start

Macoron:
### Purpose
Added antags beginning sound for traitor, blob and ops form tg-station
_Whiskey, alpha, whiskey..._

### Notes:
I had some weird bug with message serialization. For some reason PlaySound bool wasn't send to clients. I tried almost everything, but ended writing custom serialization.

CL: Added antags sounds for round start


NoooneyDude:
Would be great if we could get that bug fixed. I have a custom spawn sound for wizards, too. Well, nothing special.

Edit: oh I see, you did end up finding a solution. Nice!

PR #5532, opened 2020-11-06
Revert removal of ProjectVersion.txt

### LINKED ISSUE #5471: Name check can be bypassed by point/disarm and probably other actions
## Environment
- Game Version: Build 5440
- OS Version: 

Pointing at or trying to disarm lets you reveal the original name of person. This should not be the case if they have fake ID equipped.

### Steps to reproduce

1. Equip ID with a different name or PDA (to appear as Unknown)
2. Point at yourself (reveals original name)
3. Ask someone to disarm you (reveals original name)

Messages like knocking on windows, attacking etc leak your real name to others as well.

NoooneyDude:
- Reverts the removal of ProjectVersion.txt.
- Fixes NRE on PushPull parentContainer setting when PushPull is null.
- Changes the `ExpensiveName()` extension for player GameObjects, favouring the visible name instead. Fixes #5471.
    - If this has not fixed every instance, please create a new issue with a list of the instances.
    - Converts several instances where this is not desirable into more explicit name calls.

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Antagonists/Antags/Fugitive.cs
+++ b/UnityProject/Assets/Scripts/Systems/Antagonists/Antags/Fugitive.cs
@@ -53,7 +53,7 @@ namespace Antagonists
 				" I need to move in the shadows and keep out of sight," +
 				" I'm not going back.");
 
-			_ = StationWarning(newPlayer.ExpensiveName());
+			_ = StationWarning(newPlayer.Player().Name);
 
 			return newPlayer;
 		}
--- a/UnityProject/Assets/Scripts/Util/SweetExtensions.cs
+++ b/UnityProject/Assets/Scripts/Util/SweetExtensions.cs
@@ -61,9 +61,9 @@ public static class SweetExtensions
 		}
 
 		var player = go.Player();
-		if (player != null && !String.IsNullOrWhiteSpace(player.Name))
+		if (player != null && !String.IsNullOrWhiteSpace(player.Script.visibleName))
 		{
-			return player.Name;
+			return player.Script.visibleName;
 		}
 
 		return go.name.Replace("NPC_", "").Replace("_", " ").Replace("(Clone)","");
--- a/.gitignore
+++ b/.gitignore
@@ -77,7 +77,6 @@ UnityProject/ProjectSettings/UnityAdsSettings.asset
 .vs/
 .idea/
 workspace.xml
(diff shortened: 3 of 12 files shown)


UnityProject/Assets/Scripts/Systems/Antagonists/Antags/Fugitive.cs
```diff
@@ -53,7 +53,7 @@ public override GameObject ServerSpawn(PlayerSpawnRequest spawnRequest)
 				" I need to move in the shadows and keep out of sight," +
 				" I'm not going back.");
 
-			_ = StationWarning(newPlayer.ExpensiveName());
+			_ = StationWarning(newPlayer.Player().Name);
```
corp-0, line 56: this one was supposed to be ic name

UnityProject/Assets/Scripts/Util/SweetExtensions.cs
```diff
@@ -61,9 +61,9 @@ public static string ExpensiveName(this GameObject go)
 		}
 
 		var player = go.Player();
-		if (player != null && !String.IsNullOrWhiteSpace(player.Name))
+		if (player != null && !String.IsNullOrWhiteSpace(player.Script.visibleName))
 		{
-			return player.Name;
+			return player.Script.visibleName;
```
corp-0, line 66: just wondering, what happens with spectator players? Do they have visibleName?

Fogapod:
@NoooneyDude ~~you need to add this step back for tests to work
https://github.com/unitystation/unitystation/pull/5490/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88L100~~

Wait, seems like tests fail is unrelated, so that step was useless from beginning?

UnityProject/Assets/Scripts/Systems/Antagonists/Antags/Fugitive.cs
NoooneyDude: So it's fine if it shows up as `Unknown`? Didn't seem right to me.

Edit: oh, this won't get the username, that's `Username`. `Name` should get the character's actual name, and not the apparent name.

UnityProject/Assets/Scripts/Util/SweetExtensions.cs
NoooneyDude: I'm not terribly sure. Looks like, in any case, it defaults to " " if the name hasn't been set.

PR #5538, opened 2020-11-06
Fixes the syndicate shuttle having their lockers empty.

### LINKED ISSUE #5514: Spawned in scenes have initialisation issues
Spawned in scenes like the nuclear operatives shuttle/base and wizard scene have these issues:

-Populators do not work for mapped items, eg lockers populators not working.

-Items mapped on top of lockers do not go into the lockers on spawn.

image

### Steps to reproduce

1. Respawn self as wizard/nuclear operative
2. See lockers have no contents
3. See items on top of lockers/crates which should be in them at spawn

Aranclanos:
### Purpose
Fixes the syndicate shuttle having their lockers empty.
The RoundStarted event will now be broadcasted after the gamemode subscenes load, making OnSpawnServer() being called on their objects.
Fixes #5514

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/GameManager.GameMode.cs
+++ b/UnityProject/Assets/Scripts/Managers/GameManager.GameMode.cs
@@ -44,7 +44,7 @@ public partial class GameManager
 	public RoundState CurrentRoundState
 	{
 		get => currentRoundState;
-		private set
+		set
 		{
 			currentRoundState = value;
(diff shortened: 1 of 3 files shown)


ThatDan123:
Not sure if this will fix scenes which are loaded after round start though? Eg when respawning a syndie in a non syndie round

Aranclanos:
It won't

Aranclanos:
(follow up pr later)

Issue #1981, opened 2019-06-30
Pushing isn't as smooth as pulling


Jibbajabbafic:
# Description
Pushing an object should be as smooth as pulling it, which is the current behaviour in /tg/:
### TG pulling + pushing:
![gjdYWs4CE9](https://user-images.githubusercontent.com/4493127/60395331-ce61ef80-9b29-11e9-9abf-9405606588fa.gif)

### Our pulling + pushing:
![bSHRDQYVox](https://user-images.githubusercontent.com/4493127/60395377-492b0a80-9b2a-11e9-91f6-4c3f2529087a.gif)

## Note
The tile by tile behaviour would make sense for really heavy objects, but then the pulling behaviour should also be the same.

You should be able to push/pull something like a janitor cart around smoothly though, so perhaps it could depend on it's size or a new flag.

## Category
TODO

## Steps to Reproduce
Please enter the steps to reproduce the bug or behaviour:
1. Pull object
2. Notice it's smooth
3. Push object
4. Notice it's done tile by tile

fomalsd:
Current behaviour is actually intended, see discussion in discord
https://discordapp.com/channels/273774715741667329/312454684021620736/484743463577976835
Allows to i.e. push things into space without following them

Jibbajabbafic:
I realise it's intentional, but after playing some /tg/ the smooth pushing feels a lot nicer and more responsive IMO

fomalsd:
how about making it only act smoothly when pushing something off in pull mode?

PerfectTangent:
fixed by #8335

PR #5545, opened 2020-11-07
Stops pull when you push an object into blocked tile

### LINKED ISSUE #2214: Pushing containers into walls does not stop pulling
# Description
Grabbing a container and pushing it into a wall should stop the user from pulling it.

## Category
TODO

## Steps to Reproduce
1. Grab a crate or any container
2. Start pulling
3. Try to push the crate into a wall
4. You will not stop pulling the crate

Ruzihm:
Pushing an object you are pulling now makes you stop pulling it, even if it is pushed into a wall or other object.

Removes some redundant pushable checks.
Makes it consistent that only one object at most is told to be pushed (does not break buckled pushing)
Closes #2214

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
@@ -734,30 +734,19 @@ public partial class MatrixManager : MonoBehaviour
 				}
 			}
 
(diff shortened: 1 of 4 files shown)


Ruzihm:
Note: The way this is implemented is opposed to #1981 as it breaks pull state on *any* push on the pulled object.  But the logic for determining if the player should rather follow the pushed object would probably go where the edits in this PR are in the PlayerSyncs. It would also require making sure the player can follow.

PR #5546, opened 2020-11-07
guntrigger additions

AutumnNova:
add a clientside method
add the ability to modify if a non-clumsy player shoots themself when using a clumsy weapon (used for foamforce weapons)
's basically it


NoooneyDude:
Reverting this because it introduces a significant issue - guns do not work at all when on headless. This issue is not fixed by PR #5549.

AutumnNova:
's gonna make fixing the issues that my current pr fixes and this pr shoulda fixed a pain in the fuckin ass, but you do you

PR #5548, opened 2020-11-07
Fixes mid-round syndie & wizard subscenes initialization

Aranclanos:
### Purpose
Syndicate and wizard subscenes that load midround will now have their objects properly set up.

This means lockers will be populated and all other stuff that objects do at start

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.SceneList.cs
+++ b/UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.SceneList.cs
@@ -219,7 +219,7 @@ public partial class SubSceneManager
 	public IEnumerator LoadSyndicate()
 	{
 		if (SyndicateLoaded) yield break;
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.SceneList.cs
```diff
@@ -217,37 +219,47 @@ IEnumerator ServerLoadAwaySite(SubsceneLoadTimer loadTimer)
 	public IEnumerator LoadSyndicate()
 	{
 		if (SyndicateLoaded) yield break;
+		var pickedMap = additionalSceneList.defaultSyndicateScenes.PickRandom();;
```
corp-0: codacy is crying because the double ``;``

Aranclanos: how did that end up there

Issue #5160, opened 2020-09-18
First Implementation of the Wizard Antagonist B : 100


PerfectTangent:
## Description
Wizards of the Wizard Federation are a classic round antagonist who's signature feature is casting spells for fun and profit. They spawn inside a special wizard's shuttle and gain a spellbook with which they can memorize a variety of spells to assist in completing their objectives.

In /tg/, wizards have a few unique objectives, but the majority of them are exact copies of traitor objectives, with the exception of never having to escape on the evac shuttle (for obvious reasons). They have an allotment of 10 spell points to spend on spells from their spellbook. Some spells require the wizard to be wearing their wizard's robes to cast and others may be cast without the wizard's robes.

The intention is to make the wizard a playable antagonist in time for Spooktober this year (roughly a month from now). This does not mean that ALL things on this list must be completed by then, only enough to make the wizard _playable_ (and prefereably not broken or imba).

## Important Features:

- [x] A spell memorization system and spell points.

- [x] Wizard's robes and hat functionality.

- [x] An initial list of 10 spells to learn and cast.

- [x] The Wizard's ship mapped and added to scenes.

- [x] The teleportation scroll that the wizard uses to board the space station.

## Secondary Features that would be awesome to have:

- [x] Wizard Apprenticing - bestowing limited spellcasting to a willing apprentice. 

- [x] At least two Wizard Artifacts.

- [ ] A variety of wizard robes to choose from for customization purposes.

- [x] Spawnable and/or discoverable single-use spellbooks to grant magical power to non-wizards.

## A proposed list of spells that could be (more-or-less) easily implemented:

- [ ] Magic Missile 
- [x] Force Wall (already half way implemented with the mime)
- [x] Teleport
- [x] blink
- [ ] Repulse
- [x] Lightning Bolt
- [ ] Ethereal Jaunt
- [x] Lesser Summon Guns
- [x] Fireball
- [ ] Knock
- [ ] Smite _or_ Blind (Smite is a classic wizard spell, but it's also a bit lame and there would not be very many counters to it at the moment, so I suggest Blind as an alternative)

Bod9001:
added bounty thanks from Jack Trachea for the direct donation

Bod9001:
asked in DMs who'd be interested in doing it NoooneyDude, said they would be good for doing it

NoooneyDude:
### Development status
Ordered chronologically.

#### Stage One - PR #5274
[...]

PR #5561, opened 2020-11-08
Adds the wizard's lightning bolt spell, and the magic mirror

### LINKED ISSUE #4613: Change Paper Text box to TMP
Paper currently uses unity Text not TMP Text so has limited tag options.

https://docs.unity3d.com/Packages/com.unity.ugui@1.0/manual/StyledText.html

compared to:

http://digitalnativestudios.com/textmeshpro/docs/rich-text/

NoooneyDude:
- Adds the electrical arc effect. Uses [LightningBolt](https://assetstore.unity.com/packages/tools/particles-effects/lightning-bolt-effect-for-unity-59471) asset.
- Adds the lightning bolt spell. Continues #5160.
lightning bolt
- The wizard now spawns with a random name.
- Adds the magic mirror. Enables the wizard to change their name.
    - Includes wizard scene changes to reflect the addition.
- Paper can now fit in the ID slot.
- Wizards now spawn with an identity certificate instead of a station ID card.
- The Wizard 101 paper that spawns in the pocket is now populated.
- Paper item now uses TextMeshPro instead of default Unity Text. Fixes #4613

CL:Added lightning bolt spell and the magic mirror.


corp-0:
Codacy is complaining but it has a point. Unity uses IEnumerators in a hacky way to make code run "asynchronically". In this case we can ignore the complain.

Ruzihm:
@corp-0  Nothing is calling `StartCoroutine` on `MobSprite.StopElectrocutedOverlayAfter`, and nothing is yield returning with it so codacy is right. it doesn't need to return `IEnumerator` because is is never used as a coroutine.

Contrast with `PlayerSprites.StopElectrocutedOverlayAfter`, which does get called as a coroutine.

I'm guessing it is meant to be called with `StartCoroutine` in this case.

NoooneyDude:
Ah. Good catch! I'll commit a fix.

UnityProject/Assets/Scripts/NPC/MobSprite.cs
```diff
@@ -172,12 +173,50 @@ public void SetBurningOverlay()
...
+		public void EnableElectrocutedOverlay(float time = -1)
+		{
+			if (electrocutedOverlay == null)
+			{
+				AddOverlayGameObjects();
+			}
+
+			if (time > 0)
+			{
+				StopElectrocutedOverlayAfter(time);
```
Ruzihm:
```suggestion
				StartCoroutine(StopElectrocutedOverlayAfter(time));
```

Ruzihm (changes requested):

UnityProject/Assets/Scripts/NPC/MobSprite.cs
Ruzihm: Ah, so that's how to suggest changes. neat!

NoooneyDude: Cheers!

corp-0:
my bad. I thought it was complaining because the return value was never used, but it was actually the whole thing never being called. Hooray for codacy

UnityProject/Assets/Scripts/NPC/MobSprite.cs
```diff
@@ -172,12 +173,50 @@ public void SetBurningOverlay()
...
+		/// Disables the electrocuted overlay for the mob.
+		/// </summary>
+		public void DisableElectrocutedOverlay()
+		{
+			if (electrocutedOverlay == null)
+			{
+				return;
+			}
+
+			burningOverlay.StopOverlay();
```
NoooneyDude:
```suggestion
			electrocutedOverlay.StopOverlay();
```

PR #5549, opened 2020-11-07
fix clientside prediction issue that caused client to see non-existant projectiles

AutumnNova:
also fixed a thing that allowed you to overcharge batteries using the plasma refill script

PR #5562, opened 2020-11-08
Revert "guntrigger additions"

NoooneyDude:
Reverts unitystation/unitystation#5546

#5546 introduces an issue where no guns at all work while on headless. The issue is not fixed by #5549.


AutumnNova:
#5549 and #5546 use the same branch, if aran tested #5549 and found only issues with the pka and energyguns then i dont think this is particularly likely

PR #5564, opened 2020-11-08
Tool to disable additional scenes for temporary builds

Macoron:
### Purpose
Right now unitystation take a lot time to build. Mostly because we have about six stations and many subscenes (asteroids, away sites, etc). Full build can take up to 20+ minutes and this is really annoying for fast networking tests.

This PR adds new editor window to disable most of the scenes from build settings for fast debugging. It will drastically improve build time (from 20 to 5 minutes for me). I hope, compilation will be much faster with addressable, but it's a good temporary solution for now.

**Remember to discard any changes in the ProjectSettings and other config files. Don't commit them!**


NoooneyDude:
Awesome, this'll be really convenient. Thank you!

UnityProject/Assets/Scripts/Core/Editor/DisableNonEssentialScenes.cs
```diff
@@ -0,0 +1,215 @@
...
+/// <summary>
+/// This editor tool allows to disable some scenes from build
+/// It will drasticly improve build time
+/// </summary>
+public class DisableNonEssentialScenes : EditorWindow
+{
+	private static List<string> AdditionalScenesToRemove = new List<string>()
+	{
+		"Fallstation Centcom",
+		"Fallstation Syndicate"
```
NoooneyDude, line 17: ~~I think with `NaughtyAttributes` you can serialize refererences to scenes. Would that be a better solution?~~
It seems "WizardShip and "WizardAsteroid" additional scenes remain enabled. Is this correct?

NoooneyDude:
Tested, works fine. Have you considered adding an option to enable Quick Load as on the GameManager prefab? That way you don't need any asteroids, additional scenes or centcomm.

UnityProject/Assets/Scripts/Core/Editor/DisableNonEssentialScenes.cs
Macoron: Yeah, in ideal situation, one of them should be disabled. I just haven't done yet, because they need some tweaks in additional scenes scriptable object. Same with Fallstation Centcom-Syndicate.
It's relatively easy to add text field or NaughtyAttributes for additional scene to disable. Just haven't done it yet.

Bod9001:
all good to merge?

Macoron:
> Tested, works fine. Have you considered adding an option to enable Quick Load as on the GameManager prefab? That way you don't need any asteroids, additional scenes or centcomm.

Oh, I didn't know about this toggle, thanks. Yeah, it's good idea to add "super fast" mode with one station only as a separate toggle. May address it in next PR.
Though it would be better to move Quick Load toggle from GameManager prefab to json server config. Changing prefabs and scriptable objects always fill kinda dirty.

NoooneyDude:
> Oh, I didn't know about this toggle, thanks. Yeah, it's good idea to add "super fast" mode with one station only as a separate toggle. May address it in next PR.
> Though it would be better to move Quick Load toggle from GameManager prefab to json server config. Changing prefabs and scriptable objects always fill kinda dirty.

Ah, just FYI, Quick Load skips the lobby, so you get straight into the game.

PR #5568, opened 2020-11-09
Can use objects on walls from an extra tile away

### LINKED ISSUE #4720: The ability to use objects that occupy the same tile as a wall (buttons, switches, APCs) from an extra tile away.
## Description
In /tg/, you could use objects that occupied the same tile as a wall such as light switches, buttons, and APCs from an extra tile away due to a quirk of how they were mapped. This was a very useful quirk since those things often appeared across tables and what-not and the ability to use them from an extra tile away made them still convenient to use. However, you have to be directly adjacent to the wall said objects are on in Unitystation to use them which causes some issues.
image
For example in the above image, the switches for the blast doors and mass driver are on a wall. If this were /tg/, you could just click them from two tiles south or southeast across the table. However in Unitystation you have to be right next to said switches to use them you have to climb onto the table to open the blast doors and activate the mass driver, which proves annoying.

Ruzihm:
Adds the ability to use objects that occupy the same tile as a wall from an extra tile away.

Closes #4720

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Input System/InteractionV2/Validations.cs
+++ b/UnityProject/Assets/Scripts/Core/Input System/InteractionV2/Validations.cs
@@ -372,8 +372,13 @@ public static class Validations
 	/// <param name="targetVector">the delta vector representing how distant the interaction is occurring</param>
 	/// <param name="interactDist">the horizontal or vertical distance required for out-of-reach</param>
 	/// <returns>true if the x and y distance of interaction are less than interactDist</returns>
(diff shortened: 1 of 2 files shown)


PetMudstone:
Does this also apply to window tiles? Occasionally wallmounts can be found on window tiles.
Oh, and I'd suggest using "CL:" to add a changelog for future reference.

Ruzihm:
@PetMudstone Oh, neat. 

No, just walls. Should I include any other kinds of tiles other than walls or windows?

PetMudstone:
I think just windows and walls should be fine.

PR #5569, opened 2020-11-09
Removes some BVOld prefabs from BoxStation

Akbadain29:
### Purpose
Fixes up BoxStation's Armory so all the guns aren't BVOld, also switches all the chairs over to the new prefabs too. Also some lights.


corp-0:
that's a lot of reverts, lol. Investigate how to use interactive rebase to override your git log by squashing and renaming commits, its quite useful.

Merging :)

PR #5570, opened 2020-11-09
Makes objects on window tiles reachable from one more tile.

Ruzihm:
see #5568
Also includes some code styling & method renaming for making distinct from similar methods

CL: Can use objects on walls and windows from an extra tile away


Ruzihm:
@PetMudstone Is that the correct use of CL, btw?

PetMudstone:
Should be in regards to automatically adding the CL line to the changelog. However, you might want to state your changes in the past sense. For example, "Made it possible to to use objects on walls and windows from an extra tile away."

PR #5571, opened 2020-11-09
Photocopier mapping

PetMudstone:
### Purpose
- Adds more photocopiers to maps, replace photocopiers in maps.
- Photocopiers now have boards that can be made.
- Window sprite layer order moves. Windows will now render over standard objects. Relevant to directional windows.
- Detective's locker now comes with detective's holster.


Bod9001:
tests

PR #5572, opened 2020-11-09
Replaced all the BVOld Chairs and some of the other prefabs on SquareStation

Akbadain29:
### Purpose
Removes some BVOld Chairs and other prefabs on SquareStation and replaces them with their current iteration.


AutumnNova:
this is probably going to cause scene conficts with #5571

Akbadain29:
I'd better wait before trying to do the other ones then.

PR #5579, opened 2020-11-09
Fixes job-type chat spam on interactions

Aranclanos:
### Purpose
The player's ID job will no longer be spammed on most interactions towards listeners.
ID cards will now change their object name based on the information of the card instead of changing the player's name.
Cleaned up IDCard code, removed most initialization call spam.

### DIFF
--- a/UnityProject/Assets/Scripts/Items/IDCard.cs
+++ b/UnityProject/Assets/Scripts/Items/IDCard.cs
@@ -165,23 +165,20 @@ public class IDCard : NetworkBehaviour, IServerInventoryMove, IServerSpawn
 	public void SyncIDCardType(IDCardType oldCardType, IDCardType cardType)
 	{
 		idCardType = cardType;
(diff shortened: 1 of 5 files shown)


UnityProject/Assets/Scripts/Items/IDCard.cs
```diff
@@ -137,68 +126,46 @@ public void OnInventoryMoveServer(InventoryMove info)
...
+		itemAttributes.ServerSetArticleName(newName);
 	}
 
 	public void SyncIDCardType(IDCardType oldCardType, IDCardType cardType)
 	{
-		EnsureInit();
 		idCardType = cardType;
-		IDCardType cType = IDCardType;
-		switch (cType)
+		switch (idCardType)
```
corp-0: add the default case for safety. You can log an error and use the standard sprite if the code ever got that point.

Aranclanos:
done

PR #5600, opened 2020-11-12
add more gasses into filters

AutumnNova:
also comes with a slightly improved pipe dispenser color system because yes


corp-0:
you had chance to build and test? I can't do it on this laptop.

PR #5601, opened 2020-11-12
just some fallstation fixes, nothing too big

AutumnNova:
fixes some incorrectly connected light tubes and adds a few more light tubes in rather dark areas


corp-0:
```
FallStation: "BVOld_BVOld_BVOld_BVOld_BVOld_BVOld_BVOld_BVOld_BVOld_BVOld_BVOld_APC BOTANY" has a null value in the list.
```

only one problem, might be something you or connected to another apc

PR #5604, opened 2020-11-12
Fixes speech inside a locker reaching all players.

Aranclanos:
### Purpose
Fixes speech and other messages appearing if the originator is inside a locker.
Fixes players inside lockers being unable to hear or see anything around them.
Changes most chat display messages calls to use a simpler version, just sending the message and the originator object.


Aranclanos:
solved conflicts

corp-0:
this will probably make #5593 conflict but I think the changes are easy enough to solve on the other pr

PR #5614, opened 2020-11-14
implement a system that makes it possible to set matrixes as not console operable

AutumnNova:
maybe

### DIFF
--- a/UnityProject/Assets/Scripts/Shuttles/MatrixMove.cs
+++ b/UnityProject/Assets/Scripts/Shuttles/MatrixMove.cs
@@ -177,7 +177,7 @@ public class MatrixMove : ManagedNetworkBehaviour
 	/// </summary>
 	public bool Initialized => clientStarted && receivedInitialState;
 
-	[FormerlySerializedAs(NoConsole),Tooltip("Disable the ability for players to use a shuttleconsole to control this matrix")]
+	[FormerlySerializedAs("NoConsole"),Tooltip("Disable the ability for players to use a shuttleconsole to control this matrix")]
 	public bool IsNotPilotable = false;
 
 	public override void OnStartClient()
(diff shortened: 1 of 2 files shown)


corp-0:
do all matrixes have the MatrixMove component? Wouldn't not adding the component be enough to mark a matrix as not piloteable?

AutumnNova:
as far as i know things like the cargo shuttle and ems need that component to function

corp-0:
ah, that makes sense

UnityProject/Assets/Scripts/Shuttles/ShuttleConsole.cs
```diff
@@ -72,7 +72,11 @@ private IEnumerator InitMatrixMove()
 					Category.Matrix);
 			}
 
-			if (ShuttleMatrixMove != null)
+			if (ShuttleMatrixMove.NoConsole == true)
```
corp-0:
```suggestion
			if (ShuttleMatrixMove.NoConsole)
```
When checking a bool, you don't really need to compare it to true.

UnityProject/Assets/Scripts/Shuttles/MatrixMove.cs
```diff
@@ -177,6 +177,9 @@ public class MatrixMove : ManagedNetworkBehaviour
 	/// </summary>
 	public bool Initialized => clientStarted && receivedInitialState;
 
+	[Tooltip("Disable the ability for players to use a shuttleconsole to control this matrix")]
+	public bool NoConsole = false;
```
corp-0: Although the tooltip is great, maybe this name is not the best to understand what it does at first glance. Change it to something like ``IsPilotable``. You would need to invert the logic check if you choose that name.

PR #5618, opened 2020-11-14
Remove redundant newlines in discord webhook messages

Fogapod:
You don't actually need all these.
\`\`\`\nTEXT\`\`\` will render fine without newlines.

Discord automatically ads newline after codeblock closes, so you don't need \n when merging messages either. Currently 2! redundant newlines are placed: 
image
(one from the end of codeblock and other from joining messages)

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Chat/DiscordWebhookMessage.cs
+++ b/UnityProject/Assets/Scripts/Core/Chat/DiscordWebhookMessage.cs
@@ -181,7 +181,7 @@ namespace DiscordWebhook
 					break;
 				}
 
-				msg += queue.Dequeue() + "\n";
+				msg += queue.Dequeue();
 			}
 
 			var payLoad = new JsonPayloadContent()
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Core/Chat/DiscordWebhookMessage.cs
```diff
@@ -181,7 +181,7 @@ private void FormatAndSendMessage(string url, Queue<string> queue)
 					break;
 				}
 
-				msg += queue.Dequeue() + "\n";
```
ThatDan123: Not all messages are in code blocks, so this will need to be kept

Fogapod: Hmm, this will mean that codeblock messages will still have annoying newline then.
The solution would be check if previous symbol is ` ?

corp-0: how is it annoying when it makes no difference?

Fogapod: image
compare this to screenshot in PR message

corp-0: ah, I see. Was hard to find it at first glance

Fogapod:
I have no idea how to deal with the newline when joining messages because for some reason these low level functions are used directly from many parts of codebase (directly setting webhook url instead of some abstraction like MessageOrigin).

At least 1 newline of 2 is eliminated.

PR #5620, opened 2020-11-15
Bvold removal button

Aranclanos:
### Purpose
Lots of work here!
This simple bool on the BVOld checker component will run on editor mode and loop through all bvold objects to replace them with their assigned replacement -they must be set manually-
While some replacement links remain, this already works with the centcom station, with the exception of the flamethrower that has no replacement.
Some non-intended replacements will ocurr, for example, the plaque in centcomm is actually a cigarette butt with a different sprite and description.

More to come on follow up PRs, with actual map changes


Aranclanos:
okay okay im done from here its just manual map work after pressing the button because some of the replacements prefabs need power while the bvold doesnt and I wont connect it to a random apc, areas would help for this case!
And once thats done too, maps need testing with machines that are connected to others, like shuttle connectors and dna scanners.

Another note, a few objects dont have a replacement because it just doesn't exist, like the flamethrower, one type of shotgun, the golden detective revolver and a couple more. Those can be manually changed later.

corp-0:
Which ones need power?

Aranclanos:
this pr doesnt include map changes, but after you make the swap, its firelocks, status displays and maybe some others

corp-0:
Just to make sure, it won’t replace doors with the new doors, right?

PR #5623, opened 2020-11-15
Xenoarchaeology v1 (Telepathic artifacts and artifacts detector)

Macoron:
artifacts

### Purpose
First iteration of xenoarchaeology profession. My implementation is inspired by Baystation12 and Goons forks. Check this wiki for [more details](https://wiki.baystation12.net/Xenoarcheology). 

Right now, artifacts randomly spawns on each asteroid. Each scientist has a new artifacts detector in their pocket and lockers. When activated it prints user position and three nearest artifacts. For now excavation process isn't implemented, so artifacts just laying around.

Each artifact has activation trigger and artifacts effect. Scientist need to figure out trigger and how to use effect to help or harm station. Currently implemented triggers are:
- **Always active.** Artifact just invokes effect each ten seconds
- **Touch activation.**  Player need to touch artifact in help intent to activate it.

Right now I implemented only one effect: **telepathic messages**. This artifact broadcast different "thoughts" to their victims. They can be positive or...

### Notes:
I used Baystation12 assets to add xenoarchaeology artifacts sprites. They use same license as goon and tg. I placed all their assets in separate and add readme + license. Should be enough, but if someone knows well all this legal stuff, it would better to check this.


Macoron:
Well, after merging artifacts detector examine message behave weird. They print reports to character bubbles. That's not intended behaviour for Chat.AddExamineMsgFromServer, so I will add issue for that.
image

Bod9001:
is ok, current issue with the UI Not an issue of yours

PR #5632, opened 2020-11-16
Some More Artifacts

ThatDan123:
-Adds support for item consumption

-Adds artifact which spawns mobs

-Adds artifact which spawns items

-Teleport artifact

-Adds artifact which transmutates metal into other materials 

CL: Adds some more artifacts


ThatDan123:
~Have a weird issue where the new artifacts dont appear in the admin spawn menu, not sure why.~ nvm works fine now

PR #5642, opened 2020-11-18
Removes bvold objects from fallstation and its additional scenes

Aranclanos:
### Purpose
Removes bvold objects from fallstation and its additional scenes


corp-0:
looks like this is good to go, I'm merging

PR #5648, opened 2020-11-19
New Radial Menu

### LINKED ISSUE #5654: SubsceneTests reading all scenes in the project can result in undesired fail states
It seems like ``SubSceneTests.cs`` is trying to check all scenes in the project, which means it can fail when it shouldn't for trying to operate scenes that aren't meant to be *maps*. #5648  will introduce a fail state in the project once merged because of this.

To fix, make sure only the relevant scenes are tested, this could be defined by the folder where they are sitting.

Relevant information:
Error reported by #5648 
```
System.ArgumentException : Cannot open scene with path "Packages/com.unity.vectorgraphics/Tests/Editor/CornerTests.unity".at (wrapper managed-to-native) UnityEditor.SceneManagement.EditorSceneManager.OpenScene_Injected(string,UnityEditor.SceneManagement.OpenSceneMode,UnityEngine.SceneManagement.Scene&)
  at UnityEditor.SceneManagement.EditorSceneManager.OpenScene (System.String scenePath, UnityEditor.SceneManagement.OpenSceneMode mode) [0x00000] in <16d7f804c0bf49f2b4e46d9aa10e11d6>:0 
```

[Link to failing test's source code](https://github.com/unitystation/unitystation/blob/bab8b478a3b3df01dd3a5ce605ad537e6cf47219/UnityProject/Assets/Tests/SubsceneTests.cs#L142)

Copern:
- Adds in the new radial menu.
- Adds shortcuts to the main item ring : shift-click to examine, ctrl-click to pull, and click to pick up.
- Menu remains open when the user right clicks on an item. No longer requires the user to hold down the mouse button.
- Clicking anywhere outside of the menu will close it.
- Certain actions will leave the menu open (for example, examine) while others close it.
- If there are less items than there are segments in the radial, the segments will be transparent and inactive. Inactive segments will will not capture click actions and will allow interaction with the world.
- If there are more items than there are segments, scrolling through the items is allowed through either the mouse wheel or by dragging the item ring around.
- No longer opens up items that are underneath the inventory slots.
- The menu and its buttons are no longer destroyed and recreated every time a user opens and closes the menu.

Known Issues and/or concerns
- Codacy has a some issues with the old RightClick* classes, I've kept changes to those to a minimum for now.
- Some actions do not have icons. As a temporary solution, instead of an icon, it will show text describing the action. I'm not a spriter so someone will need to make some icons for them.
- Some items need right click appearance changes and/or sprite renderer reordering. Again, as a temporary solution, text will show instead of an icon in these cases.
- Some bugs that were in the old menu, such as being able to open up the menu over another window, are still there.
- The menu does not currently update the available actions when moving into or out of range of the items. This will be changed later.
- I wanted to add some customization options before this release. However, the options menu is a royal pain to work with so they won't make it this time. Some planned customization options include changing the number of segments, changing the scale, etc.


ThatDan123:
Tests are breaking cus of this:

System.ArgumentException : Cannot open scene with path "Packages/com.unity.vectorgraphics/Tests/Editor/CornerTests.unity".at (wrapper managed-to-native)

CdrCross:
Two good pieces of feedback from the ui/ux chat:

1. The mouse-drag rotation of the object ring should be allowed to be a bit faster/smoother.
2. It would be beneficial to add a toggle button to the radial to switch between "Show all objects" and "Show only unique objects", to allow for easier browsing through large amounts of objects for outliers.

corp-0:
I think the problem might be on us, we have no reason to try to open and test that scene``System.ArgumentException : Cannot open scene with path "Packages/com.unity.vectorgraphics/Tests/Editor/CornerTests.unity".at (wrapper managed-to-native)``

PetMudstone:
- Control with radial menu open should show admin options like Smash, Hotspot, etc. Can either be held down or a toggle. Right now, you can still see admin buttons by holding control while pressing the right mouse button, but you can't bring up the admin buttons any other way.

corp-0:
didn't someone post a picture of the radial with admin buttons?

Copern:
It works the exactly the same as it did in the old menu. Holding ctrl and right clicking brings up the menu with admin options available. I haven't changed that aspect of the RightClickManager. I've kept changes to the RightClickManager/RightClickMenuItem at a minimum and the way it works doesn't allow for updating the menu very easily (same as the old menu). That was something I was going to work on in a later iteration.

PetMudstone:
Yes, I'm aware. It just seems more natural for the control key with the radial already up to bring up admin options. It might be because the radial menu is a toggle now and that's messing me up a bit when it comes to holding the control key. Oh well.

UnityProject/Assets/Scripts/UI/Core/Animations/AnimatedRadialRotation.cs
```diff
@@ -0,0 +1,109 @@
+﻿using System;
+using UI.Core.Radial;
+using UnityEngine;
+
```
corp-0, line 4: add a namespace for any new script you add. In this case the namespace would be ``UI.Core.Animations``

UnityProject/Assets/Scripts/UI/Core/Animations/AnimationDirection.cs
```diff
@@ -0,0 +1,6 @@
+﻿public enum AnimationDirection
```
corp-0: same here, namespace.

UnityProject/Assets/Scripts/UI/Core/Animations/ReversibleObjectScale.cs
```diff
@@ -0,0 +1,76 @@
+﻿using UnityEngine;
+
```
corp-0, line 2: namespace

corp-0:
code is clean and well documented, I like what I see. I have to ask though, why are there some bvold and seemingly unrelated prefabs modified?

Copern:
Whoops, I forgot the namespaces on those classes.

As for the bvoid and prefab changes, it's because of the old RightClickAppearence component. They are all nearly pointless on those prefabs so I'll be removing them on everything that doesn't actually need it. I'll also have a simple fix for the test so it doesn't try to check package scenes.

The next commit will have an outline for the description text so it's a lot easier to see in bright areas. I've upped the max delta for the radial drag by default. I was told it was too fast before but now I guess it was too slow. 

That is one of those things that I want to add to the options menu in the future, one speed isn't going to fit everyone. As I made the radial completely grayscale, adding an option for the default button color will be there too, as well as the scaling and number of items options.

Show only unique objects will have to wait for another commit.

corp-0:
sounds wonderful, thank you very much.

PR #5656, opened 2020-11-21
allow inital filter values to be mapped

AutumnNova:
maybe
- implements a var to change what the filter is set to initally
- change the gui so it will reflect this properly, instead of always showing o2 as being fiiltered intially
- remove currently filtering as it was unused
- renamed ToMaxPressure to MaxPressure to better reflect its usage
notes: id rename the setFilterAmount(Number) to something else as its somewhat inaccurate considering it has a string passed to it, not a number, but its used in the tab prefab and i cant be bothered to update that ¯\\\_(ツ)\_\/¯

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Objects/Atmospherics/Pipes/GUI_Filter.cs
+++ b/UnityProject/Assets/Scripts/UI/Objects/Atmospherics/Pipes/GUI_Filter.cs
@@ -20,7 +20,7 @@ namespace UI.Objects.Atmospherics
 		{
 			foreach (var INFilter in Filter.CapableFiltering)
 			{
-				if (INFilter.Key == gasName.ToString()) //Checks what button has been pressed  And sets the correct position appropriate
+				if (INFilter.Key == gasName) //Checks what button has been pressed  And sets the correct position appropriate
 				{
 					((NetUIElement<string>)this[INFilter.Key]).SetValueServer("1");
 				}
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/UI/Objects/Atmospherics/Pipes/GUI_Filter.cs
```diff
@@ -90,9 +76,9 @@ public void TogglePower()
 		}
 
 
-		public void SetMaxPressure(int To)
+		public void SetMaxPressure(int Value)
```
corp-0:
```suggestion
		public void SetMaxPressure(int value)
```
I know it was ``To`` before, but we can amend those style errors now

UnityProject/Assets/Scripts/UI/Objects/Atmospherics/Pipes/GUI_Filter.cs
```diff
@@ -17,32 +16,9 @@ public class GUI_Filter : NetTab
...
-		{"NO2",Gas.Nitryl},
-		{"TRIT",Gas.Tritium},
-		{"HN",Gas.HyperNoblium},
-		{"STIM",Gas.Stimulum},
-		{"PLX",Gas.Pluoxium},
-		{"FRE",Gas.Freon},
-	};
-
-
 		public void SetFilterAmount(string Number)
```
corp-0:
```suggestion
		public void SetFilterAmount(string number)
```

PR #5658, opened 2020-11-21
Fix the NaN temperature in gas mixing.

### LINKED ISSUE #5630: NaN temperature when trying to fill an empty tank
### Environment
- Game Version: Build 5608
- OS Version: Windows 10, reproducible on Linux clients as well

### Detailed description of the bug
If an empty canister (like unlabeled yellow canisters or those created from metal) is connected to a pipe section, that section gets a NaN temperature. Any attempts at filling the pipe section would drain the gas down to 0KPa while keeping the temperature of that section at NaN. The only way to then use that pipe section normally is to reconstruct it entirely. Any canister, connected to the NaN-temperature pipe section will be immediately drained to 0KPa and then behave exactly like an empty canister. mentioned above.

### Steps to reproduce

1. Construct a pipe section with a canister connector attached, or use station atmospherics filling setup, with a pressure meter attached.
2. Attach an empty (see description above) canister to the connector, and open the valve
3. Temperature in the pipe is now NaN, as indicated by the pressure meter.
*Edit: pressure readout is actually at 0kPa, and the temperature is NaN*

Copern:
Fixes #5630

When there were gasses with a total heat capacity of 0, the temperature was determined through a division by 0 causing a NaN temperature to be set to the resulting mix.


corp-0:
very nice, thanks

PR #5662, opened 2020-11-22
Nukes will now share the same activation code.

### LINKED ISSUE #5596: Nuclear operatives receive a code for the on-station nuke and not the nuke that spawns in their shuttle.
### Environment
- Game Version: Staging 5538
- OS Version: Windows 10

For some reason, nukies receive the code for the on-station nuke instead of the nuke that comes on the syndicate shuttle. For one, the nukies got the code for the shuttle nuke before. Second, this leads to confusion among nuclear operatives who try to prime the shuttle nuke. Images below come from a nuclear emergency round.
Nukie schut
damn station nuke

### Steps to reproduce
1.  Have nukies spawn. Look at the nuke code that they are given.
2. Look at nuke code that the nukie shuttle nuke has according to overlay. It won't match what the nukies were given.
3. Look at nuke code that the on-station nuke has according to overlay. It will match what the nukies were given.

Aranclanos:
### Purpose
Nukes will now share the same activation code.
Fixes #5596

### Notes:
there's a counterargument where a syndicate player could leak the code in OOC and the crewmembers themselves will get themselves blown up, but considering this can happen right now and doesnt, I desestimate it

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Nuke.cs
+++ b/UnityProject/Assets/Scripts/Objects/Nuke.cs
@@ -62,7 +62,7 @@ namespace Objects.Command
 		public string CurrentCode => currentCode;
 
 		//Nuke code is only populated on the server
-		private int nukeCode;
+		private static int nukeCode;
 		public int NukeCode => nukeCode;
 
 		private void Awake()
@@ -82,7 +82,10 @@ namespace Objects.Command
 		[Server]
 		public void CodeGenerator()
 		{
-			nukeCode = Random.Range(1000, 9999);
+			if (GameManager.Instance.CurrentRoundState == RoundState.PreRound) //midround admin spawned nukes wont change the code
+			{
+				nukeCode = Random.Range(1000, 9999);
+			}
 			//Debug.Log("NUKE CODE: " + nukeCode + " POS: " + transform.position);
 		}
 


corp-0:
Uhm but does it make sense from an IC perspective? Wouldn't making the synd get their nuke code instead of the station's one be a better solution?

Akbadain29:
I think it would be better if the Station nuke code is transmitted on the Command Channel if the Station is set to Red Alert and is kept separate from the Syndie nuke one. That way if the admins are not on and the nukies can get the station nuke code instead.
However, since this nuke was stolen in the first place, I guess it being keyed into with the same code as the station code does make sense, as you take the nuke disk that would only work with the one code.

Aranclanos:
Seems like most people dont like the code being the same, but I'll give it till tomorrow 
dont merge meanwhile

corp-0:
very well, I will add the DNM label then.

BootlegBow:
Yeah, it makes sense for NT to make all their nukes have the same code because NT is just like that, and the nukie's nuke is a stolen NT-brand one

AutumnNova:
> Yeah, it makes sense for NT to make all their nukes have the same code because NT is just like that, and the nukie's nuke is a stolen NT-brand one

NT is just like that feels like a horribly flimsy argument, a lore based one at that
from a gameplay perspective they should be separate as it would be possible for a crewmember to be informed of the code ooc and then grief using it. it should also arguably not be the same as it allows the syndicate to use a immovable nuke should anything happen to their initial one, which is making the possibility of the station actually not being blown up smaller, it being a slim chance already

corp-0:
> it allows the syndicate to use a immovable nuke should anything happen to their initial one, which is making the possibility of the station actually not being blown up smaller, it being a slim chance already

I second this one. That's a huge change in gameplay.

Mute-Man:
what's the deal with this PR?
it's been sitting here for a month now.

corp-0:
now that #6040 was merged we can close this pr

PR #5666, opened 2020-11-23
implements functional suppressors

AutumnNova:
adds a suppressor item that can be attached to guns which have the isSuppressible bool set to true in their prefab
suppressor can be removed by holding the gun in your other hand and alt clicking on it
suppressors change the isSuppresed bool to true, preventing any messages from being sent related to you shooting
notes: i somewhat dislike needing to set a suppressor to a var in the prefab, but i couldnt think of a better method \
CL: adds functional suppressors


corp-0:
code looks good. I still would like someone to do some testing, guns are known to have regressive bugs every time someone touches them.

PetMudstone:
Built and tested as host and client. Suppressed pistols don't show action messages for either host or client when fired as is intended. Suppressors can be removed from pistols by both the host and client. Removing suppressors from pistols properly makes them un-suppressed again. Reloading works, Other ammo types work. 

Suppressed pistols for the host don't make any sound at all, but that's apparently because suppressed sound is undefined for now. However, suppressed pistols for the client do make the standard un-suppressed firing noise.

corp-0:
> Suppressed pistols for the host don't make any sound at all, but that's apparently because suppressed sound is undefined for now. However, suppressed pistols for the client do make the standard un-suppressed firing noise

Sounds like the weapon is suppresed on server but it isn't on client

corp-0:
@the-other-autumn maybe put this block https://github.com/unitystation/unitystation/blob/6c8597e1e11ab35ef19e68ed67b50d3b51f64c65/UnityProject/Assets/Scripts/Items/Weapons/Gun.cs#L715 inside an ``isServer`` if statement. I remember sound was intended to be called only from server and the method where this is evaluated seems to be called on both server and client.

Btw this method could use some refactoring, simple abstracting blocks outside of it would help a lot the readibility.

AutumnNova:
> 
> https://github.com/unitystation/unitystation/blob/6c8597e1e11ab35ef19e68ed67b50d3b51f64c65/UnityProject/Assets/Scripts/Items/Weapons/Gun.cs#L715
> inside an `isServer` if statement. I remember sound was intended to be called only from server and the method where this is evaluated seems to be called on both server and client.
> 
Are  you sure your not thinking of networked sounds? this section has been run both serverside and clientside for a while and i dont recall any issues with it

corp-0:
you're right, I was, but shouldn't this be a networked sound?

AutumnNova:
display shot is called on clients to well, display the shot, i dont see much reason to network the sounds separately?

corp-0:
to make sure everyone gets the same, I guess

corp-0:
uhm, another thing that could work is making ``IsSupressed`` a syncvar. I'm merely throwing stuff that I think could be happening based on what Pet reported.

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
```diff
@@ -170,8 +169,18 @@ public class Gun : NetworkBehaviour, IPredictedCheckedInteractable<AimApply>, IC
...
+		/// If true, displays a message whenever a gun is shot
+		/// </summary>
+		[SyncVar(hook = nameof(SyncIsSuppressed)), Tooltip("If the gun displays a shooter message")]
 		public bool isSuppressed;
+
+		/// <summary>
+		/// Enables or disables the behaviour related to applying and removing suppressors from the gun
+		/// </summary>
+		[Tooltip("If suppressors can be applied or removed")]
 		public bool isSuppressible;
```
corp-0: will ``isSuppressible`` ever change? I think this could be a private serialized field if we don't need to get the value anywhere and we don't expect the value to change after the object has spawned.

AutumnNova: several weapons start suppresed (energy crossbow)

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
```diff
@@ -327,6 +316,43 @@ public virtual bool WillInteract(AimApply interaction, NetworkSide side)
 			return false;
 		}
 
+		public virtual bool WillInteract(HandActivate interaction, NetworkSide side)
+		{
+			if (!DefaultWillInteract.Default(interaction, side)) return false;
```
corp-0:
````suggestion
```

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
```diff
@@ -327,6 +316,43 @@ public virtual bool WillInteract(AimApply interaction, NetworkSide side)
...
+			if (CurrentMagazine != null && allowMagazineRemoval && !MagInternal && side == NetworkSide.Client)
+			{
+				return true;
+			}
+			return false;
+		}
+
+		public bool WillInteract(InventoryApply interaction, NetworkSide side)
+		{
+			if (!DefaultWillInteract.Default(interaction, side)) return false;
```
corp-0:
```suggestion
			if (DefaultWillInteract.Default(interaction, side) == false) return false;
```

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
```diff
@@ -366,63 +392,44 @@ public virtual void ServerPerformInteraction(AimApply interaction)
...
+					}
+					else if (Validations.HasItemTrait(interaction.UsedObject, CommonTraits.Instance.Suppressor) && !isSuppressed && isSuppressible)
+					{
+						SyncIsSuppressed(isSuppressed, true);
+						// setting the suppressor var to the gameobject and then spawning that would be better,
+						// however the despawning sets the var to null
+						Despawn.ServerSingle(interaction.UsedObject);
+					}
+				}
+				else if (isSuppressed && isSuppressible)
```
corp-0, line 423: Not that it will make any difference in performance, but it makes more sense to me to ask if it is suppressible first.

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
corp-0: I didn't understand. You need to get this variable elsewhere? If you want to read it on a child class of ``Gun`` you could either do a public getter or make it ``protected``.

AutumnNova:
should be good now

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
```diff
@@ -172,8 +172,9 @@ public class Gun : NetworkBehaviour, IPredictedCheckedInteractable<AimApply>, IC
 		/// <summary>
 		/// If true, displays a message whenever a gun is shot
 		/// </summary>
-		[SyncVar(hook = nameof(SyncIsSuppressed)), Tooltip("If the gun displays a shooter message")]
-		public bool isSuppressed;
+		[SerializeField, SyncVar(hook = nameof(SyncIsSuppressed)), Tooltip("If the gun displays a shooter message")]
+		private bool isSuppressed;
+		public bool IsSuppressed => isSuppressed;
```
corp-0, line 177: this but with ``isSuppressible`` which I don't expect to change after being set, ever.

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
```diff
@@ -180,7 +180,7 @@ public class Gun : NetworkBehaviour, IPredictedCheckedInteractable<AimApply>, IC
 		/// Enables or disables the behaviour related to applying and removing suppressors from the gun
 		/// </summary>
 		[Tooltip("If suppressors can be applied or removed")]
-		public bool isSuppressible;
+		private bool isSuppressible;
```
corp-0, line 183: you forgot the SerializedField decorator

corp-0:
Are the oddities described by Pet solved?

PetMudstone:
Built and tested again. Sound changes work for clients too now.

corp-0:
Thank you, Mudstone.

UnityProject/Assets/Resources/ScriptableObjects/Traits/Usage/Suppressor.asset
```diff
@@ -0,0 +1,16 @@
...
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 0}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
+  m_Script: {fileID: 11500000, guid: 4ccfe138752d4b66bcb7c56a0c9769c7, type: 3}
+  m_Name: Suppressor
+  m_EditorClassIdentifier: 
+  traitDescription: This means the item has the capacity for a battery inside of
```
NoooneyDude: Perhaps a more appropriate trait description for a suppressor?

PR #5670, opened 2020-11-24
Return of PogStation, some prefab stuff.

### LINKED ISSUE #4874: Mainstations, Away Sites and Centcoms require piping changes!
After #4859 a new gas piping system was introduced and the maps need to update and integrate the new stuff.

This issue exists to keep track of the progress on this task:
- [x] BoxStationV1
- [x] FallStation
- [x] OutpostStation
- [ ] Pogstation
- [x] SquareStation
- [x] AncientTemple (Optional)
- [x] Caverna (Optional)
- [x] DarkJungle (Optional)
- [x] Palace (Optional)
- [x] Relicta (Optional)
- [x] CentCom
- [x] Fallstation CentCom
- [x] Lavaland (Optional)

Note from ThatDaniel: When you map pipes try to go through maints as well, makes them look better, and actually useful, having a loop in the main hallway is good too. See the Box station pipe mapping for a good example. This does mean it will take longer, but means it wont have to be redone later.

PetMudstone:
### Purpose
- Adds PogStation back into rotation. Fixes #4874.
  - Cleaned up BVOld objects thanks to Aranclanos.
  - Finishes the atmos pipenet.
- Adds more atmos object variants. Pre-configured filters and mixers, for example.
- Altered a material so that pipe colors actually work.

CL: Returned PogStation back to rotation.


ThatDan123:
"Altered a material so that pipe colors actually work."

What does that mean, the pipe colours worked fine for the other maps without editing the material itself.

PetMudstone:
For some reason, changing the shader the Underfloor's material (Sprite-Pixel Snap) used to the one that would allow pipes to be colored caused the material in question to be modified for some reason. The shader change only carried over if the material was left modified.

Oddly, the materials (which use the proper shader already) on other matrices on other maps can't actually be modified as far as I can tell. They're greyed out in the inspector.

corp-0:
Can someone clarify what is this material situation?

AutumnNova:
Shouldnt need to modify the material directly, the material can be changed by editing one of the components on the layer itself

Issue #5316, opened 2020-10-15
ProTip Advice and Tutoring System B: 130


CdrCross:
Update 13/11/22 - Made some changes per Max's recommendations.

## Introduction and Overview
The goal of the ProTip system is to provide new players to Unitystation responsive, automatic advice that will help them navigate our complicated game. With a flexible and easy-to-use trigger system, the ProTip system could potentially become a simple and low-maintenance means of easing the new player experience. The system is largely inspired by the Learning Helper from Rimworld. This issue very much ties in with Issue #3728 on implementing Mentors. Ideally, the system should be implemented so that non-technical contributors can write and configure new Advice. This issue has/will have a bounty of USD130.

## The GUI
The GUI should be a text box that appears in the top right of the screen, approximately taking up the same space as admin tools currently do, given that the overlap between admins and players needing advice should be approximately zero. Advice is mostly just text, but it should ideally be able to contain hyperlinks to Wiki articles on more advanced topics. Control advice should also be responsive to customized keybindings. Every Advice window must contain buttons to contact a Mentor for further questions, a button linking to the Wiki front page, an X button to close the window as well as a button to adjust Advice Frequency, see below.

When the round first starts for a new player to the server, or when a player presses the “Adjust Advice frequency”-button, the hint window should pop up in the top right, offering to tailor the level of advice to the experience level the player chooses by pressing one of the following buttons:    

1. New to Space Station 13 (All Advice)
2. New to Unitystation (Only Advice on new features in Unitystation)
3. Somewhat Experienced (Life critical Advice only)
4. Robust (No Advice)

Each piece of advice will only be shown once to a given player, and then a Help section of the Escape menu will offer a list of all advice, for players to re-read if they need to be reminded of something.

## The Triggers
Advice pop-ups should trigger based on a variety of circumstances around the player. The system should be able to detect the following:

- The player holding a specific object
- The player holding one of a class of objects (Holding any gun, for example)
- Being near a world object; contains both the object ID and trigger range (The clone scanner, for example)
- Opening a GUI (Opening the hacking screen should trigger hacking advice, for example)
- Health states (The player taking damage, going unconscious, or dying)
- The player being in a hazardous area (Extreme cold or low O2 atmosphere, for example)
- Time from player joining the server
- (Cancelled) Player actions (For walking through sequential steps, like building a wall following a girder) 
- (Cancelled) Nearby players’ health states; contains both the relevant health state and trigger range (Finding someone unconscious or dead, for example)
[...]

PR #5675, opened 2020-11-26
Add ghost role system

NoooneyDude:
- Adds `AnimateIcon` component, animates a UI gameobject. May also be useful for issue #5316.
animate icon component
- Adds ghost role system.
ghost role system in use
- Adds simple Death Squad creation for admins, to consume the ghost role system.
- Fixes NRE when the game closes for `ItemActionButton`.

CL: Added ghost role system and simple death squad.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Sprite Handler/SpriteHandler.cs
+++ b/UnityProject/Assets/Scripts/Core/Sprite Handler/SpriteHandler.cs
@@ -144,7 +144,7 @@ public class SpriteHandler : MonoBehaviour
 
 	public void ChangeSprite(int SubCataloguePage, bool Network = true)
 	{
-		if (SubCataloguePage == cataloguePage) return;
+		if (cataloguePage > -1 && SubCataloguePage == cataloguePage) return;
 
 		if (SubCataloguePage >= SubCatalogue.Count)
 		{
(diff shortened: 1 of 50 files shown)


corp-0:
*you're satiated*

UnityProject/Assets/Scripts/Core/Sprite Handler/SpriteHandler.cs
```diff
@@ -175,6 +175,7 @@ public void ChangeSprite(int SubCataloguePage, bool Network = true)
 		{
 			isPaletteSet = false;
 			PresentSpriteSet = NewspriteDataSO;
+			cataloguePage = SubCatalogue.FindIndex(SO => SO == NewspriteDataSO);
```
corp-0, line 180: what does this do?

Bod9001: so, just noting this won't Network it as if it was a catalogue change, it would network it like Some are specified a So for This Sprite handler

NoooneyDude: Yep so it was to fix an issue when you manually set SO (outside of catalogue or not) couldn't go back to the previous SO (or otherwise use ChangeSprite()) because the cataloguePage variable wasn't updated, so it thought it was already using that SO.

Could just assign -1 instead of checking to see if the new SO is already in catalogue if you are worried.

Bod9001: yeah is ok, just add a note to note it needs to be changed to a catalogue message at some point to optimise it

PR #5686, opened 2020-11-30
GasMix refactor

Aranclanos:
### Purpose
Refactors GasMixes
Removes the constant creation of new GasMixes on each operation, greatly removing the amount of GC generated.
GasMixes will now be a class.
Cleans hotspots from unused or copypasted information, they'll work as a trigger of reactions now instead.
Moves the class MixAndVolume to its own file.
Scrubbers and vents will now act slower.
Fires will now last longer and generate more temperature.
Hotspots will now spread to adjacent tiles as long as they exist.

### Notes:
There's still a lot of atmos related GC but this should tackle the biggest issues while the rest might be easier to do in a smaller PRs easier to read


corp-0:
You will get a lot of codacy complains about comparing floats to 0. Use a range or round it in the comparison.

Also, why change ``GasMix`` to ``gasMix`` ?

UnityProject/Assets/Scripts/Objects/Engineering/Reactor/ReactorGraphiteChamber.cs
```diff
@@ -165,10 +165,10 @@ public void CycleUpdate()
 				if (ReactorPipe.pipeData.mixAndVolume.Total.x > 0 &&
 					ReactorPipe.pipeData.mixAndVolume.Temperature > BoilingPoint)
 				{
-					var ExcessEnergy = (ReactorPipe.pipeData.mixAndVolume.Temperature - BoilingPoint);
-					ReactorPipe.pipeData.mixAndVolume.Remove((ReactorPipe.pipeData
-						.mixAndVolume.Total * (EnergyToEvaporateWaterPer1
-											   * ExcessEnergy * WaterEnergyDensityPer1)));
+					var ExcessEnergy = ReactorPipe.pipeData.mixAndVolume.Temperature - BoilingPoint;
+					ExcessEnergy *= EnergyToEvaporateWaterPer1 * WaterEnergyDensityPer1;
```
Bod9001, line 169: just remember to be careful here since some NAN have come from Reactor and pipes

UnityProject/Assets/Scripts/Objects/Pipes/Devices/AirVent.cs
```diff
@@ -11,7 +12,7 @@ public class AirVent : MonoPipe
 		public bool SelfSufficient = false;
 
 		// minimum pressure needs to be a little lower because of floating point inaccuracies
-		public float MaxTransferMoles = 100;
+		public float MaxTransferMoles = 10;
```
Bod9001: remember to update prefabs, also this will make scrubbers slow as balls when trying to get rid of plasma, to be fair there already pretty slow

Bod9001:
smashing

Aranclanos:
the `MaxTransferMoles = 10;` changes were some leftover from some tests, restored them back to 100
The GaxMix variable change to gaxMix because I noticed in some places it was really confusing and went on a crusade against it, might have not been the best idea to include it on this pr, a little bit too sidetracked, but oh well

corp-0:
Confusing how? For sharing the same name with the type? Maybe do it a getter then, so you prefix “Get” to it.

Public fields should use PascalCase instead of camelCase.

UnityProject/Assets/Scripts/Systems/Fluids/MixAndVolume.cs
```diff
@@ -37,7 +37,7 @@ public float Temperature
 		{
 			get
 			{
-				if (WholeHeatCapacity == 0)
+				if (Mathf.Approximately(WholeHeatCapacity, 0))
```
Bod9001, line 40: with these cases, it only needs to trigger if it is exactly 0, since there is a divide by coming up

corp-0: So this is to avoid dividing by zero? Might be a better way to handle that, maybe a try catch.

corp-0: https://docs.microsoft.com/en-us/dotnet/api/system.dividebyzeroexception?view=net-5.0

corp-0: > Dividing a floating-point value by zero doesn't throw an exception; it results in positive infinity, negative infinity, or not a number (NaN), according to the rules of IEEE 754 arithmetic. 

shieeeeet

corp-0: alright, then maybe checking if the division result is NaN https://docs.microsoft.com/en-us/dotnet/api/system.single.isnan?redirectedfrom=MSDN&view=net-5.0#System_Single_IsNaN_System_Single_

Aranclanos: up to you guys MixAndVolume.cs is not really my code, I dont have counter arguments if you want to edit it but I dont want it to hold the pr

Bod9001: humm,  would be nice and easy to fix if it's an issue so is saying it's okay

PR #5700, opened 2020-12-03
Adds the Contract of Apprenticeship

NoooneyDude:
- Adds the Contract of Apprenticeship. Continues #5160.
contract of apprenticeship
    (actual timeout period is 15 seconds)
- Fixes ghost role on server timeout triggering when ghost role had already been removed.

> Note:
> - not tested on headless; seems to be a develop issue.
> - artifact refunding is not currently implemented.

CL:Added Contract of Apprenticeship.


NoooneyDude:
Regarding the `Codacy` warnings:
- `ContractOfApprenticeship.OnGhostRoleTimeout` event _is_ actually invoked; see line 64 of the same file - you can see it working as intended in the gif.
- The float comparison warnings can be ignored as we're setting a specific integer value for two specific cases; this value was not derived from maths with float.

PR #5704, opened 2020-12-03
proper firing pin implementation

AutumnNova:
adds a firing pin trait
adds a slot for firingpins contained within guns
readds the job based limiting part of the firing pin script because someone removed it.
fixes several clientside prediction issues


Bod9001:
do all the guns come with default firing pins apart from syndicate ones?

AutumnNova:
> do all the guns come with default firing pins apart from syndicate ones?

all guns come with the default pin currently, the systems to add more pins exists, but the standard pin was the only prefab i found

Bod9001:
ok cool

PR #5705, opened 2020-12-04
Gas tank variants: oxygen, plasma tanks, and more.

PetMudstone:
### Purpose
- Adds proper variants of oxygen tanks, other gas tanks such as plasma tanks, and also non-functional jetpacks.
  - Replaces obsolete versions everywhere.
- Removes template scenes, since they just make tests angry.

### Notes:
I tried to make the contents of all the gas tanks as accurate to /tg/ as I could. Some of the values are kinda weird due to how Unitystation does atmos math, but oh well.
How much oxygen is consumed from oxygen tanks is quite different after #5686 than before. My new emergency oxygen tank got emptied in 30 seconds before that PR. Now, I estimate it will take at least 1 hour and 40 minutes.


corp-0:
what are the sounds you're adding here?

PetMudstone:
Sound files aren't being added. They're being updated for some reason. I dunno why this happens but it does. Seems to be harmless.

PR #5706, opened 2020-12-04
Feature 5309 bluespacebanana teleport on slip

### LINKED ISSUE #5309: Bluespace Banana Peels should teleport on slip
## Description

When a character slips on a bluespace nanner peel, it should teleport them randomly to a spot a short distance away. Adding this would provide Botanists with an alternative to _only_ growing lemon-nades because that's the only mutagen plant with code at the moment. It also provides a theoretical _non-violent_ method of escape for perma-brigged prisoners, as they can grow the bananas to intentionally teleport out of the brig.

Also, it'd make the clown crowd quite happy.

TheGreatShelbino:
### Purpose
Adds feature #5309 bluespace banana teleport on slip

### Notes:
Currently it does not factor in potency when calculating max distance due to no way of detecting the potency of the banana peel slipped on.


corp-0:
don't mind the test fail, that's on our side. Fix the codacy complain tho, you have an unused variable.

PR #5707, opened 2020-12-04
switch suppressors to use slots

AutumnNova:
aint much too it
CL: fixed the radius around the player that caused them to accidentally attack themselves


PetMudstone:
Tests just complain about the template scenes, which will be removed in 5705 anyways.

corp-0:
the octagon of doom is the wrong interaction when clicking the adjacent tile and hitting yourself?

AutumnNova:
it happens with guns, lowering the magnitude means the check is only done when your target is on the same tile

corp-0:
how did you get to 0.5 as the right number?

AutumnNova:
half a tile from the players position (the center of a tile) , as apposed to 1.1, which is a tile and a bit
i have tested these values

corp-0:
alrighty then, thanks. Can you mention that fix for the changelog? It is an important one some players will be happy to read it is fixed.

PR #5709, opened 2020-12-05
Player Prefab foundational re-write

jamessimo:
## Purpose
To give us a clean platform for all future health work, before player body sprites were added the same way we add clothes, now the player gets assigned a SO at spawn which will contain all the bodyparts they need and add them to the player. Bodypart prefabs now contain the health AND sprite info about a body part, so if you hack off someone's arm, you are using that arm. 

### Features:

- Limbs are now 100% controlled by Bodypart implants (removed old cloth system)
- Bodypart prefabs control gender
- Added first pass of multi races (still WIP)
- Player prefab is now completely void of any race or gender data (before the player sprite was hard-coded as human. 
- Implanting new limbs is possible on players
- Foundation is set to not allow player to interact with things if they have no hands
- Player speed is dictated by the speed of there limbs 
- Limbs can be set to Bleed or not (not yet used but foundationally it works) 
- Foundation for limbs using reagents is in, no changes from @Fortune117 original work. 

### Notes:
Lots of issues;
- Player is SLOW because for some reason player speed is not updating when implants are added after the playerprefab. needs investigation
- Genders are not yet correctly set, need to change bodypart prefabs BEFORE they are used by the player spawn. 
- Races are possible but didn't add them until I am happy with the human player logic 
- Eye and Hair code is still using the old cloth system, need all this to be attached to the head prefab on spawn
- Need to make sure the Character Creation screen can still work (It too uses the old cloth based system) 
- Player bodypart GO still have ClothItem scripts which can be removed. 
- Player (bring leg to front) code is garbage and needs a much better implementation.  (See line 235 in PlayerSprites.cs

This is not finished, however I wanted to get this out since some issues IRL are making it hard to keep up-to-date.


Bod9001:
nice will work on this thanks

PR #5710, opened 2020-12-06
Welding fuel no longer creates plasma when removed from containers.

PetMudstone:
### Purpose
- Welding fuel no longer creates plasma gas when it is removed from containers like welding tools or fuel tanks.

CL: Removed ability for welding fuel to create plasma gas when it is emptied out of containers.


corp-0:
don't worry about the example, remove the entire case.

PR #5712, opened 2020-12-07
Added in Maint Spawns for players and a Maint Teleport

Akbadain29:
### Purpose
Adds in Maint Spawns for Fugitive Role and A Maint Quantum Pad to jump between two basic maint corridors.


ThatDan123:
Did you set the quantum pad to indestructible and not deconstructable?

Akbadain29:
No, do they need to be? I thought because they aren't part of the usual ones that I didn't need to do that.

ThatDan123:
If you want them to be destroyable/ deconstructable its fine, just that the other mapped ones are prevented.

PR #5713, opened 2020-12-07
removable powercells which can be swapped out with other weapons

AutumnNova:
adds the ability to remove cells from an energy gun with a screwdriver
adds the ability to reload cells into an energy gun
adds a trait named weaponcells which dictates what can and cannot be inserted into energy firearms


Bod9001:
couldn't you just do this with a slot capacity thing?

Bod9001:
oh yeah am dumb don't listen to me

PR #5714, opened 2020-12-07
Disables quick load by default.

PetMudstone:
### Purpose
- Title. Quick load was turned on. This just turns it off.


corp-0:
thanks

PR #5716, opened 2020-12-08
some cell bugfixes

AutumnNova:
also some cleanup

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Weapons/ElectricalMagazine.cs
+++ b/UnityProject/Assets/Scripts/Items/Weapons/ElectricalMagazine.cs
@@ -16,7 +16,7 @@ public class ElectricalMagazine : MagazineBehaviour
 
 	public void Awake()
 	{
-		magType == MagType.Cell;
+		magType = MagType.Cell;
 		battery = GetComponent<Battery>();
 	}
 
--- a/UnityProject/Assets/Scripts/Items/Weapons/ElectricalMagazine.cs
+++ b/UnityProject/Assets/Scripts/Items/Weapons/ElectricalMagazine.cs
@@ -16,7 +16,7 @@ public class ElectricalMagazine : MagazineBehaviour
 
 	public void Awake()
 	{
-		isCell = true;
+		magType == MagType.Cell;
 		battery = GetComponent<Battery>();
 	}
 
@@ -43,15 +43,4 @@ public class ElectricalMagazine : MagazineBehaviour
 		double percent = (battery.Watts * 100 / battery.MaxWatts);
 		return $"It seems to be compatible with energy weapons. The charge indicator displays {Math.Round(percent)} percent.";
 	}
-
-	#if UNITY_EDITOR
-	[CustomEditor(typeof(MagazineBehaviour), true)]
-	public class MagEditor : Editor
-	{
-		public override void OnInspectorGUI()
-		{
-			DrawDefaultInspector();
-		}
-	}
-#endif
 }
--- a/UnityProject/Assets/Scripts/Items/Weapons/ElectricalMagazine.cs
+++ b/UnityProject/Assets/Scripts/Items/Weapons/ElectricalMagazine.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using UnityEngine;
+using UnityEditor;
 
 /// <summary>
 /// Gun scripts are strange, This is to make it easy to have a charging Ammo clip
@@ -7,13 +8,15 @@ using UnityEngine;
 [RequireComponent(typeof(Battery))]
 public class ElectricalMagazine : MagazineBehaviour
 {
-
+	[NonSerialized]
 	public Battery battery;
 
+	[NonSerialized]
 	public int toRemove;
 
 	public void Awake()
 	{
+		isCell = true;
 		battery = GetComponent<Battery>();
 	}
 
@@ -40,4 +43,15 @@ public class ElectricalMagazine : MagazineBehaviour
 		double percent = (battery.Watts * 100 / battery.MaxWatts);
 		return $"It seems to be compatible with energy weapons. The charge indicator displays {Math.Round(percent)} percent.";
 	}
+
+	#if UNITY_EDITOR
+	[CustomEditor(typeof(MagazineBehaviour), true)]
+	public class MagEditor : Editor
+	{
+		public override void OnInspectorGUI()
+		{
+			DrawDefaultInspector();
+		}
+	}
+#endif
 }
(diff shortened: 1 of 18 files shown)


UnityProject/Assets/Scripts/Items/Weapons/ElectricalMagazine.cs
```diff
@@ -40,4 +43,15 @@ public override String Examine(Vector3 pos)
...
 		return $"It seems to be compatible with energy weapons. The charge indicator displays {Math.Round(percent)} percent.";
 	}
+
+	#if UNITY_EDITOR
+	[CustomEditor(typeof(MagazineBehaviour), true)]
+	public class MagEditor : Editor
+	{
+		public override void OnInspectorGUI()
+		{
+			DrawDefaultInspector();
```
corp-0: I think this is not needed if all we're doing is drawing the normal inspector view

AutumnNova: as far as im aware this is the only method of overriding the inspector scripts attached to magazinebehaviour.cs

corp-0: oh, I see. The super class modified the inspector and these children classes need the normal inspector back.

UnityProject/Assets/Scripts/Items/Weapons/MagazineBehaviour.cs
```diff
@@ -270,6 +273,21 @@ public virtual String Examine(Vector3 pos)
...
+	[CustomEditor(typeof(MagazineBehaviour), true)]
+	public class MagEditor : Editor
+	{
+		public override void OnInspectorGUI()
+		{
+			DrawDefaultInspector(); // for other non-HideInInspector fields
+
+			MagazineBehaviour script = (MagazineBehaviour)target;
+
+			script.isClip = EditorGUILayout.Toggle("isClip", script.isClip);
```
corp-0: what is this for?

AutumnNova: this makes the hidden field isClip show, then the child scripts reset the inspector to default hiding that field as using it in them would break some things

corp-0: I feel like using ``ShowIf`` from NaughtyAttributes would be cleaner. 

Another way to see it is, why do the children classes need this field if they won't use it? You could delete from the super class and only add it to the children classes that make use of it.

AutumnNova: ive never managed to get showif to work
the superclass is the only thing to make use of isClip

corp-0:
```cs
[HideInInspector]
public bool isClip = false;
```

Where do you change the value of this? Why is it public?

AutumnNova: in the inspector.
it is public because gun.cs needs reference to it

AutumnNova: > I feel like using `ShowIf` from NaughtyAttributes would be cleaner.
> 
> Another way to see it is, why do the children classes need this field if they won't use it? You could delete from the super class and only add it to the children classes that make use of it.

isCell is used only by the eletrical magazine script, however moving it to the child script would result in a large quantity of unavoidable copy pasting, which is why it is in the super script in the first place. to avoid code duplication. the bool itself is atuomatically toggled by the child script. to avoid accidentally toggling it, which would break some behaviour.

corp-0: Can a weapon be cell and clip? If they can’t, then this shouldn’t be two different checkbox but a drop down with both options

AutumnNova: > Can a weapon be cell and clip? If they can’t, then this shouldn’t be two different checkbox but a drop down with both options

aight, gimme a bit to implement that then

UnityProject/Assets/Scripts/Items/Weapons/ElectricalMagazine.cs
```diff
@@ -16,7 +16,7 @@ public class ElectricalMagazine : MagazineBehaviour
 
 	public void Awake()
 	{
-		isCell = true;
+		magType == MagType.Cell;
```
corp-0: you did a comparison without if

UnityProject/Assets/Scripts/Items/Weapons/MagazineBehaviour.cs
```diff
@@ -38,12 +38,9 @@ public class MagazineBehaviour : NetworkBehaviour, IServerSpawn, IExaminable, IC
...
 
 	/// <summary>
-	///	Whether this can be used to reload other (internal or external) magazines.
+	///	The type of magazine. This effects various behaviours depending on its setting
 	/// </summary>
-	[HideInInspector]
-	public bool isClip = false;
-	[NonSerialized]
-	public bool isCell = false;
+	public MagType magType;
```
corp-0, line 43: make it private with a public read only. Otherwise other scripts are able to change your magtype which I assume is not intended.

UnityProject/Assets/Scripts/Items/Weapons/ElectricalMagazine.cs
AutumnNova: this was fixed in the very next commit

UnityProject/Assets/Scripts/Items/Weapons/MagazineBehaviour.cs
AutumnNova: its only public because eletrical magazine forces it to be a cell.
dont think i would be able to do that otherwise

UnityProject/Assets/Scripts/Items/Weapons/ElectricalMagazine.cs
corp-0: oops

UnityProject/Assets/Scripts/Items/Weapons/MagazineBehaviour.cs
corp-0: Uhmmm, do you get it outside of the class/children ? If not, it could be ``protected``

AutumnNova: gun.cs needs to get it, as mentioned previously

PR #5718, opened 2020-12-08
Fixes rounds being unplayable if any object runtimes under OnSpawnServer()

Aranclanos:
### Purpose
Fixes rounds being unplayable if any object runtimes under OnSpawnServer()


Bod9001:
hum yeah I think it's worth the performance hit

PR #5733, opened 2020-12-13
Bread Teleportation

Emma3DSS:
### Purpose
Make Bread turn into monsters at 20% chance if they go through a teleporter

### Notes:
I have done nothing but teleport bread for the last 24 hours


UnityProject/Assets/Scripts/Objects/Machines/QuantumPad.cs
```diff
@@ -96,6 +101,8 @@ private void Start()
 			{
 				LavaLandManager.Instance.LavaLandBase2Connector = this;
 			}
+
+			//int randomNumber = Random.Range(1, 5);
```
Bod9001: looks like you forgot to make it random, you can also use item tray so don't have to have a to make a class

Mute-Man: was intentionally set as unrandom for testing, since it currently isn't working (he PR'd because he can't test on his computer)
should've noted it in the commenting out of the code though

ThatDan123:
Use item traits instead of using a component as a identifier

UnityProject/Assets/Scripts/Objects/Machines/QuantumPad.cs
```diff
@@ -184,6 +191,12 @@ public void ServerDetectObjectsOnTile()
 			{
 				TransportUtility.TransportObjectAndPulled(item, travelCoord);
 				somethingTeleported = true;
+
+				if(item.GetComponent<Bread>() == true)
+				{
+					Instantiate(MutatedBread, travelCoord, Quaternion.identity);
```
ThatDan123: Dont use instantiate, use Spawn.ServerPrefab()

UnityProject/Assets/Scripts/Objects/Machines/QuantumPad.cs
```diff
@@ -184,6 +191,12 @@ public void ServerDetectObjectsOnTile()
 			{
 				TransportUtility.TransportObjectAndPulled(item, travelCoord);
 				somethingTeleported = true;
+
+				if(item.GetComponent<Bread>() == true)
```
ThatDan123: Use item traits instead, example:

if(Validations.HasItemTrait(this.gameObject, CommonTraits.Instance.SpillOnThrow))
{
}

you'll need to create the new itemtrait, theyre kept in resources/scriptableobjects/traits I think

and then assign the item trait to the bread prefab ItemAttributesV2 component

UnityProject/Assets/Scripts/Items/Food/Bread.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class Bread : MonoBehaviour
```
ThatDan123: Remove this after converting to using item traits

ThatDan123 (changes requested):
Please switch to using item traits rather than the bread monobehaviour

ThatDan123:
Either that or switch the bread script to a general script that can be called for when an object/item is teleported, taking the hard coded bread logic out of the quantum pad script.

That would probably be a better idea ^

UnityProject/Assets/Scripts/Objects/Machines/QuantumPad.cs
```diff
@@ -184,6 +187,15 @@ public void ServerDetectObjectsOnTile()
 			{
 				TransportUtility.TransportObjectAndPulled(item, travelCoord);
 				somethingTeleported = true;
+
+				if (item.TryGetComponent<Bread>(out _) && DMMath.Prob(25))
+				{
+					Spawn.ServerPrefab(MutatedBread, travelCoord);
+					Debug.Log("Mutant Bread Created");
```
ThatDan123: Also remove this log as its not needed.

Emma3DSS:
> Either that or switch the bread script to a general script that can be called for when an object/item is teleported, taking the hard coded bread logic out of the quantum pad script.
> 
> That would probably be a better idea ^

This is actually probably not the best solution since, this is meant to be a very specific easter-egg interaction and as a result isn't meant to be anything modular.

It needs to be bread specific, whether in it's component or ItemTraits as if you make it so it is simply detecting a 'food component' you would then need to add extra parameters to detect exactly what food it is, aswell as make the percentage a public or serializable integer that can be modified.

This interaction is to my knowledge supposed to be a one off, in the codebases this is present, namely BeeStation it is a one off, and I believe as a result of this, any overcomplication of the script as you have suggested would be useless.

Emma3DSS:
Yes for Bread Compression, & recycling related reasons I please request that the bread MonoBehavior be allowed to remain as I will very likely use it in the near future in combination with the addition of GoonStation/AuStation's Bread Compressor/Loafer to introduce a new fun mechanic into the game that revolves around compressing every object you can find into more and more compact bread each with a different tier that can greatly effect it's uses.
image

Bod9001:
you can have the bread script and item tray, having the item tray allows you to have different bread scripts

JukeJukestar:
Isnt 20% a lot? considering 2 people can consistently spam at either side of a quantumpad with just 20 bread on it we would have an invincible bread army in no time

NoooneyDude:
You could probably put that 20 bread into a locker, drag the locker over the pad and quickly open it before teleportation occurs - no spamming at all!

Or perhaps the bread would stack?

UnityProject/Assets/Scripts/Objects/Machines/QuantumPad.cs
```diff
@@ -184,6 +187,14 @@ public void ServerDetectObjectsOnTile()
 			{
 				TransportUtility.TransportObjectAndPulled(item, travelCoord);
 				somethingTeleported = true;
+
+				if (item.TryGetComponent<Bread>(out _) && DMMath.Prob(25))
+				{
+					Spawn.ServerPrefab(MutatedBread, travelCoord);
+					Destroy(item);
```
NoooneyDude:
```suggestion
					Despawn.ServerSingle(item);
```
Similar to `Spawn.ServerPrefab()` we have `Despawn.ServerSingle()`. These are networked and pooled.

UnityProject/Assets/Scripts/Objects/Machines/QuantumPad.cs
```diff
@@ -32,6 +33,8 @@ public class QuantumPad : NetworkBehaviour, ICheckedInteractable<HandApply>
 
 		private RegisterTile registerTile;
 
+		public GameObject MutatedBread;
```
NoooneyDude:
```suggestion
		[SerializeField]
		private GameObject MutatedBread = default;
```
Just keeps the prefab within the scope of QuantumPad.

UnityProject/Assets/Scripts/Items/Food/Bread.cs
```diff
@@ -0,0 +1,34 @@
...
+	private GameObject MutatedBread = default;
+
+	public void OnTeleportStart()
+	{
+		//Left Blank on Purpose
+	}
+
+	public void OnTeleportEnd()
+	{
+		if (DMMath.Prob(10))
```
corp-0, line 24: that's not how you type 5%

ThatDan123 (approved):

PR #5737, opened 2020-12-14
Addressable packing project additions and more converted sounds.

PetMudstone:
### Purpose
- Replaces more sounds in prefabs and scripts.
- Adds many sounds the the addressable packing project. Mainly gun sounds.
- Makes addressable sounds be set to "null" by default.


UnityProject/Assets/Scripts/Core/Addressables/SingletonSOSounds.cs
```diff
@@ -23,6 +23,9 @@ public class SingletonSOSounds : SingletonScriptableObject<SingletonSOSounds>
    public AddressableAudioSource ExplosionCreak3 = null;
    public AddressableAudioSource Notice1 = null;
    public AddressableAudioSource Notice2 = null;
+   
+   public AddressableAudioSource GunEmptyAlarm = null;
+   public AddressableAudioSource KineticReload = null;
```
NoooneyDude, line 28: Curious why `KineticReload` is in the singleton? Is it used more than just PKA?

PetMudstone: It's also used in the kinetic crusher, though admittedly that's only one other item.

NoooneyDude: Alright.

PR #5741, opened 2020-12-14
Some addressable sound conversion.

PetMudstone:
### Purpose
- Converts a large amount of .prefabs and scripts to use addressable sounds. **Note: does not affect all scripts that need to be converted, several scripts and object still require conversion.** This exists mainly so that we don't have any continued merge conflicts. Credit to Borzoi for converting most scripts beforehand.


PetMudstone:
Fixed up the moving microwaves on maps and fixed the references to microwaves and virtual containers in objects. However, I still receive this error in local tests.
Unhandled log message: '[Exception] UnityException: FindObjectsOfTypeAll is not allowed to be called from a MonoBehaviour constructor (or instance field initializer), call it in Awake or Start instead. Called from MonoBehaviour 'CentComm'.
See "Script Serialization" page in the Unity Manual for further details.'. Use UnityEngine.TestTools.LogAssert.Expect

corp-0:
changes on all maps D:

PR #5751, opened 2020-12-15
resizes pda ui and adds missing colors to assistant pda

AutumnNova:
yep, thats it.


corp-0:
can you screenshot the final look?

corp-0:
nvm, it is aight. TC counter is clipping a little bit, but it is fine for now.

PR #5753, opened 2020-12-15
Another How-To-Map Format Fix Attempt

Akbadain29:
### Purpose
Ok, I've tried different things to see what sticks for the formatting in the How-To-Map page. Fingers crossed one of these work in the actual doc, it's working when I view the preview normally.


corp-0:
yeah, html tags should work fine on the webiste. Let's test.

PR #5755, opened 2020-12-16
adds maint spawns and maint loot

AutumnNova:
also some other stuff like cargo shutters being closed


NoooneyDude:
Tests fail on:
```
FallStation: "APC" has a null value in the list.
FallStation: "APC" has a null value in the list.
FallStation: "Shutters (1)" is not assigned to "APC"
```

PR #5756, opened 2020-12-16
allows linuxserver target and cleanup

corp-0:
### Purpose
This PR will handle a "custom target" called ``linuxserver`` so we can differentitate server and player builds and only assign ``dev build`` to the server. 

Linux players have had to download a bigger file and also deal with the console spewing errors on top of the game, so this is trying to solve that.

### Notes
I also deleted all ``[Obsolete]`` methods in the file.


corp-0:
uhm, removing that "useless declaration" codacy is complaining about will make the code to not compile.

corp-0:
shameless self merge. I'm working on auto builds

PR #5761, opened 2020-12-18
OOC Http/https links

ThatDan123:
Allows admins (or all players if the config is set to allow them) to post clickable http/https links in OOC


corp-0:
what's the default for players, false?

corp-0:
why are the spaces needed now btw? It wasn't a problem in old ui

ThatDan123:
Had to change the chat front sadly which messed up the spacing for bold characters.

Tried to convert the old font to a TMP font but it wouldnt work.

ThatDan123:
And yeah, its admin only by default

PR #5762, opened 2020-12-19
Add custom flag for dev builds

corp-0:
### Purpose
My [previous atttempt](https://github.com/unitystation/unitystation/pull/5756) on differentiating linux server and clients build didn't work. This time I further cleaned up unused stuff on build script and added a custom flag. I don't know if this will work, but I don't see a lot of the flags used in our script in [here](https://docs.unity3d.com/Manual/CommandLineArguments.html) so I'm assuming they are all custom made and only parsed on our build script.


corp-0:
shameless self merge. I want to test if this is working now.

PR #5764, opened 2020-12-19
Fixes Closet, Freezer and Leaper Sprites

Akbadain29:
### Purpose
Should fix closets and freezers to make their closed door sprites reappear. I've also tried making Leaper reappear too and fixed up TestStation, some tables had gone missing.

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Items/Electricity/Medium machine connector.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Items/Electricity/Medium machine connector.prefab
@@ -46,6 +46,7 @@ SpriteRenderer:
   m_LightProbeUsage: 1
   m_ReflectionProbeUsage: 1
   m_RayTracingMode: 0
+  m_RayTraceProcedural: 0
   m_RenderingLayerMask: 4294967295
   m_RendererPriority: 0
   m_Materials:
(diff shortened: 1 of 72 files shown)


corp-0:
can you describe what you changed on the prefabs?

Akbadain29:
While the door open sprite was fine for the closets etc. as there is a reference to it in the Closet Control component, but on their Door Layer inside the prefab there was a missing reference, all that was need to be done was to update the current present field inside the SpriteHandler component to restore the closed door sprite.

Leaper was missing DirectionalSprite V2 and SpriteHandler components. Hopefully with those added, it will be visible again, I could see it when I spawned it in.

corp-0:
what about crates? I read some crates are invisible too. Do you have any idea on what could have caused the missing ref?

Akbadain29:
I'll have another look for those. It's probably the same thing.

PerfectTangent:
crate sprites are fucked. so is the reactor, turbine, boiler, water pump, the power transformer, and I believe one of the pipe colors are missing as well. There's probably more stuff, too.

Akbadain29:
These crate sprites look fine to me, nothing invisible there. Reactor sprites should be now fixed up.

corp-0:
Thanks a lot. If you don't mind, I will leave this PR open for a while, to see if we get more reports about invisible sprites.

Akbadain29:
image
The problem seems to be with however Unity is dealing with Sprites (Texture 2Ds) and how it loads (or rather doesn't load) into the SpriteHandler, converting objects to be a single sprite inside a Sprite Data SO seems to fix it.

Akbadain29:
Ok, should we be making sure that sprites are called from a Sprite Data SO and not a Texture 2D like the one above is?

NoooneyDude:
Yep, SO is preferred.

corp-0:
Brian, take a look at vending machines? Slucc said there are some invi ones

Akbadain29:
Slug said the vending machines were missing on Fall, but I've had a look and none seem to be missing their sprites.

NoooneyDude:
Good to merge?

Edit: I see scene changes so will merge earlier. Thank you.

PR #5777, opened 2020-12-24
BlobNet

ThatDan123:
Adds a blob net so that all blob structures need to be connected in some way to the blob core to function.


UnityProject/Assets/Scripts/Systems/Antagonists/Antags/Blob/BlobStructure.cs
```diff
@@ -48,9 +47,22 @@ public class BlobStructure : MonoBehaviour
...
+		[HideInInspector]
+		public List<Vector3Int> connectedPath = new List<Vector3Int>();
+
+		[HideInInspector]
+		public LineRenderer lineRenderer;
+
 		private void OnEnable()
 		{
 			integrity = GetComponent<Integrity>();
+			lineRenderer = GetComponent<LineRenderer>();
```
corp-0, line 65: any reason to cache the components OnEnable instead of on Awake like most scripts do?

ThatDan123: Dont think so, ive changed it over to use awake instead

corp-0:
let's give it a try.

waffielz:
aw yea

PR #5783, opened 2020-12-28
SeedExtractor Partial Fix to Dropping Seed Packets on Deconstruction

Akbadain29:
### Purpose
Partially fixes the SeedExtractor so now it will drop the seed packets upon deconstruction. Except when I mean seed packets I mean the variable inside SeedExtractor.cs called seedPackets and its contents being dropped on despawn. Also, the SeedExtractor only can drop produce that hasn't had the seeds extracted yet (i.e. when the device is not powered), any successful seed packets that get extracted don't get dropped.

Also adds SeedExtractor and MegaseedServitor to Botany and a cremator to the Morgue on TestStation.

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Botany/SeedExtractor.cs
+++ b/UnityProject/Assets/Scripts/Objects/Botany/SeedExtractor.cs
@@ -122,7 +122,8 @@ namespace Objects.Botany
 		[Server]
 		public void ServerPerformInteraction(HandApply interaction)
 		{
-			if (interaction.HandObject != null) {
+			if (interaction.HandObject != null)
+			{
 				var grownFood = interaction.HandObject.GetComponent<GrownFood>();
 			
 				var foodAtributes = grownFood.GetComponentInParent<ItemAttributesV2>();
@@ -150,7 +151,10 @@ namespace Objects.Botany
 
 		public bool WillInteract(HandApply interaction, NetworkSide side)
 		{
-			return DefaultWillInteract.Default(interaction, side);
+			return DefaultWillInteract.Default(interaction, side) &&
+			   interaction.TargetObject == gameObject &&
+			   interaction.HandObject != null &&
+			   interaction.HandObject.TryGetComponent<GrownFood>(out _);
 		}
 
 		/// <summary>
--- a/UnityProject/Assets/Scripts/Objects/Botany/SeedExtractor.cs
+++ b/UnityProject/Assets/Scripts/Objects/Botany/SeedExtractor.cs
@@ -125,7 +125,8 @@ namespace Objects.Botany
 			if (interaction.HandObject != null)
 			{
 				var grownFood = interaction.HandObject.GetComponent<GrownFood>();
-			
+
+
 				var foodAtributes = grownFood.GetComponentInParent<ItemAttributesV2>();
 				if (!Inventory.ServerTransfer(interaction.HandSlot, storage.GetBestSlotFor(interaction.HandObject)))
 				{
--- a/UnityProject/Assets/Scripts/Objects/Botany/SeedExtractor.cs
+++ b/UnityProject/Assets/Scripts/Objects/Botany/SeedExtractor.cs
@@ -97,6 +97,23 @@ namespace Objects.Botany
 		}
 
 		/// <summary>
+		/// Ejects all the seed packets when extractor is deconstructed
+		/// </summary>
+		[Server]
+		public void OnDespawnServer(DespawnInfo info)
+        {
+			Vector3 spawnPos = gameObject.RegisterTile().WorldPositionServer;
+			foreach (var packet in seedPackets)
+            {
+				CustomNetTransform netTransform = packet.GetComponent<CustomNetTransform>();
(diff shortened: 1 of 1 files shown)


corp-0:
indentation looks a bit off

Akbadain29:
Everything looks okay on my end, where does it look off?

corp-0:
visualization of changes in github

Akbadain29:
When I view the file in github it looks like the indentation is the same, but when I just preview it in github I can see formatting problems, no idea why that is happening. My editor is set to insert 4 spaces too.

NoooneyDude:
It seems the `IServerDespawn` interface method is implemented but the class doesn't inherit it, so it doesn't appear to be used.

Perhaps you could swap `IServerSpawn` with `IServerLifecycle`, which inherits `IServerSpawn` and `IServerDespawn`.

corp-0:
Looks like you're using spaces
image

~Uhmm, you might be able to simply add the new seeds to the list after the process is done on line 54.~
Never mind. It is actually being added in line 58, so what you did should work just fine.


I know it is not part of your initial intention but since you're there, can you fix this issue? 
https://github.com/unitystation/unitystation/blob/70d3181620ef96f847732399d0c4333992738e83/UnityProject/Assets/Scripts/Objects/Botany/SeedExtractor.cs#L123

Line 123, ``?`` operator is used and should be a normal null check instead. Using ``?`` can end in unexpected errors.

Line 20 might as well be a SerializedField so we can tweak it from inspector.

Akbadain29:
Ok, I've removed the `?` and serialized the processing time variable.

UnityProject/Assets/Scripts/Objects/Botany/SeedExtractor.cs
```diff
@@ -120,7 +121,7 @@ public void OnDespawnServer(DespawnInfo info)
 		[Server]
 		public void ServerPerformInteraction(HandApply interaction)
 		{
-			var grownFood = interaction.HandObject?.GetComponent<GrownFood>();
+			var grownFood = interaction.HandObject.GetComponent<GrownFood>();
```
corp-0: ``interaction.HandObject`` can be null if the object is interacted with an empty hand. You could avoid checking for it by making ``interaction.HandObject != null`` a condition for this interaction to trigger. You would need to make this ``ICheckedInteractable`` instead of just ``IInteractable``

corp-0: if you have any question on how to do that, just ask. We're happy to help.

Akbadain29: I think I've got it. I just have to test it out, I'll let you know how I go.

UnityProject/Assets/Scripts/Objects/Botany/SeedExtractor.cs
```diff
@@ -121,9 +122,9 @@ public void OnDespawnServer(DespawnInfo info)
 		[Server]
 		public void ServerPerformInteraction(HandApply interaction)
 		{
-			var grownFood = interaction.HandObject.GetComponent<GrownFood>();
-			if (grownFood != null)
-			{
+			if (interaction.HandObject != null) {
```
corp-0: bracket by itself on the next line

UnityProject/Assets/Scripts/Objects/Botany/SeedExtractor.cs
```diff
@@ -147,6 +148,11 @@ public void ServerPerformInteraction(HandApply interaction)
 			networkTab.ServerPerformInteraction(interaction);
 		}
 
+		public bool WillInteract(HandApply interaction, NetworkSide side)
+		{
+			return DefaultWillInteract.Default(interaction, side);
```
corp-0:
```suggestion
    return DefaultWillInteract.Default(interaction, side) &&
               interaction.TargetObject == gameObject &&
               interaction.HandObject != null &&
               interaction.HandObject.TryGetComponent<GrownFood>(out _);
```

don't directly apply the suggestion, I had to use spaces and some names might not be right.

Akbadain29:
Ok, it looks like your suggestion for WIllInteract works fine and didn't have any compiling errors.

corp-0:
> Ok, it looks like your suggestion for WIllInteract works fine and didn't have any compiling errors.

:surprisedpikachu:

Akbadain29:
Do you mind if I add in the lobby music onto this PR too? I don't like how quiet the lobby sounds on staging when I open it up.

corp-0:
sounds like a different PR to me

Akbadain29:
Alright I'll do it in the new one when this one is done.

corp-0:
I can still see spaces used instead of tabs :eyes: 
image

Akbadain29:
Ok, if that doesn't fix it I don't know what will.

Akbadain29:
image
Here is my settings so I think we are good now.

Akbadain29:
Ok OnDespawnServer has been dealt with too

PR #5784, opened 2020-12-28
Fix Centcom announcement sounds so they use addressables

corp-0:
### Purpose
Rearranged Centcom Announcement sounds so the dictionary where they were declared is inside an Awake method and thus making it viable to be used with addressables.

### Notes:
While I was on that, I moved all strings outside of the class and made static classes to hold them. In the future we might want to use the Strings namespace and its classes to build all kind of strings in the game in a unified place.


UnityProject/Assets/Scripts/Strings/ReportTemplates.cs
```diff
@@ -0,0 +1,37 @@
+﻿namespace Strings
+{
+	public static class ReportTemplates
+	{
+		public static readonly string CentcomReport = "<size=40><b>CentComm Report</b></size> " +
```
Bod9001, line 5: at some point we could drag this out to a config

PR #5785, opened 2020-12-28
fix intent keybinds

AutumnNova:
fixes intent keybinds
changes the default keybind for run back to c
changes the default keybind for quick interactions with pockets to be shift+1-3


corp-0:
It certainly works but now that I'm testing it, I feel using shift + pocket slot is a little too awkward to perform given the functionality of the shortcut (which works btw, I tested).

We might need to find a better key combo

PR #4524, opened 2020-07-02
Adds disposal stuff! Pipes, machines, pipe dispenser

NoooneyDude:
Adds disposal pipes, disposal machines and the pipe dispenser. Bored atmospheric technicians, rejoice! Fixes #2398 and fixes #2808.

Disposals

### Additions
* Adds disposal pipe objects, tiles (straight, bent, terminating, merging, splitting)
* Adds disposal bins, intakes, outlets
* Adds disposal pipe traversal logic
* Adds disposal manager to handle traversals
* Adds pipe dispenser, to dispense atmospherics and disposals components

> Disposal bin has a UI:
Disposal Bin
> Pipe dispenser has a UI:
Pipe Dispenser
> Disposal pipes and machines are mappable and de/constructable.

## Extras
* Tiles can now implement the IExaminable interface - useful for producing examine text from tile state logic.
* Tile interactions will now check each tile in the Underfloor layer, instead of only the first underfloor, for interactions.
* NetLabel NetUI component has new event for when the text changes.
* New DirectionalPassable component. Allows side passable checks to be set individually per side and will update with Directional rotation. Inherits from RegisterTile, so works best with just DirectionalPassable inheritor component on the object.

### Known issues
See issue #4654.

PR #5786, opened 2020-12-28
Comply with requested changes in #4524

NoooneyDude:
- Adds the requested changes in PR #4524. Partially fixes part 16. of #4654.
- Small tweaks to conform to project standards.

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Disposals/DisposalBin.cs
+++ b/UnityProject/Assets/Scripts/Objects/Disposals/DisposalBin.cs
@@ -35,28 +35,33 @@ namespace Objects.Disposals
 	[ExecuteInEditMode]
 	public class DisposalBin : DisposalMachine, IServerDespawn, IExaminable, ICheckedInteractable<MouseDrop>
 	{
-		const int CHARGED_PRESSURE = 600; // kPa
-		const int AUTO_FLUSH_DELAY = 2;
-		const float ANIMATION_TIME = 1.3f; // As per sprite sheet JSON file.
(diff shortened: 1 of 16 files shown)


UnityProject/Assets/Scripts/Objects/Disposals/DisposalBin.cs
```diff
@@ -35,28 +35,33 @@ public enum BinOverlaySprite
 	[ExecuteInEditMode]
 	public class DisposalBin : DisposalMachine, IServerDespawn, IExaminable, ICheckedInteractable<MouseDrop>
 	{
-		const int CHARGED_PRESSURE = 600; // kPa
-		const int AUTO_FLUSH_DELAY = 2;
-		const float ANIMATION_TIME = 1.3f; // As per sprite sheet JSON file.
+		private readonly int CHARGED_PRESSURE = 600; // kPa
```
corp-0: why can't these be configurable on inspector?

NoooneyDude: Hmm, I tend to use constants when the value, once found, shouldn't be changed unless the code changes too. Though, some of these probably could be serialized.

corp-0: in that case, make it ``const`` or was codacy complaining? It should only complain about publics const that are used in other classes.

NoooneyDude: You're right. Will fix.

PR #5787, opened 2020-12-28
An ingenious and incredibly clever way to solve an impossible problem

SluggyWuggies:
adds a single tile to outpost


corp-0:
lmao

PR #5798, opened 2020-12-31
More JoinedViewer logging

NoooneyDude:
- Adds a little more logging for JoinedViewer.
- Moves client job request out of JoinedViewer.
- Fixes wizard apprentice setup not completing.
- Fixes incorrect name posted to chat of the miner mining wizard rocks.

### DIFF
--- a/UnityProject/Assets/Scripts/Messages/Server/JoinedViewer/JobRequestFailedMessage.cs
+++ b/UnityProject/Assets/Scripts/Messages/Server/JoinedViewer/JobRequestFailedMessage.cs
@@ -1,0 +1,38 @@
+﻿using System.Collections;
+using UnityEngine;
+using Mirror;
+
+namespace Messages.Server
+{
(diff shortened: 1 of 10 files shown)


UnityProject/Assets/Scripts/Messages/Server/JoinedViewer/JobRequestFailedMessage.cs
```diff
@@ -0,0 +1,52 @@
...
+			var msg = new JobRequestFailedMessage
+			{
+				FailReason = failReason,
+			};
+
+			msg.SendTo(recipient);
+			return msg;
+		}
+
+		public override void Deserialize(NetworkReader reader)
```
Bod9001: I don't think you need these?
I think it's only needed when you doing custom serialisation

NoooneyDude: You're right. Fixed.

PR #5810, opened 2021-01-04
Modifies Documentation

Akbadain29:
### Purpose
Changes the Untiy version in the contribute guide from Unity 2020.1.6f1 to 2020.1.17f1 and messes from the bullet point formating on How-to-map, again, because I thought if I replace the html bullet tags with the four space dashes it would work, but it doesn't.


corp-0:
thank you, don't mind the pr helper.

PR #5825, opened 2021-01-06
More random events

NoooneyDude:
Adds more random events.

- ~~Bureaucratic Error: removes the max occupation cap on a random crewmember occupation (not that we don't already get this!
: )~~
- Communications Blackout: all radio channels go silent for a random duration.
- Electrical Storm: all lights in a random region on the main station get damaged and will need replacement.
- Powernet Check: each component in the station powernet will go offline for a random duration.
- Processor Overload: messages on all radio channels will become garbled for a random duration.
- Scrubber Surge: all scrubbers will eject a random reagent (will be more interesting when foaming agents work).

CL:Added a bunch more random events.

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/SceneConstruction/NestedManagers/InGameEventsManager.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/SceneConstruction/NestedManagers/InGameEventsManager.prefab
@@ -352,6 +352,58 @@ MonoBehaviour:
   maxStrength: 1500
   minTimeBetweenMeteors: 1
   maxTimeBetweenMeteors: 10
+--- !u!1 &2961505858790993329
+GameObject:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  serializedVersion: 6
+  m_Component:
+  - component: {fileID: 8107459317429369441}
+  - component: {fileID: 2610571774784088097}
+  m_Layer: 0
+  m_Name: PowernetCheck
+  m_TagString: Untagged
+  m_Icon: {fileID: 0}
+  m_NavMeshLayer: 0
+  m_StaticEditorFlags: 0
+  m_IsActive: 1
+--- !u!4 &8107459317429369441
+Transform:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 2961505858790993329}
+  m_LocalRotation: {x: 0, y: 0, z: 0, w: 1}
+  m_LocalPosition: {x: 0, y: 0, z: 0}
+  m_LocalScale: {x: 1, y: 1, z: 1}
+  m_Children: []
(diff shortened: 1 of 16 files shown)


corp-0:
I love you

UnityProject/Assets/Scripts/Systems/InGameEvents/InGameEventScripts/EventBureaucraticError.cs
```diff
@@ -0,0 +1,31 @@
...
+			}
+
+			if (FakeEvent) return;
+
+			base.OnEventStart();
+		}
+
+		public override void OnEventStartTimed()
+		{
+			OccupationList.Instance.Occupations.Where(o => o.IsCrewmember).PickRandom().SetLimit(-1);
```
corp-0: could also choose another role to reduce the limit, maybe

NoooneyDude: How do you mean?

corp-0: well, descriptions says "shortage on others". Does ss13 just give a single role unlimited slots?

NoooneyDude: Oh I getcha. https://tgstation13.org/wiki/Random_events#Bureaucratic_Error seems to imply one role. I can change it, or the text, if you think it'll be better.

NoooneyDude: The text also seems to imply that there can be only one "overflow role", though I've seen situations where there have been more than one.

corp-0: uhhhm, wiki says it shifts the unlimited role so you should give assistants a limit.

UnityProject/Assets/Scripts/Systems/Occupations/Occupation.cs
```diff
@@ -126,4 +126,12 @@ public enum OccupationDifficulty
 	[SerializeField] private bool isTargeteable=true;
 
 	public bool IsTargeteable => isTargeteable;
+
+	/// <summary>
+	/// Set the maximum simultaneous players with this occupation. Set to -1 for unlimited.
+	/// </summary>
+	public void SetLimit(int newLimit)
+	{
+		limit = newLimit;
```
corp-0: just realized something. This won't work as expected because changes to scriptable objects are persistent and will persists through game sessions.

NoooneyDude: Ah. Can't have that.

corp-0: a way around this problem is to load values from the scriptable object into whatever class that uses it and only then modify as you please, the value from the class not the scriptable object.

NoooneyDude: That won't be terribly trivial. I'll take a look at it tomorrow.

NoooneyDude:
I've removed the bureaucratic event for now.

corp-0 (approved):

PR #5827, opened 2021-01-06
Examination window steal interaction

### LINKED ISSUE #5809: Make new UI playable: steal objects using the examination window
While using the new UI, the examination window that appears when examining other players should be used to steal/loot objects from their inventory.

To consider this issue solved, the solution should consider range at which players can steal objects, availability of the action (are they cuffed? Are they missing their arms?), progress bar for the action to be completed.

Aranclanos:
### Purpose
Adds the interaction to steal and apply items for the examination window, with a progress bar and local messages.
Fixes #5809

application takes 1 second and the stealing takes 3 seconds

### DIFF
--- a/UnityProject/Assets/Scripts/Messages/Client/OtherPlayerSlotTransferMessage.cs
+++ b/UnityProject/Assets/Scripts/Messages/Client/OtherPlayerSlotTransferMessage.cs
@@ -1,0 +1,101 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using Messages.Client;
+using UnityEngine;
+public class OtherPlayerSlotTransferMessage : ClientMessage
+{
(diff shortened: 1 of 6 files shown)


corp-0:
> application takes 1 second

what's application? Putting items?

Aranclanos:
yeah

corp-0:
I will test this shortly. Is the taking/putting items range different than the examination window range?

Aranclanos:
yup, theres a CanApply() around there in both client and server

corp-0:
I think you need to check again range after the progress bar is completed. I could start stealing in range then alt+tabed to the other account and moved away. I still got my glasses stolen like 6 tiles away.

Aranclanos:
thanks, fixed

PR #5838, opened 2021-01-08
Add Radiation Storm event

NoooneyDude:
- Adds the Radiation Storm event. Spawns a bunch of dangerous radiation sources around the station that the crew must avoid.
- Adds the Fugitives event. Creates a ghost role instance where ghosts can choose to become a fugitive. Removes fugitive from preferences.

CL:Added the Radiation Storm and Fugitives event.


NoooneyDude:
Want to test this on staging today. Merging before tests complete (not much to go wrong I think).

PR #5857, opened 2021-01-11
Fixes #5801, vendor content saved to restock canister after deconstructing

### LINKED ISSUE #5801: Infinite vendomat refill
Step 1:

1) Go to a vending machine, on fallstation.
2) buy whatever
2) Use the screwdriver on the machine twice, once to unscrew and once to screw it back
3) Notice the vending machine is fully stocked.

*note: on some other stations (square) some vending machines cannot be dissasembled

correct behaviour: consumed items items should be stored in the canister

jackpotttt:
### Purpose
Fixes #5801

### Notes:
When a vendor is deconstructed, its items are saved to the vendor restock item and then reloaded if the machine is rebuilt with that restock item. If the restock item hasn't been spawned in, the items are temporarily saved in the machineframe until the restock item is spawned

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Others/VendingRestock.cs
+++ b/UnityProject/Assets/Scripts/Items/Others/VendingRestock.cs
@@ -1,9 +1,18 @@
-﻿using System.Collections;
+﻿using Objects;
+using System.Collections;
 using System.Collections.Generic;
 using UnityEngine;
 
-//TODO: No point to this. Delete it and add Pickupable to VendingRestock, use ItemTraits instead of this marker component
 public class VendingRestock : Pickupable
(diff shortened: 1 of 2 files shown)


corp-0:
>If the restock item hasn't been spawned in

Under what conditions does this happen?

jackpotttt:
> Under what conditions does this happen?

When first deconstructing the vending machine with a screwdriver. 

All of the vending machines I've tested did not contain any existing parts by default. They're only created when you use a crowbar on the machine frame to remove them

corp-0:
I see. I feel like we will need to refactor how that works in the future, for the now this is a good solution. Thanks.

Issue #5832, opened 2021-01-08
Space Exterior Spawns Added to Maps


Mute-Man:
### Description
A new "Space Exterior" spawnpoint was added in #5825 to trigger spawns on ingame events, for spacecarps, xenos, and the like.
Since it is an entirely new spawnpoint, at least one needs to be added to every map.

* [x] Add "Space Exterior" spawnpoint to Square Station
* [x] Add "Space Exterior" spawnpoint to Fall Station
* [x] Add "Space Exterior" spawnpoint to Outpost Station
* [x] Add "Space Exterior" spawnpoint to Box Station
* [x] Add "Space Exterior" spawnpoint to Pog Station
* [x] Notify Mappers working on new Maps/Stations to add "Space Exterior" spawnpoints to them

NoooneyDude:
Related to #4719.

PR #5867, opened 2021-01-15
Spawnpoint prefabs and Box modifications.

### LINKED ISSUE #5582: Missing wire at Box
### Environment
- Game Version: last, build 5538
- OS Version: Windows 10

At coordinates -191, 45 (medbay) if I am not mistaken missing low voltage wire so part of the medbay is not powere at all
<!-- What is broken? What should be happening instead? Include as much informations as possible. Images, gifs, or youtube videos are welcome! --!>

PetMudstone:
### Purpose
- Adds individual prefabs for all spawnpoints.
- Adjusts spawnpoints on Boxstation, adds many maint and external space spawnpoints. Checks Boxstation tickbox in #5832
- Adds a missing wire in Boxstation's medical, resolves #5582.

### Notes
- All spawnpoints are invisible by default in the scene view and always invisible during gameplay. However, if you turn on Gizmos in the inspector you will see textures appropriate to the kind of spawnpoint you're using. For example, a spawnpoint that's set to spawn assistants will show the sprite of an assistant.


ThatDan123:
You can set the spawnpoint layer to Ghost, and then only ghosts will see them.

Though not sure that outside the editor they will need to be seen so they could just have a gizmo picture instead.

corp-0:
I think the gizmo image is cool but it only shows when the component generating the gizmo is expanded (not collapsed). Another option is implementing ``IServerSpawn`` and make the sprite of ``SpriteRenderer`` null.

corp-0:
No wait, that wouldn't work because the change is not networked and using resources to network that change sounds like a big waste. Let's go with the gizmo solution.

corp-0:
oh no, a conflict

corp-0:
update the pr description since the note suggests spawnpoints are visible during gameplay

PR #5870, opened 2021-01-16
Add slight delay between item selection in radial.

Copern:
Getting back into things. A suggestion relayed to me was adding a brief delay to item selection on the radial.
If an item is currently selected in the radial menu, there will be a brief delay before another item is selected if you hover over it.
This gives you a brief window in which to move the mouse from a selected item over to the actions without accidentally selecting another item.
This only affects the item ring, the action ring is unaffected by this change.


corp-0:
GIF 16-01-2021 3-50-20
Works flawleslly, thank you very much

PR #5877, opened 2021-01-20
Readd self-examination and add wounds descriptions

corp-0:
### Purpose
Readd self-examination interaction. It will return the classic description you found when examininig others as seen on this recent screenshot from tg:
unknown

Adds wounds descriptions for each body part to the new expanded view of the examination window. The numbers that determine the message might need a little tweaking in the future.

Adds a short one word status description to be seen on the status part of the new examination window.

### Notes:
I added two new extensions. 
Capitalize() is a string extension that will, welp, capitalize the word so the first letter of the string is Upcase.
IsBetween() is a IComparable extension that will return true if the IComparable is between the given range. This works for all IComparables, like numbers (of any type) and dates.

### Changelog:
CL: Readds self-examination
CL: Adds new wounds description per body part visible on the new extended view of the examination window.

### DIFF
--- a/UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
+++ b/UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
@@ -2,12 +2,14 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 using Systems.Atmospherics;
 using Light2D;
 using UnityEngine;
 using UnityEngine.Events;
 using Mirror;
 using UnityEngine.Profiling;
+using WebSocketSharp;
 
 /// <summary>
 /// The Required component for all living creatures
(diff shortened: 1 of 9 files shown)


UnityProject/Assets/Scripts/Health/Living/LivingHealthBehaviour.cs
```diff
@@ -1048,13 +1054,49 @@ public string GetExamineText()
...
 		}
 
 		return healthString;
 	}
 
+	public string GetShortStatus()
+	{
+		if (IsDead)
+		{
+			return "DEAD";
```
Bod9001, line 1077: could add extra if for is SSD and dead?

corp-0: Yes I can. I didn't know if we wanted to add information that specific, is it in character information to know how someone is dead and has no soul?

corp-0: done

corp-0:
wait until #5879 is merged because it this will conflict then. I don't mind solving the conflict.

corp-0:
I'm gonna do a gamer move and speed merge this so we can trigger a build sooner

PR #5880, opened 2021-01-21
add ability to choose a pseudonym

corp-0:
### Purpose
This PR will add the ability for certain clothing to hide the real identity of a player and show a pseudonym/artistic name instead. Useful for stuff like miming and clowning.

### How it works
A new string variable called ``Pseudonym`` is added to ``Mind`` object. When the player wears a clothing piece with the ``WearablePseudonym`` component, it will try to retrieve the pseudonym from the character's mind. If no pseudonym is found, it will prompt the player to type in a new one and then store it on mind for the rest of the round. The pseudonym will be the visible name for the player as long as they are wearing the clothing piece with the component.

### Changelog
CL: Adds ability to use a pseudonym/artistic name for clowning or miming.


corp-0:
closing since artistic names are not a popup anymore, not even in ss13.

PR #5886, opened 2021-01-22
Breathing in plasma causes toxin damage, Fixes #5176

### LINKED ISSUE #5176: Breathing plasma gas should cause toxin damage
## Description

When plasma gas is breathed, it should deal toxin damage proportional to the concentration in the air. It's not damage from exposure, so those with internals are safe.

AtnerNT:
Addresses #5176, if atmospheric plasma is above a certain threshold, players will begin to take toxin damage proportional to the amount of plasma in the air they are breathing.


corp-0:
I think this works good while we don't have proper species and different interactions with gases. In the near future, this shouldn't be hardcoded. I will merge, unless @Bod9001 has an objection

PR #5887, opened 2021-01-23
Fixes an oversight from my last PR (Plasma Toxin Damage)

AtnerNT:
An oversight in my plasma-breathing PR meant that in atmospheres containing more than 501kPa of Plasma, the resulting player would suffer no toxin damage from breathing it in. It's one line and an easy fix, just an oversight.


UnityProject/Assets/Scripts/Health/RespiratoryHealth/RespiratorySystem.cs
```diff
@@ -164,6 +164,7 @@ private float HandleBreathing(GasMix gasMix)
 			plasmaDamage = (plasmaPressure - PLASMA_SAFE_MAX) / 5;
 
 			if ((bloodSystem.ToxinLevel += plasmaDamage) <= 100) bloodSystem.ToxinLevel += plasmaDamage;
```
corp-0: use ``{`` and ``}``. Brackets should be on new line by their own.

corp-0: I feel weird doing assignation inside an ``if`` condition. That's me personally though, let me see what other reviewers think of it.

PR #5888, opened 2021-01-24
Fix radial camera reference

Copern:
Tested in unity player by logging in, back out, and back in a few times. Before change I would get an NRE logging back in as the branch held a reference to the camera that was destroyed.


corp-0:
this should fix the nre that made subsequent rounds not trigger the radial menu when right clicking.

PR #5889, opened 2021-01-24
Partially converting closets and crates to use sprite handler. Renaming.

PetMudstone:
### Purpose
- Makes closets and crates mostly use the sprite handler. Weld overlays still don't use it however.
- Renames many of the file names of closets and lockers to make more sense.
- Adds a setting for whether or not you want to hide the lock light when a container is opened. Meant for crates.


corp-0:
some big changes here, please don't merge until thoroughly tested or wait until new update to main servers is released.

PetMudstone:
Can confirm that everything works while connected as client now.

PR #5890, opened 2021-01-24
Squarestation locker/crate/suitstorage fixes

PerfectTangent:
### Purpose
- Fixes **all** closets, crates. and suit storage units with messed up or missing sprites.
- lengthens a couple airlocks that were 1 tile wide to be 2 tiles wide so bodies, canisters, and so on can be dragged through them.
- slightly changes the layout of the cargo crates room so it's not as cramped and adds 3 more crates.
- does _not_ add space exterior spawns. will do that next.

### Notes:
Because I am a dumb, it overwrites #5878. I will readd the plants in the next pr that includes the space exterior spawns.


corp-0:
whoever merges this, please **squash and merge**

PR #5892, opened 2021-01-25
Adds UnRealStation, first unfinished version, with Matrix problems.

Akbadain29:
### Purpose
This is will probably fail the tests, but I've PR'd to deal with the above. Specifically, the problem I'm having is that none of the matrices are lining up properly, which is recking transitioning onto the shuttles and the disposals tubes. Can you see what I've done wrong?


wraith-54321:
new map, shiny

Akbadain29:
I tried messing with the tilemap anchor in the tilemap component inside the layers of the matrix of the mining shuttle. This was to try and remove the offset by adjusting the position of the layers to where the shuttle should be. It wasn't fixed it at all. I've doubled checked what is done on existing maps and their layering is set out the exact same way.

The command console's initial round reports it prints out are ejected with a 1 tile offset, which doesn't make much sense either.

Akbadain29:
Ok, I've moved the matrices around a bit to see if that helps and fixed more of the tests up. I've still got to deal with the missing light tube references, but because they have got double-ups, I'm going to have to go through and relabel so I figure out which ones need fixing. I'll do it that later when I get a chance.

corp-0:
did that fix the offset?

Akbadain29:
It doesn't seem to have done anything, which is a pain.

Akbadain29:
If it's alright with you, once I've fixed up the light references, can it be merged but not included in the rotation until we sort out why it's being a pain?

corp-0:
as long as it doesn't make tests fail, sure.

Akbadain29:
Cool, thanks

Akbadain29:
Some of the lights don't seem to be connected up and some haven't been put in, I'll do that later. All the tests pass now though, so when it's done please merge it.

Akbadain29:
Also the matrix offset is still there, I still have no idea why that is happening.

corp-0:
> UnRealStation: "APC_Hallway_South_2" has a null value in the list.

The only problem you have right now before merging

corp-0:
alright so before I merge, you sure this isn't enabled on rotation?

Akbadain29:
image
Yeah. I ticked that.

corp-0:
I meant the map rotation, since the state of the map is currently broken

Akbadain29:
It's definitely not in map.json

PR #5896, opened 2021-01-26
Improves to symptoms and player feedback

corp-0:
### Purpose
- [x] Makes plasma only damage when its concentration is enough to be visible. Adds symptoms for when plasma is still no visible, so they work as a hint of the danger.
- [x] Reduces the chances of spawning with common allergies.

### Changelog
CL: Improves plasma intoxication symptoms. Plasma will only damage when its concentration is enough to be visible but still can be perceived before that.
CL: Makes common allergies a little less common.


corp-0:
as it is right now, breathing plasma damage is rather underwhelming. It took me 5 minutes of inhaling pink confetti to go from 0 toxin to 100 and finally get in critical condition.

corp-0:
do not merge, I will make some changes and also add radiation symptoms.

corp-0:
I dropped radiation symptoms, it is just annoying to work with. I will give it another try when we get new helf.

PR #5903, opened 2021-01-27
Fixes duplicated bundles

Bod9001:
why has no one done this yet


corp-0:
I did on my local but got different errors so I didn't want to risk it

PR #5907, opened 2021-01-27
Fixing plasma cutters.

PetMudstone:
Purpose
- Makes plasma cutters use fusion cell ammo and alters power cells a bit. This fixes the ability for plasma cutters to fire more than once.

CL: Fixed bug where plasma cutters couldn't fire more than once.


corp-0:
2 minutes SPEED MEEERGE

PR #5908, opened 2021-01-28
Remove some unneeded files and fixes some errors

Bod9001:
saves tiled changes that aren't ready to be applied later until the scene is loaded, 

This happens when tiled changes are being performed to a matrix when Client is loading scenes, this is a nice and simple fix so is ok

Fixes tostring that causes massive ram leak from massive packet

tries to fix race conditions on reagent container atmospheric thread


UnityProject/Assets/Scripts/Messages/Server/UpdateTileMessage.cs
```diff
@@ -10,14 +11,64 @@ public class UpdateTileMessage : ServerMessage
...
+		}
+		else
+		{
+			var tileChangerManager = NetworkObject.GetComponent<TileChangeManager>();
+			tileChangerManager.InternalUpdateTile(Position, TileType, TileName, TransformMatrix, Colour);
+			TryDoUndone();
+		}
+	}
+
+	public void TryDoUndone()
```
corp-0: the method name is weird

UnityProject/Assets/Scripts/Messages/Server/UpdateTileMessage.cs
```diff
@@ -10,14 +11,64 @@ public class UpdateTileMessage : ServerMessage
...
+		{
+			NetworkObject = null;
+			LoadNetworkObject(DelayedStuff[i].TileChangeManager);
+			if (NetworkObject != null)
+			{
+				var tileChangerManager = NetworkObject.GetComponent<TileChangeManager>();
+				tileChangerManager.InternalUpdateTile(DelayedStuff[i].Position, DelayedStuff[i].TileType,
+					DelayedStuff[i].TileName, DelayedStuff[i].TransformMatrix, DelayedStuff[i].Colour);
+				DelayedStuff.RemoveAt(i);
+				i--;
```
corp-0, line 65: Codacy might be right, your loop is increasing but you're decreasing inside it. Is it intended?

Bod9001: well I increment it down after removing element so,  it still in the same relative place
so If I remove zero it will still be at zero in the next pass

corp-0: Gotcha

UnityProject/Assets/Scripts/Messages/Server/UpdateTileMessage.cs
corp-0: What is it doing or undoing?

Bod9001: changes that couldn't be applied when it was sent,it's trying to do them again

PR #5909, opened 2021-01-28
Allow Xenos to become queens

corp-0:
### Purpose
This expands the xeno lifecycle in an attempt to preserve the perfect life form. Once spawned, xenos will check the amount of queens in the current round and in case the set cap (defaults to 1) isn't reached, they will try to become a queen themselves.

### Notes:
Minor changes:
- Converted death and random hostile sounds to Addressables, I set the Xeno sounds, the rest need setting.
- Might have fixed Queens not laying eggs.
- Increased the hugger cap for queens to 20 (it was 8). They weren't laying eggs because the huggers in caverna were enough to reach the cap.

CL: Xenomorphs can become Queens now if there are no Queens in the round.
CL: Fixed Queens not laying eggs.

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Mobs/NPC/Resources/Hostile/Xenomorph/NPC_Facehugger.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Mobs/NPC/Resources/Hostile/Xenomorph/NPC_Facehugger.prefab
@@ -70,7 +70,7 @@ SpriteRenderer:
   m_AutoUVMaxAngle: 89
   m_LightmapParameters: {fileID: 0}
   m_SortingLayerID: -229502157
-  m_SortingLayer: 16
+  m_SortingLayer: 21
   m_SortingOrder: 11
   m_Sprite: {fileID: 21300016, guid: 4866054b12864074ca72678ebef13a7c, type: 3}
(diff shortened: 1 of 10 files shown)


PerfectTangent:
Might run into some weird edge cases where drones won't level up because the extra scenes like away sites start with queens. Caverna being the main one that comes to mind

corp-0:
Yep. It happened already with huggers as stated in the description. If there is a queen in caverna, then the xenos will never become one. The cap can be tweaked per xeno if we wanted.

PerfectTangent:
Could you make it check by matrix?

PerfectTangent:
Could also make the xeno queen in caverna not count towards the cap.

corp-0:
I think I like the last better. I can add a toggle box that would avoid that particular instance of the ai to count.

Mute-Man:
what does space count as? is it its own matrix? 
if queens were counted by matrix, which matrix would xenos in space count as being on

corp-0:
space is part of main station matrix afaik

Mute-Man:
wack

corp-0:
Checked caverna. It doesn't directly spawn mobs but uses a system to spawn mobs only after someone put a feet on the matrix so I can't really change anything about how they count.

Bod9001:
good to merge?

corp-0:
imo ye

Issue #5807, opened 2021-01-03
Addressables Sound Conversion - The Final Lap


PerfectTangent:
this issue covers the remaining work to be done left over from #5702. A handful of scripts must be edited to get sounds converted and working in the development build again

these are the remaining scripts that must be addressed as far as we are aware:

- [x]  PlayerScript.cs
- [x]  Antagonist.cs
- [x]  AntagManager.cs
- [x]  AntagBannerMessage.cs
- [x]  SoundAmbientManager.cs
- [x]  SpellBook.cs
- [x]  Jukebox.cs
- [ ]  Microwave.cs (unchecked) - should be _random_ punch sound on both closing and opening
- [ ]  Anything still using the Audio Source component
- [ ]  Anything that Jester has commented in
- [ ]  Anything with the "RRTL" comment in its code (from Bod)
- [x] PDA sound effects
- [x] lobby music player
- [x] Disposal Bin recharge sound
- [x] Reimplement footstep sounds

others that still may need to be looked at -BVOld_StillBeatingHeart, All-in-one Grinder, lobby clown honk

- [ ] set up a build process for pointing the catalogues towards the CDN for the official Builds
and a way for catalogues to be updated on the CDN

wraith-54321:
nice

corp-0:
would be handy to have the list of scripts here.

PerfectTangent:
PDA sounds were updated with #6078

corp-0:
only last task missing but that one isn't trivial and should have its own issue. Closing.

PR #5910, opened 2021-01-28
Add missing footsteps sounds

### LINKED ISSUE #5813: Reintegrate the ability to play footsteps
Since we moved sounds to be addressable, the footsteps logic was discarded (which to be honest is good, it was terrible).

To consider this issue solved, the solution should create a new system to play footsteps, being able to play them on any matrix and check for the different footstep sounds that could play (normal, clownstep, asteroids, sand, water, etc).

corp-0:
### Purpose
Added the missing footsteps sounds. I also deleted step changer and the need for a field in Mind.cs. Instead we will use traits in the items that can change the footstep sound.
fixes #5813 

### Also
Updated AntagBanner to use Addressable sounds. The sounds are missing tho, so they won't play anything until they are added to the addressable project and set in their respective Antagonist SO. Progress toward #5807


UnityProject/Assets/Scripts/ScriptableObjects/Audio/FloorSounds.cs
```diff
@@ -1,14 +1,18 @@
...
-	public List<AddressableAudioSource> Barefoot = new List<AddressableAudioSource>();
-	public List<AddressableAudioSource> Claw = new List<AddressableAudioSource>();
-	public List<AddressableAudioSource> Shoes = new List<AddressableAudioSource>();
+	[CreateAssetMenu(fileName = "FloorSounds", menuName = "ScriptableObjects/FloorSounds")]
+	public class FloorSounds : ScriptableObject
+	{
+		public List<AddressableAudioSource> Barefoot = new List<AddressableAudioSource>();
+		public List<AddressableAudioSource> Claw = new List<AddressableAudioSource>();
+		public List<AddressableAudioSource> Shoes = new List<AddressableAudioSource>();
+		public List<AddressableAudioSource> Clown = new List<AddressableAudioSource>();
```
Bod9001, line 13: the sounds clown shoes would make would be defined on the clown shoes not the tile

corp-0: It is an easier way to override it. Say you want to make clown play water splashes when stepping on water tile, you just assign the water splashes on its sound category and forget about doing exceptions.

corp-0: if I’m not mistaken even ss13 defined the sounds in the tiles themselves

Bod9001: well, there's a bool to make the tile Override any custom sounds like Clown  to play its own

corp-0: but then you have to individually set overrides in an SO and assign it to the step changer which will store a variable in Mind.cs . It was wacky when I first did step changer, it wasn't supposed to stay like that.

Bod9001: It seems simple though, if you want clown shoes to make clown sounds add the class StepChanger  to clown shoes and it works, if the tile however has a custom sound and tile wants to override then it can,

corp-0:
closing and pushing it to the backburner. I'll leave the branch open if anyone wants to pick up the ready to use scriptable objects and implement special shoes overwritting the normal sounds.

PR #5920, opened 2021-01-30
Various Radial QoL features

Copern:
### Purpose
The radial menu will no longer move around as the player moves around. The menu will open just as it did before, but the menu itself is locked in place (until reopening). Instead, the branch will grow/shrink as the player moves away/toward the origin of the tile clicked. This keeps the goal of making sure the menu doesn't open up outside of the screen and showing the player where they clicked, while getting rid of menu chasing and jankiness if moving around. 
The branch origin now centers on the tile clicked and it will continue to follow the tile as the player moves. 
The branch can follow moving matrices. 
The branch will appear under the UI if a tile instead of a UI slot is clicked. 
The branch lines will hide if you get too close to the origin, the origin square will continue to show but it will be under the menu. Clicking a UI element will have the same behaviour as before (doesn't change).

Two keybind shortcuts were added (can be changed in options). Shift-Q to scroll forward and Shift-E to scroll backward are the defaults.

### Known Issues:
The branch currently doesn't account moving matrix offsets or rotation so the branch origin may appear to jump as the matrix moves.


corp-0:
> so the branch origin may appear to jump as the matrix moves

can you provide a gif on how this looks?

Copern:
Looks like this for now until I handle offsets. I am moving down in a shuttle.

https://user-images.githubusercontent.com/2001506/106348802-db067680-627d-11eb-9c80-f52efa98703d.mp4

corp-0:
I see, yes it is acceptable, at least the radial is not behind the shuttle

PR #5930, opened 2021-02-02
Fixes human UI right side alerts on clients.

### LINKED ISSUE #5902: O2, fire, pressure icons are invisible when playing as client
### Environment
- Game Version: 20012600
- OS Version: NA

Danger notification icons are invisible on latest develop version when you are playing as client.

### Steps to reproduce
1. Join staging
2. Jump out an airlock
3. Notice  you're dying because of pressure and no o2, but there is no icons

Aranclanos:
### Purpose
Fixes human UI right side alerts on clients.
Fixes #5902

### Notes:
this temp fix doesnt seem to be fixing any present bug, I connected to staging and the UI was working fine

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Systems/ControlDisplays.cs
+++ b/UnityProject/Assets/Scripts/UI/Systems/ControlDisplays.cs
@@ -83,32 +83,7 @@ public class ControlDisplays : MonoBehaviour
 			HumanUI();
 		}
 	}
-
-	void Update()
-	{
-		TempFixMissingRightHud();
-	}
-
-	//Temp fix for strange bug where right hud is missing when joining headless server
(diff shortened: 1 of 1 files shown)


corp-0:
>  I connected to staging and the UI was working fine

I wouldn't get any alert on the right side, including hunger, o2, presure, etc.

> this temp fix doesnt seem to be fixing any present bug

so removing that piece of code fixes it? The fix was the bug?

Aranclanos:
I meant I connected to staging with this code in my own client, the fix says that its temporal and for a headless bug but I really cant find said bug.
And yes, by randomly setting `UIManager.PlayerHealthUI.humanUI = true;` here the Update() of the human interface wont enable the gameobject that holds all the alerts

PR #5936, opened 2021-02-03
Adds Ancientstation scene and Ancient Engineer Occupation. Fixes Wizard Ship sprites.

PerfectTangent:
### Purpose

- adds Ancient Engineer to JobType.cs, Ancient Engineer Occupation SO and Occupation Populator.
- Also adds Hivebot simple hostile NPCs and non-functional sleepers.
- Fixes wizard ship sprites and tables layer and adds tables.

### Notes:

Currently, the Ancient station scene doesn't do anything. It is also not a Mainstation, if that wasn't clear. This is for the Ancient Engineer ghost role.


corp-0:
Oh no, the 0,0 prefab test failed. I don't really know how to solve that one.

ThatDan123:
Is the ancient station at 0,0?

NoooneyDude:
If you want something specifically at local position 0, 0 but the test complains, just offset it by something less than 0.5 - the object will be set back to 0, 0 at runtime by default. Tiles won't be evaluated for the test.

PerfectTangent:
> Is the ancient station at 0,0?

no, it should be at -4500, 6500. The test complained that a singular prefab was at 0,0 locally inside a shuttle that's a part of the scene.

PR #5938, opened 2021-02-04
Startup time improvement: MetaDataNode server only

Aranclanos:
### Purpose
Metadata nodes will now have their neighbors and types set up on the server only, as the client does not use them, increasing startup time.


Bod9001:
thanks for doing this there is probably a bunch of other systems that need to be set server only

PR #5939, opened 2021-02-04
Fixes scrubbers initialization

Aranclanos:
### Purpose
Fixes pipe scrubbers initialization, now they'll be built properly.
Removes the temp failcheck added for the atmos thread to no fail on these failed scrubbers.

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Pipes/Devices/AirVent.cs
+++ b/UnityProject/Assets/Scripts/Objects/Pipes/Devices/AirVent.cs
@@ -22,10 +22,12 @@ namespace Pipes
 		public override void Start()
 		{
 			pipeData.PipeAction = new MonoActions();
-			registerTile = this.GetComponent<RegisterTile>();
-
-
 			base.Start();
+		}
+
+		private void Awake()
+		{
+			registerTile = GetComponent<RegisterTile>();
 		}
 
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Objects/Pipes/Devices/AirVent.cs
```diff
@@ -43,14 +45,6 @@ public override void TickUpdate()
 
 		private void CheckAtmos()
 		{
-			// FIXME I'm just handling the exception here, I'm no atmos nerd so I don't know what's happening.
-			// maybe it is just an initialization order problem?
-			if (metaNode == null)
-			{
-				Logger.LogError("Airvent found metadaNode to be null. Returning with no op.", Category.Atmos);
```
Bod9001, line 50: this may spam?

corp-0: I did that, it didn't spam for me

corp-0: It was an initialization problem so it only happened during round start, it would also break atmos on editor for me without the null check

corp-0: also Aran is removing the log, not adding it

PR #5940, opened 2021-02-04
Some more objectives

ThatDan123:
Adds two more escape objectives and some other objectives:

-Hijack, be the only alive player on the escape shuttle 
Adds emagging the escape shuttle console to set depart time to 10 seconds. Takes 20 seconds and has a 25% chance to fail (Maps will need to replace the shuttle console in the escape shuttles for the escape shuttle console prefab #5941 )

-Glorious Death, die before the end of the round

-Adds the Maroon objective, an alternative version to assassinate but you only have to stop them leaving on the escape shuttle, they can be alive at round end.

-Adds chance for some gimmick objectives which are always completed, currently only traitor has a 5% chance to get a gimmick objective.

Note to merger, squash commits pls


UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
```diff
@@ -0,0 +1,86 @@
...
+	private RegisterTile registerTile;
+
+	private void Awake()
+	{
+		registerTile = GetComponent<RegisterTile>();
+	}
+
+	public bool WillInteract(HandApply interaction, NetworkSide side)
+	{
+		if (!DefaultWillInteract.Default(interaction, side)) return false;
```
corp-0: compare to false to comply with the coding guidelines

UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
```diff
@@ -0,0 +1,86 @@
...
+		StandardProgressAction.Create(
+			cfg,
+			() => FinishHack(interaction)
+		).ServerStartProgress(ActionTarget.Object(registerTile), timeToHack, interaction.Performer);
+
+	}
+
+	private void FinishHack(HandApply interaction)
+	{
+		if (Random.Range(0, 100) <= chanceToFailHack)
```
corp-0: I'd use ``DMMath.Prob`` to keep consistence on how we roll chances

corp-0 (commented):
pretty good. What's the design idea behing gimmick objectives? Will they eventually be evaluated or are they really always true?

ThatDan123:
> pretty good. What's the design idea behing gimmick objectives? Will they eventually be evaluated or are they really always true?

They're always true, saw it on the goon hub wiki thought it might be fun to have.

corp-0:
> They're always true, saw it on the goon hub wiki thought it might be fun to have.

They sound like fun as long as player don't find out they are always true, then they could be just ignored. Anyways, let's give it a roll, do the requested style changes and I merge.

PR #5942, opened 2021-02-04
Asteroid mob spawner rework

Humonitarian:
### Purpose
- Replace the NPCSpawners with randomPrefabPools.
- Replace most Xenomorphs with Space Carps.
- Reduce the chance of large masses of mobs spawning in a small-sized asteroid.

### Notes:
- Carps currently seem to provide no items or leave no corpse when killed, might require an issue.


wraith-54321:
there are a bunch of mobs with that issue, and its more the corpse is invisible

PR #5946, opened 2021-02-05
Admin Gib Option

ThatDan123:
Adds a right click action on players to gib them for admins


PerfectTangent:
Does this remove the smite button in the admin player manager? It functionally does the same thing, just remotely.

ThatDan123:
doesnt remove it, just adds the new option to the right click menu

PR #5950, opened 2021-02-05
New asteroid and space ruins, increased carp spawnrate

Humonitarian:
### Purpose
- Added a new asteroid and two space ruins
- Increased carp spawnrate by 1.5x

### Notes:
- I found out why the NPCFactories made so many xenos. NPCs love to run away. Carp spawnrate increased by 150% as they are fairly rare near asteroids right now. Now the amount of simplemobs generated by asteroids will be similar or slightly less than what it used to be.
- Either the AI, or NPC generation will need to be changed if we want a consistent threat near the asteroids without having to generate excessive amount of mobs.
- Tables do not know how to respond to rotation. Since they are tile-based, we cannot tell them to ignore rotation like we do with objects, resulting in side-ways sprites.

### DIFF
--- a/UnityProject/Assets/Resources/ScriptableObjects/RandomItemPools/AsteroidPools.meta
+++ b/UnityProject/Assets/Resources/ScriptableObjects/RandomItemPools/AsteroidPools.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: cdb9f987475ca974e9507f2cf7c910e9
+folderAsset: yes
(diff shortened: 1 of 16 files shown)


Humonitarian:
The cause of the error is that there are some light switches in the new scenes that have no associated lights. I suppose I'll add some tomorrow, do not merge until then.

Humonitarian:
Once the checks are done it should be ready to merge

PR #5952, opened 2021-02-06
Fixes two rare cases where the atmos thread would start or break at roundstart

Aranclanos:
### Purpose
Fixes two rare cases where the atmos thread would start or break at roundstart.
Added PostRoundStarted broadcast event, to be caleld once the `OnSpawnServer()` calls finish

### Notes:
First one was if the thread started too fast and the `running = true` would not be set, making the loop of the thread never start.
Second one was having the thread also start too early, before the machines had their nodes ready (being set by `OnSpawnServer()`). In this case, now the atmos thread will start on a different broadcast event, called after all `OnSpawnServer()` calls are finished

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/EventManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/EventManager.cs
@@ -17,6 +17,7 @@ public enum EVENT
 	ChatUnfocused,
 	LoggedOut,
 	RoundStarted,
+	PostRoundStarted,
 	RoundEnded,
 	DisableInternals,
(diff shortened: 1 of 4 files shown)


ThatDan123:
Did you check whether the EVENT enum wasnt used in the inspector anywhere, as adding a new value in the middle would mess it up. I dont imagine it was, but its good to double check.

Aranclanos:
Looks fine, I just doublechecked

ThatDan123:
Cool, i'll merge then

PR #5953, opened 2021-02-06
Adds new escape shuttle console to all except PogStation.

Akbadain29:
### Purpose
Adds new escape shuttle console to all escape shuttles on MainStation scenes except PogStation.


corp-0:
thanks, merging

PR #5957, opened 2021-02-06
Clean up inspector a bit

ThatDan123:
Adds boxgroup to some of the common scripts to hide the Addressable Sound stuff

If you find any more annoying ones, or want to add new sounds then add this: [Foldout("name")]

eg:

[SerializeField]
[Foldout("name")]
private AddressableAudioSource name = null;

### DIFF
--- a/UnityProject/Assets/Scripts/Clothing/RestraintOverlay.cs
+++ b/UnityProject/Assets/Scripts/Clothing/RestraintOverlay.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using UnityEngine;
 using AddressableReferences;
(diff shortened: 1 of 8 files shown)


ThatDan123:
thx naughty attributes, sometimes it works other times it doesn't. Some of the sound stuff has hidden but on other scripts it doesnt do anything

ThatDan123:
Seems like it doesnt work on network behaviours, which is a pain

corp-0:
That happens when the component you’re editing uses a custom editor. You can solve it by finding the custom editor and change the base class + some code. See NaughtyAttributes readme for the exact class and method names.

PR #5968, opened 2021-02-08
Adds UI sprite scaling

Copern:
### Purpose
Many small sprites can be hard to see/read in the UI and radial. This adds in processing for sprites to gather scaling and offset data without the need to modify all textures and sprites. UI images can apply sprite scaling with an extension method by passing it the sprite. This will also re-center images where the sprites were off center in the UI. Sprite/texture metadata is created lazily and cached.

Adds icon scaling for the radial menu.

iconscaling
Note - Image is doubled in size to show affect. It looks better in-game. Some of the really small items actually look the same as they do in the world with scaling enabled.

Switches from using CrossFadeAlpha for highlighting in the radial menu and uses leantween instead.

### Notes:
Scaling for ItemSlots can be added with about 3 lines of code. However, some of the item slot images on the UI were made larger than the slot graphics so some images appear 2-3 pixels larger than the UI that houses the image if scaled. We may want to adjust the prefabs so the icons fit inside the slot graphics.

### DIFF
--- a/UnityProject/Assets/Assets.asmdef
+++ b/UnityProject/Assets/Assets.asmdef
@@ -17,7 +17,8 @@
         "Lighting",
         "LeanTween",
         "Unity.Addressables",
-        "Unity.ResourceManager"
+        "Unity.ResourceManager",
+        "SpriteProcessing"
     ],
     "includePlatforms": [],
     "excludePlatforms": [],
--- a/UnityProject/Assets/Resources/Prefabs/UI/RightClick/RightClickButtonBase.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/UI/RightClick/RightClickButtonBase.prefab
@@ -103,6 +103,7 @@ MonoBehaviour:
     m_ColorMultiplier: 1
     m_FadeDuration: 0.15
   divider: {fileID: 2376975122167300851}
+  selectionDelay: 0
 --- !u!114 &6868661415041801040
 MonoBehaviour:
   m_ObjectHideFlags: 0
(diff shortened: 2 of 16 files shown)


Copern:
Do not merge yet.

Bod9001:
humm,  couldn't you just make the image component bigger?

Bod9001:
so, this dynamically adjust the sprites to fit centre in the UI?

Copern:
> 
> 
> humm, couldn't you just make the image component bigger?

That is basically what this does, it scales the image component using its local scale. I could resize it instead, same difference really. But to do so, it is necessary to know the real size of a sprite. There's no built-in mechanism to tell how many transparent pixels are around the real (non-transparent) pixels of a sprite. As the pixels in a sprite can be concentrated at the bottom, top, left, or right of the sprite, simply resizing/scaling the UI image will cause the sprite to appear outside the UI. To prevent that, it also gets an offset. Might as well center it too, looks a lot nicer that way. It doesn't affect the sprites at all, just the UI image(s) that show the sprite.

The goal of this is to do it very quickly (no hiccups), avoid having to modify all sprites in unitystation, avoid creating loads of garbage, and to not load and calculate all sprites at startup. Due to the size of the texture atlases on building the client, I'm going to have to go with my original plan of using pointers. As the sprite data is created lazily and not all at once during startup, this needs to be very fast. It's hardly a difference code wise to use pointers (its already using the NativeArrays) but it will require using "unsafe." The benefit is it will no longer require a big ole array (the pixel cache) to copy to and it'll be significantly faster since it will only have to deal with the pixels it needs to check.

PR #5969, opened 2021-02-08
implement the ability to saw off guns

AutumnNova:
adds:
 a component that can be added to any gun which allows it to be sawn off using a tool (currently this tool is the welder due to the saw not being a thing yet
 the ability to saw off the improvised shotgun, riot shotgun, bartender's shotgun and the mosin
 the shotguns that can be sawn off will explode in the holder's face if they attempt to saw it while it still has ammo within
 the mosin does not have this behavior due to its physical design
 sawn off weapons will have worse camera recoil and accuracy
 sawn off weapons will be smaller in size compared to the un-shortened counterpart
 sawn off weapon has alternate sprite
changes:
 when firing a gun that has the ultra honk pin as a non-clumsy person you will actually shoot yourself in the face to reflect the message displayed
removes:
 nothin

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Weapons/SawnOff.cs
+++ b/UnityProject/Assets/Scripts/Items/Weapons/SawnOff.cs
@@ -58,42 +58,40 @@ public class SawnOff : MonoBehaviour, ICheckedInteractable<InventoryApply>
 
 	public void ServerPerformInteraction(InventoryApply interaction)
 	{
-		if (interaction.TargetObject == gameObject && interaction.IsFromHandSlot)
+		if (interaction.TargetObject != gameObject && !interaction.IsFromHandSlot) return;
+
+		//TODO: switch this trait to the circular saw when that is implemented
+		if (Validations.HasItemTrait(interaction.UsedObject, CommonTraits.Instance.Welder) && gunComp.FireCountDown == 0)
 		{
-			//TODO: switch this trait to the circular saw when that is implemented
-			if (Validations.HasItemTrait(interaction.UsedObject, CommonTraits.Instance.Welder) && gunComp.FireCountDown == 0)
+			if (isSawn)
 			{
-				if (isSawn)
+				Chat.AddExamineMsg(interaction.Performer, $"The {gameObject.ExpensiveName()} is already shortened!");
+			}
+
+			if (ammoBackfire && gunComp.CurrentMagazine.ServerAmmoRemains != 0)
+			{
+				gunComp.ServerShoot(interaction.Performer, Vector2.zero, BodyPartType.Head, true);
+				Chat.AddActionMsgToChat(interaction.Performer,
+				$"The {gameObject.ExpensiveName()} goes off in your face!",
+				$"The {gameObject.ExpensiveName()} goes off in {interaction.Performer.ExpensiveName()}'s face!");
+			}
+			else
+			{
+				Chat.AddActionMsgToChat(interaction.Performer,
+					$"You shorten the {gameObject.ExpensiveName()}.",
+					$"{interaction.Performer.ExpensiveName()} shortens the {gameObject.ExpensiveName()}");
+
+				itemAttComp.ServerSetSize(sawnSize);
+				spriteHandler.ChangeSprite(1);
+
+				// Don't overwrite recoil conf if it isn't setup
+				if (SawnCameraRecoilConfig.Distance != 0f)
 				{
-					Chat.AddExamineMsg(interaction.Performer, $"The {gameObject.ExpensiveName()} is already shortened!");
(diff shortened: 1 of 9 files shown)


UnityProject/Assets/Scripts/Items/Weapons/SawnOff.cs
```diff
@@ -0,0 +1,97 @@
...
+using AddressableReferences;
+using Mirror;
+using UnityEditor;
+using UnityEngine;
+using UnityEngine.Serialization;
+using Weapons;
+
+[RequireComponent(typeof(ItemAttributesV2))]
+[RequireComponent(typeof(Gun))]
+public class SawnOff : MonoBehaviour, ICheckedInteractable<InventoryApply>
```
corp-0, line 13: a little bit of a nitpick, but for me a good component name has "able" at the end to inform adding the component will add a particular functionality to whatever prefab you add attach it to. Also this one is specific for weapons, maybe ``SawoffableWeapon``? If that makes any sense in English.

UnityProject/Assets/Scripts/Items/Weapons/SawnOff.cs
```diff
@@ -0,0 +1,97 @@
...
+{
+
+	private ItemAttributesV2 itemAttComp =>
```
corp-0: might be worth to cache this getComponent in Awake instead of getting the component every time someone tries to access this property.

UnityProject/Assets/Scripts/Items/Weapons/SawnOff.cs
```diff
@@ -0,0 +1,100 @@
...
+			{
+				return true;
+			}
+		}
+		return false;
+	}
+
+	public void ServerPerformInteraction(InventoryApply interaction)
+	{
+		if (interaction.TargetObject == gameObject && interaction.IsFromHandSlot)
```
corp-0: this block would benefit of inverting the ifs logic to reduce the nesting.

UnityProject/Assets/Scripts/Items/Weapons/SawnOff.cs
```diff
@@ -0,0 +1,100 @@
...
+		{
+			//TODO: switch this trait to the circular saw when that is implemented
+			if (Validations.HasItemTrait(interaction.UsedObject, CommonTraits.Instance.Welder) && gunComp.FireCountDown == 0)
+			{
+				if (isSawn)
+				{
+					Chat.AddExamineMsg(interaction.Performer, $"The {gameObject.ExpensiveName()} is already shortened!");
+				}
+
+				WaitFor.Seconds(0.25f);
```
corp-0: does this work? You shouldn't be able to pause execution outside an async/coroutine

corp-0:
I saw the Sync methods but not the ServerSet methods. How do you set the synced values?

UnityProject/Assets/Scripts/Items/Weapons/SawnOff.cs
```diff
@@ -58,42 +58,40 @@ public bool WillInteract(InventoryApply interaction, NetworkSide side)
 
 	public void ServerPerformInteraction(InventoryApply interaction)
 	{
-		if (interaction.TargetObject == gameObject && interaction.IsFromHandSlot)
+		if (interaction.TargetObject != gameObject && !interaction.IsFromHandSlot) return;
```
corp-0, line 61: shouldn't these first 2 ifs be part of the WillInteract method instead? Since we do nothing when any of them fail.

UnityProject/Assets/Scripts/Items/Weapons/SawnOff.cs
```diff
@@ -58,42 +58,40 @@ public bool WillInteract(InventoryApply interaction, NetworkSide side)
...
+
+		//TODO: switch this trait to the circular saw when that is implemented
+		if (Validations.HasItemTrait(interaction.UsedObject, CommonTraits.Instance.Welder) && gunComp.FireCountDown == 0)
```
corp-0, line 64: actually do we need to check the trait again? I saw it was being checked in WillInteract

AutumnNova: it is setup so willinteract handles clientside checks and serverperform handles serverside
it's done this way in most of the other gun code due to the server needing to do several different things depending on the objects used within the interaction, so it only checks it once clientside and once serverside instead of twice serverside.
kept it this way for consistency

corp-0: afaik WillInteract runs on both client first and server later but alright. Are you ready to merge now?

AutumnNova: should be good now, yeah

PR #5970, opened 2021-02-08
Sound/Music fixes and Job Spawn Banner

ThatDan123:
CL: Fixed music and juke box

CL: Fixed bad bug in sound manager, which broke logic for changing sounds after they have started playing.

CL: Adds job banner when spawning as non-antag, though to distinguish from antag spawn there's no background colour

For example:

image
Colours can be changed in the job occupation SO, atm i just copied the colours that are used in the Job select in the lobby


ThatDan123:
Codacy is now complaining that it should be returning void, before it was complaining that it was returning void........

corp-0:
leave it as Task, I removed the pattern from codacy

corp-0:
looks good, anything missing other than uploading the bundle?

UnityProject/Assets/Scripts/Core/Lifecycle/PlayerSpawn.cs
```diff
@@ -23,13 +24,26 @@ public class SpawnEventArgs : EventArgs { public GameObject player; }
...
 	/// <returns>the game object of the spawned player</returns>
-	public static GameObject ServerSpawnPlayer(JoinedViewer joinedViewer, Occupation occupation, CharacterSettings characterSettings)
+	public static GameObject ServerSpawnPlayer(JoinedViewer joinedViewer, Occupation occupation, CharacterSettings characterSettings, bool showBanner = true)
 	{
 		NetworkConnection conn = joinedViewer.connectionToClient;
 
 		// TODO: add a nice cutscene/animation for the respawn transition
 		var newPlayer = ServerSpawnInternal(conn, occupation, characterSettings, null);
+
+		if (occupation != null && showBanner)
```
corp-0: at some point we need to reparent Occupation and Antagonist to be children of the same base class so we don't need duplicated code for stuff like this.

ThatDan123:
Good to be merged whenever

corp-0:
Ah damn, I didn't merge in time for the build. Sorry Dan.

PR #5978, opened 2021-02-10
Fixes player interactions inbetween adjacent matrixes

### LINKED ISSUE #5690: You can't interact with (click on) objects on matrices other than the one you're standing on
### Environment
- Game Version: 5608
- OS Version: Windows 10

For some reason, you can't interact with objects (.prefabs, to be exact) that are on different matrices than the one you are currently occupying. You can still interact with tiles like walls, (non-directional) windows, and floors, but you can't interact with objects like airlocks, floor tile items, and other things. This is a problem since you can't do stuff like interact with airlocks on the station from the arrivals shuttle, which is quite problematic when you're a fugitive that needs to get out or if you need to unweld the airlocks leading from the station to the arrivals shuttle.

### Steps to reproduce
1. Go onto the arrivals shuttle. Open the white airlocks that serve as the shuttle exits.
2. Try to click the red airlocks that act as entrances onto the station.
3. See nothing happen.

Aranclanos:
### Purpose
Removes the matrix check in the validation IsReachableByRegisterTiles(), enabling interactions inbetween two adjacent matrix
Fixes #5690

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Input System/InteractionV2/Validations.cs
+++ b/UnityProject/Assets/Scripts/Core/Input System/InteractionV2/Validations.cs
@@ -404,12 +404,12 @@ public static class Validations
 		if (IsNotBlocked(fromWorldPos, toWorldPos, isServer: isServer, context: context))
 		{
 			Vector3Int toWorldPosInt = toWorldPos.RoundToInt();
-	
+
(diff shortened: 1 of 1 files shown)


corp-0:
I wonder why the check was ever needed

PR #5979, opened 2021-02-10
Change MatrixInfo from struct to class

Copern:
### Purpose
While deep profiling I kept seeing some functions like WorldToLocal taking quite some time, more time than it really should have. Some of it made sense but there was definitely something off. While working on something related I noticed MatrixInfo was a struct. A struct with 14 fields (edit: was tired and misread the first time), 3 of those also structs. A struct that is copied in a lot of functions. Functions like WorldToLocal. This really should not be a struct.

Changing this to a class prevents copying all the fields each time it is passed along or returned from a function. This is a great time savings (around 30% for small functions) for anything that uses MatrixInfo. As an extreme example, there were some frames with 9700 WorldToLocal calls that took 50% of the total frame time. It was around 480+ms (with deep profiling on mind you) each time it happened. Those same 9700 calls took around 320 ms (again, deep profiling) when converting MatrixInfo to a class.

### DIFF
--- a/UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
+++ b/UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
@@ -305,7 +305,7 @@ namespace IngameDebugConsole
 
 				foreach ( var movableMatrix in matrices )
 				{
-					if ( movableMatrix.GameObject.name.ToLower().Contains( "verylarge" ) )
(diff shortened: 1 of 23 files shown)


Copern:
Do not merge. I haven't run into any issues, but since I'm changing from struct to class, I still need to go around and add null checks where appropriate.

corp-0:
Let us know when you're ready

Copern:
Got distracted by other things. Assuming this passes tests, it should be good to go.

PR #5984, opened 2021-02-11
Fixes a case of rounds being unable to restart

Aranclanos:
### Purpose
Fixes rounds being unable to restart if there is a sound still playing in a matrix.

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/SoundManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/SoundManager.cs
@@ -184,7 +184,7 @@ public class SoundManager : MonoBehaviour
 	{
 		foreach (var sound in Instance.SoundSpawns)
 		{
-			if (sound.Value == null)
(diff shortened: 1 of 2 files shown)


Aranclanos:
DNM, found more related to it

Aranclanos:
I'm still unable to find the reason of why the dictionary can still have a null value at that stage, but considering the staging release I added a quick hotfix instead

UnityProject/Assets/Scripts/Managers/SoundManager.cs
```diff
@@ -184,6 +184,10 @@ private void OnSceneChange(Scene oldScene, Scene newScene)
 	{
 		foreach (var sound in Instance.SoundSpawns)
 		{
+			if (sound.Value == null)
```
corp-0: leave a comment to make sure we remember later this is a hackfix, something with a //FIXME for example.

PR #6002, opened 2021-02-12
Move clothing SOs to new SO script, update prefabs with ClothingV2

NoooneyDude:
- New singular SO to serve sprite data for all clothing. Removes scripts responsible for the old SOs.
This SO no longer has an "adjusted" section; these are now just represented by a new SO asset.
- Updates `ClothingV2` to reflect the change. Changing the current sprite data is now similar to how it would be done in `SpriteHandler`.
- Updated all old SOs to the new version, including generating adjusted versions (identified by `_adjusted`) reference.
- Updated all references in the `ClothingV2` components to the new SOs, including adjusted SOs where relevant and available.
- `FlashLight` and facehugger scripts have been updated to take advantage of these changes.
hardhat headlamp

CL:Player lights now update sprite on player character where relevant (e.g. hardhat headlamps).

> Tested with two clients on headless; no obvious issues.


Bod9001:
just noting does, Sprite handler handle the networking for variant change?  when the character changes directions, since that's pretty "expensive" networking since you have to do it for every single sprite vs the one message saying character change direction

corp-0:
can you do separate commits for the prefabs and .cs files? I literally can't find any script to review

NoooneyDude:
> can you do separate commits for the prefabs and .cs files? I literally can't find any script to review

:excellent: :evil:

corp-0:
devilish

NoooneyDude:
I've split into two commits.

> just noting does, Sprite handler handle the networking for variant change? when the character changes directions, since that's pretty "expensive" networking since you have to do it for every single sprite vs the one message saying character change direction

I didn't change too much on the spritehandler side, just expose a method to allow changing sprite SOs and the way the spritehandler cache thing is populated in ClothingV2. You should be able to see the changes easier now - in first commit.

Bod9001:
ok cool

corp-0:
mergegegege

PR #6005, opened 2021-02-12
Provide (disabled) tool to migrate clothing SOs

NoooneyDude:
- Adds the tool used to migrate clothing SOs in PR #6002.
    - The tool is commented out as it will error if compiled on a fork where the old SO scripts have been removed, which will be the case after #6002 is merged.
    - To use, uncomment and set the relevant fields in ClothingV2 as public until you're done.

> Decided to commit it after all, as it may be useful for other forks.


corp-0:
why disabled?

corp-0:
nvm I can't read for some reason

PR #6015, opened 2021-02-13
Local profiling

Aranclanos:
### Purpose
Adds a button to the debug console, to run a local profile of 30 frames.
Moves the profiling to its own new manager, profile manager.

### DIFF
--- a/UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogManager.cs
+++ b/UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogManager.cs
@@ -589,6 +589,22 @@ namespace IngameDebugConsole
 		}
 
 		/// <summary>
+		/// Creates a local profile
+		/// </summary>
+		public void LocalProfileButtonPressed()
+		{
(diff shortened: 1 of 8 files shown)


UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogManager.cs
```diff
@@ -588,6 +588,22 @@ public void FilterErrorButtonPressed()
 			FilterLogs();
 		}
 
+		/// <summary>
+		/// Creates a local profile
+		/// </summary>
+		public void LocalProfileButtonPressed()
+		{
+			if (Debug.isDebugBuild)
```
corp-0, line 596: this won't work for the clients we distribute then

Aranclanos: yeah but thats just how it works

corp-0: I know. Still nice to have

PR #6017, opened 2021-02-13
MatrixManager Performance Improvements

ThatDan123:
Needs a very good review, probably shouldn't be merged till after next update.

-Makes every check only check nearby matrixes, does this by working out distance to the centre of the under floor layer, if distance > 300 then it'll ignore that matrix.
300 is just a semi-random guess for what the distance should be.

-Changes for singularity stuff for better performance.


ThatDan123:
didnt refresh branch so ignore commits below the Performance Stuff commit

PR #6018, opened 2021-02-13
Fixes the emergency light error spam on clients

Aranclanos:
### Purpose
Fixes emergency light script spamming errors on the console after they get removed on clients.

They were being added to the updatemanager twice


corp-0:
looks good I wonder why they didn't break stuff before if they were always like this

Issue #5937, opened 2021-02-04
Makarov pistol and grenade arming play no sound effects


PerfectTangent:
### Environment
- Game Version: latest staging (confirmed existed on several previous builds, too)
- OS Version: Windows 10

### Detailed description of bug

The makarov plays no firing sound effect and the grenade plays no arming sound effect when thrown.

### Steps to reproduce

1. get makarov
2. fire makarov
3. makarov makes no sound

1. get grenade
2. arm and throw grenade
3. grenade makes no sound until it explodes

PR #6021, opened 2021-02-14
Audio Fix

### LINKED ISSUE #5990: Many sounds are not playing consistently
### Environment
- Game Version: build 21021100

Many sound effects such as airlock opening sounds are playing inconsistently, sometimes playing and sometimes not playing.

Bobtheapple:
Monitor in Soundspawn.cs was pushing ~50% of all sounds to NonplayingSounds, rather than the intended few.

### Purpose
Currently on PR, many sounds are not getting played.  This is easily testable by spawning a Makarov pistol and firing it repeatedly.  About half of all gun shots will not be heard.  This is being caused by the monitor variable being set to true whenever a sound is played, and not being toggled back to false until a sound is pushed to NonPlayingSounds, causing about a 50% sound loss.  This change makes monitor toggle back to false every time there is a an update to SoundSpawn.cs, and will only push sounds to NonplayingSounds if they are trying to play on the same update as another sound, which I believe was the intended function.

Fixes #5990, #5937

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Sound/SoundSpawn.cs
+++ b/UnityProject/Assets/Scripts/Core/Sound/SoundSpawn.cs
@@ -72,7 +72,6 @@ public class SoundSpawn: MonoBehaviour
 		if (!AudioSource.isPlaying)
 		{
 			IsPlaying = false;
-			monitor = false;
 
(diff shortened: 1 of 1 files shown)


NoooneyDude:
Amazing. Thank you.

PR #6023, opened 2021-02-14
Changed UnRealStation and Fixes HoS Headset so it shows the right radio channels

### LINKED ISSUE #5992: The Singularity Engine should be mapped into at least 1 map
### Description

Now that the Singularity is in game, it should be added as the main power source for at least one Mainstation.

Akbadain29:
### Purpose
Updated UnRealStation to have Singlo for Engies to set up, puts the map in maps.json and fixed the HoS's Radioheadsets because they were stuck on the common encryption key not the HoS one, so they are now on the HoS one.

closes #5992

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Items/Clothing/Earwear/RadioHeadsetHoS.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Items/Clothing/Earwear/RadioHeadsetHoS.prefab
@@ -24,6 +24,11 @@ PrefabInstance:
       value: The headset of the man in charge of keeping order and protecting the
         station.
       objectReference: {fileID: 0}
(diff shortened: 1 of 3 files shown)


corp-0:
> puts the map in maps.json

this means it is available in rotation? Did you fix the weird cross matrix interaction?

Akbadain29:
Yeah, it somehow fixed itself, I could suddenly board the mining shuttle with no issue. It's wack.

Akbadain29:
It failed, why did it fail?

corp-0:
> UnRealStation: "LightBulbFixture (2)" miss switch reference.
UnRealStation: "LightBulbFixture (1)" miss switch reference.

two lights missing switch

Akbadain29:
Ok, that test is green now, I hate forgetting to tick the "Is without switch" box.

PR #6026, opened 2021-02-14
 Hopefully fixes tiles trying to be removed before the scene is loaded 

Bod9001:
yeah, this could have problems with tile being destroyed then place down, then it finally been done on client destroying the thing that had been placed down ,  should probably work this into update tile

### DIFF
--- a/UnityProject/Assets/Scripts/Messages/Server/RemoveTileMessage.cs
+++ b/UnityProject/Assets/Scripts/Messages/Server/RemoveTileMessage.cs
@@ -30,12 +30,6 @@ public class RemoveTileMessage : ServerMessage
 		LoadNetworkObject(TileChangeManager);
 		if (NetworkObject == null)
 		{
-
-		}
-
-
-		if (NetworkObject == null)
-		{
 			DelayedStuff.Add(new delayedData(Position, LayerType, RemoveAll, TileChangeManager));
 		}
 		else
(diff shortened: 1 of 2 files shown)


corp-0:
aaaaah you could have mentioned you were working on this, let me abort the oingoing build

Bod9001:
it's not a build if it doesn't get cancelled and then run again a minute later

UnityProject/Assets/Scripts/Messages/Server/RemoveTileMessage.cs
```diff
@@ -8,11 +9,57 @@ public class RemoveTileMessage : ServerMessage
...
+			RemoveAll = inRemoveAll;
+			TileChangeManager = inTileChangeManager;
+		}
+	}
 	public override void Process()
 	{
 		LoadNetworkObject(TileChangeManager);
-		var tileChangerManager = NetworkObject.GetComponent<TileChangeManager>();
-		tileChangerManager.InternalRemoveTile(Position, LayerType, RemoveAll);
+		if (NetworkObject == null)
```
corp-0, line 31: duplicated check

corp-0:
what's up with the old commits?

Bod9001:
odd, usually they don't show up it's because it's done on the same branch, but the commits should already be there so don't
 get it

corp-0:
yeah, it is because you were using the same branch for all three of those changes. Anyways, it is fine. I will squash.

PR #6027, opened 2021-02-15
Use bit check to avoid garbage with IsInLayer.

Copern:
Yoinking this fix from another PR I'm working on. Using HasFlag is unnecessary since we are only checking to see if one layer is in SpecifyLayers. This avoids boxing from using HasFlag.


UnityProject/Assets/Scripts/Tilemaps/Utils/Types.cs
```diff
@@ -59,7 +59,8 @@ public static class LTSUtil
 	public static bool IsLayerIn(LayerTypeSelection SpecifyLayers, LayerType Layer)
 	{
 		LayerTypeSelection LayerCon = LayerType2LayerTypeSelection(Layer);
-		return (SpecifyLayers.HasFlag(LayerCon));
+		//HasFlag causes boxing, since we are only checking for one layer, a simple bit check is fine here
```
corp-0: instead of explaining why it isn't using HasFlag, this comment should explain what the line is doing for contributors who don't know bit operations that well.

PR #6030, opened 2021-02-15
Artifact performance improvement

Copern:
Artifacts caused some frame drops occasionally due to checking all tiles in all matrices in a 20x20 area for players leading to roughly 10-20 thousand calls and a decent amount of garbage (in the order of about 60 kilobytes each call).

artifacts

Since it's pretty unlikely there will be 10000+ players, just check if any players are in the artifact bounds. This still creates a little garbage (all methods in PlayerList return new lists) but a couple hundred bytes is a wee bit better.

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Research/Artifacts/Effects/TelepaticArtifactEffect.cs
+++ b/UnityProject/Assets/Scripts/Objects/Research/Artifacts/Effects/TelepaticArtifactEffect.cs
@@ -37,13 +37,14 @@ public class TelepaticArtifactEffect : ArtifactEffect
 	private void IndocrinateMessageArea()
 	{
 		var objCenter = gameObject.AssumedWorldPosServer().RoundToInt();
-		var xMin = objCenter.x - auraRadius;
-		var yMin = objCenter.y - auraRadius;
-		var bounds = new BoundsInt(xMin, yMin, 0, 20, 20, 1);
-		foreach (var connected in PlayerList.Instance.InGamePlayers)
+		var hitMask = LayerMask.GetMask("Players");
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Objects/Research/Artifacts/Effects/TelepaticArtifactEffect.cs
```diff
@@ -34,21 +36,16 @@ public override void DoEffectPulse(GameObject pulseSource)
...
 	{
-		// get effect shape around artifact
 		var objCenter = gameObject.AssumedWorldPosServer().RoundToInt();
-		var shape = EffectShape.CreateEffectShape(EffectShapeType.Square, objCenter, auraRadius);
-
-		foreach (var pos in shape)
+		var xMin = objCenter.x - auraRadius;
+		var yMin = objCenter.y - auraRadius;
+		var bounds = new BoundsInt(xMin, yMin, 0, 20, 20, 1);
+		foreach (var connected in PlayerList.Instance.InGamePlayers)
```
Bod9001: think you can use physics2D circle overlap tho  I do want to convert that to a custom solution may be some time (probably won't happen)

Bod9001: hitMask = LayerMask.GetMask( "Players")
var player = Physics2D.OverlapCircleAll(transform.position, 20f, hitMask);

Copern: Good point, thanks.

PR #6032, opened 2021-02-15
Some fixes and performance stuff

ThatDan123:
Some performance stuff and NRE fixes

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
@@ -34,18 +34,7 @@ public partial class MatrixManager : MonoBehaviour
 {
 	private static MatrixManager matrixManager;
 
-	public static MatrixManager Instance
-	{
-		get
-		{
-			if (matrixManager == null)
-			{
-				matrixManager = FindObjectOfType<MatrixManager>();
-			}
-
-			return matrixManager;
-		}
-	}
+	public static MatrixManager Instance => matrixManager;
 
 	private static LayerMask tileDmgMask;
 
@@ -77,6 +66,18 @@ public partial class MatrixManager : MonoBehaviour
 		IsInitialized = true;
 
 		EventManager.Broadcast(EVENT.MatrixManagerInit);
+	}
+
+	private void Awake()
+	{
+		if (matrixManager == null)
+		{
+			matrixManager = this;
+		}
+		else
+		{
+			Destroy(this);
+		}
 	}
 
 	private void OnEnable()
@@ -681,14 +682,18 @@ public partial class MatrixManager : MonoBehaviour
 	/// Picks best matching matrix at provided coords and releases reagents to that tile.
 	/// <inheritdoc cref="MetaDataLayer.ReagentReact"/>
 	/// </summary>
-	public static void ReagentReact(ReagentMix reagents, Vector3Int worldPos)
+	public static void ReagentReact(ReagentMix reagents, Vector3Int worldPos, MatrixInfo matrixInfo = null)
 	{
 		if (CustomNetworkManager.IsServer == false)
 		{
 			return;
 		}
 
-		var matrixInfo = AtPoint(worldPos, true);
+		if (matrixInfo is null)
+		{
+			matrixInfo = AtPoint(worldPos, true);
+		}
+
 		Vector3Int localPos = WorldToLocalInt(worldPos, matrixInfo);
 		matrixInfo.MetaDataLayer.ReagentReact(reagents, worldPos, localPos);
 	}
@@ -1126,7 +1131,7 @@ public partial class MatrixManager : MonoBehaviour
 	public static Vector3 WorldToLocal(Vector3 worldPos, MatrixInfo matrix)
 	{
 		//Invalid matrix info provided
-		if (matrix == null || matrix.Equals(MatrixInfo.Invalid) || worldPos == TransformState.HiddenPos)
+		if (matrix is null || matrix.Equals(MatrixInfo.Invalid) || worldPos == TransformState.HiddenPos)
 		{
 			return TransformState.HiddenPos;
 		}
--- a/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
@@ -76,6 +76,8 @@ public partial class MatrixManager : MonoBehaviour
 		}
 		else
 		{
+			if (Application.isEditor && !Application.isPlaying) return;
+
 			Destroy(this);
 		}
 	}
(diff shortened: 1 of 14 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/TilemapDamage.cs
```diff
@@ -76,7 +78,12 @@ private float DealDamageAt(float damage, AttackType attackType, Vector3Int cellP
 		}
 
 		data.AddTileDamage(Layer.LayerType, basicTile.Armor.GetDamage(damage < basicTile.damageDeflection? 0: damage, attackType));
-		SoundManager.PlayNetworkedAtPos(basicTile.SoundOnHit, worldPosition);
+
+		if (basicTile.SoundOnHit.IsReadyLoaded)
+		{
+			SoundManager.PlayNetworkedAtPos(basicTile.SoundOnHit, worldPosition);
+		}
```
ThatDan123: Bod, this probably isnt the best way, as im not sure if this blocks the sound from trying to load the first time, so would never to true. The issue was that it was spamming load but wasnt loading for some reason, it wasnt null or anything so is a bit weird.

corp-0:
test is failing

UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
```diff
@@ -34,18 +34,7 @@ public partial class MatrixManager : MonoBehaviour
...
-		{
-			if (matrixManager == null)
-			{
-				matrixManager = FindObjectOfType<MatrixManager>();
-			}
-
-			return matrixManager;
-		}
-	}
+	public static MatrixManager Instance => matrixManager;
```
corp-0, line 37: I think FindObjectOfType is required for the singleton monobehaviours

ThatDan123: nope

UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
```diff
@@ -79,6 +68,18 @@ private IEnumerator WaitForLoad()
...
 
+	private void Awake()
+	{
+		if (matrixManager == null)
+		{
+			matrixManager = this;
+		}
+		else
+		{
+			Destroy(this);
```
corp-0, line 81: could this be the thing making tests go ape?

ThatDan123: yup, cus of execute in edit mode, trying to find a way to fix

UnityProject/Assets/Scripts/Objects/ExosuitFabricator.cs
```diff
@@ -62,10 +57,17 @@ private void Awake()
...
 		public void EnsureInit()
 		{
-			SyncSprite(ExosuitFabricatorState.Idle, ExosuitFabricatorState.Idle);
+			spriteHandler = GetComponentInChildren<SpriteHandler>();
 			materialStorage = this.GetComponent<MaterialStorage>();
 		}
 
+		private void Start()
+		{
+			if(CustomNetworkManager.IsServer == false) return;
```
corp-0: I remember there was a Server version of this event method, so you get rid of the if.

UnityProject/Assets/Scripts/Tilemaps/Utils/TileList.cs
```diff
@@ -64,16 +64,18 @@ public bool HasObjects(Vector3Int position)
...
 	{
-		return _objects.ContainsKey(position) ? _objects[position] : emptyList;
+		return _objects.TryGetValue(position, out var objectsOut) ? objectsOut : emptyList;
 	}
 
-	public IEnumerable<RegisterTile> Get(Vector3Int position, ObjectType type) {
-		if ( !HasObjects( position ) )
+	public IEnumerable<RegisterTile> Get(Vector3Int position, ObjectType type)
+	{
+		if (_objects.TryGetValue(position, out var objectsOut) == false)
```
corp-0, line 72:
```suggestion
		if (_objects.TryGetValue(position, out _) == false)
```
If you're not using the out value, you can dispose it like this.

ThatDan123: It is using the out value

UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
ThatDan123: Added a check which fixes this

UnityProject/Assets/Scripts/Tilemaps/Utils/TileList.cs
corp-0: ah yeah, got confused by the return brackets.

UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
ThatDan123: Most managers should be moved over to this system, as checking FindObjectOfType is expensive for start up. But the main reason for me switching it up was the null check on the get is very bad.

corp-0: Let's wait on what @Bod9001 has to say before marking as resolved.

UnityProject/Assets/Scripts/Tilemaps/Behaviours/TilemapDamage.cs
Bod9001: you could use a try catch to see if it's doing anything?

UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
Bod9001: This should work okay as long as it's not used before Awake, you can use the initialisation manager to ensure that it is so

UnityProject/Assets/Scripts/Tilemaps/Behaviours/TilemapDamage.cs
ThatDan123: Ive removed the check for now

Aranclanos:
are you sure about some of the checks? I didnt look deeply but it feels like the pipe and disposal ones are hiding a more in depth problem

ThatDan123:
The disposal one is just checking to see if its null when the object is destroyed, dont see why that one could be that bad to add.

The pipe check i know less about the system, but it seems like thats right as it should return its internal mixture if its not on the network?

corp-0:
are we ready to merge?

ThatDan123:
yeah, i think

corp-0:
boom, merged

PR #6033, opened 2021-02-15
Audio Parameters Were Not Serializable

### LINKED ISSUE #5981: JukeBox global sound
The jukebox music is playing globally, this is due to the spatial blend being set to 0 and not 1 for 3D. 

The audioSourceParameters in the JukeBox script is set to make it 1, but somewhere during the process to play the sound it gets set to 0. I tried to have a look and it seemed, unless I missed something, that the issue is with the PlaySoundMessage.cs net message it goes in as 1 but comes out as null.

If this isnt fixed by the time the new update needs releasing then the jukebox should be disabled until fixed.

Bobtheapple:
PlaySoundMessage attempts to serialize the AudioSourceParameters it receives through the send/sendToAll/sendToNearbyPlayers methods as part of its sending messages to clients.  However, because AudioSourceParameters was using nullable variables and non-serialized get/setters it could not be serialized due to the limitations of the engine.  This resulted in all AudioSourceParameters being ignored and reset to the default null values.  

### Purpose
This fix changes the nullable variables in AudioSourceParameters to normal variables, and instead uses the value of -1 to represent a "default" value.  All code that references AudioSourceParameters using null values has been changed to reflect this.  This change permits all related already existing code to function as intended.  This fix also adjusts ShakeParameters to also be serializable as well, as it was it always produced the same, default shake.

Fixes #5981

### Notes:
Note that this bug impacted almost all code that made changes to pitch, volume, pan, spread, and spatial blending, as any code that played sound through PlaySoundMessage was being reset to default values.  Because of this, many sounds are changed with this fix.

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/SoundManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/SoundManager.cs
@@ -611,20 +611,16 @@ public class SoundManager : MonoBehaviour
 		Instance.PlaySource(sound, polyphonic, global);
 	}
 
-	private void PlaySource(SoundSpawn source, bool polyphonic = false, bool Global = true, MixerType mixerType = MixerType.Unspecified)
+	private void PlaySource(SoundSpawn source, bool polyphonic = false, bool Global = true, MixerType mixerType = MixerType.Master)
 	{
-		if (mixerType != MixerType.Unspecified)
-		{
-			if (!Global
-			    && PlayerManager.LocalPlayer != null
-			    && (MatrixManager.Linecast(PlayerManager.LocalPlayer.TileWorldPosition().To3Int(),
-					    LayerTypeSelection.Walls, layerMask, source.RegisterTile.WorldPositionClient.To2Int().To3Int())
-				    .ItHit))
+		if (!Global
+		    && PlayerManager.LocalPlayer != null
+		    && (MatrixManager.Linecast(PlayerManager.LocalPlayer.TileWorldPosition().To3Int(),
+			    LayerTypeSelection.Walls, layerMask, source.RegisterTile.WorldPositionClient.To2Int().To3Int())
+			    .ItHit))
 			{
 				source.AudioSource.outputAudioMixerGroup = soundManager.MuffledMixer;
 			}
-		}
-
 		if (polyphonic)
 		{
 			source.PlayOneShot();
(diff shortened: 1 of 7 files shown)


corp-0:
I'm wondering, what would happen if we wanted to play a sound with lower pitch than default? Like a huge bike horn doing an infernal honk.

UnityProject/Assets/Scripts/Managers/SoundManager.cs
```diff
@@ -226,16 +225,12 @@ private void OnSceneChange(Scene oldScene, Scene newScene)
...
 		{
 			var ToReturn = NonplayingSounds[addressableAudioSource.AssetAddress][0];
 			NonplayingSounds[addressableAudioSource.AssetAddress].RemoveAt(0);
-			if (soundSpawnToken != "") //non addressables dont have a token
-			{
-				ToReturn.Token = soundSpawnToken;
-				SoundSpawns.Add(soundSpawnToken, ToReturn);
-			}
+			ToReturn.Token = soundSpawnToken;
+			SoundSpawns.Add(soundSpawnToken, ToReturn);
```
Aranclanos: could you revert this change? it's from today

Aranclanos:
also tested locally and it doesnt seem to work, something about mirror runtiming while trying to pack it

corp-0:
from Discord by Aran:
> ok I opened each file in the messages folder and only PlayerCustomisationMessage uses a class, CharacterSettings
and also sounds that are broken
just make the shakeatributes and soundatributes structs

Apparently this implementation doesn't work, Mirror isn't packaging the class properly.

Bobtheapple:
Changed AudioSourceParameters and ShakeParameters to structs, and updated all related code.  Had to completely rework SoundManager's ApplyAudioSourceParameters, thoroughly commented it to show what it can and can't do now.  Added a method ForceAudioSourceParameters which gives complete control.

Please do not merge until someone confirms that they are able to get in game with this.  I was able to build the project and tested it server side and client side, but I was able to get it working with the earlier version that used classes as well.  If this does work then I can do a lot of spring cleaning and get SoundManager to not be a jumbled mess.

Aranclanos:
works fine made a build and tested it, theres more that can be done but can wait for another pr

PR #6035, opened 2021-02-15
Fix reversed artifact logic.

Copern:
Woke up after realizing I unintentionally reversed the logic for artifacts after switching to using physics2d to search for players. Now I was activating it on dead players or if it was null. Whoops.


corp-0:
Its fine, we didn't catch it either.

PR #6036, opened 2021-02-15
Some Blob Fixes

### LINKED ISSUE #6019: Blob + Big Size + Rapid Fire Mutation = Server Lag
### Environment
- Game Version: 21021300
- OS Version: Windows 10

If the Blob reaches 300+ occupied tiles and gains the rapid fire mutation, it will attack every single adjacent tile at the same time and proceed to grind the server to a halt through the message spam to the chat.

### Steps to reproduce

1. Spawn as Blob 
2. Grow in size
3. Reroll strain so you get the Rapid Fire Mutation
4. Watch the chat receive thousands of notifications that your blob bits are attacking walls.

ThatDan123:
CL: Fixed blob spawning with multiple start scripts

CL: Switched detect objects to use physic2D instead

CL: Fixed blob sound

CL: Blob now ends round after 60 seconds rather than 180 seconds when it wins

CL: Removed blob attack messages, to reduce spam

Hopefully Fixes #6019 , test on staging if it still needs improving

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Player/Resources/Blob.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Player/Resources/Blob.prefab
@@ -118,6 +118,8 @@ MonoBehaviour:
   syncMode: 0
   syncInterval: 0.1
   playerMove: {fileID: 3920682320074774566}
+  playerScript: {fileID: 0}
+  Step: 0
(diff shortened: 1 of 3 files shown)


corp-0:
I'm not sure if I like not having the attack messages, but we can give it a try on staging and check the performance improvement. Merging.

PR #6040, opened 2021-02-16
fix the nukie code being the wrong one

### LINKED ISSUE #5596: Nuclear operatives receive a code for the on-station nuke and not the nuke that spawns in their shuttle.

AutumnNova:
closes #5596

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Antagonists/Antags/NuclearOperative.cs
+++ b/UnityProject/Assets/Scripts/Systems/Antagonists/Antags/NuclearOperative.cs
@@ -16,11 +16,15 @@ namespace Antagonists
 
 			//send the code:
 			//Check to see if there is a nuke and communicate the nuke code:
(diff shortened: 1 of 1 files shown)


corp-0:
thanks. I still think we should in the future reduce the search to nukie ship. We will see how we can optimize it.

PR #6043, opened 2021-02-16
Fixes Plasma Generator Sound

ThatDan123:
CL: Fixes Plasma Generator Sound

-Removes the unused string functions in the sound manager


PerfectTangent:
Huzzah!

PR #6047, opened 2021-02-17
Fixes doors not having the correct sprite and collider on clients

Aranclanos:
### Purpose
Fixes doors not having the correct sprite and collider on clients.

When the client asked for the message, they were only receiving it back if the door was open, also both spritehandler and one of the animator scripts were setting the sprite at start, depending on the connection to the server, they could set it after the message arrives


corp-0:
I’m confused. The prefab you modified is from new doors but both scripts are from old doors

Aranclanos:
My tests were with firelocks, who are under this prefab but dont replace the old door controllers.
I deactivated the spritehandler bool that changes the sprite at init to a predeterminated one.

corp-0:
will you fix new doors and add them here or should I merge this now?

Aranclanos:
mergerino, no plans on fixing new doors soon either

PR #6049, opened 2021-02-17
fix players being able to revive themselves when they shouldn't

corp-0:
### Purpose
Players were able to revive themselves even when they weren't admin and the server config didn't allow it because of a bad check for server. This PR fixes the issue while keeping the intended behaviour.


corp-0:
selfmergerino

PR #6056, opened 2021-02-18
Lobby Music Autoplay and Various Addressable Audio fixes

Bobtheapple:
### Purpose
This PR addresses two sets of problems, and since they deal with related code I have included them together.  First, the PR addresses several underlying problems that was preventing Lobby Music from playing properly:

1. MusicManager.isLobbyMusicPlaying() as written would always return true, preventing Lobby Music from autoplaying.

2. SongTracker was calling for music to begin playing before Addressables had finished loading, throwing an exception.

3. MusicManager and SoundTracker had no null handling in case of failure.

4. AddressableReference had no means of identifying a non-null, non-valid Addressable Address.  This was solved by adding a new method, HasValidAddress(), which can be safely called to check whether an addressable can be found at its address.

5. Several LogError calls where attempting to use information from an addressable whose data could not be read to report that its data could not be read, throwing exceptions that broke the await statements.

This PR addresses all of the above issues, and additionally makes changes to GetAddressableAudioSourceFromCache and EnsureAddressableAudioSourceFromCache.  

1. Separating these methods did nothing, as the former was simply an await statement for the latter to finish.  Since any method calling the former would have to have an await statement to await the former awaiting the latter, one might as well just skip the middleman.

2. EnsureAddressableAudioSourceFromCache was not actually doing its namesake, it was building a cache but ultimately did nothing with it.  This PR implements that functionality, in testing I observed an approximate 30% reduction in paused Audio Sources with these changes.

### Notes
Lobby Music is now working for me consistently after build, but not the first run when using "Build And Play" in UnityProject.  I suspect there may still be another underlying problem related to the Addressable Catalogue code that I could not find.

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/LobbyManager/MusicManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/LobbyManager/MusicManager.cs
@@ -116,7 +116,7 @@ namespace Audio.Containers
 		{
 			if (Instance.currentLobbyAudioSource != null
 			    && Instance.currentLobbyAudioSource.isPlaying
-			    || !(SunVox.sv_end_of_song((int) Slot.Music) == 0))
+			    || (SunVox.sv_end_of_song((int) Slot.Music) != 0))
 			{
 				return true;
 			}
--- a/UnityProject/Assets/Scripts/Managers/LobbyManager/MusicManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/LobbyManager/MusicManager.cs
@@ -80,6 +80,11 @@ namespace Audio.Containers
 			StopMusic();
 			if (currentLobbyAudioSource == null) Init();
 			var audioSource = await SoundManager.GetAddressableAudioSourceFromCache(new List<AddressableAudioSource>{audioClips.GetRandomClip()});
+			if(audioSource == null)
+			{
+				Logger.LogError("MusicManager failed to load a song, is Addressables loaded?", Category.Addressables);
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Managers/LobbyManager/MusicManager.cs
```diff
@@ -111,7 +116,7 @@ public static bool isLobbyMusicPlaying()
 		{
 			if (Instance.currentLobbyAudioSource != null
 			    && Instance.currentLobbyAudioSource.isPlaying
-			    || !(SunVox.sv_end_of_song((int) Slot.Music) == 1))
+			    || !(SunVox.sv_end_of_song((int) Slot.Music) == 0))
```
corp-0: Why did this change from 1 to 0?

Bobtheapple: Basically, because it was backwards.  sv_end_of_song returns 0 if the song is still playing, 1 if the song has ended, as in is not playing (I know, its weird, but it makes sense... kind of).  As it was, this statement would return true if either the current lobby audiosource was playing or SunVox was not playing, ie it would be true whether a song was playing or not, and could only be false if both SunVox was playing and a song wasn't playing, ie it could never be false.

corp-0: fair enough

PR #6078, opened 2021-02-20
Update PDA Sounds to Addressable

Bobtheapple:
### Purpose
Changes PDALogic to use AddressableAudioSources, fixes a bug where GUI_PDASettingMenu.setRingtone was allowing empty strings, and removed now unused placeholder overloads in SoundManager

### Notes:
Currently, there is only one ringtone enabled for PDAs, TwoBeep.  I tested the code with other sound effects and it worked, so we should be fine to expand the list.  Also, as implemented the ringtone that you need to enter into the PDA is the file name of the prefab and is case sensitive, we may want to change that.

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Items/PDA/GUI_PDASettingMenu.cs
+++ b/UnityProject/Assets/Scripts/UI/Items/PDA/GUI_PDASettingMenu.cs
@@ -19,7 +19,7 @@ namespace UI.Items.PDA
 
 		public void SetRingtone(string ringtone)
 		{
-			if (ringtone == default) return;
+			if (ringtone == "") return;
 
 			if (controller.PDA.IsUplinkCapable)
 			{
--- a/UnityProject/Assets/Scripts/UI/Items/PDA/GUI_PDASettingMenu.cs
+++ b/UnityProject/Assets/Scripts/UI/Items/PDA/GUI_PDASettingMenu.cs
@@ -19,7 +19,7 @@ namespace UI.Items.PDA
 
 		public void SetRingtone(string ringtone)
 		{
-			if (ringtone == "") return;
+			if (String.IsNullOrEmpty(ringtone)) return;
 
 			if (controller.PDA.IsUplinkCapable)
 			{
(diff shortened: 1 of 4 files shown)


corp-0:
how do players set their ringtone now? A dropdown? How can they input the uplink code now?

UnityProject/Assets/Scripts/UI/Items/PDA/GUI_PDASettingMenu.cs
```diff
@@ -19,7 +19,7 @@ public void OnPageActivated()
 
 		public void SetRingtone(string ringtone)
 		{
-			if (ringtone == default) return;
+			if (ringtone == "") return;
```
corp-0: I thought ``default`` for strings was ""

corp-0: another way to check it would be
```cs
if (ringtone.IsNullOrEmpty()) return;
```

Bobtheapple: It wasn't working as "== default", updated to use IsNullorEmpty

Bobtheapple:
Nothing has changed in the interface, players still type in their desired ringtone and can put in the uplink code.  If the player puts in a ringtone that isn't in their PDA's ringtones list, the ringtone is not changed, this is to prevent players from using any sound effect in the game as their ringtone.

corp-0:
cool cool. Now I wonder if using literally any sound in the game was intended or an overlook.

Bobtheapple:
I can set it that way if that's what we want, or have some PDAs have access to all sounds, but I don't think it was like that on tgstation.

corp-0:
fair enough, let's go with the restricted list first and discuss on it.

NoooneyDude:
> cool cool. Now I wonder if using literally any sound in the game was intended or an overlook.

was intended, hah (with the intention of limiting later)

UnityProject/Assets/Scripts/UI/Items/PDA/GUI_PDASettingMenu.cs
NoooneyDude: Yeah so it turns out `default` for string is actually `null`, not an empty string. Wonder where else I've assumed that.

corp-0: :eyes:

PR #6083, opened 2021-02-21
Updates Mirror

### LINKED ISSUE #5765: Updating Mirror
All the network messages will now need to be structs, see example:

https://mirror-networking.com/docs/Articles/Guides/Communications/NetworkMessages.html

Our current system of using the classes; ClientMessage/ServerMessage will need changing as they inherit from NetworkMessage.

Also the custom changes we made to mirror need to be kept in. To see the custom changes download the current version of mirror that we use: https://github.com/vis2k/Mirror/releases/tag/v11.4.2.  Then use a program like winmerge to see the changes with our current version stored in the Standard Assets folder and try to port the changes over to the version of mirror we are upgrading to.

Updating mirror is necessary if we want to use any version of unity after 2020.1

Though if we wanted a hacky fix so that mirror works in the next versions without a full update then this fix could be ported over:
https://github.com/vis2k/Mirror/pull/2497

Though since mirror is quite out of date it would be preferable if it was properly updated instead.

ThatDan123:
CL: Updates Mirror to github version 32.0.0

-All net messages are now structs, all 154 of them

fixes #5765

Works in editor, just build testing now

~Also completely removed all the custom unity station code in the mirror code, i can add it back but we might be ok without it~

Custom code has been kept for now, might be able to removed when we try the prox stuff. 

And should all be labelled with:

//CUSTOM UNITYSTATION CODE//


corp-0:
what did our custom code do?

ThatDan123:
im actually readding it as it seems like it is necessary, its for observers and stuff

corp-0:
the new version was added to UnityProject/Assets ? I thought packages were only cached in a particular folder where the package manager install stuff.

ThatDan123:
ah my mistake its not a proper package

ThatDan123:
~Current problems is that it takes for ever to join a headless server as client and freezes, everything is also invisible and theres errors for spawning in prefabs~

corp-0:
oof

ThatDan123:
Btw codacy just is gonna hate this no matter what, as theres stuff in the mirror code its complaining about

ThatDan123:
Built tested with headless and clients, all seems good. Ready to be reviewed.

ThatDan123:
If anyone has time could they build as well as this is quite a critical update so should be thoroughly tested.

corp-0:
it is hard to review because of the amount of files and mirror and our files are together in the same commit.

UnityProject/Assets/Scripts/Messages/Server/AdminTools/AdminChatNotifications.cs
```diff
@@ -10,27 +10,36 @@
...
+		public AdminChatWindow TargetWindow;
+		public int Amount;
+		public bool ClearAll;
+		public bool IsFullUpdate;
+		public string FullUpdateJson;
+	}
 
-	public override void Process()
+	//This is needed so the message can be discovered in NetworkManagerExtensions
+	public AdminChatNotificationsNetMessage IgnoreMe;
```
Bod9001: so, couldn't you abstract this away  setting up ServerMessage<T>
When you decorate it you just have to do AdminChatNotifications : ServerMessage<AdminChatNotificationsNetMessage>
public AdminChatNotificationsNetMessage IgnoreMe;
And that gets turned into public T IgnoreMe; in ServerMessage/base message/whatever

ThatDan123: bit confused what you mean, that variable is there so i can link the struct to the class. I dont think you'd be able to abstract it away. I am no c# wizard, if you can give an example, i might be misunderstanding.

ThatDan123: yeah sorry i cant read, am tired. Dont think i could have abstracted that, i think having that empty variable is the only way to detect it for the handler as i need that specific type. Its a bit convoluted but it means that we can have generic message classes and avoid some duplicate code.

Bod9001: ok cool

ThatDan123:
DNM until after the next update, people can work on scenes if they want/need to, its just a one click fix i need to do so dont worry about making this have conflicts.

Copern:
There is a way to detect the type within NetworkManagerExtensions, avoid the casting within Process<T>, and remove the need to iterate the fields and have an ignored instance variable. As it is, `Process<T>` is kind of misusing and defeating the point of generics as you are casting to the concrete message type in the concrete classes. 

It will require a little bit of work, so apologies for the incoming wall of text. First change `GameMessageBase : MessageBase` to `GameMessageBase<T>: MessageBase where T : NetworkMessage` (may want to constrain it to `struct` too). Do similar for ClientMessage and ServerMessage. In GameMessageBase, Process can then be changed to `public virtual void Process( NetworkConnection sentBy, T msg )` and PreProcess to `public virtual void PreProcess(NetworkConnection sentBy, T b)`. 

Now for example, AdminBwoinkMessage can be changed to `public class AdminBwoinkMessage : ServerMessage<AdminBwoinkMessage.AdminBwoinkMessageNetMessage>`. To avoid the extra long generic argument you have a couple options. Use `using static AdminBwoinkMessage` at the top or rename AdminBwoinkMessageNetMessage to just NetMessage since it is already nested within a class. Now Process in `AdminBwoinkMessage` can be `public override void Process(AdminBwoinkMessageNetMessage msg)` or better `public override void Process(NetMessage msg)` if you go that route. 

In NetworkManagerExtensions, instead of iterating the fields, you can get the BaseType of the type. You can then call GetGenericArguments on that base type and, in this case, use the first type. It should give you the full type, such as `AdminBwoinkMessage.AdminBwoinkMessageNetMessage` (or `AdminBwoinkMessage.NetMessage`) that can be used to make the generic method with. Note that MakeGenericType accepts multiple parameters so you could use `mi.MakeGenericMethod(type, netMessageType)`. Now RegisterHandler could be `public static void RegisterHandler<T, U>(bool isServer) where T : GameMessageBase<U>, new() where U : NetworkMessage, new()`. Hopefully this helps you and thanks for going through the trouble of updating to the new Mirror.

ThatDan123:
Maybe im just misunderstanding, first time ive done anything with generics, but now that NetworkMesage has to be inherited by a struct and cant be abstract in the new mirror version so why would this work?
```GameMessageBase<T>: MessageBase where T : NetworkMessage```  as GameMessageBase is a class and abstract?

Also MessageBase no longer exists, was replaced by NetworkMessage

Copern:
Apologies, I was partly looking at what was already in unitystation and didn't see when you had made the change to the GameMessageBase class. So instead of what I had before it would be `public abstract class GameMessageBase<T> where T : struct, NetworkMessage` or the appropriate adjustment based on what you have now. ClientMessage and ServerMessage would change also.

ThatDan123:
thx for the help, good thing this isn't needed anytime soon. 

Gonna take me another 3 hours to go through all those messages, the price of having good code :p

Copern:
Feel free to drop me a message in discord if you run into an issue when switching the message classes to generics.

ThatDan123:
This is where the fun begins: 

image

ThatDan123:
I'll only cry a bit

image

ThatDan123:
Really strange but the asset store version that i try to import is actually the complete wrong version, even though when i import it says its the right version. So i got the new version from the mirror github instead, big problem though is that for some reason it will just fail to load a scene on start and will stop playmode but only sometimes.

Bod9001:
odd so just got to fix that and secns and it should be good?

ThatDan123:
i managed to fix that issue, but i found another while testing: 

When spawning stuff in it seems to move another object to the spawn location, and ive got no idea why thats happening

Bod9001:
could be that net ID overlaps?

Bod9001:
humm , so shall we merge this before health?

ThatDan123:
Up to you, are we ready to push the proper build?

ThatDan123:
Still got the issue of when objects spawn they sometimes move an object thats already in the scene.

ThatDan123:
Think ive fixed the issue now, so should be good to merge whenever

PerfectTangent:
this won't overwrite previous mapping changes, will it? Just want to make sure.

ThatDan123:
It wont override

PR #6093, opened 2021-02-22
Proposal: Sound Manager Rework

Bobtheapple:
This is a proposal for a possible rework of SoundManager.cs and all code that calls it.  Since this is a stylistic choice, it is ultimately a matter of taste.  As such, I'll leave it to the reviewers to determine whether this is something we want or not.

This code does not address any bug (though see "Followup" below), nor does it change functionality in anyway.  It also will not change memory usage, as SoundManager always uses AudioParameters and ShakeParameters anyway, and they have been converted into structs now.

### Purpose
Currently, SoundManager is called in a wide variety of ways, and uses overload methods to handle all of the different calls.   SoundManager needs overloads to handle code calling it with:
 - The sound with a pitch change and values for the shake to play
 - The sound with a pitch change and ShakeParameters
 - The sound with AudioParameters and values for the shake to play
 - The sound with AudioParameters and ShakeParameters
 - A list of sounds with a pitch change and values for the shake to play
 - A list of sounds with a pitch change and ShakeParameters
 - A list of sounds with AudioParameters and values for the shake to play
 - A list of sounds with AudioParameters and ShakeParameters

This PR simplifies this by requiring code calling SoundManager to always declare an AudioParameter or ShakeParameter struct before using SoundManager (if changes those parameters are desired), cutting down to just two cases per method, one for just the sound and one for lists.

### Why would we want this?
The main advantage of this change is the idea of deliberate coding, writing code that makes the reader know for sure what you are doing is on purpose.  Take for example the old code:
`SoundManager.PlayNetworkedAtPos(Spray2, startPos, 1, sourceObj: interaction.Performer);`

Spray2 already has a pitch of 1.  It is unclear whether the writer was trying to make sure that this always will play Spray2 with a pitch of 1 because it had a different pitch in the past or might have a different pitch in the future, or whether the old code required a pitch declaration, or whether the writer thought they needed to declare some pitch, and 1 is a good default.  Compare that to the new code:

`AudioSourceParameters audioSourceParameters = new AudioSourceParameters(pitch: 1);`
`SoundManager.PlayNetworkedAtPos(Spray2, startPos, audioSourceParameters, sourceObj: interaction.Performer);`

And it becomes clear that the writer yes, deliberately wanted the pitch to be set to 1, and that to be the only change to the sound.  Why else would they declare an AudioSourceParameters to do so?  (The author adding a comment explaining why they wanted it to be 1 would a good inclusion).

Or this code:
`SoundManager.PlayNetworked(SingletonSOSounds.Instance.ClownHonk, Random.Range(0.2f,0.5f),true,true);`

At first glance, it isn't even clear what this is doing.  The first 'true' is for the boolean polyphonic, the second true is for a boolean "shakeGround", to turn on shaking for the sound.  It can be followed by two optional parameters that set the intensity and radius of the shake, which the writer did not define, leaving them to the default 64 and 30 (you would only know that's what the values are for the shake if you dove into SoundManagers web of method overloads though).  Presumably that is what the writer wanted, but there is also some other explanations.  SoundManager.Play() has a boolean for "global" between polyphonic and shakeGround, maybe the writer thought they were defining that.  Or maybe the writer wanted a different shake, but didn't know what parameters they wanted yet so left it, but because it is so easy to miss that second true in the statement, it was forgotten and no one went to put in values for it.  Compare this to the new code:

`AudioSourceParameters audioSourceParameters = new AudioSourceParameters(pitch: Random.Range(0.2f,0.5f));`
`ShakeParameters shakeParameters = new ShakeParameters(true, 64, 30);`
`SoundManager.PlayNetworked(SingletonSOSounds.Instance.ClownHonk, audioSourceParameters, true, shakeParameters);`

This makes it very clear what the writer wanted to do, and the shake values they are using.  If the shake values are too large/too small, it is very easy to find the source and fix it.

To use the new code, it requires a couple of extra lines of code to use SoundManager, but this is only true if you are changing the AudioSourceParameters or ShakeParameters, otherwise the calls are unchanged.  This also serves as a way to ask the writer "Are you sure you want to change the pitch from the default?"  To cut down on hard coded pitch changes.

### Other Changes
 - Pitch was the most common call of AudioSourceParameters, so I moved it to the first variable of the constructor and made the rest optional, so one can write AudioSourceParameters(1) to get a pitch changed to 1.
 - MusicLibrary was removed from SoundManager.  Nothing was using this code, and it wasn't working anyway (the list of music wasn't being populated).  If this is something we need, it should be handled by MusicManager probably.
 - Made an overload for GetAddressableAudioSourceFromCache to let it handle non-lists.
 - PlayAtPosition had some weirdness, I separated it into two methods for now, one the static void, now called PlayAtPositionAttached, and the other the task, and I changed the parameter declarations to be consistent.

### Followup
Before merging this PR, it is a good idea if we could go through and make sure that all of the hardcoded pitch change declarations are intentional.  This is also a good idea if the PR is rejected.  This PR only updates all of the code to use the new standard, it did not change any SoundManager calls.  The ones that jumped out at me in particular are the following:
 - MakeCommandReport in CentComm.cs is setting the pitch to 1 of all update sounds, regardless of type.
 - TableBuilding.Assemble.cs is setting pitch to 1 for all assembly sounds, regardless of type.
 - CustomNetTransform.Move.OnHit.cs is setting pitch to 1 for all hits, regardless of type
 - Horn.cs and ExplosionUtils.cs have calls that set the pitch to -1.  This will play the sound backwards (or not at all, depending on Unity version)


corp-0:
I like the proposal and would like to see it merged.

> Horn.cs and ExplosionUtils.cs have calls that set the pitch to -1. This will play the sound backwards (or not at all, depending on Unity version)

more than likely a left over of -1 being the way to tell the method to use the default pitch of the sound.

You seem to have unrelated commits.

UnityProject/Assets/Scripts/Core/Transform/CustomNetTransform.Move.cs
```diff
@@ -6,6 +6,7 @@
 using Objects;
 using Objects.Disposals;
 using Random = UnityEngine.Random;
+using Assets.Scripts.Messages.Server.SoundMessages;
```
corp-0: all scripts are under ``Assets/Scripts`` so you can get rid of that part when doing namespaces.

UnityProject/Assets/Scripts/Health/Living/HumanHealth/PlayerHealth.cs
```diff
@@ -356,9 +357,11 @@ protected override void PainfulElectrocution(Electrocution electrocution, float
 		// TODO: Add sparks VFX at shockSourcePos.
 		SoundManager.PlayNetworkedAtPos(SingletonSOSounds.Instance.Sparks, electrocution.ShockSourcePos);
 		Inventory.ServerDrop(itemStorage.GetActiveHandSlot());
+		
 		// Slip is essentially a yelp SFX.
+		AudioSourceParameters audioSourceParameters = new AudioSourceParameters(pitch: UnityEngine.Random.Range(0.9f, 1.1f));
```
corp-0: you changed the range in this instance.

UnityProject/Assets/Scripts/Core/Transform/CustomNetTransform.Move.cs
Bobtheapple: That's the namespace that AudioSourceParameters is using, weirdly.  Should I change that?

corp-0: yes. The person who did it probably uses VS and VS auto suggest the whole path.

UnityProject/Assets/Scripts/Health/Living/HumanHealth/PlayerHealth.cs
```diff
@@ -356,9 +357,11 @@ protected override void PainfulElectrocution(Electrocution electrocution, float
 		// TODO: Add sparks VFX at shockSourcePos.
 		SoundManager.PlayNetworkedAtPos(SingletonSOSounds.Instance.Sparks, electrocution.ShockSourcePos);
 		Inventory.ServerDrop(itemStorage.GetActiveHandSlot());
+		
 		// Slip is essentially a yelp SFX.
+		AudioSourceParameters audioSourceParameters = new AudioSourceParameters(pitch: UnityEngine.Random.Range(0.8f, 1.2f));
```
ThatDan123: range change here too

UnityProject/Assets/Scripts/Health/Living/HumanHealth/PlayerHealth.cs
```diff
@@ -406,8 +409,10 @@ private IEnumerator ElectrocutionSequence()
 		yield return WaitFor.Seconds(timeBeforeDrop); // Instantly dropping to ground looks odd.
 													  // TODO: Add sparks VFX at shockSourcePos.
 		registerPlayer.ServerStun(ELECTROCUTION_STUN_PERIOD - timeBeforeDrop);
+
+		AudioSourceParameters audioSourceParameters = new AudioSourceParameters(pitch: UnityEngine.Random.Range(0.9f, 1.1f));
```
ThatDan123: and here

Bobtheapple: Fixed, thanks

Bobtheapple:
Sorry about the weird list of commits, I was working on this off and on while I was doing other fixes, and my coding style doesn't jive well with GitHub, so I have commits to undo commits and merges and weird stuff... is there a good way to clean that up to make things clearer?

I don't think Honk.cs has been changed since I started working on SoundManager, I was the one that changed the default to -1 from null briefly, and that was only like, a day or so before I updated to a struct.  My guess is they wanted a deep, low pitch honk there for the 'critical honk'.

corp-0:
>  is there a good way to clean that up to make things clearer?

I think it is fine this time since we already merged your other changes and I can just squash this pr. In the future try to make fresh new branches for every new feature/fix you intend to pr.

corp-0:
>  My guess is they wanted a deep, low pitch honk there for the 'critical honk

This is probably it. Maybe that's why I was confused when I asked about doing deep sounds using negative numbers. In my mind a pitch value of ``0`` meant whatever original pitch was and negative numbers made the pitch go into lower frequencies, as in adding the value to the sound pitch. Probably because that's how it works in most DAWs.

corp-0:
test are failing apparently because compiling errors.

Bod9001:
as long as you can play a sound without having to specify AudioSourceParameters  if you just wanted to play with default parameters

Bobtheapple:
Yeah, AudioSourceParameters are an optional parameter.  For day-to-day uses of SoundManager, where you just want to play the sounds as they are in the prefab, nothing has changed, you don't have to include anythign or write any extra code.

Bobtheapple:
How pitch works in Unity is that how the audioclip sounds normally is equal to 1, if you want to shift up an octave by one you multiply by 2, down an octave you divide by 2.  To change the tone or key, you multiply or divide by 1.059463 (or to be super accurate, 2 root 12) per semitone shift (this is because there are 12 semitones in an octave, and 1.059463^12 = 2).  Thus for example, the mouseAI code for its squeak gives it a pitch of Random.Range(.6f, 1.2f), which means it can squeak in a range of 1 full octave, centered (roughly) 3 semitones lower than the soundclips default.

So if we wanted the critical honk to be 1 octave lower, it would be a pitch of 0.5, two octaves would be 0.25, etc.

corp-0 (approved):

PR #6095, opened 2021-02-22
Soundmanager LogError change

Bobtheapple:
Updated SoundManager's Log Errors to be more verbose and cover more situations.

### Purpose
Currently, SoundManager Error Logging does not handle situations when the Addressable Audio Sources it recieves:
- Are a non-null, default Addressable Audio Source
- Have a null address
- Have an address set to the string "null"
- Are sent in the form of a list

This PR changes it to cover those scenarios.


Bod9001:
would this collide with the other PR?

Bobtheapple:
Yes, but the other PR wont be done for a little while.  Gilles asked me to run this one out separately so we can get it up faster.

PR #6096, opened 2021-02-22
Overflow protection for tile damage

ThatDan123:
Just in case theres a loop, where a tile is spawned and then spawns the one that spawns itself


corp-0:
> a tile is spawned and then spawns the one that spawns itself

Mind = blown, I did not understand what is happening there

ThatDan123:
I dont think theres a loop problem but for example, floor1 turns into floor2 when destroyed, but then floor2 accidently is set to turn to floor1 if it is damaged enough. And somehow even though the damage should reduce it still forever loops.

ThatDan123:
But i dont think this is the crash issue, i did double check the tiles i set up.

corp-0:
why is the limit 10?

ThatDan123:
10 is a nice number, it's divisible by 2, and 5. :p

In all seriousness, i dont think theres a tile damage route which is bigger than like 6 which i think it what RWalls is.

I can change it to a const if you think i should.

corp-0:
yes, that way the number is identified and we can easily find it and change it if we ever do something that should exceed that limit.

ThatDan123:
cool, done

PR #6099, opened 2021-02-23
Fixes ghost positioning related interactions

Aranclanos:
### Purpose
The server will now always update its own values of PlayerSync.Client, fixing ghost serverside positioning and its position related interactions.

PlayerSync is a mess serverside with what it claims as clientside only variables being a key into the whole system to work, let alone the class being almost 2400 lines long inbetween three files of partial classes.
It might need a rewrite or rework

its important to notice that the clientside of PlayerSync changes the transform.position of an object, which is fed into the registertile, making it a serverside alteration


corp-0:
I wasn't expecting one line change with that description

Aranclanos:
I guess its just ranting

PR #6106, opened 2021-02-23
ExplosionNode Violating Thermodynamics Fix

Bobtheapple:
Currently, ExplosionNode.cs is written so if an explosion does not expend energy by damaging a tile or object, it continues to expand outward with the same energy forever. This spams the server with TileMapDamage.DealDamageAt calls after an explosion, as each tile it passes through is checked to see if it can be damaged.  A test reactor explosion on Square Station resulted in 388012 calls, only 15600 of which were non-null tiles.

### Purpose
This PR attempts to alleviate this problem by implementing a minimum energy expenditure per tile expanded on an explosion.  With the variables currently set, this gives all tiles a minimum hp of 1 for purposes of explosions.  This reduces the number of calls of TileMapDamage.DealDamageAt with the same explosion to 119828, 8818 are non-null, cutting the tiles checked by 70% and the number of sounds attempting to play by half.  Examining the explosion pattern, the change in explosion radius is actually very minor, many of the explosion lines sent out have little damage remaining, see attached image (warning, high res).
ExplosionNodeComparison

The explosion still lags the server, but this should reduce it significantly, maybe preventing crashes.  I've tested other numbers as well, but obviously as you increase the minimum expenditure, the explosion radius goes down as well.  Here is the result of setting it to `if(EnergyExpended <= 0.375f) EnergyExpended = 10f;`
ExplosionNode10
With these settings, the server was barely affected, but at the cost of a much smaller explosion.

Also included some error logging tweaks.

### Physics
As an aside for those who may be wondering about the realism of this change, while objects in space do not change velocity over time in space (and that code is unaffected by this, any object propelled through space by the explosion will move at its speed until it hits something else), explosions will degrade rapidly as they expand into a vacuum, as the gas that is expanding (the source of the concussive blast and conveyor of thermal energy) becomes extremely diffuse as it reaches equilibrium with the vacuum.
One could argue that the explosion's damage value is also encompassing fast moving tiny debris like screws and glass shards not otherwise covered by the code, but one could then also argue that as you get farther from the explosion, the probability of you being hit by that debris falls off, and the easiest way to factor that is to have damage fall off over distance.

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/SoundManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/SoundManager.cs
@@ -118,12 +118,12 @@ public class SoundManager : MonoBehaviour
 		}
 		if (string.IsNullOrEmpty(addressableAudioSource.AssetAddress))
 		{
-			Logger.LogWarning($"{addressableAudioSource.AudioSource.name} has a null address, and could not be played!", Category.Addressables);
+			Logger.LogWarning("SoundManager received a null address for an addressable, look at log trace for responsible component", Category.Addressables);
 			return null;
 		}
 		if (addressableAudioSource.AssetAddress == "null")
 		{
-			Logger.LogWarning($"{addressableAudioSource.AudioSource.name} has an address set to the string null, and could not be played!", Category.Addressables);
+			Logger.LogWarning("SoundManager received an addressable with an address set to the string 'null', look at log trace for responsible component", Category.Addressables);
 			return null;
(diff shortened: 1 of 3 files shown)


Bobtheapple:
Probably should also have a minimum damage threshold for playing sounds now that I think of it, probably will add that next.  
Edit: Added it, set it to 1 because I think that's the smallest amount of damage a normal weapon can do.

PR #6109, opened 2021-02-24
fixes incorrect plasma damage values (and also adds gas mask functionality and carbon dioxide suffocation)

VernaSaurus:
_probably_ works, i wasn't able to test it very well because unity is a bitch


UnityProject/Assets/Scripts/Items/Traits/CommonTraits.cs
```diff
@@ -69,6 +69,7 @@ public class CommonTraits : SingletonScriptableObject<CommonTraits>
 	[BoxGroup("Materials")] public ItemTrait OrePlasma;
 
 	[BoxGroup("Clothing")] public ItemTrait Mask;
+	[BoxGroup("Clothing")] public ItemTrait GasMask;
```
NoooneyDude, line 72: Instead of adding this trait, which is rather uncommonly used, to the `CommonTraits` SO, I suggest serializing reference to it directly in `RespiratorySystem`.

VernaSaurus: aighty, will do

NoooneyDude:
I'm thinking, it isn't very sustainable adding all the different gas effects we want into the same class. Some sort of abstraction here would be ideal to make it scalable.

Have you considered something to this end?

UnityProject/Assets/Scripts/Items/Traits/CommonTraits.cs
VernaSaurus: ...actually, there are a bunch of significantly less common items that use itemtraits that are placed in CommonTraits, like "Emag" and "Hatchet," both of which are only used by one item. GasMask is used by a total of 5 items

VernaSaurus:
not really i kinda just wanted to make a quick fix and be done with it

UnityProject/Assets/Scripts/Items/Traits/CommonTraits.cs
NoooneyDude: Yes, and it annoys me to no end ; )

NoooneyDude: GasMask is used by 5 items? Are you sure? I could only find two references here.

VernaSaurus: ...it occurs to me that I was also the one to add _both_ of those traits... man, i really am a shitcoder huh

VernaSaurus: > GasMask is used by 5 items? Are you sure? I could only find two references here.

i made sure all five gas masks were using the GasMask trait

NoooneyDude: Oh right, but only one script.

NoooneyDude: Speaking of that - do the items not use inheritance? Why can't you get away with assigning it to just the base prefab?

VernaSaurus: they don't use inheritance which is super obnoxious but i didn't feel like changing all of them at the time

NoooneyDude: Ah I see 👍. That's a little annoying for you. All good then.

corp-0:
> I'm thinking, it isn't very sustainable adding all the different gas effects we want into the same class. Some sort of abstraction here would be ideal to make it scalable.
> 
> Have you considered something to this end?

like previous plasma damage, this is just a facade of proper breathing simulation until we get new helf and work on that system instead.

UnityProject/Assets/Scripts/Health/RespiratoryHealth/RespiratorySystem.cs
```diff
@@ -224,14 +298,48 @@ private float HandleBreathing(GasMix gasMix)
...
+		plasmaConsumed = gasMix.GetMoles(Gas.Plasma) * AtmosConstants.BREATH_VOLUME;
+
+		// the player is wearing a gas mask and there are less than 25 moles of plasma in the air
+		if (gameObject.Player() != null && gasMix.GetMoles(Gas.Plasma) < 25)
+		{
+			if (gameObject.Player().Script.Equipment.GetClothingItem(NamedSlot.mask).gameObject.GetComponent<ItemAttributesV2>() != null)
+			{
+				if (gameObject.Player().Script.Equipment.GetClothingItem(NamedSlot.mask).gameObject.GetComponent<ItemAttributesV2>().HasTrait(CommonTraits.Instance.GasMask))
+				{
+					if (DMMath.Prob(90))
```
corp-0: extract from here into its own method so you can reuse. Extract the ifs into a single bool property so you can simply do
```cs
if (isWearingGasMask)
{
    HandleGasMask(gasMix); 
    return;
}
```

corp-0:
I feel like gas mask messages and logic should go in the gas mask itself

corp-0:
added a DNM since this is still wip

VernaSaurus:
should be ready to test and merge

UnityProject/Assets/Scripts/Health/RespiratoryHealth/RespiratorySystem.cs
```diff
@@ -224,14 +321,19 @@ private float HandleBreathing(GasMix gasMix)
...
 	/// </summary>
 	/// <param name="plasmaAmount"></param>
-	private void HandleBreathingPlasma(float plasmaAmount)
+	private float HandleBreathingPlasma(GasMix gasMix)
 	{
+		float plasmaPressure = gasMix.GetPressure(Gas.Plasma);
+
+		float plasmaConsumed = 0;
+
+		plasmaConsumed = gasMix.GetMoles(Gas.Plasma) * AtmosConstants.BREATH_VOLUME;
```
LemonInTheDark, line 330: If I'm understanding what you're doing here correctly it would most likely be cleaner to make the input gasmix scale, rather then doing it each time you want to consume a gas.

UnityProject/Assets/Scripts/Health/RespiratoryHealth/RespiratorySystem.cs
```diff
@@ -268,6 +370,177 @@ private void HandleBreathingPlasma(float plasmaAmount)
...
+		{
+			// if there is too much CO2 in the air, but there isn't too much plasma in the air
+			if (gasMix.GetMoles(Gas.Plasma) < 25)
+			{
+				GasMix gasMix2 = gasMix;
+				gasMix2.RemoveGas(Gas.CarbonDioxide, 30);
+				HandleBreathingCarbonDioxide(gasMix2);
+				return true;
+			}
+			else if (gasMix.GetMoles(Gas.Plasma) >= 25)
```
LemonInTheDark: Isn't the check here redundant?

UnityProject/Assets/Scripts/Health/RespiratoryHealth/RespiratorySystem.cs
VernaSaurus: this is simply a copy of what oxygen inhalation does. it takes the current amount of the gas and multiplies it by how big your lungs are. it does this for both oxygen and plasma.

UnityProject/Assets/Scripts/Health/RespiratoryHealth/RespiratorySystem.cs
```diff
@@ -268,6 +370,177 @@ private void HandleBreathingPlasma(float plasmaAmount)
...
+				{
+					gasMix.RemoveGas(Gas.Plasma, plasmaBreathedWithMask);
+					registerTile.Matrix.MetaDataLayer.UpdateSystemsAt(registerTile.LocalPositionClient, SystemType.AtmosSystem);
+				}
+				HandleBreathingCarbonDioxide(gasMix2);
+				return true;
+			}
+		}
+		// if there is too much plasma in the air
+		if (gasMix.GetMoles(Gas.Plasma) >= 25)
```
LemonInTheDark, line 467: Won't both this block and the one above it run if there's too much plasma and co2 in the air? Why do the check on the inside in the first place? Can't you just check plasma, do the plasma things, then check co2 and do the co2 things?

UnityProject/Assets/Scripts/Health/RespiratoryHealth/RespiratorySystem.cs
VernaSaurus: good point, i'll remove the redundancies.

UnityProject/Assets/Scripts/Health/RespiratoryHealth/RespiratorySystem.cs
VernaSaurus: i believe you missed the "return" present in the block above. but you do make a good point.

UnityProject/Assets/Scripts/Health/RespiratoryHealth/RespiratorySystem.cs
LemonInTheDark: I'll admit it's a bit hard to tell since I don't know the codebase, but `HandleBreathingPlasma` expects the gasmix we're breathing from right? Like the tile or gas tank? Why not scale that gasmix before passing it in, so we can treat it as the whole "breath" without needing to worry about scaling it?

VernaSaurus: mainly due to the pressure thing. the "pressure" part is necessary to figure out whether you should die or not, hence it should calculate from the tile or tank's pressure. however, the amount consumed by breathing needs to be scaled.

VernaSaurus:
alright, all of lemon's suggestions have been resolved

UnityProject/Assets/Scripts/Health/RespiratoryHealth/RespiratorySystem.cs
LemonInTheDark: Ah I see you're handling the damage to the body/lungs in the same bit as the breathing.

VernaSaurus: The reason this is so... unrefined is because this is largely a placeholder until actual body parts are implemented. For now this will have to do.

VernaSaurus:
this just in, if this goes through then our gas masks will officially be superior to /tg/'s

UnityProject/Assets/Scripts/Items/Traits/CommonTraits.cs
corp-0: I thought you were going to remove the trait from the common traits list.

VernaSaurus: i'm not entirely sure where to put it

LemonInTheDark:
> 
> 
> this just in, if this goes through then our gas masks will officially be superior to /tg/'s

reeeeeeeeeeeeeeeeeeee

UnityProject/Assets/Scripts/Items/Traits/CommonTraits.cs
corp-0: Honestly I don't think it is needed at all. You could instead of getting the trait, try to get the gas mask component from the clothing item in gas slot like this ``TryGetComponent<GasMask>(out var gasMask)``. Add all those messages to that component and let the gas mask handle the filtering of the air, so you could change effectiveness of a particular gas mask or custom messages or whatever. I can guide you on discord on doing that.

VernaSaurus: there is no gas mask component

UnityProject/Assets/Scripts/Health/RespiratoryHealth/RespiratorySystem.cs
LemonInTheDark: F

UnityProject/Assets/Scripts/Items/Traits/CommonTraits.cs
VernaSaurus: i might add that in a later PR but for now i'm just going to be done with this

UnityProject/Assets/Scripts/Health/RespiratoryHealth/RespiratorySystem.cs
```diff
@@ -268,6 +370,146 @@ private void HandleBreathingPlasma(float plasmaAmount)
...
+		}
+
+		if (carbonPressure < CARBON_DIOXIDE_WARNING_LEVEL)
+		{
+			CarbonSuffocationTimer = 0;
+			return (carbonPressure > 0);
+		}
+		return false;
+	}
+	private bool HandleWearingGasMask(GasMix gasMix)
```
corp-0, line 455: what does the return value mean for this method?

UnityProject/Assets/Scripts/Health/RespiratoryHealth/RespiratorySystem.cs
```diff
@@ -268,6 +370,146 @@ private void HandleBreathingPlasma(float plasmaAmount)
...
+		{
+			return true;
+		}
+
+		if(gasMix.GetMoles(Gas.Plasma) == 0 && gasMix.GetMoles(Gas.CarbonDioxide) == 0)
+		{
+			return false;
+		}
+
+		if (DMMath.Prob(90))
```
corp-0, line 492: this bit is weird, why are you returning !filtered?

UnityProject/Assets/Scripts/Health/RespiratoryHealth/RespiratorySystem.cs
VernaSaurus: the return value for HandleWearingGasMask or the value for HandleBreathingCarbonDioxide?

for HandleWearingGasMask, the return value tells the breathing script whether or not to go through with running the other checks.
for HandleBreathingCarbonDioxide, it's a bool that's used in the rare case that _only_ carbon dioxide is breathed in, to tell the rest of the program that you did, in fact, breathe that cycle

UnityProject/Assets/Scripts/Health/RespiratoryHealth/RespiratorySystem.cs
VernaSaurus: you make a good point, seems i mistakenly told it to return the wrong value. i will remedy this really quick

VernaSaurus: thanks for catching that

VernaSaurus:
ok NOW it should be good to go

PR #6118, opened 2021-02-25
Radiation penetration is now affected by tile stats and bundle is given a unique name each build to prevent collisions

Bod9001:
yay thread safe tile access

local bundles are given a unique name to prevent cached collisions tho it did break again but restarting the built client fixed it for some reason

NOTE: someone needs to go through and set appropriate values for tiles  I did it for the basic walls and reinforced walls and plasma windows

### Changelog:
CL: radiation now penetrates differently depending on the tile

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Tiles/BasicTile.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Tiles/BasicTile.cs
@@ -34,6 +34,10 @@ public abstract class BasicTile : LayerTile
 	[Tooltip("Can this tile be mined?")] [FormerlySerializedAs("Mineable")] [SerializeField]
 	private bool mineable = false;
 
+
+	[Tooltip("RadiationPassability 0 = 100% Resistant")] [SerializeField]
+	public float RadiationPassability = 1;
+
 	/// <summary>
 	/// Can this tile be mined?
 	/// </summary>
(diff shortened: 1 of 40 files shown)


UnityProject/Assets/Scripts/Tilemaps/Tiles/BasicTile.cs
```diff
@@ -34,6 +34,10 @@ public abstract class BasicTile : LayerTile
 	[Tooltip("Can this tile be mined?")] [FormerlySerializedAs("Mineable")] [SerializeField]
 	private bool mineable = false;
 
+
+	[Tooltip("RadiationPassability 0 = 100% Resistant")] [SerializeField]
+	public float RadiationPassability = 1;
```
corp-0, line 39: should the range be from 0 to 1 then?

Bod9001: humm You can't have a range for floats sadly has to be int?

corp-0: https://docs.unity3d.com/ScriptReference/RangeAttribute.html according to this you can

Bod9001: must have forgot to included the float value 0,1 doesn't work unless you have 0f, 1f

Issue #5906, opened 2021-01-27
Add sound to food processors. 


Lennoxite:
Right now food processors are fully functional gameplay wise, but are lacking a sound and the shaking animation that occurs while it's actively processing food.

The basic functions for the animation are already in FoodProcessor.cs, but due to a LeanTween build error it was commented out. 

Right now there is a syncvar that changes to true while the processor is on and false while the processor is off, and is used to play a looping sound, which was taken from Microwave.cs. However in light of the new sound addressable system both food processors and microwaves do not play sounds while active, probably meaning the function needs to be adjusted to fit the new system.

- [x] add shaking animation
- [x] add sound effect

corp-0:
Animation is done, now it is missing the looped sound. AFAIK the generator has a looped sound so whoever wants to implement it can take a look there.

PetMudstone:
There is a sound that plays when using the food processor, just tested it.

PR #6119, opened 2021-02-25
Created a shake effect and added it to food processors.

MaxIsJoe:
### Purpose
Adds a shake effect that is flexible enough to be used by (almost*) anything. Also this has been added to the food processor as it [required animations that work](https://github.com/unitystation/unitystation/issues/5906).

### Notes:
(*) : I intended to make this script for stationary objects, meaning objects that won't move. While it does work to some degree on moving objects I did notice some minor bugs that I can't replicate now even though I have not changed the code's behavior in anyway since I experienced them. I added a note ontop of Shake.cs to point out that this is intended for non-moving objects so others can be aware of this and for anyone smarter than me in the future to fix these bugs if they ever occur again.

### Changelog:

- Created a shake effect. (Shake.cs)
- Added the shake effect to food processors.

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Objects/Kitchen/FoodProcessor.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Objects/Kitchen/FoodProcessor.prefab
@@ -129,6 +129,7 @@ GameObject:
   - component: {fileID: 680863957139992462}
   - component: {fileID: -5821624160305598259}
   - component: {fileID: -770403792292043468}
+  - component: {fileID: 8428099613412960887}
   m_Layer: 11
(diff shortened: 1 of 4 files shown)


corp-0:
did you successfuly test this with a server and a client? The person who did the food processor had troubles trying to network the animation before.

MaxIsJoe:
> did you successfuly test this with a server and a client? The person who did the food processor had troubles trying to network the animation before.

Just tested it out, yes it works.
if anyone wants to double test this to confirm that it works that would be great as well.

MaxIsJoe:
Tested and works. Ready to merge 👍

PR #6120, opened 2021-02-25
PKA shot adjustments. CMO radio headset fix.

PetMudstone:
### Purpose
- Fixes the available channels in the CMO's headset.
- PKA shots converted to be normal (not-hitscan) projectiles instead of beams. This means PKAs can be used to mine again.

### Notes
- Since PKA shots are now projectiles, the pressure based damage scripts no longer work properly. PKA shot damage reduced to prevent players from using PKAs to murder people in pressurized enviroments for now.


corp-0:
10 damage is reasonable tho cargo won't be happy when they findout they can't open weapon crates in the void anymore.

PR #6128, opened 2021-02-26
Field Gen bug fix

ThatDan123:
Fixes being able to move around and activate unwelded generators

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Engineering/FieldGenerator.cs
+++ b/UnityProject/Assets/Scripts/Objects/Engineering/FieldGenerator.cs
@@ -272,7 +272,7 @@ namespace Objects.Engineering
 						break;
 					}
 
-					if (objects.Count > 0)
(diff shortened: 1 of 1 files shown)


corp-0:
Isn't there a wrenchable component that could handle this stuff?

ThatDan123:
There is but i decided not to use it, this issue still wouldve been there as i just forgot to add a check for if it was wrenched.

Bod9001:
does this fix a bug with being able to move the Field Gen  while on?

corp-0:
> There is but i decided not to use it, this issue still wouldve been there as i just forgot to add a check for if it was wrenched.

I just think standardizing common stuff like wrenching machinery/furniture is better. You could always expose a bool in the wrenchable component to get IsWrenched and keep all the rest.

corp-0:
alright I'm merging because we need to drop a new release with fixes so we can keep working with health in develop branch, but think about it. Standardizing good.

Issue #4995, opened 2020-08-30
Objects with the Directional Rotates Children component behave oddly.


PetMudstone:
## Description
Objects with the Directional Rotates Children component behave oddly in that their sprites seem to reset to a certain direction (apparently, the default direction defined in the inspector rather than altered directions done in scenes) under multiple circumstances. When you duplicate objects in a scene their sprites appear to be set to a different direction then they should be. However, they still maintain the same direction in terms of physics, if that makes sense. For example, a directional window can appear to be facing south but in reality it will be facing west and physically behave as such.

The directional changes also appear to happen after updates. This results in all the directional windows, windoors, and other objects (usually) facing south while in reality facing other directions.

### Steps to Reproduce
1.  In a scene, select a directional window that is not facing the same direction as its parent .prefab and duplicate it.
2.  Notice the new window's sprite appears to be facing south. despite the inspector stating otherwise.
3.  Go into play mode, find the weird window, notice how it behaves as though it were facing the direction the inspector says it is instead of as if it was facing the direction the sprite is.

CdrCross:
I suspect this is still at play. See this:
https://i.imgur.com/3Xna1Dq.gif

PR #6144, opened 2021-02-28
Windoor and Directional Firelock Rotation

### LINKED ISSUE #5778: Windors have not correct directions
image
image

Bobtheapple:
Went through and rotated all of the Windoors and directional firelocks on the remaining MainStations.
Fixes #5778
Note that #4995 is still an issue, I simply manually went through and properly rotated the children.


corp-0:
```
OutpostStation: "APC" has a null value in the list.
OutpostStation: "APC" has a null value in the list.
SquareStation: "APC - RND" has a null value in the list.
```
You have some apc's with null values

corp-0:
This is fine to merge as soon as it passes tests

Bobtheapple:
The null entries on the APCs appears to be unrelated to my changes.  My best guess was that some object was connected to the APCs in the past, and the object was removed (or its usage of the script was) but the APC was not updated.  When I rotated the windoors, the APC updated and started throwing the error.  Unfortunately, I have no idea what the object was.  For now, I just removed the null entries.

The APCs in question were Outpost's Robotics APCs, at (40,43) and (41,45), and Square Station's RND APC at (65,62), so it has something to do with Robotics/RND.

I also went through and named every APC in Outpost Station so that things are easier to find.  Some of the name choices might not be correct, but at least each APC has a unique name now so we can find problems easier.

corp-0:
> I also went through and named every APC in Outpost Station

Thank you!

Issue #4901, opened 2020-08-17
Pickaxes and shovels should have variable speeds, using the Tool script.


PetMudstone:
## Description
In /tg/, pickaxes and shovels are classified as tools (under TOOL_MINING and TOOL_SHOVEL respectively). It is also possible for their variants in /tg/ to have variable tool speeds. However, this is not possible in Unitystation. You can not adjust the speed at which a pick mines or the speed at which a shovel digs, which makes the new pick and shovel in #4902 identical in function to their parents aside from damage and size changes. I think the existing tool script might be good for such a purpose. 

- [x] pickaxe variable speeds
- [ ] shovel variable speeds

PR #6161, opened 2021-03-03
Variable Pickaxe Speeds

PerfectTangent:
### Purpose
This PR addresses the pickaxe half of #4901. Rather than mining speed being static, each tile can have a configurable mining speed and the quality of the pickaxe being used modifies that speed.

### Notes:

- Tiles have a configurable mining time. Default is the same as what the static mining time was (5 seconds).
- Pickaxes have a mining speed modifier
- Different pickaxes also had their damage numbers modified to mirror their /tg/ counterparts (most were already correct)
- This update also works with PickaxeMineable objects
- The Mining Vendor prefab now includes a single silver plated pickaxe for one lucky miner.

Tested smashing rocks and rock walls in editor, and it works. Thanks for the help, @corp-0!

### Changelog:

CL: Different tiers of pickaxes have varying mining speeds.

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Tool/Pickaxe.cs
+++ b/UnityProject/Assets/Scripts/Items/Tool/Pickaxe.cs
@@ -8,6 +8,11 @@ using AddressableReferences;
 public class Pickaxe : MonoBehaviour, ICheckedInteractable<PositionalHandApply>
 {
 	[SerializeField] private AddressableAudioSource pickaxeSound = null;
+
+	[Tooltip("How much does this tool multiply mining time")] [SerializeField]
+	private float timeMultiplier = 1f;
+	public float TimeMultiplier => timeMultiplier;
+
 	private static readonly StandardProgressActionConfig ProgressConfig =
 		new StandardProgressActionConfig(StandardProgressActionType.Construction, true);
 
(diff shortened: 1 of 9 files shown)


PerfectTangent:
while I was looking through /tg/'s repo I found this gem in the mining_tools.dm

`//metal needed for a crowbar and for a knife, why the FUCK does a knife cost 6 metal sheets while a crowbar costs 0.025 sheets? shit makes no sense fuck this`

UnityProject/Assets/Scripts/Items/Tool/Pickaxe.cs
```diff
@@ -8,6 +8,11 @@
 public class Pickaxe : MonoBehaviour, ICheckedInteractable<PositionalHandApply>
 {
 	[SerializeField] private AddressableAudioSource pickaxeSound = null;
+
+	[Tooltip("How much does this tool multiply mining time")] [SerializeField]
```
Bod9001, line 13: Could you put this in the toolutil , so you just add the tool component to the pickaxe and make it call the toolutil function you told with the function you want to activate once it's finished
void ProgressComplete()
		{stuff you want to do when finished}
ServerUseToolWithActionMessages(interaction, default time, text for action, ect. ect. , ProgressComplete )

ServerUseToolWithActionMessages is a good example of the lambda function thing

PerfectTangent: Done, but there's some kind of issue with it that I can't identify.

NoooneyDude: Some kind of issue? Something preventing merge?

UnityProject/Assets/Scripts/Items/Tool/Pickaxe.cs
```diff
@@ -25,40 +29,67 @@ public bool WillInteract(PositionalHandApply interaction, NetworkSide side)
...
-		var bar = StandardProgressAction.Create(ProgressConfig, ProgressComplete)
-			.ServerStartProgress(interaction.WorldPositionTarget.RoundToInt(),
-				5f, interaction.Performer);
-		if (bar != null)
-		{
-			SoundManager.PlayNetworkedAtPos(pickaxeSound, interaction.WorldPositionTarget, sourceObj: interaction.Performer);
-		}
+		ToolUtils.ServerUseToolWithActionMessages(
+			interaction, calculatedMineTime,
+			$"You start mining the {objectName}...",
```
NoooneyDude, line 56: `objectName` isn't set. It _does_ get set in the other ServerPerformInteraction, however. You should be able to get the name of the tile from the interaction.

NoooneyDude: While you're at it, consider rescoping the `objectName` variable to within those methods; don't think it needs to be a class field.

corp-0: >  however. You should be able to get the name of the tile from the interaction

TargetObject is null for some reason which is preventing the tile region to trigger. Also you can't get the tile name from the interaction afaik, there is a ``DisplayName`` field in ``BasicTile`` but I rarely see it filled.

NoooneyDude: Right, tiles aren't objects. This is `PositionalHandApply`, ~you're probably after `TileApply`~. EDIT: oh, sorry, nevermind.

PR #6171, opened 2021-03-04
RWall Health Change

ThatDan123:
Changes Rwall health from 500 to 100 due to the new damage layers being spawned on tile destruction.


corp-0:
should the other phases of rwall construction also be changed?

ThatDan123:
They are set to 100 and I think there's 6 layers so its now 600 health instead of 1000.

corp-0:
alright then, merging

PR #6174, opened 2021-03-05
Super matter

ThatDan123:
CL: Adds working supermatter
CL: Adds reflectors

-Use screwdriver on reflectors to change angle, ALT click to change angle in reverse direction

-Doesnt add anomalies but the system is there to spawn them.

-This is mainly moving the Tg system to c#, cleaned up a bit. Hopefully the math works out correctly and i havent messed anything up :p


corp-0:
are those variables codacy is complaining about meant to be constants?

ThatDan123:
nah, but ive made them local now

UnityProject/Assets/Scripts/Health/Objects/Integrity.cs
```diff
@@ -94,6 +94,11 @@ public class Integrity : NetworkBehaviour, IHealth, IFireExposable, IRightClicka
 	[Tooltip("The explosion strength of this object if is set to explode on destroy")]
 	public float ExplosionsDamage = 100f;
 
+	[SerializeField]
+	private bool doDamageMessage = true;
+
+	public bool DoDamageMessge => doDamageMessage;
```
corp-0: typo, you're missing an ``a`` :B

UnityProject/Assets/Scripts/Objects/Engineering/Reflector.cs
```diff
@@ -0,0 +1,391 @@
...
+			DownGradeState(true);
+		}
+
+		#region HandApply
+
+		public bool WillInteract(HandApply interaction, NetworkSide side)
+		{
+			if (DefaultWillInteract.HandApply(interaction, side) == false) return false;
+
+			if (Validations.HasItemTrait(interaction.HandObject, CommonTraits.Instance.Welder)) return true;
```
corp-0, line 126: shouldn't use ``||`` here? Any check under the welder won't happen if the object doesn't have the welder trait.

UnityProject/Assets/Scripts/Objects/Engineering/Reflector.cs
```diff
@@ -0,0 +1,391 @@
...
+			Chat.AddExamineMsgFromServer(interaction.Performer, "The reflector needs to be constructed first");
+		}
+
+		//Deconstruction
+		private void TryWrench(HandApply interaction)
+		{
+			if (isWelded)
+			{
+				//Needs to be unwelded before deconstruction
+				Chat.AddExamineMsgFromServer(interaction.Performer, "The reflector needs unwelded first");
```
corp-0:
```suggestion
				Chat.AddExamineMsgFromServer(interaction.Performer, "The reflector needs to be unwelded first");
```

UnityProject/Assets/Scripts/Objects/Engineering/Reflector.cs
```diff
@@ -0,0 +1,391 @@
...
+				else
+				{
+					Chat.AddExamineMsgFromServer(interaction.Performer, $"You need {diamondsNeeded} diamond sheets to build a box reflector.");
+				}
+			}
+		}
+
+		#endregion
+
+		#region Mathhelpers
```
corp-0: these two should probably go in a utils class instead of this component.

UnityProject/Assets/Scripts/Objects/Engineering/Reflector.cs
ThatDan123: If it doesnt have the welder trait then it'll go to the next If statement, it only returns true if it is a welder. Does remind me though that i forgot to add a check for the welder to be on

UnityProject/Assets/Scripts/Objects/IBumpObject.cs
```diff
@@ -0,0 +1,9 @@
+﻿using UnityEngine;
+
+namespace Objects
+{
+	public interface IBumpObject
+	{
+		void BumpAble(GameObject bumpedBy);
```
corp-0: maybe the method should be called ``OnBump``

UnityProject/Assets/Scripts/Objects/Engineering/Reflector.cs
corp-0: ah yeah, you're not returning if it doesn't have it.

Cboyjet4:
Keep in mind, we probably need to have functioning air alarms/scrubbing and vent settings, to have a proper SM engine.

UnityProject/Assets/Scripts/Objects/IBumpObject.cs
```diff
@@ -0,0 +1,9 @@
+﻿using UnityEngine;
+
+namespace Objects
+{
+	public interface IBumpObject
```
corp-0: I don't know why I didn't mention it before, but the interface should be called ``IBumpableObject`` to be ultra complacent with naming convention.

> Another common interface naming convention is used when an interface declares only one key method, such as Draw() in the above example. The interface name is then formed as an adjective by adding the "...able" suffix. So, the interface name above could also be IDrawable. This convention is used throughout the .NET Framework.

https://en.wikibooks.org/wiki/C_Sharp_Programming/Interfaces

ThatDan123: cool, ive renamed it

PR #6175, opened 2021-03-05
Changed CNT to a SyncVar

ThatDan123:
Syncing clientstate with a sync var now instead of net messages.

-Saves on message spam as before i was seeing three messages being sent about the same thing in a frame, and now a message will only send if it is different.

-Doesnt change any prediction stuff as the update is being triggered in the same places as it was before

-This will be good for when trying out the prox stuff as the state will auto send for new observers


Bod9001:
nice, at some point will probably need to implement a current state store to allow for network bubbling

UnityProject/Assets/Scripts/Core/Transform/CustomNetTransform.cs
```diff
@@ -144,6 +146,13 @@ private MotionStateEnum MotionState
 	public TransformState PredictedState => predictedState;
 	public TransformState ClientState => clientState;
 
+	private void Awake()
+	{
+		registerTile = GetComponent<RegisterTile>();
+		itemAttributes = GetComponent<ItemAttributesV2>();
+		syncInterval = 0f;
```
Bod9001, line 153: does this need to be manually set on stuff?

Bod9001: syncInterval = 0f;

ThatDan123: it would be if i didnt put that, making it 0 means it'll send the update immediately if the value was changed

PR #6187, opened 2021-03-07
Gun fire delay, grille, and wall mount adjustments.

PetMudstone:
### Purpose
- Replaces the FireRate variable in Gun.cs with FireDelay, which is the time in seconds that a gun can fire again as opposed to FireRate being how many times a gun can fire in a second. Done as FireDelay is more similar to how /tg/ (and CM) does it.
  - Tweaks for fire delays across the board.
- Makes broken grilles less durable. Stops bullets from colliding with them.
- _Hopefully_ makes wall-mounted objects no longer block projectiles
- Makes air meters no longer block movement.

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
+++ b/UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
@@ -130,10 +130,10 @@ namespace Weapons
 		public AddressableAudioSource DryFireSound;
 
 		/// <summary>
-		/// The amount of times per second this weapon can fire
+		/// The time in seconds before this weapon can fire again.
 		/// </summary>
-		[Tooltip("The amount of times per second this weapon can fire")]
-		public double FireRate;
+		[Tooltip("The time in seconds before this weapon can fire again.")]
+		public double FireDelay = 0;
 
 		/// <summary>
 		/// If suicide shooting should be prevented (for when user inadvertently drags over themselves during a burst)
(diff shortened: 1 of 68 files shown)


UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
```diff
@@ -130,10 +130,10 @@ protected StandardProgressActionConfig ProgressConfig
 		public AddressableAudioSource DryFireSound;
 
 		/// <summary>
-		/// The amount of times per second this weapon can fire
+		/// The time in seconds before this weapon can fire again.
 		/// </summary>
-		[Tooltip("The amount of times per second this weapon can fire")]
-		public double FireRate;
+		[Tooltip("The time in seconds before this weapon can fire again.")]
+		public double FireDelay = 0;
```
corp-0: is zero a good default delay?

PetMudstone: Honestly, no. Will change to 0.5 right now (which is similar to how fast guns with 0 delay fire in /tg/ due to lag and stuff, actually).

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
```diff
@@ -842,7 +842,7 @@ private void DequeueAndProcessServerShot()
 			if (shooter == PlayerManager.LocalPlayer)
 			{
 				//this is our gun so we need to update our predictions
-				FireCountDown += 1.0 / FireRate;
+				FireCountDown = FireDelay;
```
corp-0: I remember something about Autumn saying this should keep the +=

PetMudstone: Plus sign added (properly this time).

Bod9001:
can't work out why tests are failing Shall I merge?

ThatDan123:
maybe open and run it in editor just to make sure there isnt a compilation error

corp-0:
if it has been reran and still failed, then most likely it has compiling errors.

PR #6205, opened 2021-03-08
Readded blood effects

ThatDan123:
Readded blood effects and moved sickness to base health script instead of player


corp-0:
nice. In all honesty I want PlayerHealth to vanish and make all creatures follow the same rules.

PR #6215, opened 2021-03-09
NPC AI Fixes and adjustments

Bobtheapple:
Fixed Hostile AIs to play sounds properly, using an async rather than a coroutinue to reduce allocations.  Removed many uses of transform references and replaced them with register tile references.

I also rebased MobMeleeAttack on MobMeleeAction and  FaceHuggerAi on GenericHostileAI.   These scripts had a lot of overlap, doing this simplified them greatly.  MobMeleeAction was made more configurable by subdividing CheckForTargetAction into methods for validating the target, always acting on target, acting on players, acting on npcs, and acting on tiles.  This will allow programmers to more easily add mobs that act generally like another mob, but interact differently with one group of targets.  Also, in the process some AI bugs were potentially fixed.

Fixed some health stuff from the health update.

### DIFF
--- a/UnityProject/Assets/Scripts/NPC/AI/Friendly/CatAI.cs
+++ b/UnityProject/Assets/Scripts/NPC/AI/Friendly/CatAI.cs
@@ -90,7 +90,7 @@ namespace Systems.MobAIs
 				if (coll.GameObject == null) continue;
 
 				if (coll.GameObject != gameObject && coll.GameObject.GetComponent<MouseAI>() != null
-				                                  && !coll.GameObject.GetComponent<LivingHealthBehaviour>().IsDead)
+				                                  && !coll.GameObject.GetComponent<LivingHealthMasterBase>().IsDead)
 				{
 					return coll.GameObject.GetComponent<MouseAI>();
 				}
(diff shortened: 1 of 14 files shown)


UnityProject/Assets/Scripts/NPC/AI/Friendly/CatAI.cs
```diff
@@ -90,7 +91,7 @@ private MouseAI AnyMiceNearby()
 				if (coll.GameObject == null) continue;
 
 				if (coll.GameObject != gameObject && coll.GameObject.GetComponent<MouseAI>() != null
-				                                  && !coll.GameObject.GetComponent<LivingHealthBehaviour>().IsDead)
+				                                  && !coll.GameObject.GetComponent<LivingHealthMasterBase>().IsDead)
```
corp-0: are mobs using the new health classes now? I think you need to add raceData and the list of organs, I might be wrong though.

UnityProject/Assets/Scripts/NPC/AI/MobMeleeAction.cs
```diff
@@ -9,20 +9,20 @@
...
 	public class MobMeleeAction : MobFollow
 	{
 		[Tooltip("If a player gets close to this mob and blocks the mobs path to the target," +
-		         "should the mob then focus on the human blocking it?. Only works if mob is targeting" +
-		         "a player originally.")]
+				 "should the mob then focus on the human blocking it?. Only works if mob is targeting" +
+				 "a player originally.")]
 		[SerializeField]
-		private bool targetOtherPlayersWhoGetInWay = true;
+		public bool targetOtherPlayersWhoGetInWay = true;
```
corp-0: if your intention was to make these accesible to the children classes, use ``protected`` instead. If you're getting this from another place, add a public property that returns the private value.

Bobtheapple:
I wasn't aware that we'd be keeping both health engines, that certainly complicated things a bit.  Now the code will be able to handle that.

ThatDan123:
We're not planning to keep both systems, it just the mobs haven't been transferred over

Bobtheapple:
Ah, okay, well this code will be pretty easy to simplify down when the change happens then, and I made it so that it will work on npcs that have either health system.

Bod9001:
is good to merge?

PR #6217, opened 2021-03-09
Electrocution Damage

ThatDan123:
-Readded electrocution for players
-Examine text for non players readded
-Made supermatter trigger sound every 15 seconds now

Supermatter sounds are still broken on staging, havent got any idea why, followed all the steps to add sounds and everything looks right in the addressables project

### DIFF
--- a/UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
@@ -2,6 +2,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 using Systems.Atmospherics;
 using Chemistry;
 using Health.Sickness;
@@ -755,10 +756,59 @@ namespace HealthV2
 			healthStateController.SetFireStacks(0);
 		}
 
-		public virtual string GetExamineText()
+		#region Examine
+
+		public string GetExamineText()
 		{
-			return "Weeee";
+			if (this is PlayerHealthV2)
+			{
+				// Let ExaminablePlayer take care of this.
+				return default;
+			}
+
+			// Assume animal
+			var healthString = new StringBuilder("It is ");
+
+			if (IsDead)
+			{
+				healthString.Append("limp and unresponsive; there are no signs of life");
+
+				healthString.Append("...");
+
+				return healthString.ToString();
+			}
+
+			healthString.Append($"{ConsciousState.ToString().ToLower().Replace("_", " ")} and ");
+
+			var healthFraction = OverallHealth / maxHealth;
+			if (healthFraction < 0.2f)
+			{
+				healthString.Append("heavily wounded.");
+			}
+			else if (healthFraction < 0.6f)
+			{
+				healthString.Append("wounded.");
+			}
+			else
+			{
+				healthString.Append("in good shape.");
+			}
+
+			if (respiratorySystem != null && respiratorySystem.IsSuffocating)
+			{
+				healthString.Append("having trouble breathing!");
+			}
+
+			// On fire?
+			if (FireStacks > 0)
+			{
+				healthString.Append("And is on fire!");
+			}
+
+			return healthString.ToString();
 		}
+
+		#endregion
 
 		#region Sickness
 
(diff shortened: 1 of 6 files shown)


corp-0:
> Supermatter sounds are still broken on staging, havent got any idea why, followed all the steps to add sounds and everything looks right in the addressables projec

do they work when local testing?

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -755,11 +756,60 @@ public void Extinguish()
...
+			}
+
+			// Assume animal
+			var healthString = new StringBuilder("It is ");
+
+			if (IsDead)
+			{
+				healthString.Append("limp and unresponsive; there are no signs of life");
+
+				healthString.Append("...");
```
corp-0: you could add the three dots to the line before this one, no?

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -755,11 +756,60 @@ public void Extinguish()
...
+				healthString.Append("wounded.");
+			}
+			else
+			{
+				healthString.Append("in good shape.");
+			}
+
+			if (respiratorySystem != null && respiratorySystem.IsSuffocating)
+			{
+				healthString.Append("having trouble breathing!");
```
corp-0, line 797: this will appear next to the previous line, right after the period.
```suggestion
				healthString.Append(" Having trouble breathing!");
```

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -755,11 +756,60 @@ public void Extinguish()
...
+			}
+
+			// On fire?
+			if (FireStacks > 0)
+			{
+				healthString.Append("And is on fire!");
```
corp-0:
```suggestion
				healthString.Append(" And is on fire!");
```

UnityProject/Assets/Scripts/HealthV2/Living/PlayerHealthV2.cs
```diff
@@ -196,5 +199,166 @@ private void ApplyStartingAllergies()
 		}
 
 		#endregion
+
+		#region Electrocution
```
corp-0, line 203: I heavily dislike how hardcoded this is and much prefer the interface approach it had before when Nooney first PRed.

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
ThatDan123: theres already a space before it in: 

```healthString.Append($"{ConsciousState.ToString().ToLower().Replace("_", " ")} and ");```

File: UnityProject/Assets/Scripts/HealthV2/Living/PlayerHealthV2.cs (hunk 4 above)
Comment (ThatDan123) [reply UnityProject/Assets/Scripts/HealthV2/Living/PlayerHealthV2.cs:203]: I just copied it over from the old health, i'll put it on the to do list

File: UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs (hunk 3 above)
Comment (NoooneyDude) [inline UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs]:
```suggestion
				healthString.Append("and is on fire!");
```
suggestion V2 ; )

ThatDan123: humm, but the last added string has punctuation at the end, so it would be And

NoooneyDude: Which last string? You mean the having trouble breathing? It is overwritten if on fire, as I consider that to be of more importance.

NoooneyDude: Sorry... nevermind. That used to be the old behaviour.

PR #6219, opened 2021-03-09
MaterialStorage refactor

Aranclanos:
### Purpose
Refactors the MaterialStorage class to have less bloat, be more organized and remove copypasted code in autolathe and exosuit fabricator code.
The crafting manager will now handle the type of materials that all machines will accept all across the board, setting a common base for the material silo.

On a player perspective there are no noticeable changes


UnityProject/Assets/Scripts/Objects/ExosuitFabricator.cs
```diff
@@ -47,43 +47,26 @@ public enum ExosuitFabricatorState
...
-					if (Validations.HasItemTrait(interaction.HandObject, material))
-					{
-						InsertedMaterialType = material;
-						return true;
-					};
-				}
+				return true;
 			}
-
 			return false;
```
joelhed, line 70: Can be simplified further to
```c#
return InsertedMaterialType != null;
```

PR #6223, opened 2021-03-10
fixes a few issues with surgery

Bod9001:
fixes error with null health
you now can implant in root body parts
you can cancel surgeries with cautery

### Changelog:
CL: you can cancel surgeries with cautery


UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -357,6 +357,11 @@ public float GetTotalBurnDamage()
 			return toReturn;
 		}
 
+		public void AddBodyPartToRoot()
+		{
+
+		}
```
ThatDan123, line 363: Empty method?

Bod9001: ok will note

PR #6224, opened 2021-03-10
Mineral silos

Aranclanos:
### Purpose
Adds the material silo, also known as the ore silo. This early implementation of the machine, if added on the maps, will link autolathes and exosuit fabricators with itself, making them use the minerals inside the silo.
The silo has an infinite storage space.
Storing minerals on machines linked to the silo will store them in the silo.
In case of the silo not existing or being destroyed, the machines will use their own storage.
The original name is ore silo but they dont actually store ores so I assumed a namechange would be fitting.

### Notes
There's no way to remove minerals from the silo using the silo itself as it has no interface
It is impossible to build one
Admin spawned silos won't link existing machines and will only link with ones spawned afterwards


PerfectTangent:
First, rad.
Can multi tools be used to change the links?
What happens if there's more than 1 silo at round start?

Aranclanos:
no multitool implementation at the moment
the one who gets the Awake() called last will be the one picked to be the station's silo

Bod9001:
looks like compile fail?

PR #6245, opened 2021-03-14
New Health Commenting and Tweaks

Bobtheapple:
### Purpose
With the new health system update, I noticed some understandable confusion regarding how everything works now.  I've decided to go through and comment the new code to help clarify as much as possible the new systems, also fixed some camelCase/PascalCase and tweaked some of the wording/grammar.  In the process, I've encountered a few bugs and oversights which I've also addressed

### Notes:
This is a large chunk of the commenting out of the way, hopefully this will help cut down on the confusion.  There is still a lot to go through, but I figured I'd put up what I have for now.  If I have misunderstood how anything works, don't hesitate to correct the comments

### Changelog:
(Roughly in order of significance)
General
* ImplantUpdate and ImplantPeriodicUpdate were changed to longer take a HealthMaster as a parameter.  If an update requires the passing of a Health Master, an organ should always pass its own, not one it receives.  Otherwise you could end up with a case where someone's heart literally beats for another.

Lungs, RespiratorySystem, and RespiratoryInfo
CL: Made lungs no longer expel desired reagent from the blood stream
CL: Generalized a several things that were hard coded for oxygen and co2.
* Moved HandleBreathing to RespiratorySystemBase
* Separated Breathe into three methods, TryBreathing(), BreatheIn(), and BreatheOut()
* Added a new method on RepiratorySystemBase, GasExchange(), to handle swaping gases from atmos to blood
* Removed RespiratoryInfo, all of its functions are now handled by the lung itself as per the comments

BodyPartDamage.cs
CL:  Added the ability to split damage among contained organs to TakeDamage(), removed oxy damage as per comment to prevent child organs from suffering greatly magnified oxy damage during suffocation

HealthStateController.cs
CL: Made the overallHeathSync initialize as MaxHealth, rather than a hard coded 100.

LivingHealthMasterBase.cs
* Commented out the unused AvailableBlood and bodyPartContainers to reduce confusion
CL: Fixed a bug where ChangeFireStacks could set fireStacks to less than on fire or more than maxFireStacks

UI_HeartMonitor.cs
CL:  Made reported positive wound states relative to max health to be consistent with examine
CL:  Fixed a bug where heart monitor would not detect states lower than injured

CharacterSettings.cs
CL: Added Themself pronoun

BodyPart.cs
* HealthMaster, IsSurface, Storage, BodyPartType, and LimbSpriteData PascalCase was corrected

RootBodyPartContainer.cs
* IsSurfaceSprite PascalCase was corrected
* NewSpite changed to NewSprite
* Reorganized TakeDamage() and changed SplitDamage to damageSplit

BodyPartBlood.cs
*  isBloodReagentConsumed PascalCase was corrected

CirculatorySystemBase.cs
* UseBloodPool changed to UsedBloodPool


ThatDan123:
Test fail reason:

Can't load asset HumanRespiratoryInfo.asset. Maybe linked ScriptableObject script is missing?

  Expected: <empty>
  But was:  < "Assets/Resources/ScriptableObjects/Health/RespiratoryInfo/HumanRespiratoryInfo.asset" >at

corp-0:
Havent read the code yet because I’m on my phone but I think organs taking oxydamage is intended, it simulates one of the most important tasks of blood, that is moving oxygen in blood stream.

Bobtheapple:
Sorry, didn't make that note clear.  What TakeDamage() does is deals damage to a body part and then also does that damage to the body part's children, and it is called whenever damage is to be applied to a body part.  All body parts automatically deal themselves damage when they lack oxygen.  So without this change, when low on oxygen all body parts would take damage, then all children of body parts would take damage again, then if its a child of a child it would take damage a third time, so that body parts take oxygen damage equal to N*Damage, where N is their depth.  (Less then that technically due to the armor calculation)  This change makes it so when each body part is low on oxygen it damages itself and only itself.

corp-0:
Ahh, I understand now.

> (Less then that technically due to the armor calculation)

That damage is supposed to be of type Internal, so it is not reduced by armor.

corp-0:
> Made lungs no longer expel desired reagent from the blood stream

what does this mean?

Bobtheapple:
How it was set up was that lungs would dump all gases (including oxygen) from the blood they were processing each time it exhaled, and then take in the as much oxygen as possible (up to the maximum amount the blood can hold) from the atmosphere when it inhaled.  This meant that as soon as the creature entered a low oxygen environment, the lungs would start dumping oxygen and taking in nothing, effectively becoming a detriment to survival.  In real life, the lungs do not do this do, you generally retain all the oxygen remaining in your bloodstream until it is converted into CO2 due to how the concentration gradients work, holding your breath, etc.  This change makes so that the lungs check to make sure that when they dump gases they don't dump the oxygen (or whatever the blood reagent that the lungs are configured to work off of).  They still only fill up to the blood capacity.

In testing this didn't seem to affect the survival time very much, the main limiting factor in surviving in low oxygen appears to be the rate that all you body parts consume your oxygen pool, not the rate the lungs was dumping it.  In theory though this lets a creature survive in space for a little longer and recover from low oxygen a little faster, and makes having multiple lungs more of a benefit and less of a problem.

UnityProject/Assets/Scripts/HealthV2/Living/Respiration/RespiratorySystemBase.cs
```diff
@@ -100,83 +84,82 @@ private void MonitorSystem()
 			}
 		}
 
-		private bool Breathe(IGasMixContainer node)
+		public void GasExchange(GasMix atmos, ReagentMix blood, ReagentMix toProcess, bool takeFromBlood = false)
```
corp-0, line 87: the if feels weird. We only use one of the arguments of the function or the other based on that if. Wouldn't an overload make more sense here?

corp-0:
I updated your description. Use ``CL:`` only for changes that are relevant for the end user.

Bobtheapple:
Hmmm, has more weirdness going on then I thought, converting this to a draft while I work out a few things.

Bobtheapple:
Will be back with something a lot more thorough soon

PR #6250, opened 2021-03-15
Gas Overlay Fix

ThatDan123:
Allows for multiple gas overlays again, woo

CL: Fixed multiple overlays not working
CL: Moved tile reactions to atmos simulation
CL: Moved gas overlay setting to atmos simulation

Squash merge this


Bod9001:
does this move over wind as well?

Bod9001:
press the wrong button xD

ThatDan123:
Didnt move wind over, or hotspots as im not exactly sure how those systems work. Though they probably can be now and the reaction manager would be obsolete. I might do it in a PR after this one.

Bod9001:
the only problem I see is moving items but that might be okay

PR #6252, opened 2021-03-15
adds crawling

### LINKED ISSUE #4960: The ability to lie down and crawl on command using the REST button.
## Description
Now that health has been merged, resting as a feature needs to be developed for a variety of activities.

In /tg/, you can lie down on the floor by clicking on the REST button. Surgery should not be conducted on standing individuals, and should require them to be resting. The surrender emote also requires resting for it to be properly implemented.

image

Lying down on the floor allows you to crawl around slowly. A special crawling sound plays when you do this. While lying down, projectiles will fly over you unless whoever is firing them clicks directly on you. You can also pass through certain objects while lying down. For example, you can only pass onto the same tile airtight flaps are on if you are lying down.

Airtight_Flaps_Prone_Passage

You can click the REST button again while lying down to get back up onto your feet after a second or two. You will fail to get up if you abruptly move, such as if you're on an active conveyor belt.
Related to #4665.

Resting will also allow players to properly benefit from resting in beds, providing anyone with a rudimentary treatment method. Resting is also closely related with #6238

Bod9001:
fixes #4960
### Notes:
the bullets passing over and healing isn't implement it the rest is

### Changelog:
CL: you can now crawl around


Aranclanos:
can probably remove ServerLayDown() and ServerStandUp() and its variants

Bod9001:
need the ServerStandUp since I can't pass arguments with it with the progress bar

UnityProject/Assets/Scripts/HealthV2/Living/PlayerHealthV2.cs
```diff
@@ -119,6 +120,18 @@ protected override void Gib()
 			playerNetworkActions.ServerSpawnPlayerGhost();
 		}
 
+		bool RegisterPlayer.IControlPlayerState.AllowChange(bool rest)
+		{
+			if (ConsciousState == ConsciousState.CONSCIOUS)
```
corp-0: why not return ``ConsciousState == ConsciousState.CONSCIOUS`` ?

PR #6260, opened 2021-03-16
Fixed objects not despawning properly

ThatDan123:
-Fixed issues whether objects were not moved to hidden pos when despawned on client

-Hopefully fixed that spamming mouse util error

-Fixed mob sounds trying to play on client when it was a server only method

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Input System/MouseUtils.cs
+++ b/UnityProject/Assets/Scripts/Core/Input System/MouseUtils.cs
@@ -94,8 +94,8 @@ public static class MouseUtils
 			//get the "parent" game object of each of the hit renderers
 			//for a sprite renderer, the parent is the object that has a RegisterTile.
 			//for a tilemap renderer, the parent is the oject that has a Matrix
-			.Select(r => r is TilemapRenderer ? r.GetComponentInParent<InteractableTiles>()?.gameObject :
-				r.GetComponentInParent<RegisterTile>()?.gameObject)
+			.Select(r => r is TilemapRenderer ? r.GetComponentInParent<InteractableTiles>().OrNull()?.gameObject :
+				r.GetComponentInParent<RegisterTile>().OrNull()?.gameObject)
 			//each gameobject should only show up once
 			.Append(IInteractableTiles).Distinct();
 	}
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/Core/Input System/MouseUtils.cs
```diff
@@ -94,8 +94,8 @@ public static IEnumerable<GameObject> GetOrderedObjectsAtPoint(Vector3 worldPoin
 			//get the "parent" game object of each of the hit renderers
 			//for a sprite renderer, the parent is the object that has a RegisterTile.
 			//for a tilemap renderer, the parent is the oject that has a Matrix
-			.Select(r => r is TilemapRenderer ? r.GetComponentInParent<InteractableTiles>().gameObject :
-				r.GetComponentInParent<RegisterTile>().gameObject)
+			.Select(r => r is TilemapRenderer ? r.GetComponentInParent<InteractableTiles>()?.gameObject :
```
corp-0:
```suggestion
			.Select(r => r is TilemapRenderer ? r.GetComponentInParent<InteractableTiles>().OrNull()?.gameObject :
```

UnityProject/Assets/Scripts/Core/Input System/MouseUtils.cs
```diff
@@ -94,8 +94,8 @@ public static IEnumerable<GameObject> GetOrderedObjectsAtPoint(Vector3 worldPoin
...
+				r.GetComponentInParent<RegisterTile>()?.gameObject)
```
corp-0:
```suggestion
				r.GetComponentInParent<RegisterTile>().OrNull()?.gameObject)
```

corp-0 (commented):
Remember you can't ``?`` Unity Objects. Copern added ``OrNull()`` extension for that.

ThatDan123:
Its not a unity object, components arent objects, so its fine to use ? when checking.

corp-0:
How come? If you're getting it with ``GetComponent`` ?

ThatDan123:
if i'd done r?.GetComponentInParent<RegisterTile>(), then yeah r is an object so would bypass the unity check

corp-0:
Aren't Monobehaviour also derived from UnityObject?

corp-0:
Let me open it in Rider

ThatDan123:
hum, maybe it is then. Rider didnt complain like it usually does when you use ? for unityobjects

ThatDan123:
better safe than sorry

corp-0:
image
It probably was Rider getting confused with the syntax there, maybe LINQ or something. But yeah, it does complain when you elvis Monobehaviours.

corp-0:
when were objects not despawning properly?

ThatDan123:
The problem was more to do that the objects weren't being sent to the hidden pos before they were set inactive during the despawn, the added delay means that the move message will be sent before.

PR #6261, opened 2021-03-16
Fixes ApplyDamageAll damage spread

Aranclanos:
### Purpose
Fixes the logic of ApplyDamageAll, now it'll properly spread the damage on each limb in the case that a bodypart container has more than one limb.

### Notes:
because each bodypart container only has one limb, this has no noticeable changes

### DIFF
--- a/UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
@@ -443,21 +443,21 @@ namespace HealthV2
 		public void ApplyDamageAll(GameObject damagedBy, float damage,
 			AttackType attackType, DamageType damageType, bool damageSplit = true)
 		{
-			float BodyParts = 0;
 			if (damageSplit)
 			{
-				BodyParts += RootBodyPartContainers.Sum(Container => Container.ContainsLimbs.Count());
+				float bodyParts = RootBodyPartContainers.Sum(Container => Container.ContainsLimbs.Count());
+				damage /= bodyParts;
 			}
 
 			foreach (var Container in RootBodyPartContainers)
 			{
 				if (damageSplit)
 				{
-					Container.TakeDamage(damagedBy, damage *  (Container.ContainsLimbs.Count / BodyParts) , attackType, damageType);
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -443,21 +443,21 @@ public void CalculateOverallHealth()
 		public void ApplyDamageAll(GameObject damagedBy, float damage,
 			AttackType attackType, DamageType damageType, bool damageSplit = true)
 		{
-			float BodyParts = 0;
 			if (damageSplit)
 			{
-				BodyParts += RootBodyPartContainers.Sum(Container => Container.ContainsLimbs.Count());
+				float bodyParts = RootBodyPartContainers.Sum(Container => Container.ContainsLimbs.Count());
```
Bod9001, line 448: this doesn't really make sense with bullets lets say if there was two torsos, both would get damaged from one bullet somehow?

Aranclanos: yes, its ApplyDamageAll, grabs the whole body and applies the dmg value, splitting it or not is up to the bool argument

Bod9001: oh sorry Miss red

Bod9001: ok, reading through and doing the maths doesn't this do it already,
so 10 body parts 100 damage, with container with two body parts
original
100*(2/10 ) = 20
new
10 = 100/10
10  * 2 = 20

so, systems are exactly same?

Aranclanos: Yes! Just simpler to read, now, for the else case it avoids multiplicative dmg since it also carries out the flag

Bod9001: just saying for the one 
Container.TakeDamage(damagedBy, damage, attackType, damageType, true);
if there's two limbs the damage would be split across an equally, tho depending on the attack type I of them could get the full brunt hummm

PR #6262, opened 2021-03-16
fixes requested by server hosts

Bod9001:
fixes server getting stuck into weird state if restart is triggered before round start
added try catch to web request


Aranclanos:
can it still get stuck if an admin restarts before roundstart?

Bod9001:
nope

Bod9001:
I think

PR #6263, opened 2021-03-16
Fixes the dev spawner freezing lag on empty search

Aranclanos:
### Purpose
Fixes the dev spawner searching with an empty or really short string.

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Systems/AdminTools/DevTools/GUI_DevSpawner.cs
+++ b/UnityProject/Assets/Scripts/UI/Systems/AdminTools/DevTools/GUI_DevSpawner.cs
@@ -64,16 +64,17 @@ public class GUI_DevSpawner : MonoBehaviour
     {
 	    if (searchWhileTyping)
 	    {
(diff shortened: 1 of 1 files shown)


Pockets-byte:
wait so this is just the nuclear option?....
well I guess I'll take what I can

corp-0:
what you mean?

Pockets-byte:
I was sorta hoping dev spawner was fixed so it handled having to load every prefab in the game didn't cause massive slowdown, rather than just removing that ability.

Aranclanos:
well it was a bug, from design it was never inteded to show all loaded prefabs (it doesnt show ALL prefabs)
do you want to have the entire list?

Pockets-byte:
It's always useful in my area to know and be able to use every usable item in the game. which is why I go through the list of items I can spawn every once in a while 
I haven't done it recently because the dev spawner list has been increasing exponentially in length

Aranclanos:
it uses Resources.FindObjectsOfTypeAll() so objects not present in the scene wont show up, but we also have to re-do this to stop using the resources folder, maybe a proper dev spawner with categories could be used

PR #6266, opened 2021-03-18
Refactor CargoData to use SO categories and orders

corp-0:
### Purpose
Refactors CargoData to use individual scriptable objects to add categories and orders inside each category. 

This is good because:
* It is less likely to have conflicts when working with CargoData
* You can reorder the categories in CargoData very easily so it changes how it is displayed in the console.
* You can reorder the orders inside each category so it changes how it is displayed in the console.
* Adding, removing or modifying the orders is easier to digest than having all of them at once in the tiny space that is inspector.

This is bad because:
* All categories and orders are a single file, so we have more files now. How bad is that?

### Notes:
* I also removed Mirror unit tests.
* Removed the script ThatDan did to detect cargo exploits. I instead made each order calculate its sell price OnValidate, so the editor is aware of the price as soon as they modify an order.


corp-0:
it says Scene Changes because I deleted two scenes when deleting Mirror tests. No project scene was changed!

UnityProject/Assets/Scripts/UI/Objects/Cargo/CargoData.cs
```diff
@@ -1,13 +1,13 @@
...
 using UnityEngine;
 
 namespace Systems.Cargo
 {
 	public class CargoData : ScriptableObject
 	{
-		//Stores all possible supplies broken into categories
-		public List<CargoOrderCategory> Supplies = new List<CargoOrderCategory>();
+		[ReorderableList]
+		[Tooltip("Stores all possible supplies broken into cateogires")]
```
Bod9001: cateogires xD

corp-0: Lol, nice catch

corp-0: done

PR #6267, opened 2021-03-18
Pickup animation will now play on all clients

MaxIsJoe:
### Purpose
title

### Changelog:
CL: pickup animations will now play on all clients.

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Inventory/Pickupable.cs
+++ b/UnityProject/Assets/Scripts/Systems/Inventory/Pickupable.cs
@@ -161,15 +161,13 @@ public class Pickupable : NetworkBehaviour, IPredictedCheckedInteractable<HandAp
 		var ps = interaction.Performer.GetComponent<PlayerScript>();
 		var extendedRangeOnly = !ps.IsRegisterTileReachable(cnt.RegisterTile, true);
 
-		//Start the animation on all clients.
-		StartCoroutine(PickupAnim(interaction.Performer.gameObject));
-
-		//Start the animation on the server.
-		LeanTween.move(gameObject, interaction.Performer.transform, pickupAnimSpeed);
-		LeanTween.scale(gameObject, new Vector3(0, 0), pickupAnimSpeed);
+		//Start the animation on the server and clients.
+		PickupAnim(interaction.Performer.gameObject);
+		RpcPickupAnimation(interaction.Performer.gameObject);
 		yield return WaitFor.Seconds(pickupAnimSpeed);
 
 		//Make sure that the object is scaled back to it's original size.
+		RpcResetPickupAnim();
 		LeanTween.scale(gameObject, new Vector3(1, 1), 0.1f);
 
 		//if it's in extended range only, then we will nudge it if it is stationary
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Systems/Inventory/Pickupable.cs
```diff
@@ -205,6 +208,15 @@ private IEnumerator serverPreformInteractionLogic(HandApply interaction)
...
 
+	[ClientRpc]
+	private IEnumerator RpcPickupAnimation(Transform interactor)
+	{
+		LeanTween.move(gameObject, interactor, pickupAnimSpeed);
+		LeanTween.scale(gameObject, new Vector3(0, 0), pickupAnimSpeed);
+		yield return WaitFor.Seconds(pickupAnimSpeed);
+		LeanTween.scale(gameObject, new Vector3(1, 1), 0.1f);
+	}
+
```
ThatDan123, line 228: I dont think that'll start the coroutine on client, maybe switch it to:

<code>

        [ClientRpc]
	private void RpcPickupAnimation(Transform interactor)
	{
		StartCoroutine(RpcPickupAnimation(interaction.Performer.transform));
	}

       private void PickUpAnimation(Transform interactor)
       {
            LeanTween.move(gameObject, interactor, pickupAnimSpeed);
		LeanTween.scale(gameObject, new Vector3(0, 0), pickupAnimSpeed);
		yield return WaitFor.Seconds(pickupAnimSpeed);
		LeanTween.scale(gameObject, new Vector3(1, 1), 0.1f);
       }
</code>

<strike>-Also means you can remove the duplicated animation in serverPreformInteractionLogic and start the server animation with StartCoroutine(RpcPickupAnimation(interaction.Performer.transform)); as well </strike>

 Actually might need that delay so the object isnt hidden before animation has finished.

ThatDan123: Also i dont think you'll be able to send Transform in the Rpc, think you'll be able to send gameObject and then do gameObject.Transform on client side.

MaxIsJoe:
Needs some testing now on more than two clients.

UnityProject/Assets/Scripts/Systems/Inventory/Pickupable.cs
```diff
@@ -167,6 +167,9 @@ private IEnumerator serverPreformInteractionLogic(HandApply interaction)
 
 		LeanTween.scale(gameObject, new Vector3(1, 1), 0.1f);
 
+		//Start the animation on all clients.
+		StartCoroutine(PickupAnim(interaction.Performer.gameObject));
```
ThatDan123: Might be better to move this to before the server animation as well

MaxIsJoe: Did you test it?

corp-0:
I don't like how the animation code is repeated for client and server, it could be encapsulated in a single method called by both server and the RPC method.

MaxIsJoe:
> I don't like how the animation code is repeated for client and server, it could be encapsulated in a single method called by both server and the RPC method.

done.

UnityProject/Assets/Scripts/Systems/Inventory/Pickupable.cs
```diff
@@ -205,6 +208,25 @@ private IEnumerator serverPreformInteractionLogic(HandApply interaction)
...
+	private void PickupAnim(GameObject interactor)
+	{
+		LeanTween.move(gameObject, interactor.transform, pickupAnimSpeed);
+		LeanTween.scale(gameObject, new Vector3(0, 0), pickupAnimSpeed);
+	}
+
+	[ClientRpc]
+	private void RpcPickupAnimation(GameObject interactor)
+	{
+		LeanTween.move(gameObject, interactor.transform, pickupAnimSpeed);
```
corp-0: You're still repeating the code, I intended to mean something like this:

```suggestion
		PickUpAnim( interactor);
```

MaxIsJoe: done

PR #6269, opened 2021-03-19
Circulatory and Respiratory Rework

### LINKED ISSUE #6259: Health Analyser UI
OH GOD the text is pretty hard to read, Improve it with the UI, and have a button to dump text version into chat as memory/print out piece of paper with information on,

fill out moore with desired look

Bobtheapple:
### Purpose
Completely reworks the Circulatory and Respiratory system to be more functional and in line with their design philosophy.

### Notes:
This PR covers a lot of things, so I'll gloss over the smaller fixes.  Essentially, this addresses a problem I saw in BodyPartBlood, where it had a two competing ideas of how it was supposed to work, in simplistic terms whether body parts were supposed to store only reagents, or whether body parts were supposed to store a blood/reagent mix.  This PR resolves that conflict, having them use a blood/reagent mix, and in that process I had to rework a lot of stuff. 

 I took a lot of effort to try to get the numbers right where I could, this means that humans in game roughly consume the same amount of oxygen per second as they do in real life (0.001 mols/breath, 1 breath/4seconds), oxygen concentration hovers at 98-99% in ideal conditions, blood level is at 5L, and a bunch of other little stuff.
 
Edit: Efficiency multipliers are back, and damage is more normalized, just had to do some math.

Fixes #6259, copied the layout of the old TGStation analyzer.  Its still missing a lot of features (species, body temp, bleeding notification, blood type, and treatment advice), but I figured those can wait for after health is finished.

### Changelog:
The code is very generalized, but for the sake of clarity I will use common terms in place of generalized terms, such as "oxygen" to refer to all "Required Reagents"

In General
 CL: Respiration and circulation will now work much more realistically
Lots and lots of pascal/camel case fixes, grammar and wording changes, added a bunch of comments, some other minor fixes I've probably forgotten.

 
BodyPart.cs
 CL: Body Parts will now initialize their blood after they know what kind of blood they should have
 
BodyPartBlood.cs
 CL: Body Parts are now filled with blood, not air.
 CL: Body Parts now have their initial owner's blood type.
 CL: Body Parts consume oxygen more realistically.
 CL: Oxy damage and healing now scales with blood oxygen saturation.
 CL: Blood flow now scales with number of hearts/heart rate.
- Split modifiers from BodyPartBlood to its own partial class.

Part of the old system was that it consumed all of the oxygen in a volume of the internal blood tank, regardless of how much oxygen was there.  This meant that organ with 10 times the oxygen in it would consume 10 times as much, an organ with 10 times the container size would consume 1/10 as much, and other weirdness. Now the organ consumes a set amount of oxygen each tick, and takes scaling damage based on the concentration of oxygen.  Body parts were also sending out blood as part of their normal tick, but receiving blood as part of a heart beat event.  This meant that a faster heart rate or more hearts would have no impact on the organ whatsoever, so long as its minimum needs were met.

BodyDamage.cs
 CL: Body parts are better at assigning damage to sub body parts
 - (I removed the oxy damage change I had from before, it turns out I was mistaken about how oxy damage worked, and regardless its all different now)

Bones.cs
 CL: Bones will no longer slowly fill your body with blood until you have so much you can't breathe.

CirculatorySystemBase.cs
 CL: Blood will now start oxygenated

BloodType.cs
 CL: Fixed a bug where blood could hold double the oxygen it should
 CL: Bodies will no longer work with a different species blood

Brain.cs
 CL: Brains are now part of the body too!

Heart.cs
 CL: Hearts will no longer destroy excess blood
 
Kidney.cs
 CL: Kidneys will no longer destroy their owner's blood
 
Lung.cs
 CL: Lungs will now work more realistically.
 CL: Lungs will no longer break if too many different things are dissolved in the bloodstream

HealthScanner.cs
 CL: Health Scanner is much more useful

ReagentMix.cs
 - Fixed a bug where subtract returned the wrong value
 - Added GetPercent

LivingHealthMasterBase.cs
 CL: Fixed a bug where ChangeFireStacks could set fireStacks to less than on fire or more than maxFireStacks

CharacterSetting.cs
 - Added themself pronoun

CPRable.cs
 CL: CPR no longer violates thermodynamics, and works the way you think it would.

Other:
 CL: Players will once again fall down when unconscious.


corp-0:
To make a changelog line you just start a line typing CL: and your message. No code block or _ needed. Also remember to only register to changelog stuff that is relevant for the final user and use verbs in past tense like “added” or “fixed” preferably. 

Example of changelog line:
CL: added new organ: Brain.

UnityProject/Assets/Scripts/HealthV2/Living/CirculatorySystem/BodyPartBlood.cs
```diff
@@ -8,192 +8,271 @@
...
-				TotalModified *= Mathf.Max(0, Modifier.Multiplier);
+				if (isBloodCirculated)
+				{
+					HealthMaster.CirculatorySystem.ReadyBloodPool.TransferTo(BloodContainer.CurrentReagentMix, BloodStoredMax);
+				}
+				BloodContainer.ContentsSet = true;
+			}
+			if (bloodType == null)
+			{
+				bloodType = HealthMaster.CirculatorySystem.BloodType;
```
Bod9001, line 107: humm What's the desired outcome if you attach plasma man arm to you,
I would expect it to work as long as you have plasma in your bloodstream,
if you have plasma in your bloodstream then it takes it from the blood and goes this is good,
I suppose organs don't care about the blood in them, they just care if it has the gas they require/reagent

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
```diff
@@ -148,65 +192,91 @@ void Awake()
 		{
 			foreach (BodyPartSprites b in GetComponentsInParent<BodyPartSprites>())
```
Bod9001: Look like this got left in somehow

UnityProject/Assets/Scripts/Health/CPRable.cs
```diff
@@ -62,15 +62,17 @@ private void ServerDoCPR(GameObject performer, GameObject target,BodyPartType Ta
...
 		foreach (var BodyPart in health.GetBodyPartsInZone(TargetBodyPart))
 		{
 			var heart = BodyPart as Heart;
 			if (heart != null)
 			{
 				heart.Heartbeat(HeartbeatStrength);
 			}
+
+			// Still violates thermodynamics, but meh.
+			BodyPart.BloodContainer.CurrentReagentMix.Add(health.CirculatorySystem.CirculatedReagent,
```
Bod9001: you could check for lung and then do it to that?

Bod9001:
would be nice bringing back the multiplier for lungs and heart but I can see the hard part, with the spiral of death, maybe we can make it so that lungs don't take oxygen damage and Hart immediately takes from ready blood pool instead of having to pump to itself?

UnityProject/Assets/Scripts/HealthV2/Living/CirculatorySystem/BodyPartBlood.cs
Bobtheapple: Actually, as I have it implemented the default body types only take required reagent from the blood type they are assigned (as part of ConsumeReagents),  Assuming plasmamen use plasmamen blood that carries plasma instead of oxygen, for a human to have a plasmaman arm they would need plasmaman blood in their body as well as human blood, otherwise the arm would take oxy damage.  I was thinking we wanted implant rejection, but I can fix this if we don't.

Of course, any body part can override how they use reagents, so the plasmaman arm could be set to ignore what blood it has and only check for oxygen.  That will definitely be useful for synthetic arms that are supposed to be one-size-fits-all.  We could also have surgery change what kind of blood the body part runs off of (bloodtype), without changing what kind of reagents it desires (requiredreagent).  Of course, human blood is very bad at carrying plasma compared to oxygen.  

Or I could change the check to look for any kind of blood, but then you would be able to get transfusions from any blood type and still get some functionality.  Of course, different blood are better/worse at handling different reagents, so it wouldn't exactly remove the mis-match type problem.  No matter what direction we go the human kidney would have to be replaced (or its code reworked), it removes plasma and foreign blood right now.

I guess, end of the day, with current code you could graft the plasma arm on and hook yourself up to an IV drip bag of plasma blood to avoid the kidneys removing it all and plasma because your lungs wont work with the plasma blood.

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
Bobtheapple: Oops, yeah, that was from when I needed to figure out what was going on with the brain not being in the body.  Removed.

UnityProject/Assets/Scripts/Health/CPRable.cs
Bobtheapple: There is quite a bit of delay between the lung pulling in oxygen and body parts receiving them, so I was thinking this way would make CPR more effective, but that would be more realistic, yeah.  Also, gave me a good idea, one moment.
Edit: Done, I'll need to do a build to test if full efficiency is too strong, that will take a bit.

UnityProject/Assets/Scripts/HealthV2/Living/CirculatorySystem/BodyPartBlood.cs
Bod9001: yeah, so what is the point of blood type then other than lungs for like balancing chemicals right?

Bobtheapple:
And got that too, just adjusted them to a square root curve and tweaked some damage numbers, feels pretty good now.  Its a little funky how oxy damage works now, like you can dive through an airlock, go into another airlock, climb in, do a little victory dance, and suddenly your legs start to give out from under you... you need to lie down for a minute... fall unconscious...  Its pretty realistic, but very different from how video games normally work.

Oh, to be clear on the level of robustness allowed with the current numbers, I tested them on square station sprinting from the Bar to Arrivals without internals.  You can make it, so long as you aren't blown back by the airlock on the other side, and you will recovery once in.  But if you are foolish enough to try a return trip before you've healed all of your oxy damage, you will die.

Bobtheapple:
Hmm, having some issues with consciousness states on the build, and CPR isn't effective, don't merge, needs some more bug hunts

UnityProject/Assets/Scripts/HealthV2/Living/CirculatorySystem/BodyPartBlood.cs
Bobtheapple: The point of the blood type is to make it very difficult to use other people's body parts in your body, this is what it does in real life.  To simulate this, I have it so that all body parts only pull oxygen from the body's native blood type, and take damage when their native blood type is low on oxygen.  I think having the lungs refuse to oxygenate foreign blood, the kidney removing foreign blood, and your body parts not pulling oxygen from foreign blood, along with the part only working off of foreign blood simulates body part rejection quite well.  To not need constant external help you would need a transfusion of 5L of foreign blood, foreign lungs, remove your kidneys, a second heart, and you would still run into problems of your blood being 50% dilute.  I haven't tested any of this because non-human blood isn't done yet, so the body may just fail even then.  If we do want rejection, and this isn't sufficient, additional code can be added to handle rejection once we have something to work with, ie when a race that uses another blood type is done.  If we want rejection, but this is too much, my above suggestions of a body part overriding how it handles reagents to be more universal or surgery setting the blood type of parts would alleviate it somewhat.

If that is NOT what we want, if we want foreign body parts to be usable freely, then yes, blood types are pointless and the code should be rewritten entirely.  Frankly, at that point, body parts should not have blood in them at all, and only reagents, and the heart should pump reagents to them, and blood would just be a number that determines how effective the heart and lungs are, and does oxy damage to the whole body when its low.

Edit Edit Edit: Perceiving some issues, gimme a bit to work this through

Bobtheapple: Okay, I had misunderstood something there, blood was still generic, ie there wasn't a different reagent for each kind of blood.  This has been fixed now, BloodType is the actual reagent being pumped, BloodType.Type is that blood's blood type (A, B, O, etc).  Currently there is no code to handle a mismatch in BloodType.Type, that can be added later once everything else is working.

corp-0: I kinda like the idea of organ rejection and making cross specie implants not ideal or harder to make them work although leaving the option and mechanics so they are possible.

Bod9001:
is it good now?

Bobtheapple:
This push should make it good enough to merge.  CPR works and falling down works.  There is still something off on how I'm calculating consumption, and the damage formulas could use some tweaking, but that's minor and the number of other things this changes means it should probably get added before I spend forever fine tuning things.

UnityProject/Assets/Scripts/HealthV2/Living/PlayerHealthV2.cs
```diff
@@ -119,12 +115,10 @@ protected override void Gib()
 			PlayerMove.PlayerScript.pushPull.VisibleState = false;
 			playerNetworkActions.ServerSpawnPlayerGhost();
 		}
-
-		bool RegisterPlayer.IControlPlayerState.AllowChange(bool rest)
```
Bod9001, line 123: I fix this in another PR

Bod9001: https://github.com/unitystation/unitystation/blob/4afe058c81b47744396ddd1113b7daf3835cc152/UnityProject/Assets/Scripts/HealthV2/Living/PlayerHealthV2.cs#L123

Bod9001:
all good to merge?

Bobtheapple:
Yeah

PerfectTangent:
did the rework of the health analyzer text add oxyloss to it?

PR #6273, opened 2021-03-19
better sprite handling

Bod9001:
if it can't pass a Sprite change it saves it for later, also simplified the de-serialisation side
fixed a bug where floating animation tried to play when character wasn't set, 
stop clients from receiving character data when it does not need it
fixes surgery
fixes laying down or triggering properly


UnityProject/Assets/Scripts/Messages/Server/SpritesMessages/SpriteUpdateMessage.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Text;
+using Google.Protobuf.WellKnownTypes;
```
corp-0: owo wats this

UnityProject/Assets/Scripts/Player/PlayerEffectsManager.cs
```diff
@@ -22,6 +22,7 @@ private void Awake()
 
     private void Update()
     {
+	    if (PlayerManager.PlayerScript?.PlayerSync == null ) return;
```
corp-0:
```suggestion
	    if (PlayerManager.PlayerScript.OrNull()?.PlayerSync == null ) return;
```
PlayerScript is a ManagedNetworkBehaviour which I assume inherits from UnityEngine.Object

UnityProject/Assets/Scripts/Messages/Server/SpritesMessages/SpriteUpdateMessage.cs
Bod9001: I think it got confused with the different type rider did

UnityProject/Assets/Scripts/Messages/Server/SpritesMessages/SpriteUpdateMessage.cs
```diff
@@ -36,193 +40,156 @@ public override void Process(NetMessage msg)
...
 			{
+
+				ProcessingData.Clear();
+				ProcessingData.AddRange(UnprocessedData);
+				UnprocessedData.Clear();
+				foreach (var UnprocessedDataSingle in UnprocessedData)
+				{
+					ProcessSection(0, 0, UnprocessedDataSingle);
+				}
+
```
ThatDan123: Clearing it then looping through???

Bod9001: oh yeah, good catch xD

PR #6277, opened 2021-03-20
Working Crayons and Spray Cans

ThatDan123:
CL: Added working crayons and spray cans

image

image

-Crayons cannot be used on walls only floors, spray cans can be used on walls and floors
-Crayons and cans have 30 charges, this PR does not implement can refilling
-Crayon graffiti picker UI

-Sadly i had to make two sets of overlay tiles one for walls and one for floors, due to layering. As you cant dynamically change the layering by instantiating a new SO and just changing that value as client requires that the tile SOs exist on their side for syncing

-This means if you want to add more graffiti types you need to add them for both walls and floors and the lists must be in the same order as it expects the same index values, see the GraffitiCategoryList and GraffitiList SOs


Bod9001:
had a look through and it all looks good

Bod9001:
I suppose we never gonna use the extra layers on the time up since there's no way of rendering just one layer of tiles, since I was planning that was how the multi-Z level stuff at work but doesn't seem like it would work, due to rendering and underfloor being a pain

Bod9001:
humm, as for having a limit of 20 watts stop it going past 20 if you just continuously add one to a tile

UnityProject/Assets/Scripts/Items/Tool/CrayonSprayCan.cs
```diff
@@ -0,0 +1,471 @@
...
+
+					UseAndCheckCharges(interaction);
+
+					//Should only ever be one of the overlay
+					return;
+				}
+			}
+
+			//Only allow 5 graffiti overlays on a tile
+			if (graffitiAlreadyOnTile.Count > 5)
```
Bod9001, line 232: ok cool beans

PR #6279, opened 2021-03-21
Fixes field generator

ThatDan123:
-Fixes field generator prefab
-Fixes field generator cargo data


Bod9001:
"Assets/Scenes/MainStations/SquareStation.unity: FieldGenerator (UnityEngine.Transform) is at 0,0 Please update the prefab/update the map/revert "
humm, don't think Anyone is working on that scene currently?

PR #6280, opened 2021-03-21
Moved Scriptable Objects out of Resources

ThatDan123:
-Moved all non singleton scriptable objects out of the resources folder

-Made new singletons for rightclickactions and chemistry reagents(for scrubber event)


Bod9001:
How should we merge this sense is a few other PRs with SOs?

ThatDan123:
If this is merged first then when they merge they'll just need keep their changes and then move the files they've changed to the new folder. 

Or those PRs can be merged first, then in this PR i'll move them. Up to you.

Bod9001:
some stuff going to take awhile to merge so I'll merge this first

PR #6283, opened 2021-03-22
Fixes Crayon Eating

ThatDan123:
Fixes crayon eating blocking graffiti selector


corp-0:
I feel like moving the edible component to the top of the component list would have been easier. This introduced a lot of dup code.

Issue #4145, opened 2020-05-05
Outpost and Box Updates


Ranzarth:
## Description

Decided to do problems, suggestions, and improvement tickets on every map including outpost and also box. Most of the outpost stuff I wrote here is for me to remember to get to at a later date.

### Outpost

- [x]  Bridge needs to be redone, too big.
- [ ]  Telecomms needs to be readded.
- [ ]  Gravity Room needs to be added.
- [x]  Medical needs some tweaks to the design.
- [x]  Engineering needs a renovation, looks weird with the hacked on sec office.
- [x]  Science needs to be extended, some of the rooms like toxins and robotics are too small.

### Box

- [x] Self destruct terminal should be moved to the vault
- [x] Bridge needs blast doors on its two side entrances
- [x] A button connecting the front doors of medical should be placed on the front desk
- [x] A button inside should also be connected to the front doors so anyone can let themselves out.
- [x] Needs a syndicate outpost
- [x] Still has the old V1 Syndicate shuttle

Comment below if something should be added/removed/changed

ThatDan123:
Box
-Bridge needs blast doors on its two side entrances

The reason i didnt map those was when i did it caused a really weird looping air sound, needs to be investigated more as it might be fixed.

ThatDan123:
Fixed up boxstation, blast doors seem to be ok now.

NoooneyDude:
**_Complete_** For boxstation, morgues exist in security and robotics, but not in medical's main morgue room.

corp-0:
Closing as the issue is quite old and has no progress. There is also a new mapping issue template that is better to track, since it is a single problem per issue.

If anything of this list is still a problem, it should be reported using said template.

PR #6284, opened 2021-03-22
Many small fixes and updates to Outpost's rooms

PerfectTangent:
### Purpose

This PR continues work on #4145 and makes some general modifications and improvement to Outpost station.

changes of significance:

- reworked atmos's foyer area to have a bit more space and not feel so squeezed in.
- shrunk the size of the bridge as described in #4145
- enlarged robotics as described in #4145. It's about double in size, now, with separate places for surgery and manufacturing
- added a maintenance door to the north of Atmos
- added a sec door in the Detective's room to let them walk straight into security without having to walk around through the front door.
- added some extra lighting in places in medical, botany, security, atmos, bathrooms and near arrivals
- added a security post next to arrivals
- rotated some consoles to have more appropriate facings
- slightly rearranged the lawyer's office so that it can support 2 attornies at law (probably is too small, though. Might move the office to near wherever the Courtroom ends up going, but that's for the next PR).
- added secondary entrance to cargo that exists to bring crates through for delivery
- replaced all instances of surgery_12 with operating tables
- fixed an issues in the command shuttle and the two docked utility shuttles that caused them to be depressurized (engine prefabs are not air tight).
- fixed a few mismatched tiles
- fixed about 15 orphaned pipes, mis rotated pipes, or missing pipes in the atmos pipe network
- added a missing tiny fan to one of the doors in the command shuttle
- found and deleted some assets that lacked sprites in the scene (portable air pumps and scrubbers)


PerfectTangent:
This is a draft PR because there is something very strange going on with Outpost's UnderFloor layer. It's not clear to me what is the problem, but it appears that the original UnderFloor layer has somehow had all of the pipes and wires on it become disconnected. 

Any new wires and pipes I mapped into the scene instead went on to a new UnderFloor (1) layer that spontaneously popped up. Deleting this layer cause the tile palette tools to stop functioning. Deleting the original UnderFloor layer doesn't seem to do much at all; none of the original wires and pipes in the scene are deleted with it. Same with hiding it in the hierarchy.

underfloor

Bod9001:
try deleting the rogue layer then restarting the editor/loading a different map then loading them up again

corp-0:
````
OutpostStation: "APC - Atmos02" has a null value in the list.
OutpostStation: "APC - Atmos02" has a null value in the list.
OutpostStation: "APC - Atmos02" has a null value in the list.
````

PerfectTangent:
Alright, fixed the UnderFloor layer issue. Thanks, Bod. Your suggest helped me solve it. I also fixed the null values. 

If the tests pass, then it should be good to merge. Then I can wash my hands of this.

PR #6295, opened 2021-03-25
Thread safe list wrapper with no allocations on iteration

Aranclanos:
### Purpose
Adds ThreadSafeList, a thread safe list wrapper that can be iterated without allocations.
Fixes the atmos manager removing a random pipe from its list instead of the assigned one.

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/AtmosManager.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/AtmosManager.cs
@@ -23,7 +23,8 @@ namespace Systems.Atmospherics
 
 		public HashSet<PipeData> inGameNewPipes = new HashSet<PipeData>();
 		public HashSet<FireAlarm> inGameFireAlarms = new HashSet<FireAlarm>();
-		public ConcurrentBag<PipeData> pipeToAdd = new ConcurrentBag<PipeData>();
+		private ThreadSafeList<PipeData> pipeList = new ThreadSafeList<PipeData>();
+		private GenericDelegate<PipeData> processPipeDelegator;
 
 		public static int currentTick;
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/AtmosManager.cs
```diff
@@ -23,7 +23,8 @@ public class AtmosManager : MonoBehaviour
 
 		public HashSet<PipeData> inGameNewPipes = new HashSet<PipeData>();
 		public HashSet<FireAlarm> inGameFireAlarms = new HashSet<FireAlarm>();
-		public ConcurrentBag<PipeData> pipeToAdd = new ConcurrentBag<PipeData>();
+		private ThreadSafeList<PipeData> pipeList = new ThreadSafeList<PipeData>();
```
Bod9001, line 26: isn't the problem with reagent mix being modified unsafely

InvalidOperationException: Collection was modified; enumeration operation may not execute.
System.Collections.Generic.Dictionary`2+Enumerator[TKey,TValue].MoveNext () (at <fb001e01371b4adca20013e0ac763896>:0)
Chemistry.ReagentMix.get_WholeHeatCapacity () (at /root/UnityProject/Assets/Scripts/Systems/Chemistry/ReagentMix.cs:61)
Chemistry.ReagentMix.get_InternalEnergy () (at /root/UnityProject/Assets/Scripts/Systems/Chemistry/ReagentMix.cs:79)

Aranclanos: Yes, but its unrelated to the pipe list. If everything works out, every list that needs to be threaded and iterated could be moved to this new one as long as its acceptable the way it handles iterations. 
No dev will have to type lock while using them

Issue #6276, opened 2021-03-20
Fix player sprite and creation issues and allow the character customization window to save multiple characters B:150


PerfectTangent:
### Description

**This is a health-related bounty.** This bounty requires a quality assurance pass on the player sprites in the character customization window and in game. Currently on develop, a variety of sprites are not displaying correctly,

known issues include;

- sprites not displaying when they should from certain character facings both in game and in the customization window
- sprites are offset or overlapping other layers in strange ways
- sprites are not coloring correctly (for example, the lizard tail)
- characters are occasionally appearing as disembodied floating skulls
- certain limbs or extra bits (moth antennas) are not appearing removed when they should be after surgical removal
- characters are spawning or respawning with the wrong or default skin color.
- ghost sprites are not visible

The randomize button in the character customizer also currently cannot randomize hair, facial hair, and socks. This needs to be reimplemented.

This bounty also calls for adding the feature to save and load multiple characters in the character customization window, saving all details about them.

## What is needed to fulfill the bounty:

- [x] fix all visual and layering issues with player sprites in the customizer and in game
- [x] reimplement the remaining broken features of the randomize button
- [x] add the ability to save and draw up multiple characters to the character customization window.
- [x] Make ghost sprites visible again

MaxIsJoe:
Since I'm already working on #6271 I'll claim this bounty.
That PR will still be pushed however once it's been reviewed and tested by Gillies, bod or dan.

MaxIsJoe:
Tick one off for this.

Now I'm going to work on ghosts and see why they're not showing up on clients most of the time.
Also if anyone has steps to reproduce this issue please send it to me, I've been killing myself both on local servers and on staging to see what's wrong and I can't replicate this issue.

MaxIsJoe:
The character creator task can be ticked off now

PerfectTangent:
this bounty has proven to be particularly nasty to sort out, and is being raised to 125.00 USD.

PerfectTangent:
After 16 PRs, this bounty is being marked as complete. there are a couple remaining bugs, like the disembodied skull bug, but the vast majority of the problems covered by this bounty have been fixed, assuming #6518 is merged.

Issue #6195, opened 2021-03-08
Health rework issues live on staging


Aranclanos:
WHEN YOU CHECK SOMETHING OFF REFERENCE THE PR THAT FIXED IT.

Roughly ordered by priority.

## BUGS

- [ ] ~~the bright red vision constricting crit overlay persists after death sometimes.~~ - needs reproduction, existed before health
- [ ] ~~bodies are getting stuck in place after death, becoming undraggable.~~ - existed before health
- [ ] when you are stunned or slipped there's no body drop sound.
- [ ] on spawn, initial movement is slow (as in #6552)
- [ ] moth wings and antenna are not rendering in the correct spot from certain sprite facings, both in the character customization window and in game. Sometimes they are completely absent. -  **Progress made with #6329, #6333, and #6361, but still not fully fixed.**
- [x] excessive blood production (see #6493) - mysteriously fixed, likely by #6634 and #6620
- [x] Welding things with welders is causing people to suffocate to death, even with internals, _possibly, **needs reproduction.**_. wtf, man. - **Fixed with #6470** 
- [x] when examining a character, the text is not describing whether they are hurt or dead. **fixed by #6519**
- [x] health analyzer doesn't track oxyloss damage anymore - **presumably fixed by #6471**
- [x] the character preview in the character selector is not rendering pixel perfect. - **fixed by #6440 and #6512** 
- [x] character sprites in the character customization window may not be rendering pixel perfect - **fixed by #6313, #6440, and #6512**
- [x] Some interactions still use the old LivingHealthBehaviour to apply damage, see in the comment below for more info
- [x] health doll doesnt update?
- [x] players spawning with a different character sheet - **possibly fixed with #6202**
- [x] hair names dropdown list having long names, ex: "head hair_hair_bob" **Fixed by #6356**
- [x] Human Lungs gasps! message - **Fixed with #6196**
- [x] o2 alert, pressure alerts and temperature alerts dont show up
- [x] ghosts are invisible - **Fixed with #6312**
- [x] arms and hands dont have cloth sprites when your character is looking to the sides - **Fixed with #6207**
- [x] human skin can be any color. - **Fixed with #6282**
- [x] lizard tail is not correctly coloring to match the body - **Fixed with #6357**
- [x] NPC mobs can't seem to be hurt or hurt plays in return - **Fixed with #6206** 
- [x] brains have no examine text - **Fixed with #6301** 
[...]

PR #6298, opened 2021-03-25
A Fix to the ghost visibility issues.

MaxIsJoe:
### Purpose

Fixes an issue reported in #6195 and ticks off one box in #6276 

### Changelog:
CL: Fixes ghost sprites visibility issues.

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/SceneConstruction/NestedManagers/AdminManager/AdminManager.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/SceneConstruction/NestedManagers/AdminManager/AdminManager.prefab
@@ -27,7 +27,8 @@ Transform:
   m_LocalRotation: {x: -0, y: -0, z: -0, w: 1}
   m_LocalPosition: {x: 230.5842, y: 320.23022, z: 0}
   m_LocalScale: {x: 1, y: 1, z: 1}
-  m_Children: []
+  m_Children:
+  - {fileID: 744200998588035166}
   m_Father: {fileID: 0}
   m_RootOrder: 0
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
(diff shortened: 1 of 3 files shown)


Aranclanos:
How to properly fix bugs:
step 1: find out whats causing it
step 2: fix it with that knowledge

I dont want to see any coroutines that just generate more problems down the line merged

MaxIsJoe:
> How to properly fix bugs:
> step 1: find out whats causing it

Can't be recreated easily and every time I ask someone about it they just give me possible scenarios for what could cause it, so I'm going with the most likely scenario as to why this is happening.

Aranclanos:
sometimes its hard

MaxIsJoe:
It's been confirmed by @Aranclanos that this issue happens on headless servers.
However I have yet to find this issue again while testing using the latest PRs in develop.

Could it be fixed? Needs more testing and for others to also hop on and test this possible fix I made.

Aranclanos:
problem still exists on yesterday's staging build for admins only.
it's heavily related to https://github.com/Aranclanos/unitystation/blob/6aabf31842bed562c051d914cb92198ce7ba66cc/UnityProject/Assets/Scripts/Messages/Server/AdminTools/AdminEnableMessage.cs addition and the following changes during the mirror update on this file.

You have all the tools needed to recreate it now to perform a proper fix

MaxIsJoe:
This solution I pushed actually fixes this issue, tested and works on headless and localserver.

We can merge now.

Aranclanos:
This is still not good enough, which itemslot is the admin trying to fetch and why is he missing it?

MaxIsJoe:
> This is still not good enough, which itemslot is the admin trying to fetch and why is he missing it?

there are no itemslots to begin with so I have no idea why it's trying to fetch itemslots,
Either way this kind of change is needed to avoid issues like this in the future and to tell people what's the problem. Someone else can worry about getting rid of this error log later, right now pushing a fix is more important.

Bod9001:
Aranclanos is adamant

MaxIsJoe:
> Aranclanos is adamant

And I did what my bounty told me to do, I fixed an issue and made a PR for it.
It's not a game breaking problem so I don't need to cover something outside of my bounty.

Aranclanos:
That just makes it worse, you're just placing a null check for something that shouldn't be EVER null for a bounty, which should require more effort than this, lazy coding shouldn't be rewarded.
And this is after I spent hours doing the research for you when you claimed it was impossible to recreate it, doing builds and testing, pointing out the exact problems and where to continue investigating towards the real cause of the issue. 

Stop being lazy, find the real reason of the bug and fix it.

PerfectTangent:
> > Aranclanos is adamant
> 
> And I did what my bounty told me to do, I fixed an issue and made a PR for it.
> It's not a game breaking problem so I don't need to cover something outside of my bounty.

you do have to complete it to our standards to qualify for the bounty and Aranclanos is a code reviewer. If he sees a problem consider taking the advice.

MaxIsJoe:
> That just makes it worse, you're just placing a null check for something that shouldn't be EVER null for a bounty, which should require more effort than this, lazy coding shouldn't be rewarded.

A fix is a fix, regardless of what you did to fix that issue.

> Stop being lazy, find the real reason of the bug and fix it.

I'm not lazy coding, I'm just fixing a very specific issue and I don't need to go beyond what's needed to be done. If I wanted to lazy code I would have done what Bod did with `characterroll()` and just commented out the entire thing and called it a day.
Actually no one on this project has the right to criticize or call anyone lazy for ANYTHING when the health update was merged in it's broken state because of one the reasons was "Bod was sick of working on it" and Gillies tried defending this terrible decision by ignoring their argument and trying to use the fact that me and them were frustrated by how low quality this merge was and how a lot of the code was terrible and undocumented against them saying "have a civil discussion or leave".

It's 100% hypocritical to tell anyone on this project that they're lazy when the entire feature freeze that we're in is because of the result of people being lazy as well.

Am I going to keep pointing out how terrible this decision is? Yes. And I will continue doing so in every given moment until the steam release. Anyways, going to go find a "fix with more effort" and if I don't find it by Monday, we merge without questions.

PerfectTangent:
> It's 100% hypocritical to say that anyone on this project is lazy when the entire feature freeze that we're in is because of the result of people being lazy as well

Two wrongs don't make a right. If we don't improve our standards, then the same thing will happen again.

> Anyways, going to go find a "fix with more effort" and if I don't find it by Monday, we merge without questions.

I'm afraid you don't get to make that decision, Max. I think it is an agreeable decision. But let's make something clear; bounties are a privilege that is administered and provided by us. It is our decision if and when things get merged and when things count as completed for a bounty, as stipulated in the contribution guidelines.

NoooneyDude:
I agree with Aranclanos. Have you seen the Madagascar movies? This is like the penguin smashing the blinking warning indicator to "solve" the problem. 

Edit: well I guess you did log it, so it's better than that.

MaxIsJoe:
I have no idea why, I shouldn't even have to wonder why, but headless straight up refuses to spawn in an itemstorage prefab for clients no matter what. This behavior is not present at all on localhost and their clients and no amount of testing or digging through code can explain why headless refuses to do something that works normally on localhost and their clients.

The fix I pushed makes the prefab already there from the start and referenced inside the admin manager, is this a bad solution? the results say no but my third eye that can see in the future can see Arancianos that this is unacceptable because i'm somehow still haven't solved the issue of headless refusing to spawn the itemstorage by it self yet everyone on discord seems to be scratching their heads in confusion because of this weird behavior.

Aranclanos:
closing in favour of #6312

PR #6299, opened 2021-03-26
war op fixes

AutumnNova:
fix two bugs and account for another possible bug


Aranclanos:
Please be more descriptive on what you're fixing

AutumnNova:
> Please be more descriptive on what you're fixing

only changes in the pr are actually assigning the tc trait, setting TcIncrement to the right thing and accounting for some possible fuckery related to withdrawing when their is less then TcIncrement TC avaliable

PR #6302, opened 2021-03-27
Fixes V2 doors being always closed on clients.

Aranclanos:
### Purpose
Fixes V2 doors being always closed on clients.

makes the door update message use the V2 system


Bod9001:
Just need to implement hacking then it's done

PR #6304, opened 2021-03-27
Headless Server Health Update Error Fix

ThatDan123:
-Fixes headless server error when trying to run client side health code

### DIFF
--- a/UnityProject/Assets/Scripts/HealthV2/Living/HealthStateController.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/HealthStateController.cs
@@ -79,7 +79,8 @@ namespace HealthV2
 		{
 			if (DollDataChanged)
 			{
-				SyncHealthDoll(healthDollData, JsonConvert.SerializeObject(CurrentHealthDollStorage));
+				healthDollData = JsonConvert.SerializeObject(CurrentHealthDollStorage);
+				DollDataChanged = false;
 			}
 		}
 


Aranclanos:
wouldnt removing the [Client] fix that

ThatDan123:
It would but then it would also run this which doesnt need to be run on a headless server.

`
CurrentHealthDollStorage = JsonConvert.DeserializeObject<HealthDollStorage>(healthDollData);
			for (int i = 0; i < CurrentHealthDollStorage.DollStates.Count; i++)
			{
				UIManager.PlayerHealthUI.bodyPartListeners[i].SetDamageColor(CurrentHealthDollStorage.DollStates[i].damageColor.UncompresseToColour());
				UIManager.PlayerHealthUI.bodyPartListeners[i].SetBodyPartColor(CurrentHealthDollStorage.DollStates[i].bodyPartColor.UncompresseToColour());
			}
`

Aranclanos:
will the editor and non headless servers stop updating the doll color now?

ThatDan123:
nope, as you connect as local host in editor so it would run the client hook

ThatDan123:
I think...right?

Aranclanos:
okay apparetly the server never sets DollDataChanged on his own character to true so the stuff in LateUpdate() doesn't even run so its irrelevant

ThatDan123:
its set to true on line 142

Aranclanos:
yeye but that function doesnt run if the PlayerHealthUI is the localplayer

ThatDan123:
Yeah cus of the check in the PlayerHealthUI, it wont run on localplayer, but this still is syncing for other clients for their health updates so is still needed

Bod9001:
do you need to add is server == false instead of the hook?

Bod9001:
inside of the hook*

PR #6312, opened 2021-03-29
Fixes admins ghost sprite visibility

Aranclanos:
### Purpose
Fixes admins ghost sprite visibility.
Removes CustomNetSceneChecker from the AdminGhostItemStorageHolder since they're in a manager, not a scene.


corp-0:
what was the fix? I only see prefab changes

corp-0:
> Removes CustomNetSceneChecker from the AdminGhostItemStorageHolder since they're in a manager, not a scene.

ah

PR #6316, opened 2021-03-31
Fixes Racks not properly despawning when deconstructed or destroyed. #6190

### LINKED ISSUE #6190: Racks are not properly despawning when deconstructed or destroyed.
### Environment
- Game Version: latest staging

when a rack is disassembled or destroyed, it does not properly despawn, and still persists and blocks passage. It can even be seen in the item wheel.

### Steps to reproduce

1. destroy or disassemble a rack
2. observe how the rack sprite disappeared, but no rack parts were dropped
3. try to move through the space, observe how you cannot.
3. right click on the tile and see that there is still a rack there

JesterX666:
### Purpose
Fixes #6190

### Notes:
Simple one.  Reattached the rackPart prefab to the property that was supposed to hold it.  Also, I made the property private, serializedField and added a tooltip.

### Changelog:
CL:Bugfix: Racks are now spawning rack parts when deconstructed.

### DIFF
--- a/UnityProject/Assets/Resources/Prefabs/Objects/Construction/Rack.prefab
+++ b/UnityProject/Assets/Resources/Prefabs/Objects/Construction/Rack.prefab
@@ -14,7 +14,8 @@ MonoBehaviour:
   m_EditorClassIdentifier: 
   syncMode: 0
   syncInterval: 0.1
-  rackParts: {fileID: 0}
+  rackParts: {fileID: 1483045645304554680, guid: a9f1aff2f68e03a43b3ccd4f75c3adee,
+    type: 3}
 --- !u!1001 &6647483129041870009
 PrefabInstance:
   m_ObjectHideFlags: 0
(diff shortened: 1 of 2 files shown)


PerfectTangent:
we recently saw reports that this is apparently happening with a number of things. 

See here;
https://discordapp.com/channels/273774715741667329/668605946691452949/825022637758742539

it may be that this problem is of a larger scope than we initially realized.

JesterX666:
> we recently saw reports that this is apparently happening with a number of things.
> 
> See here;
> https://discordapp.com/channels/273774715741667329/668605946691452949/825022637758742539
> 
> it may be that this problem is of a larger scope than we initially realized.

RackParts have been lost from rack.prefab in commit 5d02945433e772f9ca9b47637d86d78db0b79469 (october 7, 2020)
Which is a table + rack only commit.  It was not lost due to a commit that affected other things such as the link you described.
Alas, the reason for girders and such is another problem in itself.

PR #6319, opened 2021-03-31
Armor reimplementation & formula rework & penetration

### LINKED ISSUE #6288: Armor formula rework and reimplementation B:55
## Description

**This is a health related bounty.** This bounty calls for adding armor back into the game, which was commented out in the great health purge due to incompatibility with the new changes. Since armor needs to be readded to the game, now is a good opportunity to rework how it is calculated.

the old formula for armor was purely additive. Multiple armor sources protecting a particular body part would be added together, expressed as a percent, and multiplied against the damage. Example: if someone had 20 armor from one piece and 50 armor from another, they were added to make 70 armor, and that armor would reduce all damage to that body part by 70%.

Simply adding the armors together results in a couple undesireable situations, namely that you can get enough armor together to equal 100 or greater armor, in which case you are totally immune to damages of that type of armor.

The new formula should instead be proportionally added: a source of 20 armor reduces incoming damage by 20% and the second source of 50 armor should reduce the leftover damage afterwards by an additional 50%. 

_Initial damage * armor A, take the remaining damage * armor B, take the remaining remaining damage * armor C = damage dealt._

When expressed as an process of repeated multiplication, it does not matter which armor reduction is calculated first; you will always get the same results. This system also still allows devs and admins to make armor pieces that are totally immune to damage if that is desired by setting the armor value directly to 100. 

This formula rework also won't have much of an effect on the current game balance because there are not that many ways to stack armor at this time, but it will be a much easier thing to do in the future, so this formula helps future proof armor balance a bit.

Second, a important feature of weapons is the ability to penetrate armor. This is essentially anti-armor, subtracting from each source of armor before the armor is used in the same fashion that armor subtracts from the total damage.

## What is required to fulfill the bounty:

- [x] rework armor back into the game using the new formula described above
- [x] add armor penetration as a stat to weapons and projectiles.

Lizurt:
### Purpose
fixes #6288

<details>
<summary>Highlights</summary>

- Damage to a body part protected by multiple armor is now calculated using the formula: `InitialDamage * armor1 * armor2 * ... * ArmorN = damageDealt` instead of `InitialDamage * (Armor1 + Armor2 + ... + armorN)`.
- `WearableArmor` now protects specific body parts with specific protection options. This makes it possible to make modular armor.
- `WearableArmor`'s armor bit mask (enums) have been replaced with `BodyPart`'s enum.
- The armor of the body parts now extends not only to the organs(for example, it is logical that the xenomorph's exoskeleton will be stronger than a pathetic human skin).
- Bullects can now penetrate armor
</details>

It was tested on headless server and seems to work fine.

### Changelog:

CL: Armor system was reworked


corp-0:
> WearableArmor's armor bit mask (enums) have been replaced with BodyPart's enum.

does this mean armor can only protect one body part at a time? The intention of the bitflags was recreating how they behave in byond where you actually use bitflags to determine what body parts a particular armor piece protects.

corp-0:
Nevermind just read it is using a list now

Lizurt:
Should I edit prefabs or assets here for armor penetration and armor or is it better to do it in a separate PR?

corp-0:
Uhm, you can do it here but if you can bundle the prefab changes in a commit separated from code changes, I'd love you to all eternity.

Lizurt:
It looks like that's all. The PR is ready for review

UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartDamage.cs
```diff
@@ -8,18 +8,19 @@ namespace HealthV2
 	public partial class BodyPart
 	{
 		/// <summary>
-		/// The amount damage taken by this is modified by
+		/// The armor of the clothing covering a part of the body, ignoring selfArmor.
 		/// </summary>
-		[Tooltip("The amount that damage taken by this is modified by")]
-		public Armor BodyPartArmour = new Armor();
+		[Tooltip("The armor of the clothing covering a part of the body, ignoring selfArmor.")]
+		public readonly LinkedList<Armor> ClothingArmors = new LinkedList<Armor>();
```
corp-0: from what I understand, this field is not supposed to be set in inspector, make it ``[NonSerializable]`` or better yet, make it private and expose an interface with the intended interaction (getting, setting with validation, ect)

Lizurt: Yes, you're right, this field isn't for the inspector.

Maybe it's better to make this field `internal`? Because as I noticed in this project, it is normal to practice public fields without getters and setters.

corp-0: At least the old code yes, uses and abuses of public fields. Our contribution guides say to prefer public properties for private fields, and a lot of our code is not complying but I always try to persuade contributors to make new code comply.

UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartDamage.cs
```diff
@@ -8,18 +8,19 @@ namespace HealthV2
...
 
 		/// <summary>
+		/// The armor of the body part itself, ignoring the clothing (for example the xenomorph's exoskeleton).
 		/// The amount damage taken by body parts contained within this body part is modified by
 		/// increases as this body part sustains more damage, to a maximum of 100% damage when the
 		/// health of this part is 0.
 		/// </summary>
-		[Tooltip("The amount that damage taken by contained body parts is modified by")]
-		public Armor SubOrganBodyPartArmour = new Armor();
```
Bod9001, line 22: There is technically 3 Armors , clothing, body part itself and then protection of sub body parts

Bod9001: so,  this is so you can implement Gollum's that take less damage from certain sources inherently,
 and the sub body armour is for stuff like a skull that would protect the organs below it

Lizurt: `SelfArmor` and `SubOrganBodyPartArmour` are united. You can still make Gollum, give his skull `SelfArmor`, and then other body parts.

UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartDamage.cs
Lizurt: Done

PerfectTangent:
How is negative **armor** handled? Does it increase damage taken in the same manner that positive armor reduces it?

Bod9001:
You mean negative armour, humm yeah multiplies it

UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartDamage.cs
Bod9001: but you can't separate how much damage is done to the body part itself versus how much damage is transferred to the sub- organs?

Lizurt: I can by assigning different armor values to the body part and its innards since the organs are also body part

Lizurt: I think I understood you, I see the problem now. I'll fix it

Lizurt: > I think I understood you, I see the problem now. I'll fix it.

~~Or wait, no 😄 . I don't need to set organ's armor, but I use `selfArmor` of the body part to protect its organs.~~

UPD: Nevermind, I still can't "separate how much damage is done to the body part itself versus how much damage is transferred to the sub- organs"

Lizurt: Done.

(on the third try I finally understood what you meant. I feel stoopid)

corp-0:
do we ever want to allow negative damage tho? I think healing should be done using its own interface rather than making negative damage.

corp-0:
nevermind I just read Bod's answer.

UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartDamage.cs
Bod9001: looks good

Bod9001:
dam it, forgot to check before merging xD, ClothingArmors  isn't used anywhere

Bod9001:
it's okay it's used I was just a dumb dumb

PR #6322, opened 2021-04-01
Making ScriptableObject tests work again.

PetMudstone:
### Purpose
- Makes the tests for scriptable objects work again by pathing to the proper folder.
- Fixes typos in the test names.


corp-0:
it should check on both since we still have some ScriptableObjects in Resources, the singleton ones.

Issue #6091, opened 2021-02-22
Remove Connected Players On Stat Tab


ThatDan123:
The tab only shows in game players and their job, however that should only be in the crew manifest which only select jobs have.
Can lead to exploits where people look at stats tab and then the player list to find out if its a nukie round

![image](https://user-images.githubusercontent.com/56727168/108755004-98635180-753e-11eb-8206-f13211214c5f.png)

If people want to see OOC names, then the player list button should be used instead.

![image](https://user-images.githubusercontent.com/56727168/108755024-9f8a5f80-753e-11eb-9835-61d12cc9cdf0.png)

ThatDan123:
Or just move it to a new nettab for a new console

PR #6324, opened 2021-04-02
Hides player names and player jobs from the stats tab

MaxIsJoe:
### Purpose
Title. solves #6091 

### Notes:
It's cleaner to just display all players inside that tab instead of creating an entire new window for it.

### Changelog:
CL: Antags and character names will no longer be displayed inside of the stats tab.


ThatDan123 (changes requested):
Dont need to show player usernames at all now as they are shown in the player list.

Also you can remove the networking parts which send the names and jobs as they shouldnt be needed anymore

MaxIsJoe:
> Dont need to show player usernames at all now as they are shown in the player list.
> 
> Also you can remove the networking parts which send the names and jobs as they shouldnt be needed anymore

I've removed usernames from the stats tab as you requested, however removing the networking parts broke the game for me so I guess they're still needed somewhere. Either way PR is done, ready to merge.

ThatDan123 (approved):

PR #6326, opened 2021-04-03
Adds ID and PDA clothing overlays, fixes some issues and other cool stuff

MrAsbestos:
### Purpose
This PR adds ID and PDA overlays, this means that when you wear an ID or a PDA, it will show up on your character!
Also fixes some issues related to overlay rendering, tank rotations and other cool stuff, check it out in the changelog.

### Changelog:
CL: Added overlays for every ID and PDA in the game.
CL: Nuclear devices now have proper names and descriptions.
CL: You can now wear emags in your ID slot (why? why the hell not? plus it's literally an ID, you can wear it)
CL: Fixed some rotation issues with tanks (they used turn in incorrect ways for some reason, for example if you pulled it to the right, the sprite would go up)
CL: Fixed some sprite overlay rendering issues (back items used to go on top of items, so if you held a sniper rifle, your backpack straps would go through it, also head items rendered on top of back items, so if you weared a sniper rifle on your back, the sniper rifle would go inside your head item, that is fixed now.)
CL: Fixed sprayer bottles not having any sounds.

a-a- and that's all folks!!


MrAsbestos:
wut, how did it fail? huh

MrAsbestos:
ah nvm, it seems that the helper just pooped itself lol

PR #6328, opened 2021-04-05
cleanup gun scripts

AutumnNova:
hopefully this should make things a bit easier to understand
feel free to suggest other changes that could improve readability


corp-0:
you removed that editor hack to hide fields?

AutumnNova:
> you removed that editor hack to hide fields?

yeah dont need to use that anymore, was messy anyway

Bod9001:
looks like it doesn't compile?

corp-0:
you probably forgot to import the namespace somewhere

AutumnNova:
compiles fine for me?

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
```diff
@@ -92,25 +93,25 @@ protected StandardProgressActionConfig ProgressConfig
 		/// <summary>
 		/// Whether the gun uses an internal magazine.
 		/// </summary>
-		[HideInInspector, Tooltip("Effects if the gun will use an internal mag")] // will be shown by the code at the very end, if appropriate
+		[HideIf("SmartGun"), Tooltip("Effects if the gun will use an internal mag")]
```
corp-0:
```suggestion
		[HideIf(nameof(SmartGun)), Tooltip("Effects if the gun will use an internal mag")]
```

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
```diff
@@ -92,25 +93,25 @@ protected StandardProgressActionConfig ProgressConfig
...
 		public bool MagInternal = false;
 
 		/// <summary>
 		/// If the gun should eject it's magazine automatically (external-magazine-specific)
 		/// </summary>
-		[HideInInspector, Tooltip("If the gun should eject an empty mag automatically")] // will be shown by the code at the very end, if appropriate
+		[HideIf("MagInternal"), Tooltip("If the gun should eject an empty mag automatically")]
```
corp-0:
```suggestion
		[HideIf(nameof(MagInternal)), Tooltip("If the gun should eject an empty mag automatically")]
```

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
```diff
@@ -1105,23 +1097,15 @@ public enum WeaponType
 		Burst = 2
 	}
 
-#if UNITY_EDITOR
-	[CustomEditor(typeof(Gun), true)]
-	public class GunEditor : Editor
+	/// <summary>
+	///	//TODO: write a half decent summary
```
corp-0: to do it e.e

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
```diff
@@ -1105,23 +1097,15 @@ public enum WeaponType
...
+	/// </summary>
+	public enum TriggerBehaviour
```
corp-0: this enum represents particular classes, doesn't it? I think we should test using the type itself intead of enum values you have to update every time you add a new strategy.

corp-0 (changes requested):
a couple changes for the sake of future proofing. I wish I was more knowledgeable with this weapon system, it really looks like it needs a deep clean up in architecture.

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
AutumnNova: shit i totally forgot about that, my bad

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
AutumnNova: the only reason this enum exists is because returning a number isn't particularly clear as to what the fuck it does

corp-0: oh, the strategies are just numbers?

AutumnNova: a method in a component on the firing pin is called, the method returns a value which effects what will happen, be it nothing, working normally or playing the bike horn sfx

corp-0: why don't we make the pin act by itself instead of telling weapon how to act?

AutumnNova: i assume you mean either calling a method on the guns component, which would require creating a method that only contains the switch case or alternatively moving over all firing behaviour to the pin itself, which is a huge amount of work

AutumnNova:
> a couple changes for the sake of future proofing. I wish I was more knowledgeable with this weapon system, it really looks like it needs a deep clean up in architecture.

Ill be entirely honest with you, i don't think much can be removed without removing behaviour

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
corp-0: I think the later is the way to go. Instead of switching what the pin returns, you have a base abstract pin class and make the rest inherit from it. Add an abstract trigger method and call that from Gun.cs with a reference to the abstract class so it works whatever pin you have attached.

corp-0: I understand all of this is a lot of work. This PR intends to clean these scripts and honestly one of the biggest problems with Gun.cs is that it does way too many things. If you're not willing to do that level of clean up, it is aight, we can do baby steps until that class nice and tidy.

AutumnNova: i dont see how it would be possible to even seperate them into seperate components or something
the if2 interactions are 1/4th of the script alone

corp-0: well, the pin behaviours sound like a good start. You would still be calling them from the IF2 methods inside Gun.cs but what pulling the trigger does will be called directly from pin object with the structure I talked about above. If there is something I'm not seeing, let me know.

corp-0:
cool, now just to make sure we comply with best practices, make the abstract class I talked about yesterday. The idea is making classes have dependencies on abstracts/interfaces rather than concrete classes. This is quite easy to do and you won't lose anything, you just make a new 

```cs
public abstract class PinBase: MonoBehaviour
{
    public abstract void ClientBehaviour();
    public abstract void ServerBehaviour();
}
````
And then you make GenericPin, ClownPin, HonkPin or any other pin that overrode GenericPin methods inherit from that one. 

This will ensure the inheritance is flexible enough so we can keep adding pins without worrying about undesired behaviour and make Gun class only depend on an abstract rather than a concrete.

**Review by corp-0 — CHANGES_REQUESTED** [review]
cool, now just to make sure we comply with best practices, make the abstract class I talked about yesterday. The idea is making classes have dependencies on abstracts/interfaces rather than concrete classes. This is quite easy to do and you won't lose anything, you just make a new 

```cs
public abstract class PinBase: MonoBehaviour
{
    public abstract void ClientBehaviour();
    public abstract void ServerBehaviour();
}
````
And then you make GenericPin, ClownPin, HonkPin or any other pin that overrode GenericPin methods inherit from that one. 

This will ensure the inheritance is flexible enough so we can keep adding pins without worrying about undesired behaviour and make Gun class only depend on an abstract rather than a concrete.

**Comment by AutumnNova** [comment]
dont merge this just yet, still need to fix a few things

File: UnityProject/Assets/Scripts/Items/Weapons/FiringPins/PinBase.cs [hunk 5]
```diff
@@ -0,0 +1,70 @@
...
+
+		protected JobType GetJobClient()
+		{
+			return PlayerManager.LocalPlayerScript.mind.occupation.JobType; 
+		}
+
+		protected void ClumsyShotServer(AimApply interaction)
+		{
+			//shooting a non-clusmy weapon as a clusmy person
+			int chance = rnd.Next(0 ,2);
```
corp-0: use ``DM.Prob()`` to test chances instead

ThatDan123:
Sorry to fix the conflicts you'll need to merge upstream, keep your changed files, then move them out of the old prefab folder and into the new prefab folder that is out of resources.

AutumnNova:
should be good to go now

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
```diff
@@ -287,6 +280,7 @@ private void Init()
 			}
 
 			Inventory.ServerAdd(Spawn.ServerPrefab(pinPrefab).GameObject, pinSlot);
+			FiringPin.gunComp = this.GetComponent<Gun>();
```
corp-0, line 283: just ``this`` should be enough since ``this`` is the Gun component, it didn't work like that?

corp-0 (approved):

corp-0:
seems like it isn't compiling

Issue #6150, opened 2021-03-01
Add the random bounty system and make the Cargo Console Emaggable bounty B:140


PerfectTangent:
### Description

**This Bounty calls for the addition the random bounty system and emagging to the cargo console as well as allowing admins to generate custom bounties.**

Bounties are a way for Cargo to make additional money by snooping through the station or asking other departments for assistance. Each bounty has a name, description, reward in credits, and a priority marking. High priority bounties multiply the reward, with the default multiplier being 2.

Bounties are split into categories to get a distribution of bounties across different departments.

The bounty console lists all of the available bounties. It can print paper listing every bounty available. Bounties that have been fulfilled can be claimed with the console - money is not added to the cargo account until they are claimed. After 5 bounties are claimed, additional ones are added to the list. Currently, it is not possible to add all bounties that exist in /tg/ because several of them require other system such as science to be developed, but we should add all of them that we can.

When the Cargo console is Emagged, it should allow players to purchase special crates that they can not typically purchase. (asthetically, if you want my 2c, these options should be colored red as opposed to the cargo console's typically yellow appearance.)

Lastly, Admins should have a system that allows them to generate brand new bounties with custom text, item fulfillment requirements, and variable credit payouts.

### what is required to fulfill the bounty:

- [x] add the bounties tab to the Cargo console. (added in #6330)
- [x] add an addition to the event system for randomly drawing bounties from a pool. Admins should be able to disable the generation of new bounties if they so wish, as well.
- [x] create a method for admins to generate custom bounties.
- [x] let the cargo console sell additional items after being Emagged. (added in #8632)

### Notes

Bounty Generation in /tg/:
https://github.com/tgstation/tgstation/blob/fa47976cbedc44cf41751d5a0db5701f69412eaa/code/modules/cargo/bounty.dm
Bounty Console:
https://github.com/tgstation/tgstation/blob/2029163d335356dc5f78ddf04f0750c4fa06c429/code/modules/cargo/bounty_console.dm
List of Bounties:
https://github.com/tgstation/tgstation/tree/2591ce17a6609002109fc8eafa48ae0479fc0b8e/code/modules/cargo/bounties

I guess you could say this is a _bounty bounty...._

Aranclanos:
gonna tackle this one

Aranclanos:
![image](https://user-images.githubusercontent.com/701959/114322156-0a92f400-9af5-11eb-98de-1d9ce545408c.png)

[...]

PR #6330, opened 2021-04-05
Adds the base for cargo bounties

Aranclanos:
### Purpose
Adds cargo bounties base, three different bounties for cargo techs to complete from roundstart.
Moves the CargoData SO and script to their correct folders.
Once arriving centcomm the cargo shuttle will check the contents of each wrapped objects for processing and ignore the wrapping object price values.
Adds the bounty window to the cargo console UI.
Reworks the cargo console UI, making it simpler and easier to understand and fixing some bugs when two consoles were used at the same time.
Adds the base for the emagging of the cargo console.

makes progress on #6150.

### Notes:
To check for completion status of the bounties, they'll check for itemattributes traits
The three base bounties are to get 5 wrenches, 5 welders and 5 crowbars. They reward 200 points.


UnityProject/Assets/Scripts/Managers/CargoManager.cs
```diff
@@ -303,18 +308,66 @@ public void DestroyItem(ObjectBehaviour item, HashSet<GameObject> alreadySold)
 			export.Count += count;
 			export.TotalValue += credits;
 
+			var itemAttributes = item.GetComponent<ItemAttributesV2>();
+			if (itemAttributes)
+			{
+				foreach (var itemTrait in itemAttributes.GetTraits())
```
corp-0, line 314: can you reformulate this logic to reduce nesting? Maybe use ``Validations.HasItemTrait`` instead of iterating all traits or even extract some portions into private methods.

UnityProject/Assets/Scripts/Systems/Cargo/CargoData.cs
```diff
@@ -6,8 +6,8 @@ namespace Systems.Cargo
 {
 	public class CargoData : ScriptableObject
 	{
-		[ReorderableList]
```
corp-0, line 9: why did you remove the tooltip and reorderable attribute?

Aranclanos:
updated

UnityProject/Assets/Scripts/Managers/CargoManager.cs
```diff
@@ -223,11 +219,20 @@ public void OnShuttleArrival()
...
+			OnShuttleUpdate.Invoke();
 		}
 
-		public void DestroyItem(ObjectBehaviour item, HashSet<GameObject> alreadySold)
+		public void ProcessCargo(ObjectBehaviour item, HashSet<GameObject> alreadySold)
 		{
+			var wrappedItem = item.GetComponent<WrappedBase>();
+			if (wrappedItem)
+			{
+				var wrappedContents = wrappedItem.GetOrGenerateContent().GetComponent<ObjectBehaviour>();
```
corp-0: this ObjectBehaviour can be null

PerfectTangent:
the bounties should appear in the status tab under the centcom messages whenever they are fulfilled.

PerfectTangent:
the bounties themselves need to have an entry for the name of the bounty and a separate entry that can be filled with flavor text. Since the flavor text can get pretty long for some bounties, I think the description should be viewed by clicking the bounty listing in the bounty tab; so, a bounty would display in the tab as `200 credits - send 10 pens` and when you click on the bounty, it would expand to display the full description: `200 credits - send 10 pens - We are hosting the intergalactic pen balancing competition. We need you to send us some standardized ball point pens.`

Since bounties can be filled out over multiple shipments, I think we need to see the progress made on a bounty. If you have to ship 5 wrenches and you ship 2 wrenches, it should show in the bounty 2/5 wrenches sent.

Issue #6097, opened 2021-02-22
Windows don't have breaking sounds


MrAsbestos:
### Detailed description of bug
When you break a window, it is completely silent

### Steps to reproduce
1.  Find any window
2.  Break it
3.  Notice how it doesn't make any sound

ewy1:
I'm working on a generalized solution for this that should allow breaking sounds for all tiles.

MrAsbestos:
Oooh cool

If I remember well, I partially fixed the window breaking sounds, but there was something weird, window breaking sounds only played when they were unanchored, very weird...

NoooneyDude:
Unanchored windows are gameobjects, while installed ones are tiles. The two types are handled quite differently.

NoooneyDude:
Presumably fixed by #6345.

PR #6345, opened 2021-04-10
Feature: tile destruction sounds

ghost:
### Purpose
A simple way to add tile destruction sounds to basic tiles through the editor. Currently only windows have breaking sounds. This is a very small change; the commits should speak for themselves.

I wrote this following issue #6097

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
@@ -2,7 +2,6 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
-using Light2D;
 using UnityEngine;
 using UnityEngine.Events;
 using Debug = UnityEngine.Debug;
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
```diff
@@ -1272,6 +1273,12 @@ public void RemoveTileWithlayer(Vector3Int position, LayerType refLayer)
 
 				if (TileLcation != null)
 				{
+					if (TileLcation.Tile is BasicTile tile && tile.SoundOnDestroy.Count > 0)
+					{
+						// Because this function is called through a network message, the sound does not have to
+						// be played through the network, as all clients will receive the network message.
+						SoundManager.PlayAtPosition(tile.SoundOnDestroy.RandomElement(), position);
+					}
```
ThatDan123: Dont think this is the best place to put this as it would play for any type of removing/changing a tile, check out TilemapDamage.cs around line 96

ewy1:
Very well! I had placed it there before but moved it because this'd marginally save networking, but I understand this is a lot more logical.

corp-0:
I'm lost, we already had this feature in game. When was it removed?

CdrCross:
> 
> 
> I'm lost, we already had this feature in game. When was it removed?

The PR is in response to Issue #6097. Prolly should be labelled as a bug fix rather than a feature.

ewy1:
I couldn't find any existing way to add breaking sounds so I went ahead and made my own. There are hit sounds; maybe that's causing confusion? Or maybe I missed tile breaking sounds entirely? Feel free to correct me.

corp-0:
> There are hit sounds; maybe that's causing confusion?

Oh yes, that seems to be the case

PR #6346, opened 2021-04-10
Self sufficient vent & scrubbers GC reduction

Aranclanos:
### Purpose
Self sufficient vents and scrubbers will no longer create a new gasmix on each tick update, reducing the GC that they generate.

### Notes:
The amount of GC is highly tied to each map, based on how many of these self sufficient pipes exist in the scene


Bod9001:
looks good

PR #6352, opened 2021-04-11
Fixed Autolathe destruction bug

### LINKED ISSUE #6348: Invincible Circuit Imprinter
### Environment
- Game Version: 21040900
- OS Version: MacOSX

Circuit Imprinter is invincible to admin destruction tools and even the "Smash" variable simply creates a machine frame behind the circuit imprinter.

Screen Shot 2021-04-10 at 2 02 22 PM

### Steps to reproduce
1. Place Circuit Imprinter
2. Attempt to destroy it

Semuca:
### Purpose
Fixes #6348. When the server hooks were being removed, Autolathes/Circuit Imprintors with no materialStorageLink would cause an error in the despawning process. 

### Changelog:
CL: Fixed Autolathes not being destroyed when not connected to a material silo or storage.

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Machines/Autolathe/Autolathe.cs
+++ b/UnityProject/Assets/Scripts/Objects/Machines/Autolathe/Autolathe.cs
@@ -174,7 +174,10 @@ namespace Objects.Machines
 				StopCoroutine(currentProduction);
 				currentProduction = null;
 			}
-			materialStorageLink.Despawn();
+			if (materialStorageLink != null)
+			{
+				materialStorageLink.Despawn();
+			}
 		}
 
 		public void PowerNetworkUpdate(float Voltage) { }


corp-0:
why was materialStorageLink null though?

corp-0:
> CL: Added null check to Autolathe's materialStorageLink during destruction.

change this line to something that makes more sense for the final user, maybe something like "Fixed autolathes sometimes not despawning when destroyed".

Semuca:
> why was materialStorageLink null though?

My bad, I'll clarify. materialStorageLink is from a recent update where lathes can have their own material silos or storages, separate from the lathe. With machines that are newly made, or just not connected to a seperate silo, they won't have a materialStorageLink by default. I've also updated the changelog, I didn't realise the changelog was directly sent to stationHub.

corp-0:
wonderful, thank you very much

PR #6354, opened 2021-04-11
Updates Prefabs for Scrubber and Vents, as well as on some Maps

Akbadain29:
### Purpose
Updates Prefabs for Scrubber and Vents, some maps still need to be moved over to use AirVent and AirScrubber prefabs to account for the Air Monitors. This PR fixes up Box, Square and UnReal. Pog is fine, but Fall and Outpost need to be updated at some point in the future.


Akbadain29:
Oh dear.
> PogStation: "AirScrubber (14)" miss APC reference.
PogStation: "AirVent (3)" miss APC reference.
PogStation: "AirScrubber (9)" miss APC reference.
PogStation: "AirVent (19)" miss APC reference.
PogStation: "AirVent (7)" miss APC reference.

I haven't even committed that one lol.

> BoxStationV1: "APC" has a null value in the list.
Pity that all of the APCs are now called APC, I'll have to label them all and see if I can find the APC in question.

Akbadain29:
Ok, tests should pass now and I've also fixed up Pog so the tests clear. Someone is working on a differently named Pog map which isn't the one modified here.

PR #6356, opened 2021-04-11
Fixed dropdown names for hair, facial hair, etc.

MaxIsJoe:
### Purpose
Part of the #6276 bounty

### Changelog:
CL: Fixes dropdown names.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Sprite Handler/SpriteDataSO.cs
+++ b/UnityProject/Assets/Scripts/Core/Sprite Handler/SpriteDataSO.cs
@@ -12,6 +12,8 @@ public class SpriteDataSO : ScriptableObject
 	public bool IsPalette = false;
 	public int setID = -1;
 
+	public string DisplayName;
+
 	[System.Serializable]
 	public class Variant
 	{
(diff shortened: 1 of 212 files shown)


UnityProject/Assets/Scripts/Core/Sprite Handler/SpriteDataSO.cs
```diff
@@ -12,6 +12,8 @@ public class SpriteDataSO : ScriptableObject
 	public bool IsPalette = false;
 	public int setID = -1;
 
+	public string DisplayName;
```
Bod9001, line 15: why not just rename file?

MaxIsJoe: 1. Underwear, Undershirts, etc already have some sort of display name that let's them display correctly in game while hair, facial hair, lizard tails, etc don't. This makes it so everything is equal.
2. Because that's a terrible fix.

Bod9001: Uses less obscure names and Makes it easy to find in project I don't get what the problem is

PR #6357, opened 2021-04-12
Lizard tails will now match their skin colors.

MaxIsJoe:
### Purpose
Part of the #6276 bounty
Adds in a bool that lets us check of this character's customizations mimic their skin tone colors instead of having customizable ones.

### Notes:
The hair part is intentional.

### Changelog:
CL: Lizards customizations will now copy their skin tone's color.


UnityProject/Assets/Scripts/Player/PlayerHealthData.cs
```diff
@@ -39,6 +39,8 @@ public class RaceHealthData
 	public BloodType BloodType;
 
 	public ImplantProcedure RootImplantProcedure;
+
+	public bool BodyPartsShareSameSkinColor = false;
```
Bod9001: why not have that on the body part customisation? instead

UnityProject/Assets/Scripts/UI/Systems/Lobby/CharacterCustomization.cs
```diff
@@ -61,7 +61,7 @@ public class CharacterCustomization : MonoBehaviour
 		public CharacterDir currentDir;
 
 		[SerializeField] private List<Color> availableSkinColors;
-		private CharacterSettings currentCharacter;
+		public CharacterSettings currentCharacter;
```
corp-0: do you need to set this from outside or only read it from outside?

corp-0: https://softwareengineering.stackexchange.com/questions/161303/is-it-bad-practice-to-use-public-fields

UnityProject/Assets/Scripts/Player/PlayerHealthData.cs
MaxIsJoe: What about body parts that are shared between all races? for example, hair and facial hair are treated as body parts and are shared between all races; if you want it to share colors with your race's skintone you'll run into a problem where enabling this on the hair causes it to happen across all races.

This is a better solution because at the moment there is no race or use case what-so-ever on US13 or SS13 where a race has some body parts/hair that shares their skin tone while others don't. (Unless you count 18+ servers such as Skyrat where they give you the option to sync your gentiles colors with your skin tone).

UnityProject/Assets/Scripts/UI/Systems/Lobby/CharacterCustomization.cs
MaxIsJoe: done.

UnityProject/Assets/Scripts/Player/PlayerHealthData.cs
Bod9001: ok, let me get good example, 
let's say if you want your body markings on lizards to be different colour from your skin tone?

MaxIsJoe: Are body markings customizable on /TG/? If yes, then I'll change it so we have rules applied to related body parts only.
If not, then we can start a poll on the server to see if we should keep this solution or not as I do feel like Lizards look amazing with matching colors.

PetMudstone: I just checked and body markings for lizards do not have customizable colors, just different sprites.

MaxIsJoe: Good. Can I have my matching colored lizards now pls 
image

corp-0: > at the moment there is no race or use case

I'd rather have future proofing

Bod9001: yeah, Future proofing is better

MaxIsJoe:
Sorry for being inactive on this, I ran into a lot of troubles fixing another issue at the request of @NoooneyDude and I've been busy with Ramdan and studying.
I'll be back at this in a day or two

UnityProject/Assets/Scripts/Player/PlayerHealthData.cs
MaxIsJoe: Done and I even got my matching colored lizards.

MaxIsJoe:
fuck you git

MaxIsJoe:
fuck this, I'm making a new PR.

PR #6368, opened 2021-04-15
Reagents garbage collection fix part one

Aranclanos:
### Purpose
ReagentMix will no longer be an IEnumerable
SerializableDictionary dictionary will now be public as the first stepping stone to removing its IDictionary and other IEnumerable interfaces.
Reagents will now use SerializableDictionary's dictionary directly.
All these changes have the goal to reduce the usage of IEnumerables to remove the leaks from reagent code that the new health changes brought to surface.
Fixes all garbage generated by metabolismreaction Apply()
Disables reagent tests for the moment.

### Notes:
this is the first step into fixing all GC that reagents create, more prs in the future


Bod9001:
The metabolismreaction is all so in the reaction code for naomall reactions. Code looks good

PR #6373, opened 2021-04-16
Radiation collectors use localposition instead of worldposition

### LINKED ISSUE #6350: PogStation tile bug
### Environment
- Game Version: 21041100
- OS Version: MacOSX

There is a tile on PogStation in the energy ball/singularity engineering area that has a tile on it, but you can't wrench down a rad collector down on it.

Screen Shot 2021-04-10 at 7 26 04 PM

### Steps to reproduce

1. Go to the specific energy ball/singularity area in engineering on PogStation
2. Try to wrench down a rad collector on the specified tile

Semuca:
### Purpose:
Fixes #6350. 

### Explanation:
On Pogstation the Worldposition is off from the actual station by a small amount. When wrenching the radiation collectors, they would test the tile one up and to the right instead of the tile that they were actually on.

### Notes:
Seeing as both lockers and radiation collectors use this interaction, maybe we should consider just making a standalone script that makes pushable objects locked when a certain item is used.

### Changelog:
CL: Fixed radiation collectors not being wrenched properly


corp-0:
There is the wrenchable component, can you try converting this interaction to use that component instead?

Semuca:
Sure, I'll get on that

Semuca:
I'm going to close this for now, and open a new fix later.

PR #6387, opened 2021-04-17
Adds in some improvements and fixes to emotes

MaxIsJoe:
### Purpose

Part of #6276 and should two issues in fix an issue in #6195

Adds in an improvement I've suggested me and gillies a while ago to make crits check available in the base EmoteSO script + added in a check for some emotes to avoid them from doing backflips while crawling + fixes an issue where the player becomes upside down while surrendering and crawling at the same time.

### Notes:

Requires testing by other people on more than one client, pls.

### Changelog:

CL: Fixed an issue where players become upside down when surrendering.
CL: You will no longer be able to do some emotes like backflips while crawling.
CL: Critical condition texts are no longer exclusive to gendered emotes, all emotes use them now.
CL: ActionText will no longer get rotated.

### DIFF
--- a/UnityProject/Assets/Scripts/ScriptableObjects/RP/EmoteSO.cs
+++ b/UnityProject/Assets/Scripts/ScriptableObjects/RP/EmoteSO.cs
@@ -60,7 +60,7 @@ namespace ScriptableObjects.RP
 
 		public virtual void Do(GameObject player)
 		{
-			if(allowEmoteWhileCrawling == false && CheckIfPlayerIsCrawling(player))
+			if(allowEmoteWhileCrawling == false && CheckIfPlayerIsCrawling(player) == true)
 			{
 				FailText(player, 0);
 				return;
--- a/UnityProject/Assets/Scripts/ScriptableObjects/RP/EmoteSO.cs
+++ b/UnityProject/Assets/Scripts/ScriptableObjects/RP/EmoteSO.cs
@@ -58,21 +58,27 @@ namespace ScriptableObjects.RP
 		[SerializeField]
 		private Vector2 pitchRange = new Vector2(0.7f, 1f);
 
+		protected enum FailType
+		{
+			Normal,
+			Critical
+		}
+
 		public virtual void Do(GameObject player)
 		{
-			if(allowEmoteWhileCrawling == false && CheckIfPlayerIsCrawling(player) == true)
-			{
-				FailText(player, 0);
-				return;
-			}
 			if (allowEmoteWhileInCrit == false && CheckPlayerCritState(player) == true)
 			{
-				FailText(player, 1);
+				FailText(player, FailType.Critical);
+				return;
+			}
+			if (allowEmoteWhileCrawling == false && CheckIfPlayerIsCrawling(player) == true)
+			{
+				FailText(player, FailType.Normal);
 				return;
 			}
 			if (requiresHands && CheckHandState(player) == false)
 			{
-				FailText(player, 0);
+				FailText(player, FailType.Normal);
 				return;
 			}
 			else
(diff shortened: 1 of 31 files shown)


UnityProject/Assets/Scripts/ScriptableObjects/RP/EmoteSO.cs
```diff
@@ -108,5 +148,28 @@ protected bool CheckHandState(GameObject player)
 		{
 			return !player.GetComponent<PlayerScript>().playerMove.IsCuffed && Validations.HasBothHands(player);
 		}
+
+		/// <summary>
+		/// Checks if the player is in critical condition or not
+		/// </summary>
+		/// <param name="player">The player that you want to check their health.</param>
+		/// <returns>True if crit, false otherwise.</returns>
+		protected bool CheckPlayerCritState(GameObject player)
```
Bod9001, line 161: Would be nice if that there was a sanded function of this, but this is good for now

UnityProject/Assets/Scripts/ScriptableObjects/RP/EmoteSO.cs
```diff
@@ -60,6 +82,24 @@ public virtual void Do(GameObject player)
 			}
 		}
 
+		/// <summary>
+		/// Use this instead of rewriting Chat.AddActionMsgToChat() when adding text to a failed conditon.
+		/// </summary>
+		/// <param name="player">The orignator</param>
+		/// <param name="type">0: failText, 1: critViewText</param>
+		protected void FailText(GameObject player, int type)
```
corp-0: that int type is a little obscure. You can do a private enum to make it more human readable.

MaxIsJoe: Done 👍

MaxIsJoe:
I've tested everything and made sure everything works across all clients.
This PR is done and has been tested, we can merge now 👍 

P,S: I've enabled the surrender Emote since all of it's issues has been fixed + you can now stand up after laying down.

MaxIsJoe:
Bonus Fix!

Bonus Fix!

Come get your bonus fixes! Fixed another issue where ActionTexts get rotated. They no longer will.

Bod9001:
I'll wait for Giles

corp-0:
looks good, don't know if the text rotation thing works or if it is a good solution. It is setting a *right* rotation every frame, maybe the actual solution would be searching what is rotating it and make sure that doesn't happen.

MaxIsJoe:
> looks good, don't know if the text rotation thing works or if it is a good solution. It is setting a *right* rotation every frame, maybe the actual solution would be searching what is rotating it and make sure that doesn't happen.

It is the best solution that doesn't introduce more issues from my testing and it always stays at 0 no matter what.

Changing the transform's parent or playing around with freezing the rotation of the action text through other means causes the text to disappear completely, get stuck in its original place or gitter/delaying while moving.

corp-0:
> Changing the transform's parent or playing around with freezing the rotation of the action text through other means causes the text to disappear completely, get stuck in its original place or gitter/delaying while moving.

Well I first would wonder why is it rotating. Without looking at any code, I could throw a guess about how the crawling effect is applying rotation to the whole player game object instead of the sprite and since the text is following a transform inside the player game object, it rotates. If that's the case, then the real solution is to make crawling only rotate the sprite game object.

MaxIsJoe:
> > Changing the transform's parent or playing around with freezing the rotation of the action text through other means causes the text to disappear completely, get stuck in its original place or gitter/delaying while moving.
> 
> Well I first would wonder why is it rotating. Without looking at any code, I could throw a guess about how the crawling effect is applying rotation to the whole player game object instead of the sprite and since the text is following a transform inside the player game object, it rotates. If that's the case, then the real solution is to make crawling only rotate the sprite game object.

I did investigate that and apparently not, the game isn't rotating the player's object it's rotating it's sprites.
If you're going to ask "why not just change the parent then?" refer back to what I said about how changing the parent causes issues + whoever coded this was probably correct about leaving it this way because if we offset the sprites for any reason the the actionText will follow it, if we change this you'll have text floating far away from your character in-game when their sprites are offset from 0,0

Either way consider this future proofing so that the action text never gets rotated again for any reason when we decide to actually rotate the player gameObject himself.

MaxIsJoe:
This is finally it. time to merge.

PR #6389, opened 2021-04-18
Fixes editor only spawn api related runtimes 

Aranclanos:
### Purpose
Fixes spawn behavior needing the networkmanager prefab to have the list of prefabs up to date by manually updating it, avoiding important runtimes on load.

### Notes:
This only affects the editor, the list is updated when a build is made and a build itself cannot run that code due to an `#if UNITY_EDITOR` check

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Lifecycle/Spawn.cs
+++ b/UnityProject/Assets/Scripts/Core/Lifecycle/Spawn.cs
@@ -54,6 +54,7 @@ public static class Spawn
 	{
 		if (objectPool == null)
 		{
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Core/Lifecycle/Spawn.cs
```diff
@@ -54,6 +54,7 @@ private static void EnsureInit()
 	{
 		if (objectPool == null)
 		{
+			CustomNetworkManager.Instance.SetSpawnableList();
```
Bod9001, line 57: This takes like 10 seconds on start up

PR #6394, opened 2021-04-19
Pickable Mention sound

ThatDan123:
-Adds option to pick from 4 different mention sounds


corp-0:
adding a DNM because of the freeze. It should shortly be lifted, but rules are rules.

Bod9001:
Due to merge conflicts

PR #6405, opened 2021-04-22
Fixes half of bloodpump GC

Aranclanos:
### Purpose
Modifies the TransferTo() function of ReagentMix to avoid generating garbage from creating a new reagentmix and utilizing some ToArray() calls to modify and loop through dictionaries

image
some rough stats on the difference, while one has an extra human on the frame, the change is visible


ThatDan123:
Codacy complaining about float check

NoooneyDude:
Can use `Mathf.Approximately()` if appropriate.

Bod9001:
I think is good enough it's only there to stop a divide by zero error

PR #6406, opened 2021-04-23
Quick fix to character customization menu

Semuca:
### Purpose
Character menu dropdowns broke with new raycasting, so I fixed it


MaxIsJoe:
Thank you, this has been bothering me for a while.

PR #6407, opened 2021-04-23
Memory Profiler

ThatDan123:
CL: Adds the unity memory profiler package

Admins can trigger a memory profile for the server, the files are large so do not spam this.

Clients can also do memory profile but only if they have a development version like the server does.


Bod9001:
looks good

PR #6409, opened 2021-04-24
Fixed UI Icons remaining after death

Semuca:
### Purpose
When you died, action icons like toggling the PDA light would remain as you were a ghost, and even remain when you respawned, so I modified the scriptable object so they despawn on death. Only certain items needed this change, because other items like bags have icons that only appear when the item is in your hands and you automatically drop what you're holding when you die when you die (disabling the icon).

### Changelog:
CL: Fixes action icons not disappearing after death

### DIFF
--- a/UnityProject/Assets/ScriptableObjects/UIActionsGraphics/RestraintOverlay.asset
+++ b/UnityProject/Assets/ScriptableObjects/UIActionsGraphics/RestraintOverlay.asset
@@ -20,7 +20,7 @@ MonoBehaviour:
   description: You're handcuffed and can't act. If anyone drags you, you won't be
     able to move. Click the alert to free yourself.
   Sprites:
-  - {fileID: 0}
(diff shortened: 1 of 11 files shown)


UnityProject/Assets/ScriptableObjects/UIActionsGraphics/RestraintOverlay.asset
```diff
@@ -12,13 +12,20 @@ MonoBehaviour:
...
+  actionType: 0
+  activeSpriteIndex: 1
   callOnClient: 1
   callOnServer: 0
   actionName: Handcuffed
   description: You're handcuffed and can't act. If anyone drags you, you won't be
     able to move. Click the alert to free yourself.
   Sprites:
-  - {fileID: 11400000, guid: 0c3c1db3702490542a12d241056cb8ed, type: 2}
+  - {fileID: 0}
```
corp-0: is this removing a sprite that was set previously?

Semuca:
I messed up my branch while I was rebasing, I'll close this and resubmit.

PR #6412, opened 2021-04-24
Turns Off Some Damage Messages

ThatDan123:
CL: Disables damage message on Singularity, field generator, and tesla ball

-Moves some prefabs into their correct folder


corp-0:
disable how? I don't see any code changes

PR #6420, opened 2021-04-26
Players should no longer be able to join the round before finishing loading their game.

### LINKED ISSUE #5512: Clients are allowed to join the round before they finished loading
Players can open their game and latejoin a round before most maps loaded, causing some annyoing heavy lag until the client finishes loading everything

### Steps to reproduce
1. Start a server and a round
2. Open new client
3. Instantly join the round before most maps load
4. Experience lag

MaxIsJoe:
### Purpose
Fixes a lot of errors and possibly some other bugs that appear when players rejoin the game.
It freezes the spawn logic completely until everything is loaded. (or more accurately, it waits until all tiles are loaded because they take the longest to load and players CAN get into the game before the tiles where they are haven't even loaded yet)

### Notes:
Testing required in case we're freezing this in the wrong function.
Any other new things to wait for would also be great if someone pointed them out.

closes #5512 

### Changelog:

CL: Players will be unable to join the game until they have finished loading their game.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Lifecycle/PlayerSpawn.cs
+++ b/UnityProject/Assets/Scripts/Core/Lifecycle/PlayerSpawn.cs
@@ -235,6 +235,22 @@ public static class PlayerSpawn
 	/// <param name="body">to transfer into</param>
 	public static void ServerRejoinPlayer(JoinedViewer viewer, GameObject body)
 	{
+		RpcAwaitPlayerLoading(viewer, body);
+	}
+
+	[TargetRpc]
+	public static void RpcAwaitPlayerLoading(JoinedViewer viewer, GameObject body)
+	{
+		while(TileManager.TilesToLoad > TileManager.TilesLoaded)
+		{
+			Logger.Log($"Loading tiles.. {TileManager.TilesToLoad} / {TileManager.TilesLoaded} loaded.");
+		}
+		CmdPlayerFinishedLoading(viewer, body);
+	}
+
+	[Command]
+	public static void CmdPlayerFinishedLoading(JoinedViewer viewer, GameObject body)
+	{
 		var ps = body.GetComponent<PlayerScript>();
 		var mind = ps.mind;
 		var occupation = mind.occupation;


ThatDan123 (changes requested):
This logic isnt going to work as the PlayerSpawn class is static and not a network behaviour, so mirror doesnt know the RPC or command exists.

So all the code is just being run on server as it was before.

A pause before loading might help fix some errors, but that would need proper testing and a more suitable place to put it.

MaxIsJoe:
> This logic isnt going to work as the PlayerSpawn class is static and not a network behaviour, so mirror doesnt know the RPC or command exists.
> 
> So all the code is just being run on server as it was before.
> 
> A pause before loading might help fix some errors, but that would need proper testing and a more suitable place to put it.

I thought I was pausing it already?
Where exactly should we pause this? In what script?

ThatDan123:
> I thought I was pausing it already?

This PR is not doing anything, the While is being run on the server as you cant do networking in this class

> Where exactly should we pause this? In what script?

The problem is I'm not sure what errors you are trying to fix, there are no tile errors that I'm aware of which are being caused by the initialisation of the tile manager (which would have already initialised before joining the server as it is also in the lobby managers). So I cant really tell you where to put the pause or if its even needed and that there's a separate issue instead.

The only bug thing related to tiles, I can think of, is when rejoining you sometimes spawn in what appears to be space, that's due to the client only waiting for the station scene to load before switching views. Where that code is maybe would be a good place to stick a small pause, but would need testing.

MaxIsJoe:
> This PR is not doing anything, the While is being run on the server as you cant do networking in this class

How is it not? It's checking something from the client and once it's done it tells the server to let them join.
And "you can't do networking in this class" is an impossible thing because this entire class is based around networking.

> The problem is I'm not sure what errors you are trying to fix, there are no tile errors that I'm aware of which are being caused by the initialisation of the tile manager (which would have already initialised before joining the server as it is also in the lobby managers). So I cant really tell you where to put the pause or if its even needed and that there's a separate issue instead.
> 
> The only bug thing related to tiles, I can think of, is when rejoining you sometimes spawn in what appears to be space, that's due to the client only waiting for the station scene to load before switching views. Where that code is maybe would be a good place to stick a small pause, but would need testing.

The game takes a longer time spawning in tiles which causes a lot of errors and bugs when the game starts so it needs to be stopped until everything is loaded. 
Also like I said in the notes, if there is any other stuff that we need to check for then I'm open for suggestions. But I don't get your comment about it being impossible to do networking in that at all.

ThatDan123:
> How is it not? It's checking something from the client and once it's done it tells the server to let them join.
> And "you can't do networking in this class" is an impossible thing because this entire class is based around networking.

The playerspawn class contains no networking, it is all server side methods called by the server. 

It is not a network behaviour so mirror doesnt know it exists, if it did the commands wouldnt work anyway as they can only be called on the local player object as that only has authority. The RPC doesnt have the target connection parameter so would be erroring as well, but it doesnt as mirror doesnt know to check it as you cant do networking in this type of class.

> The game takes a longer time spawning in tiles which causes a lot of errors and bugs when the game starts so it needs to be stopped until everything is loaded.
> Also like I said in the notes, if there is any other stuff that we need to check for then I'm open for suggestions. But I don't get your comment about it being impossible to do networking in that at all.

If "The game takes a longer time spawning in tiles which causes a lot of errors and bugs when the game starts so it needs to be stopped until everything is loaded." is the case can you do a stopwatch on the tile initialisation and record the time as it really shouldnt take long.

MaxIsJoe:
> If "The game takes a longer time spawning in tiles which causes a lot of errors and bugs when the game starts so it needs to be stopped until everything is loaded." is the case can you do a stopwatch on the tile initialisation and record the time as it really shouldnt take long.

No need for a stopwatch, just open staging inside the editor and watch how slow things spawn inside the scene tree

MaxIsJoe:
Made this a draft until I figure out how we can properly fix this

MaxIsJoe:
Just realized that this is more complicated and confusing than what I'm capable off.
Doesn't help that some of the code behind it is confusing and all over the place. Closing this and I'm going to move on to other stuff.

PR #6425, opened 2021-04-27
Multitool text improvement, NetToggle workaround

NoooneyDude:
- Improves the interaction text of the multitool.
- Adds an optional workaround for NetToggles where the toggles are assigned a group triggering listeners multiple times.
    - Updates the pipe dispenser to use this workaround.

CL: Better multitool text.

#### Test Guide

Precondition | Expectation | Result
--- | --- | ---
Use a multitool on one electrical wire, multiple wires, no wires | The multitool should output the appropriate results into the chat with the new strings, with the text being understandable |
Use a multitool to connect slave devices to a master device and then clear the buffer | The multitool should post the new messages for each action to the chat |
Use the pipe dispenser and operate the category, colour and category section toggles | The toggles should change state and the appropriate page should be activated | 

> Note: the `Test Guide` is an optional process we'd (or I'd) like to try, where the PR creator creates a bunch of relevant test cases so that others can verify the quality of the PR without necessarily having coding knowledge and without having to study the changes. It is the PRer's responsibility to ensure good test coverage, though everyone is encouraged to add more test cases in a followup comment.
>
> The test coverage doesn't need to be exhaustive; just cover the major, important things that may impact the player. The more complex the PR, the more necessary good testing is, and so this simple peer testing process will hopefully improve the quality of our game without too much overhead. This PR is pretty simple and could possibly forego testing, but serves as a good example.
>
> The tester can copy the table into a followup comment. Then for each test case, in the `Result` column, the tester will add a ✅, :x:, or ⚠️, where the first indicates a pass, the second a fail, and the warning could mean a potential issue or some other indeterminate response. Extra info can be written here, too, such as failure reproduction steps, etc.
>
> When performing a retest, you can just edit your existing table comment.
>
> It's worth noting this is not the same as a _code review_, which is a separate process. Hence the two labels `Status: Awaiting Testing` and `Status: Awaiting Review`.

### DIFF
--- a/UnityProject/Assets/Prefabs/UI/Tabs/Resources/PipeDispenser/ColorToggle.prefab
+++ b/UnityProject/Assets/Prefabs/UI/Tabs/Resources/PipeDispenser/ColorToggle.prefab
@@ -12,7 +12,6 @@ GameObject:
   - component: {fileID: 8727998958661407103}
   - component: {fileID: 5649171872587022298}
   - component: {fileID: 1540758683622087102}
-  - component: {fileID: 4570004714174650456}
   - component: {fileID: 5594961893074090065}
   m_Layer: 5
   m_Name: ColorToggle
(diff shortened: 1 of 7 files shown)


corp-0:
Precondition | Expectation | Result
--- | --- | ---
Use a multitool on one electrical wire, multiple wires, no wires | The multitool should output the appropriate results into the chat with the new strings, with the text being understandable | ✅
Use a multitool to connect slave devices to a master device and then clear the buffer | The multitool should post the new messages for each action to the chat |✅
Use the pipe dispenser and operate the category, colour and category section toggles | The toggles should change state and the appropriate page should be activated | ✅

All seems to work as intended and the text is a lot better now, thanks. I'm merging :B

PR #6431, opened 2021-04-28
Fixes action icons not despawning after death

Semuca:
Purpose
When you died, action icons like toggling the PDA light would remain as you were a ghost, and even remain when you respawned, so I modified the scriptable object so they despawn on death. Only certain items needed this change, because other items like bags have icons that only appear when the item is in your hands and you automatically drop what you're holding when you die when you die (disabling the icon).

Changelog:
CL: Fixed action icons not disappearing after death


corp-0:
can you briefly explain how the fix works?

NoooneyDude:
Listens to the networked event local player died, disabling the action button when triggered.

@Semuca what happens if the player's corpse is revived?

Semuca:
Yup, from my testing it works when a player is revived

NoooneyDude:
> Yup, from my testing it works when a player is revived

Awesome, thank you.

PerfectTangent:
the buckled icon failed to disappear when I aghosted, when I killed myself, and when I respawned. Tested on staging.

Issue #6427, opened 2021-04-27
Mobs turn invisible on death if they're not facing south.


PetMudstone:
### Environment
- Game Version: 21042400
- OS Version: Windows 10

### Detailed description of bug

The vast majority of mobs have their sprites disappear when they die while not facing south. The mobs themselves do not despawn as can be seen by right clicking on the tile that they occupy. In fact, it is possible to pull dead mobs using right click's pull option even if they are invisible. Notably, pulling invisible mobs south will make their sprite appear again and it won't turn invisible again if you pull it in other directions.

Hypothesis for why this happens: Sprite data SOs can have sprites for multiple directions. Most mobs have directional sprites while alive. However, the vast majority of mobs have dead sprites that aren't directional and only have a southern sprite. So if a mob is facing west, east, or north and dies the game will try to use a directional sprite that doesn't actually exist. 
The reason why dead mobs don't turn invisible while facing south is because south is the default direction for sprites, so therefore dead sprites without additional directional sprites will only have southern sprites.

### Steps to reproduce
1. Kill a mob somehow, make sure it isn't facing south while it dies and doesn't actually despawn on death.
2. Watch it turn invisible. Right click on the spot where it died to confirm that the animal is still present.
3. Stand adjacent to the spot where the mob died and right click on it. Pull the dead mob.
4. Move south while pulling the mob, watch as it becomes visible again.

NoooneyDude:
~After #6437 is merged, it might be more appropriate to form a list of which mobs need configuring to the new mob sprite state system on `MobSprite`, along with any other remaining system issues.~

NoooneyDude:
~Related to #5325.~

PR #6437, opened 2021-04-29
Refactor AI initialisation, more flexible dead/prone states on MobSprite

NoooneyDude:
- Refactors mob AI initialisation to make it more clear when the mob is spawned and when the AI behaviour begins.
    - Fixes various mobs with broken AI.
- Fixes the name and description texts for plasma crate being swapped.
- Enables more sprite configuration flexibility on `MobSprite`. This is to address #6427.
    - Refactors knocked down rotation to `MobSprite`.
    - Sprite state for death or being knocked down can now be represented by either a sprite index, rotation value or no change.

> Concerns: may reintroduce the issue described in #6383. Not tested.
> May be a few other sprite state bugs remaining. Inconclusive, will wait on more usage from #6427.
> `NaughtyAttributes`' `BoxGroup` and `ShowIf` don't appear to be working on `MobSprite`.
> Only `God` mob prefab has been configured to reflect the changes.

### DIFF
--- a/UnityProject/Assets/Prefabs/Mobs/NPC/Resources/Friendly/NPC_GOD_Friendly.prefab
+++ b/UnityProject/Assets/Prefabs/Mobs/NPC/Resources/Friendly/NPC_GOD_Friendly.prefab
@@ -93,8 +93,7 @@ MonoBehaviour:
   syncMode: 0
   syncInterval: 0.1
   maxHealth: 100
-  canBreathAnywhere: 0
-  cloningDamage: 0
+  canBreatheAnywhere: 1
   bloodSystem: {fileID: 0}
   brainSystem: {fileID: 0}
   respiratorySystem: {fileID: 0}
@@ -104,9 +103,6 @@ MonoBehaviour:
   - {fileID: 186469482417783052}
   allowKnifeHarvest: 0
   butcherResults: []
-  RTT: 0
-  RadiationStacks: 0
-  deadState: 0
 --- !u!114 &114555785714625820
 MonoBehaviour:
   m_ObjectHideFlags: 0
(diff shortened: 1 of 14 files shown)


corp-0:
NaughtyAttributes doesn't work in components that use a custom editor. I don't know if that's the case for Mobsprite but if it is, we can make it work with some tweaks on the custom editor.

> Most of the attributes are implemented using Unity's CustomPropertyDrawer, so they will work in your custom editors. The attributes that won't work in your custom editors are the meta attributes and some drawer attributes such as ReorderableList, Button, ShowNonSerializedField and ShowNativeProperty.
If you want all of the attributes to work in your custom editors, however, you must inherit from NaughtyInspector and use the NaughtyEditorGUI.PropertyField_Layout function instead of EditorGUILayout.PropertyField.

From the [Readme ](https://github.com/dbrizov/NaughtyAttributes) (it seems like they made it even more compatible in the latest version, previously it wouldn't work at all without the described tweaks.)

NoooneyDude:
> NaughtyAttributes doesn't work in components that use a custom editor. I don't know if that's the case for Mobsprite but if it is, we can make it work with some tweaks on the custom editor.
> 
> > Most of the attributes are implemented using Unity's CustomPropertyDrawer, so they will work in your custom editors. The attributes that won't work in your custom editors are the meta attributes and some drawer attributes such as ReorderableList, Button, ShowNonSerializedField and ShowNativeProperty.
> > If you want all of the attributes to work in your custom editors, however, you must inherit from NaughtyInspector and use the NaughtyEditorGUI.PropertyField_Layout function instead of EditorGUILayout.PropertyField.
> 
> From the [Readme ](https://github.com/dbrizov/NaughtyAttributes) (it seems like they made it even more compatible in the latest version, previously it wouldn't work at all without the described tweaks.)

That's interesting. Thank you. I don't think `MobSprite` itself has associated editor scripts, but maybe `NetworkBehaviour`, which it inherits from, does?

PR #6448, opened 2021-05-01
Gun and PKA Fix

### LINKED ISSUE #6428: Guns can no longer be fired as a client.
### Environment
- Game Version: 21042400
- OS Version: Windows 10

For some reason, guns can no longer be fired as a client. Hosts are still able to fire them however and don't receive the errors in the console below. 
Firing guns as a client creates NREs in the console. Notably, you get different kinds of NREs with the same gun depending on whether you try to shoot using harm intent or not. 
Meanwhile, trying to fire an unloaded gun (either without a power cell or magazine) creates a yellow warning in the console. This warning is not affected by what intent you use when  trying to fire.

Screenshot of console after trying to fire an Imperator. Yellow warnings come from trying to fire it while it has no power cell, Red warnings come from trying to fire it with a power cell. Middle red warning comes from trying to fire it on any intent other than harm intent while bottom red warning comes from trying to fire it on harm intent.
image

### Steps to reproduce
1. Find a gun while connected to a server.
2. Click with it on something in order to attempt firing it. 
3. Fail. Check console for errors that come after clicking.

ThatDan123:
CL: Fixes guns not validating correctly on client due to reference not set
CL: Fixes PKA not working due to projectile not being set

Fixes #6428

~Doing build to test atm.~

Tested and guns are working now

### DIFF
--- a/UnityProject/Assets/Prefabs/Items/Mining/ProtoKineticAccelerator.prefab
+++ b/UnityProject/Assets/Prefabs/Items/Mining/ProtoKineticAccelerator.prefab
@@ -25,8 +25,6 @@ MonoBehaviour:
   SpawnsCasing: 0
   MagInternal: 1
   SmartGun: 0
-  CurrentRecoilVariance: 0
-  FireCountDown: 0
   FiringSoundA:
     SetLoadSetting: 0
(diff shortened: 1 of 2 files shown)


corp-0:
Huh,   I was sure I suggested exactly that in Autumn's pr.

PR #6454, opened 2021-05-03
Enabling Physics2D for ambience colliders

### LINKED ISSUE #6132: Spaceship ambience is missing.
An ambience that used to play when you were inside a spaceship or the station (or basically a place that has atmos) is missing, this ambience really helps fill the silent spaces, and keeps you immersed into the game.

### Steps to reproduce
1.  Boot up game
2.  No spaceship ambience

pyrosshade:
* Physics2D simulation mode changed from script to update
* Physics2D y axis gravity changed to 0

### Purpose
Enables collision simulation, in this case so that AmbientSoundArea colliders can be checked for/from to enable/disable station ambience.

### Notes:
Performance before and after the changes were profiled and performance of rotating/piloting the main station matrix was tested. Performance difference was found to be negligible in both cases, most likely since anything that used to rely on Physics2Dsimulation was changed to a different system.

### DIFF
--- a/UnityProject/ProjectSettings/Physics2DSettings.asset
+++ b/UnityProject/ProjectSettings/Physics2DSettings.asset
@@ -3,8 +3,8 @@
 --- !u!19 &1
 Physics2DSettings:
   m_ObjectHideFlags: 0
-  serializedVersion: 4
-  m_Gravity: {x: 0, y: -9.81}
+  serializedVersion: 5
+  m_Gravity: {x: 0, y: 0}
   m_DefaultMaterial: {fileID: 0}
   m_VelocityIterations: 8
   m_PositionIterations: 3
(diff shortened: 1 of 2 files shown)


UnityProject/ProjectSettings/Physics2DSettings.asset
```diff
@@ -3,8 +3,8 @@
 --- !u!19 &1
 Physics2DSettings:
   m_ObjectHideFlags: 0
-  serializedVersion: 4
```
Bod9001, line 6: why the serialisation version change?

pyrosshade: That was something automatic on the part of unityengine, the changes to project settings were made through the project settings window. If need be i can make a commit changing back the serializedVersion.

pyrosshade:
Presumably fixes #6132

Bod9001:
ok looks good

PR #6463, opened 2021-05-04
Fixes the singulo not killing players

### LINKED ISSUE #6462: Players are invincible, even from Lord Singulo
### Environment
- Game Version: Staging 21050308
- OS Version: Irrelevant

This was encountered on the staging server. Some of the staging testers noticed that they were unable to kill one another. Even being thrown into singularity doesn't do anything.

Possibly related to the fire damage issue described in #6398.

image

### Steps to reproduce

1. Join a game, preferably with a client connected as well.
2. Try to kill each other.
3. Throw yourself into Lord Singulo.
4. Observe that you cannot die no matter what.

Aranclanos:
### Purpose
Fixes #6462
Fixes gibbing not ghosting out players.

### DIFF
--- a/UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
@@ -64,11 +64,6 @@ namespace HealthV2
 					{
 						OnConsciousStateChangeServer.Invoke(oldState, value);
 					}
-
-					if (value == ConsciousState.DEAD)
-					{
-						Death();
-					}
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


Korywon:
This does not completely resolve #6462, only progress towards it. There is still the issue of players not being able to receive damage from anything. This should only count as progress towards the issue.

Aranclanos:
Damage still works, if players noticed that they weren't taking damage from space after stepping inside the singulo, is that their characters were actually in hiddenpos serverside.
However, I also just briefly tested shooting and stepping on space as a client and server, damage works just fine both ways.

Korywon:
Ah okay, sounds good. Just making sure that we're covering all bases. That would explain why there was no damage being done.

waffielz:
f

PR #6465, opened 2021-05-05
Hotfixes half of the memory leaks inbetween rounds

Aranclanos:
### Purpose
Hotfixes MetaDataNodes persisting through rounds, removing half of the memory leaks.
The main issue persists and this hotfix arguably can be removed in the future after the game's matrixes and its contents properly gc.

### Notes
This does not fix the original issue, its only a patch


corp-0:
leave a TODO comment explaining where is the issue and how this should be removed in the future

PR #6466, opened 2021-05-05
Fix cable cutting window, cable duplication, and cable construction costs

### LINKED ISSUE #6458: Cutting cables causes a ArgumentException and cable cutting window is broken
### Environment
- Game Version: 0.4.1
- OS Version: Pop OS 20.10 (Linux)

Attempting to cut cables will cause an ArgumentException. The cables will rarely cut depending on the map (strange, I know). It appears that it's being caused by a failure to check for null when iterating through tiles. Also there is no reference to the cable cutting window prefab.

This issue was probably caused by an attempt to address #2373.

Full stack trace right below:

```
ArgumentException: The Object you want to instantiate is null.
UnityEngine.Object.Instantiate[T] (T original) (at /home/bokken/buildslave/unity/build/Runtime/Export/Scripting/UnityEngineObject.bindings.cs:277)
LoadCableCuttingWindow.OpenCableCuttingWindow () (at Assets/Scripts/UI/Objects/Engineering/Cable/LoadCableCuttingWindow.cs:106)
InteractableTiles.EnableCableCuttingWindow () (at Assets/Scripts/Tilemaps/Behaviours/InteractableTiles.cs:239)
InteractableTiles.Interact (PositionalHandApply interaction) (at Assets/Scripts/Tilemaps/Behaviours/InteractableTiles.cs:207)
InteractionUtils.CheckInteractInternal[T] (IBaseInteractable`1[T] interactable, T interaction, NetworkSide side, System.Boolean& wasClientInteractable) (at Assets/Scripts/Core/Input System/InteractionV2/InteractionUtils.cs:112)
InteractionUtils.ClientCheckAndTrigger[T] (IBaseInteractable`1[T] interactable, T interaction) (at Assets/Scripts/Core/Input System/InteractionV2/InteractionUtils.cs:46)
MouseInputController.CheckHandApply (UnityEngine.GameObject target) (at Assets/Scripts/Core/Input System/MouseInputController.cs:427)
MouseInputController.CheckClick () (at Assets/Scripts/Core/Input System/MouseInputController.cs:356)
MouseInputController.CheckClickInteractions (System.Boolean includeAimApply) (at Assets/Scripts/Core/Input System/MouseInputController.cs:266)
MouseInputController.CheckMouseInput () (at Assets/Scripts/Core/Input System/MouseInputController.cs:181)
MouseInputController.LateUpdate () (at Assets/Scripts/Core/Input System/MouseInputController.cs:108)
```

### Steps to reproduce

1. Host/join a game.
2. Find or place down some cables.
3. Attempt to cut wires.
4. Either you rarely cut the cables successfully or you hit a "The object you're trying to instantiate is null."
5. Also observe that the cable cutting window doesn't pop up.

### LINKED ISSUE #6460: Laying down cables with length of 2 only consumes 1
### Environment
- Game Version: 0.4.1
- OS Version: Pop OS 20.10 (Linux)

Duplication bug. If you lay down a cable with a length of 2, then cut it, you'll notice that it'll only take one to lay and you get two back. This allows you to effectively duplicate cables.

### Steps to reproduce

1. Grab cables.
2. Lay down a cable with at least a length of 2.
3. Cut the cable.
4. Observe that it takes 1 to lay down the cable but you get 2 in return.

### LINKED ISSUE #6469: Cable cutting window does not pop up when tiles has pipes
### Environment

- Game Version: 0.4.1
- OS Version: Pop OS 20.10 (Linux)

Cutting cables on tiles that has pipes will not have the cable cutting window to pop up.

### Steps to reproduce

1. Lay down some cables on a tile with pipes.
2. Lay down some cables on a tile without pipes.
3. Attempt to cut cables on both these tiles.
4. Observe that cutting cables on a tile with pipes will not allow the cable cutting window to pop up.

Korywon:
### Purpose

Fixes a whole lot dealing with the cable cutting window, the amount of cable coils consumed, and duplication issue.

Fixes #6458.
FIxes #6460.
Fixes #6469.

### Notes:

`cableCuttingWindowPrefab` was never assigned, which causes the `Instantiate()` call to raise an `ArgumentException`. This in turn breaks the cable cutting window from popping up, and ultimately results in the user being unable to cut electrical cables. Simply enough, the solution is to just load the resource. We ideally only want to do it once (hence the `null` check).

After fixing the cable cutting window, I noticed that cables are immediately cut without the cable cutting window popping up if build cables on a cell with other non-electrical wire tiles, such as pipes. It only checks the first tile. I fixed it so that we search all tiles in cell at a given position before determining whether or not to open the cable cutting window.

I then fixed the cable duplication issue and did some basic arithmetic to make sure that we only consume the difference needed to build over overlapping cables.

There are some other behaviors I want to work on, but since we're in a feature freeze, I'll hold it off for now.

### Changelog:

CL [Fix]: Fixed issue of not being able to cut electrical cables.
CL [Fix]: Fixed cable coil duplication and consumed amount.
CL [Fix]: Fixed cable cutting window not showing when cutting cables on cells with non-electrical tiles.

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Engineering/CableCoil.cs
+++ b/UnityProject/Assets/Scripts/Items/Engineering/CableCoil.cs
@@ -16,6 +16,12 @@ namespace Objects.Electrical
 		public GameObject CablePrefab;
 		public PowerTypeCategory powerTypeCategory;
 
+		/// <summary>
+		/// Message that is send to the client when attempting to build a cable
+		/// that already exists at a given position.
+		/// </summary>
+		private const string MSG_BUILD_CONFLICT = "There is already a cable at that position.";
+
 		public Connection GetDirectionFromFaceDirection(GameObject originator)
 		{
 			var playerScript = originator.GetComponent<PlayerScript>();
@@ -86,18 +92,15 @@ namespace Objects.Electrical
 				Connection wireEndB = interaction.WireEndB;
 
 				// Make sure that the ending connection actually exists.
-				if (wireEndB != Connection.NA)
+				if (IsLegalConnection(wireEndA, wireEndB))
 				{
-					// Check if the connection is legal.
-					if (!IsLegalConnection(wireEndA, wireEndB))
-					{
-						return;
-					}
-
+					// Grab a list of electrical connections from the matrix at
+					// the given location.
 					List<IntrinsicElectronicData> eConnList =
 						interaction.Performer.GetComponentInParent<Matrix>().GetElectricalConnections(localPosInt);
 
-					// TODO: There is a LOT of repeat code here.
+					// Find any cables on the matrix that conflicts with our
+					// proposed connections.
 					// TODO: What's the point of the nested for loop?
 					foreach (IntrinsicElectronicData eConnI in eConnList)
 					{
@@ -105,9 +108,7 @@ namespace Objects.Electrical
 						{
 							if (eConnI.WireEndA == wireEndB || eConnI.WireEndB == wireEndB)
 							{
-								Chat.AddExamineMsgToClient("There is already a cable at that position");
-								eConnList.Clear();
-								ElectricalPool.PooledFPCList.Add(eConnList);
+								MsgAndAddToPool(ref eConnList, MSG_BUILD_CONFLICT);
 								return;
 							}
 
@@ -117,16 +118,12 @@ namespace Objects.Electrical
 								{
 									if (eConnI.WireEndA == eConnJ.WireEndA || eConnI.WireEndB == eConnJ.WireEndA)
 									{
-										Chat.AddExamineMsgToClient("There is already a cable at that position");
-										eConnList.Clear();
-										ElectricalPool.PooledFPCList.Add(eConnList);
+										MsgAndAddToPool(ref eConnList, MSG_BUILD_CONFLICT);
 										return;
 									}
 									else if (eConnI.WireEndA == eConnJ.WireEndB || eConnI.WireEndB == eConnJ.WireEndB)
 									{
-										Chat.AddExamineMsgToClient("There is already a cable at that position");
-										eConnList.Clear();
-										ElectricalPool.PooledFPCList.Add(eConnList);
+										MsgAndAddToPool(ref eConnList, MSG_BUILD_CONFLICT);
 										return;
 									}
 								}
@@ -134,24 +131,48 @@ namespace Objects.Electrical
 						}
 					}
 
-					eConnList.Clear();
-					ElectricalPool.PooledFPCList.Add(eConnList);
-					// Builds the cable using the position and connections.
+					MsgAndAddToPool(ref eConnList, null);
 					BuildCable(localPosInt, wireEndA, wireEndB, interaction);
 				}
 			}
 		}
 
 		/// <summary>
+		/// Adds examine message to the client (if provided), clears the
+		/// electrical connections list, and adds it to the electrical pool.
+		/// </summary>
+		/// <param name="eConnList">List of electrical connections. Will not add to electrical pool if null.</param>
+		/// <param name="addMsg">Message to show as an examine message. Will not show anything if null.</param>
+		private void MsgAndAddToPool(ref List<IntrinsicElectronicData> eConnList, string msg)
+		{
+			if (msg != null)
+			{
+				Chat.AddExamineMsgToClient(msg);
+			}
+
+			if (eConnList != null)
+			{
+				eConnList.Clear();
+				ElectricalPool.PooledFPCList.Add(eConnList);
+			}
+		}
+
+		/// <summary>
 		/// Checks if the connection ends are legal. In this case, high-voltage
-		/// cables cannot cannot diagonally. All other connections are legal.
+		/// cables cannot cannot diagonally and no connections are illegal. All
+		/// other connections are legal.
 		/// </summary>
 		/// <param name="wireEndA">The first connection we're building from.</param>
 		/// <param name="wireEndB">The second connection we're building to.</param>
 		/// <returns>false if attempting to connect high voltage cables
-		/// diagonally, true otherwise.</returns>
+		/// diagonally or no connection, true otherwise.</returns>
 		private bool IsLegalConnection(Connection wireEndA, Connection wireEndB)
 		{
+			if (wireEndA == Connection.NA && wireEndB == Connection.NA)
+			{
+				return false;
+			}
+
 			if (CableType == WiringColor.high)
 			{
 				switch (wireEndA)
--- a/UnityProject/Assets/Scripts/Items/Engineering/CableCoil.cs
+++ b/UnityProject/Assets/Scripts/Items/Engineering/CableCoil.cs
@@ -57,7 +57,7 @@ namespace Objects.Electrical
 
 		public void ServerPerformInteraction(ConnectionApply interaction)
 		{
-			var cableCoil = interaction.HandObject.GetComponent<CableCoil>();
+			CableCoil cableCoil = interaction.HandObject.GetComponent<CableCoil>();
 			if (cableCoil != null)
 			{
 				Vector3Int worldPosInt = Vector3Int.RoundToInt(interaction.WorldPositionTarget);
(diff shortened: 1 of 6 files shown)


Korywon:
Do not merge yet. Going to implement the ability to cut single wires without the window as a QOL improvement.

Aranclanos:
afaik we're trying to stop using Resources.Load and other variants to remove assets from the resources folder

Korywon:
Good point. I'll see what I can do. Some people pointed towards the usage of the UIManager. I'll get that after I get the little QOL done.

corp-0:
I don't get it, why downcast the reference?

Korywon:
Do you mean changing `cableCuttingWindowPrefab` from `CableCuttingWindow`  to `GameObject`? If so, it's because the reference is a prefab and the `CableCuttingWindow` class requires `RectTransform`. I'm assuming that the `RectTransform` is derived. I did try leaving the type as `CableCuttingWindow` but no joy.

corp-0:
uhm, I think the instantiation of the object could be solved adding it to UIManager, like someone suggested you before. I wonder if there is a bettter pattern to follow when doing UI in unity.

Korywon:
Yeah there was mention of the UIManager starting to become a little too clustered.

image

image

Probably good enough to throw into UiManager for now, but going forward, we might want to rethink about the UI. Maybe even generalizing/standardizing UI so that we can dynamically create windows. Maybe even creating a script or some kind of toolkit so we don't have so many custom windows and prefabs. Something like how desktop apps use GTK/GDK or or web apps using component libraries.

Still learning the ins and outs of how UI is handled for both unity and the game, but I'm up for ideas.

Korywon:
Did some more work on this. Cleaned up quite a bit, added a LOT more documentation, and simplified/consolidated the logic into smaller, more manageable methods. Last step is to fix the scrollbar on the cable cutting window and cut if there's only one cable.

I think I'll create another issue to get the cable cutting window moved onto the UI manager. This PR is already big enough as it is.

UnityProject/Assets/Scripts/Items/Engineering/CableCoil.cs
```diff
@@ -55,180 +61,277 @@ public bool WillInteract(ConnectionApply interaction, NetworkSide side)
...
+				eConnList.Clear();
+				ElectricalPool.PooledFPCList.Add(eConnList);
+			}
 		}
 
-		public void FindOverlapsAndCombine(Vector3 position, Connection WireEndA, Connection WireEndB,
-			ConnectionApply interaction)
+		/// <summary>
+		/// Checks if the connection ends are legal. In this case, high-voltage
+		/// cables cannot cannot diagonally and no connections are illegal. All
```
NoooneyDude:
```suggestion
		/// cables cannot connect diagonally and no connections are illegal. All
```

UnityProject/Assets/Scripts/Items/Engineering/CableCoil.cs
```diff
@@ -55,180 +61,277 @@ public bool WillInteract(ConnectionApply interaction, NetworkSide side)
...
+		{
+			ElectricalManager.Instance.electricalSync.StructureChange = true;
+			FindOverlapsAndCombine(position, wireEndA, wireEndB, interaction);
+		}
+
+		/// <summary>
+		/// Finds overlapping wires and combines them to create a new
+		/// connection.
+		/// </summary>
+		/// <param name="position">Position of the cell we're building thecable at.</param>
```
NoooneyDude:
```suggestion
		/// <param name="position">Position of the cell we're building the cable at.</param>
```

UnityProject/Assets/Scripts/Items/Engineering/CableCoil.cs
```diff
@@ -55,180 +61,277 @@ public bool WillInteract(ConnectionApply interaction, NetworkSide side)
...
+		/// <param name="wireEndA">The first connection we're building from.</param>
+		/// <param name="wireEndB">The second connection we're building to.</param>
+		/// <param name="interaction">Object to allow us to interact with the server.</param>
+		public void FindOverlapsAndCombine(Vector3 position, Connection wireEndA, Connection wireEndB,
+			ConnectionApply interaction)
+		{
+			// If the wire ends are overlapping...
+			if (wireEndA == Connection.Overlap | wireEndB == Connection.Overlap)
```
NoooneyDude:
```suggestion
			if (wireEndA == Connection.Overlap || wireEndB == Connection.Overlap)
```

NoooneyDude:
Superb efforts in refactoring and documenting the code! Though, if I may, some of the comments may perhaps be a little too overbearing:

```
// Now, we loop through each electrical connection.
for (int i = 0; i < eConns.Count; i++)
```

Too many comments can make the code harder / take longer to read. Just a heads-up for next time ; )

UnityProject/Assets/Scripts/Tilemaps/Behaviours/InteractableTiles.cs
```diff
@@ -188,33 +188,53 @@ public string Examine(Vector3 pos)
...
 	}
 
+	/// <summary>
+	/// Interaction callback to attempt to interact with tiles using a HandApply
+	/// interaction. We check for both basic tiles and under floor tiles and
+	/// attempt to apply its interaction. In the event that an electric cable
+	/// tile is encountered, we open the cable cutting window.
+	/// </summary>
+	/// <param name="interaction">Position of interaction with a hand</param>
+	/// <returns></returns>
```
NoooneyDude:
```suggestion
```

Korywon:
Thanks! And fair point. That comment was placed there before I was able to untangle everything. There definitely is a fine balance.

Are there any other comments you found to be too obvious or not very useful?

NoooneyDude:
Sure. Here are a few more examples:
```
// First, make sure we actually found electrical connections.
if (eConns != null)
```
```
// If the tile we're looking at is a basic tile...
if (tile is BasicTile basicTile)
```
```
// If the the tile is something that's supposed to be underneath floors...
if (basicTile.LayerType == LayerType.Underfloor)
```
```
// Then we loop through each under floor layer in the matrix until we
// can find an interaction.
foreach (BasicTile underFloorTile in matrix.UnderFloorLayer.GetAllTilesByType<BasicTile>
```

If you think it's ready you can mark as ready. 

I have a suggestion for the tile interaction that you may or may not be interested in tackling, but I'll create a new issue for it.

UnityProject/Assets/Scripts/Items/Engineering/CableCoil.cs
```diff
@@ -55,180 +61,277 @@ public bool WillInteract(ConnectionApply interaction, NetworkSide side)
...
 
+		/// <summary>
+		/// <para>Facilitates the interaction of the cable coil in hand.</para>
+		/// First checks if the connection we're trying to make is legal, then
+		/// checks against all electrical connections on the cell in the matrix
+		/// to see if the connection already exists. If everything is good,
+		/// this proceeds to add the new cable to the electrical pool and
+		/// builds the cable.
+		/// </summary>
+		/// <param name="interaction"></param>
```
NoooneyDude:
```suggestion
```

UnityProject/Assets/Scripts/Items/Engineering/CableCoil.cs
```diff
@@ -55,180 +61,277 @@ public bool WillInteract(ConnectionApply interaction, NetworkSide side)
...
+		/// <para>Destroys the electrical connection (if we're given one) and
+		/// then adds a new electrical connection by adding an electrical
+		/// node. We then calculate and consume a number of cables from the
+		/// hand.</para>
+		/// </summary>
+		/// <param name="position">Position of the electrical connection.</param>
+		/// <param name="eConn">Electrical connection we're replacing.</param>
+		/// <param name="wireEndA">Connection direction of one end.</param>
+		/// <param name="wireEndB">Connection direction of the other end.</param>
+		/// <param name="interaction"></param>
```
NoooneyDude:
```suggestion
```

NoooneyDude:
I can't wait for you to mark as ready - merging now! : )

NoooneyDude (changes requested):

Korywon:
Whoops, I was asleep around the time you suggested changes. Thanks for catching those typos! I was going to implement the single cable cutting feature, but that's alright. I'll open up another PR for it. :smile:

I'll take a look at the issue you popped up,

Korywon:
As a side note, did we change the way the actions work? I don't see my comments on the changelog.

NoooneyDude:
> As a side note, did we change the way the actions work? I don't see my comments on the changelog.

That's odd. I'm not sure myself why the changelog comments don't exist.

Korywon:
Oh, I think the actions were changed to `CL [type]: comment` rather than `CL: comment`. I changed my PR to fit that. Do you mind adding it to the CL if possible?

PR #6468, opened 2021-05-06
Custom Net Transform Packet Optimisations

ThatDan123:
-Sync's the clientState Transform struct as individual sync vars instead of the whole struct as the whole struct was sent even if one one value was changed.

-Adds log to failed admin check

-Fixes cant spawn error for food processor sprite as it had a net identity where it shouldnt

Has been built tested with headless and two joining clients throwing items and pushing stuff around


Bod9001:
looks good, can't test now though

PR #6470, opened 2021-05-06
Fixes welders turning air into space

Aranclanos:
### Purpose
Fixes items inside storages not changing its matrix when the storage itself does.

The welder (or any firesource) was using its original matrix's ReactionManager, reaching for empty spots resulting in space atmos.


corp-0:
awesome. I just have a style concern with that Matrix property set, it has quite lengthy logic. I'd move it out to a method and prettify the property like this:

```cs
public Matrix Matrix 
{
    get => matrix;
    private set => SetMatrix(value);
}
```

corp-0:
Meh, I'm mergin. I can do some janitorial work later myself and try cleaning up style and shit.

waffielz:
lmao thanks for doing this, it was super annoying before this

Issue #4292, opened 2020-05-19
Gamer Nicknames  and Text is Allowed.


Call-Me-You:
First of all, you can put a cap, spacebar, another cap, delete spacebar and get all caps nicknames.
![image](https://user-images.githubusercontent.com/62477979/82324894-12304280-99e3-11ea-9737-ea13694aadbb.png)
Second of all, you can copy-paste nicknames when making an account. //this is not the accounts name, there is 'some' protection. Still looks cool tough.
![image](https://user-images.githubusercontent.com/62477979/82325276-a3071e00-99e3-11ea-95b4-df567f95898e.png)
Third of all, you can do space trumps walls of text by copy pasting text into the chat box.
![image](https://user-images.githubusercontent.com/62477979/82325804-7c95b280-99e4-11ea-8555-0220b69a1bda.png)

//This has been in the game for a LONG time, surprised nobody made a #issue (probably, none with name in their name were)

Call-Me-You:
Oh, and if i remember correctly, you can get blank captains messages by sending some unicode characters.

nBova:
I've been looking at this bug for a little bit today and was able to replicate the first problem you outlined, and put out a fix for it in my [own branch](https://github.com/nBova/unitystation/tree/issue-4292-fix). 

I haven't been able to replicate the other two issues you've outlined though, would someone else be able to take a look at that and double check for me? It's possible they've been fixed since this issue was posted almost a year ago now.

Just a sidenote: My fix doesn't prevent the user from going cap, space, cap, space, delete spaces. It processes the string afterwards by setting the first character in a name to uppercase, and the rest to lowercase. Works with names that have spaces.

nBova:
Nevermind, I was misunderstanding the last two problems. I put out a fix  for them in the branch I mentioned previously. Making a PR for the bug

Kehren:
![image](https://user-images.githubusercontent.com/79835205/118737186-fdcc9300-b84c-11eb-8a13-c2784342763b.png)
![image](https://user-images.githubusercontent.com/79835205/118737203-0624ce00-b84d-11eb-9b52-7df74cd88be5.png)

> #6472 

technically fixed it, tough issue is still very open lol
#6478

MaxIsJoe:
This has been fixed in #6617

PR #6472, opened 2021-05-06
Issue 4292 fix

nBova:
### Purpose
Prevents users from setting all caps names, and prevents users from spamming chat and chat bubbles vertically.
Fixes issue #4292.

### Changelog:
CL: Users can no longer set all caps character names
CL: Newlines and return characters are removed from chat messages so chat isn't spammed vertically

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Chat/ChatUI.cs
+++ b/UnityProject/Assets/Scripts/Core/Chat/ChatUI.cs
@@ -347,7 +347,7 @@ public class ChatUI : MonoBehaviour
 
 	private void PlayerSendChat(string sendMessage)
 	{
-		sendMessage = sendMessage.Replace("\n", " ").Replace("\r", " ");
+		sendMessage = sendMessage.Replace("\n", " ").Replace("\r", " ");  // We don't want users to spam chat vertically
 		if(selectedVoiceLevel == -1)
 			sendMessage = "#" + sendMessage;
 		if(selectedVoiceLevel == 1)
--- a/UnityProject/Assets/Scripts/Core/Chat/ChatUI.cs
+++ b/UnityProject/Assets/Scripts/Core/Chat/ChatUI.cs
@@ -347,6 +347,7 @@ public class ChatUI : MonoBehaviour
 
 	private void PlayerSendChat(string sendMessage)
 	{
+		sendMessage = sendMessage.Replace("\n", " ").Replace("\r", " ");
 		if(selectedVoiceLevel == -1)
 			sendMessage = "#" + sendMessage;
 		if(selectedVoiceLevel == 1)
--- a/UnityProject/Assets/Scripts/UI/Systems/Lobby/CharacterCustomization.cs
+++ b/UnityProject/Assets/Scripts/UI/Systems/Lobby/CharacterCustomization.cs
@@ -1169,7 +1169,7 @@ namespace UI.CharacterCreator
 
 		private string TruncateName(string proposedName)
 		{
-			proposedName = TI.ToTitleCase(proposedName.ToLower());
+			proposedName = textInfo.ToTitleCase(proposedName.ToLower());
 			if (proposedName.Length >= CharacterSettings.MAX_NAME_LENGTH)
 			{
 				return proposedName.Substring(0, CharacterSettings.MAX_NAME_LENGTH);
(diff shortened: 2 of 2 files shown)


UnityProject/Assets/Scripts/Core/Chat/ChatUI.cs
```diff
@@ -347,6 +347,7 @@ public void OnClickSend()
 
 	private void PlayerSendChat(string sendMessage)
 	{
+		sendMessage = sendMessage.Replace("\n", " ").Replace("\r", " ");  // We don't want users to spam chat vertically
```
corp-0, line 350: wait what... People could type "\n" in chat and it was parsed as a new line?

UnityProject/Assets/Scripts/UI/Systems/Lobby/CharacterCustomization.cs
```diff
@@ -1165,6 +1166,7 @@ public void OnManualNameChange()
 
 		private string TruncateName(string proposedName)
 		{
+			proposedName = System.Threading.Thread.CurrentThread.CurrentCulture.TextInfo.ToTitleCase(proposedName.ToLower());
```
corp-0: I'm not sure we want this restriction. What about machine-like names when we have borgs and the like?

UnityProject/Assets/Scripts/Core/Chat/ChatUI.cs
nBova: I could go into notepad and type a word or sentence, hit enter (inserting a return character), and keep doing that then copy and paste all that into the chat window. Then it'd display past its the bounds of the chat bubble and chat box.

UnityProject/Assets/Scripts/UI/Systems/Lobby/CharacterCustomization.cs
corp-0: Also the way you get to ``ToTitleCase`` looks convoluted. You could have just imported the ``System.Globalization`` namespace by adding ``Using System.Globalization`` to the top of the file.

corp-0 (changes requested):

UnityProject/Assets/Scripts/UI/Systems/Lobby/CharacterCustomization.cs
nBova: That's a fair point, I guess I was just going off of the original bug report and the fact that the text box doesn't let you enter multiple capital characters in a row. You had to enter a capital, space, another capital, delete spaces

UnityProject/Assets/Scripts/Core/Chat/ChatUI.cs
corp-0: interesting. Yes, this should do it.

nBova:
Recent push addresses the awkward route to the ToTitleCase method. I know there are some reservations about preventing users from having all caps names, let me know what the outlook on that is. The character name input box doesn't allow multiple capitals in a row so that would need to be addressed.

UnityProject/Assets/Scripts/UI/Systems/Lobby/CharacterCustomization.cs
```diff
@@ -119,6 +120,8 @@ public class CharacterCustomization : MonoBehaviour
 		private CharacterSettings lastSettings;
 		private int currentCharacterIndex = 0;
 
+		private TextInfo TI = new CultureInfo("en-US", false).TextInfo;
```
corp-0:
```suggestion
		private TextInfo textInfo = CultureInfo.CurrentCulture.TextInfo;
```
Does this work?

corp-0:
I think we can use the Title Case for now and think how we want to validate names based on species later.

UnityProject/Assets/Scripts/UI/Systems/Lobby/CharacterCustomization.cs
nBova: Just had to change line 1172 to reflect that as well but looks good!

corp-0: wonderful, I'm merging after the tests are done.

corp-0 (approved):

PR #6473, opened 2021-05-07
Cookable Components on all Foods

### LINKED ISSUE #5480: Add Cookable component to all foods
- All foods should be given the `Cookable` component. This determines how much cooking (in terms of time) is required before the food item transforms into the assigned cooked prefab.
- Cooked foods should also have this component, which will transform into a burned mess if cooked for too long.
    - That means something raw placed into the microwave could turn into a burned mess if the microwave timer is set too high.
- Most foods should probably have the default cook time of ten seconds, for the chef's convenience, though perhaps some exotic foods should have a different time.

evh5516:
Added Cookable component to all the Foods with default timer of 10 and Cooked Prefab of Burned Mess

### Purpose
Fixes #5480
All foods have been given the Cookable component

### Changelog:
CL: All Cook Time values are set to 10
CL: All Cooked components are set to Burned Mess


ThatDan123:
Is there not a common prefab base which can have the component put on instead?

evh5516:
Looks like there is and I missed it. Would all the other prefabs inherit from it directly? Sorry I'm not used to working on prefabs.

ThatDan123:
> Looks like there is and I missed it. Would all the other prefabs inherit from it directly? Sorry I'm not used to working on prefabs.

They should if the other food prefabs have been set up right.

Double check when adding it to the base that it doesnt override the current food prefabs that already have that component and have unique values

evh5516:
Moved the component to just the food prefab base. I didn't see any conflicts or overrides, but it is possible I missed it.

NoooneyDude:
Hmm, the raw food items should still have the reference to the cooked food set, right? Should be raw > cooked > burned mess. As far as I understand the changes, won't this mean a raw steak would turn into a burned mess?

corp-0:
are all foods coockable?

NoooneyDude:
> Hmm, the raw food items should still have the reference to the cooked food set, right? Should be raw > cooked > burned mess. As far as I understand the changes, won't this mean a raw steak would turn into a burned mess?

Yep, raw meats should be assigned the cooked variants in the `Cookable` component (the `Raw Meat` component should be deprecated as of #6481). Same goes for eggs and likely other items. `BurnedMess` can have the component removed. Drinks, condiments, probably don't need the component, either.

corp-0:
are drinks children of _FoodBase?

NoooneyDude:
> are drinks children of _FoodBase?

Yep.

corp-0:
oof. What should we do? Make a different base?

NoooneyDude:
> oof. What should we do? Make a different base?

No, it's fine - I think all drinks inherit from an intermediate base, so just need to remove the component on that.

vlandemart:
Well, I see no problem if we can put anything into a microwave and it will result in a burnt mess (unless it was a pre-defined recipe).
Do we want to merge it, or still keep it abandoned?

NoooneyDude:
IIRC the microwave looks for the new cookable component first, before trying the old way. Merging this'll break existing recipes.

PR #6476, opened 2021-05-09
Overlay fixes

### LINKED ISSUE #6429: Overlays don't match sprite rotation
### Environment
- Game Version: 21042400
- OS Version: Windows 10

When electrocuted or set on fire, the overlay will not match the player's rotation, whether flipping or falling on the ground.

I tested on a build from a little while ago, and the bug didn't show up on that. From this, I think it's rather likely that this bug was caused by #6387.

### Steps to reproduce

1. Spawn as Engi
2. Hack a door without gloves to get electrocuted
3. Set yourself on fire with a plasma flood. Escape the flood (This is easiest on SquareStation) and do flips / drop to the floor

Semuca:
### Purpose
This PR fixes bugs that I noticed while looking at plasma fires, like player overlays, tile overlays, and Unity throwing modified collection errors in TileList.cs.

### Explanation:
Screenshot 2021-05-09 194229
Overlays are now stored under the sprite gameobject in players, so they rotate whenever the player body rotates. This fixes #6429.

Tile overlays like windows smashing were added too many times because the system was not checking if an overlay was already added correctly. I overwrote the .Equals() function so they can compare in a more functional way.

Tile overlays now have a const which determines the maximum overlays allowed per tile, and some weird code has been cleaned up. Previously, the z position of overlays was increased in this order: 1, 2, 4, 7, 11. Now, they'll increase in a normal ascending order (1, 2, 3, etc).

### Changelog:
CL: Fixed fire and electrocution overlay not matching the player's rotation
CL: Fixed tiles having too many overlays

### DIFF
--- a/UnityProject/Assets/Scripts/Player/PlayerSprites.cs
+++ b/UnityProject/Assets/Scripts/Player/PlayerSprites.cs
@@ -91,7 +91,8 @@ public class PlayerSprites : MonoBehaviour
 
 	public bool RootBodyPartsLoaded = false;
 
-	public GameObject OverlaySprites;
+	[SerializeField]
+	private GameObject OverlaySprites;
 
 	protected void Awake()
 	{
--- a/UnityProject/Assets/Scripts/Player/PlayerSprites.cs
+++ b/UnityProject/Assets/Scripts/Player/PlayerSprites.cs
@@ -91,6 +91,8 @@ public class PlayerSprites : MonoBehaviour
 
 	public bool RootBodyPartsLoaded = false;
 
+	public GameObject OverlaySprites;
+
 	protected void Awake()
 	{
 		directional = GetComponent<Directional>();
@@ -127,7 +129,7 @@ public class PlayerSprites : MonoBehaviour
 		if (engulfedBurningOverlay == null)
 		{
 			engulfedBurningOverlay =
-				Instantiate(engulfedBurningPrefab, transform).GetComponent<PlayerDirectionalOverlay>();
+				Instantiate(engulfedBurningPrefab, OverlaySprites.transform).GetComponent<PlayerDirectionalOverlay>();
 			engulfedBurningOverlay.enabled = true;
 			engulfedBurningOverlay.StopOverlay();
 		}
@@ -135,14 +137,14 @@ public class PlayerSprites : MonoBehaviour
 		if (partialBurningOverlay == null)
 		{
 			partialBurningOverlay =
-				Instantiate(partialBurningPrefab, transform).GetComponent<PlayerDirectionalOverlay>();
+				Instantiate(partialBurningPrefab, OverlaySprites.transform).GetComponent<PlayerDirectionalOverlay>();
 			partialBurningOverlay.enabled = true;
 			partialBurningOverlay.StopOverlay();
 		}
 
 		if (electrocutedOverlay == null)
 		{
-			electrocutedOverlay = Instantiate(electrocutedPrefab, transform).GetComponent<PlayerDirectionalOverlay>();
+			electrocutedOverlay = Instantiate(electrocutedPrefab, OverlaySprites.transform).GetComponent<PlayerDirectionalOverlay>();
 			electrocutedOverlay.enabled = true;
 			electrocutedOverlay.StopOverlay();
 		}
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
@@ -42,6 +42,8 @@ namespace TileManagement
 		public Layer[] LayersValues { get; private set; }
 		public ObjectLayer ObjectLayer { get; private set; }
 
+		//Determines the maximum amount of overlays allowed on a tile
+		private const int OVERLAY_LIMIT = 20;
 
 		public List<Layer> ffLayersValues;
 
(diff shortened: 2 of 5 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
```diff
@@ -954,15 +956,15 @@ public bool HasTile(Vector3Int position, LayerType layerType)
 
 			TileLocation tileLocation = null;
 			OverlayTile overlayTile = null;
-			position.z = 1;
+			position.z = 0;
```
Bod9001: the reason it starts at one is because 0 is occupied by non-overlay tiles

Bod9001: but since it's checking for a OverlayTile it doesn't technically clash

Semuca: Since it's incremented at the beginning of the while loop anyway the first position checked is when z = 1

ThatDan123:
"some weird code" who would do such a thing, thanks for fixing it.

corp-0:
> so they rotate whenever the player body rotates

shouldn't flames always be upright?

UnityProject/Assets/Scripts/Player/PlayerSprites.cs
```diff
@@ -91,6 +91,8 @@ public class PlayerSprites : MonoBehaviour
 
 	public bool RootBodyPartsLoaded = false;
 
+	public GameObject OverlaySprites;
```
corp-0: is any other script going to interact with this field? If not, make it private and add ``SerializeField`` attribute instead of making it public.

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
```diff
@@ -1165,15 +1167,15 @@ public bool HasOverlay(Vector3Int position, LayerType layerType, OverlayTile ove
...
 
 			if (Layers.TryGetValue(layerType, out var layer))
 			{
-				//Only check 20 as its unlikely to ever have more than 20 overlays
+				//Go through overlays under the overlay limit. The first overlay checked will be at z = 1.
```
corp-0, line 1169: this loop seems to be used a lot and only slightly different each time. Can you extract it into another method to avoid duping code?

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
corp-0: then the intention would be read better if you started at one and incremented the z at the end

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
Semuca: Not really, because each loop does something different depending on the function it's in

Semuca:
> > so they rotate whenever the player body rotates
> 
> shouldn't flames always be upright?

But otherwise when the player drops the outline of the player on fire will just be left standing. If you mean wanting to have the fire rise up no matter what rotation the player is at, that's quite a big task which to me isn't all that noticeable. For now I think just rotating the overlay with the player is fine.

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
corp-0: yes, just slightly different and as you could experience yourself now when you had to make some changes, you had to do the same changes in all loops. Try extracting the part that is identical in all instances, return what the different methods need and operate with that returned value, that way if we need to change the loop we only change it in one place.

NoooneyDude:
> > > so they rotate whenever the player body rotates
> > 
> > 
> > shouldn't flames always be upright?
> 
> But otherwise when the player drops the outline of the player on fire will just be left standing. If you mean wanting to have the fire rise up no matter what rotation the player is at, that's quite a big task which to me isn't all that noticeable. For now I think just rotating the overlay with the player is fine.

Yeah some sort of dynamic flame sprite generation based on the current object sprite would be nice. Could be split into individual, small flame sprites determined by fire stacks and object sprite existence in that area. Could get pretty complicated.

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
NoooneyDude: I would suggest a `Func` delegate (take to `PlayerList` `getInternal()` and its callers for an example) however I suspect it creates garbage.

NoooneyDude: Actually should be fine if you only create the delegate objects once.

corp-0:
@Semuca do you think you can make the requested changes? Let us know if you find any trouble when trying.

Semuca:
Sorry about all that, I was up to my neck in schoolwork. I need to learn about delegate functions, but that shouldn't be too much. I'll see if I can do that change on Friday.

NoooneyDude:
> Sorry about all that, I was up to my neck in schoolwork. I need to learn about delegate functions, but that shouldn't be too much. I'll see if I can do that change on Friday.

I've decided to merge this now - probably shouldn't keep player prefab in limbo. A followup PR with the refactor would be greatly appreciated, still.

PR #6480, opened 2021-05-11
Fixes some client small freezes

Aranclanos:
### Purpose
Updates SpriteUpdateMessage to stop manually parsing the messages causing performance issues for clients.


UnityProject/Assets/Scripts/Messages/Server/SpritesMessages/SpriteUpdateMessage.cs
```diff
@@ -2,412 +2,267 @@
...
+			PresentSpriteSet,
+			VariantIndex,
+			CataloguePage,
+			AnimateOnce,
+			PushTexture,
+			Empty,
+			PushClear,
+			ClearPallet,
+			SetColour,
+			Pallet
```
NoooneyDude, line 265: I think this is Bodspeak for `palette`.

PR #6489, opened 2021-05-12
Fixes a case of clients seeing other players as floating skulls

Aranclanos:
### Purpose
Stops the server from sending sprite updates from objects that no longer exist, making clients constantly search for them.
Fixes a runtime in the server sprite update sending message, preventing clients who logged in from receiving an important chunk of these updates, resulting in players seeing others as floating skulls.

### Notes:
the list NewClientChanges could contain SpriteHandlers from objects who were deleted
I introduced this bug on my last pr and staging didn't update yet, but I understand that floating skulls were present before I started my work on SpriteUpdateMessage, so there's a chance that the problem persists and that I'm just fixing a bug nobdoy has experienced yet.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Sprite Handler/SpriteHandler.cs
+++ b/UnityProject/Assets/Scripts/Core/Sprite Handler/SpriteHandler.cs
@@ -576,6 +576,15 @@ public class SpriteHandler : MonoBehaviour
 		TryInit();
 	}
 
+	private void OnDestroy()
+	{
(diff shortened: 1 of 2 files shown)


Bod9001:
"Fixes a runtime in the server sprite update sending message" what was the problem, I can't seem to find any changes related to it?

Aranclanos:
the NewClientChanges dictionary could have an entry to an object deleted, there was a null check that I removed in my last PR

Bod9001:
ok cool

PR #6511, opened 2021-05-14
Blob Fixes and Forces Player Colour Alpha to 1

ThatDan123:
-Should fix blob UI issues

-Forces player colour alpha to 1


UnityProject/Assets/Scripts/Player/PlayerSprites.cs
```diff
@@ -382,6 +383,8 @@ public void SetSurfaceColour()
 			ColorUtility.TryParseHtmlString(ThisCharacter.SkinTone, out CurrentSurfaceColour);
 		}
 
+		CurrentSurfaceColour.a = 1;
```
Bod9001, line 397: can you fix the bug with using == / contains with colour when it needs to be a approximate check because float numbers, needs to be on the customisations sprites as well

PR #6512, opened 2021-05-14
Character preview will no longer escape outside of the window

MaxIsJoe:
CL: Character previews are now fully contained inside their preview containers. They can no longer escape.


MaxIsJoe:
why is tests failing

ThatDan123:
Turns out spritehandler controller is still needed

ThatDan123:
You can download the test results:

image

And then open index.html

corp-0:
probably, prefabs that had a reference to the deleted monobehaviour

waffielz:
finally, proper contentment for the scp

Issue #6507, opened 2021-05-14
Lag after roundstart


NoooneyDude:
There seems to be more lag than usual at round start. A cheap solution is to have most things start, say, 15 seconds earlier than roundstart (the moment players start spawning) to reduce the lag at this critical point.

ThatDan123:
Think its due to all the populator contents spawning in (lockers and bags....)

To reduce it either spawn them before start, or only spawn them in on first open (e.g opening locker for the first time), does have the down side of the first open might be slightly delayed, but would be quick afterwards

PR #6513, opened 2021-05-14
Closet Populator Optimisation

ThatDan123:
CL: Locker initial contents will now only spawn on first open instead of round start

Before this PR networked objects on square:
Before start: 4632, after start: 7351

After this PR:
4632 to 5617

1,734 less game objects! (as lockers are opened contents do spawn so that number will go down as the round goes on, but now they will only be first spawned if needed)

Should reduce lag and network usage for new players joining a server (assuming a lot of those lockers are never opened).
Also should reduce that slight delay after a round has started after the countdown reaches 0;
Might help #6507 a bit.

-Next step would be to do the same for bag populators


NoooneyDude:
I think something like this needs to be tried out on staging.

PR #6516, opened 2021-05-15
Item Storage Populator Optimiations

ThatDan123:
CL: Item Storage will only spawn stuff on first open not round start anymore

Using Square Station quick load (square is the only scene) as a test:

Before PR:
Lobby: 2513 Start: 4277

After PR:
Lobby: 2513 Start: 2591

1686 less game objects!

This will need some actual testing as item storage is used in lots of stuff unlike the locker PR which was easy to test.

If some stuff isnt spawning there is an override on the item storage component which will spawn the stuff at start like before.

Related to #6507 and #6513

-Adds tests to make sure all objects (not items) are forced to have spawn on start to avoid non spawning issues
-Also fixes cargo cash issue and adds tests for empty item traits
-Fixes singularity not locking points on stage 0

~Currently doing a build test now~
Looks like its working fine, cant find any problems.


Aranclanos:
I fear that many objects will break different type of interactions that a sandbox game like ss13 has. 
On an easy example, lets say an EMP grenade goes off next to an untouched locker and there are energy guns that spawn from it. If they're not inside the batteries of these guns are not drained and once opening the locker you'll have loaded guns.
There's many cases that inacuaracies can happen with this strange on-use spawn.

ThatDan123:
I get what you mean, but this optimisation is worth the extra effort to have to think about those niche situations and then remember to trigger the spawn manually (E.g if EMP'd)

NoooneyDude:
~Works as expected, does what it says on the tin.~

ThatDan123:
Ready to be merged

corp-0:
Huh, we don't have the "test on staging" label anymore. Anyways, I'm test merging this, if something breaks in staging remember to revert it.

PR #6519, opened 2021-05-15
Readd Player and mob examine message

ThatDan123:
CL: Readds mob examine

CL: Readds player status and soul status


corp-0:
Looks good

PR #6524, opened 2021-05-15
Null Check PickUpAnimation And Other Error Fixes

ThatDan123:
image

Was the last error before being disconnected every time, hopefully this fixes the disconnects.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Lighting/EmergencyLightAnimator.cs
+++ b/UnityProject/Assets/Scripts/Core/Lighting/EmergencyLightAnimator.cs
@@ -47,10 +47,15 @@ namespace Objects.Lighting
 		private void AnimateLight()
 		{
 			if (lightSource == null || lightSource.mLightRendererObject == null ||
-				lightSource.mLightRendererObject.transform == null)
+			    lightSource.mLightRendererObject.transform == null)
 			{
-				Debug.LogError($"{gameObject.name} had something null");
 				StopAnimation();
+
+				if (this != null && gameObject != null)
+				{
+					Debug.LogError($"{gameObject.name} had something null");
+				}
+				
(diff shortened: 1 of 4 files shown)


MaxIsJoe:
Is adding a null check really the solution here?

Also how can interactor ever be null?

ThatDan123:
It errors on join, in the RPC. Probably cus the client hasnt loaded the object before the message is sent. Not sure why its force closing the connection though.

ThatDan123:
The disconnect is cus is theres an error downstream mirror will force disconnect:

image

I managed to do a check earlier for netmessages to stop those disconnects but since this is an RPC mirror directly handles it

MaxIsJoe:
Ah, alright then.

ThatDan123:
Good to be merged

UnityProject/Assets/Scripts/Core/Lighting/EmergencyLightAnimator.cs
```diff
@@ -47,10 +47,15 @@ protected virtual void UpdateMe()
...
 			if (lightSource == null || lightSource.mLightRendererObject == null ||
-				lightSource.mLightRendererObject.transform == null)
+			    lightSource.mLightRendererObject.transform == null)
 			{
-				Debug.LogError($"{gameObject.name} had something null");
 				StopAnimation();
+
+				if (this != null && gameObject != null)
+				{
+					Debug.LogError($"{gameObject.name} had something null");
```
corp-0:
```suggestion
					Logger.LogError($"{gameObject.name} had something null");
```

corp-0: the log message seems a little ambiguous

Issue #6398, opened 2021-04-20
Eternal Flames and Fire Damage Problems B:50


PerfectTangent:
### Description

**This is a stability related bounty, and is immune to the feature freeze.** This bounty calls for fixing an issue with fires breaking and never dissipating. Quite frequently nowadays, fires will break and never end. Fires are also reportedly not doing damage to players when they should, and players are unable to resist out of burning and staying lit on fire forever. The specific causes for the bugs are not known. The issue of players not taking fire damage may be related to fires never ending or it may be a separate bug introduced with the new health content.

This bounty also calls for a small feature. Now that we have crawling in the game, resisting flames should initiate crawling as players are trying to put out the flames by rolling around on the ground.

### What is needed to fulfill the bounty:

- [x] Fix any bugs that are causing fires to never end or that cause them to cease interacting with their environment.
- [x] Fix any bugs with player characters not receiving damage while on fire
- [x] Fix any bugs with players not being able to extinguish themselves with the resist key. Resisting out of fire should also cause players to begin to crawl, as they stop, drop, and roll.

### Steps to reproduce

depiction of an endless fire on square station
![Screenshot_at_2021-04-19_13-00-53](https://user-images.githubusercontent.com/68730437/115423856-2a4cab00-a1cc-11eb-8365-18a95bb9c3cc.png)

PerfectTangent:
Bounty price set at 100.00 USD.

Semuca:
I'll take this bounty up and see how I go

PerfectTangent:
alright, you've been assigned.

Semuca:
I'm going to document this here for other devs and also myself. This error is reproducible, but in specific circumstances.

<img width="960" alt="Screenshot 2021-05-09 200823" src="https://user-images.githubusercontent.com/62051810/117572266-b0417f00-b104-11eb-9c2b-b76ec5055c66.png">
By placing the plasma canister in this specific position on SquareStation and starting a fire, it will consistently cause the fires to become stuck. Some other notes:

- If the canister gets pushed by space wind while the fire occurs, this eternal fire will break.
- If the canister gets pushed by a player after the fire has become eternal, the fire will not break. If it is pushed into an non-flame tile, it will only ignite that tile.
- The fire will hurt you and set you on fire, but it seems like it won't register with the UI on the right-hand side of the screen.
- Strangely enough, this eternal flame doesn't seem to be affected by where the fire is started.

corp-0:
[...]

PR #6527, opened 2021-05-16
Possible fix for unending plasma fires

Semuca:
### Purpose
Removes a InvalidOperationExemption error that occurred when plasma fires were around. This seems stop the unending plasma fires that could be started using my process in #6398, although I think we should wait to see if any more come up before settling this bounty.

### Notes:
Got a pretty bad cold on Friday, so I was gone for a bit. Sorry about that.

### Changelog:
CL: Possible fix for eternal fires

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Utils/TileList.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Utils/TileList.cs
@@ -159,7 +159,14 @@ public class TileList
 		}
 
 		lockedPosition = localPosition;
+		List<RegisterTile> registerTiles = new List<RegisterTile>();
 		foreach (var registerTile in Get((Vector3Int)lockedPosition))
+		{
+			registerTiles.Add(registerTile);
+		}
+
+		//To prevent modifying collection while in the loop
+		foreach (RegisterTile registerTile in registerTiles)
 		{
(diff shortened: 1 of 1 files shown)


ThatDan123:
Worth saying this function will now create GC

NoooneyDude:
You're right. Should I revert?

NoooneyDude:
Would it be possible to sort out the GC issue? Having that list as static readonly list might help.

ThatDan123:
Make a reusable array with a sensible max capacity in the reaction manager (as thats the only thing which is calling this function which is nice), maybe like 30 (not sure how many register tiles could be on a tile) and then send and reuse that array each time the function needs it.

Pretty sure that will be no GC, Aranclanos knows this stuff better though

Semuca:
Would it be valid to store the reusable array inside the ReactionManager, so that all users of the function won't have to have their own reusable list?

ThatDan123:
Yes, i think so as i dont think each reaction manager is on its own thread if you wanted to use a static array. It'll be fine to use a instance array too

Aranclanos:
don't make it static since theres one reactionmanager per matrix and make sure to Clear() it after use

Aranclanos:
actually, isnt this a new list already? cant you just loop backwards

Issue #2463, opened 2019-12-14
Remove Network Identity from Matrices and Consolidate


chairbender:
# Description
Everywhere you look, Mirror says it's not supported to have objects which have nested network identity components. All of our matrices have a NetworkIdentity, and all of the objects within them also have one. Currently it seems to work okay for us because the matrices are mapped in, but if they were converted to prefabs we'd likely have issues. And this doesn't seem like a safe situation to be in considering mirror warns against it.

Refactor so the matrices no longer need a network identity component. This will require having a way to lookup matrices by ID (which pretty much already exists in MatrixManager), then coming up with an alternative for every NetworkBehaviour that is used on the Matrix so they can be converted to MonoBehaviors. For example you could create net messages for these which send and look up the matrix ID using MatrixManager.

Or, possibly another idea that's easier and would still allow NetworkBehavior, you could possibly just create an invisible "matrix proxy" object that's in the root of the scene which you explicitly link to one of the matrices (one per matrix). It wouldn't have anything inside it, but the various networkbehaviors that currently live on the matrix could be moved here, and any manipulations it needs to perform on the matrix would be done on the matrix transform that it's linked to. So the proxy matrix would be the main point for networking / networkbehavior, but it would make changes to the actual matrix object.

And register tile already works to figure out the matrix an object is supposed to be in when the object is spawned over the network.

This should also help set the stage for turning matrices into prefabs or allowing dynamically adding / removing matrices.

Along with these, please also investigate the possibility of consolidating matrices so that there's no more need for both a parent and child matrix, as this has been creating confusion and complexity that doesn't seem like it should be needed.

ThatDan123:
Main part of this issue was done in #6531, other issues are open for the remaining parts so this can now be closed

PR #6531, opened 2021-05-16
Remove Network identity From Matrix v2

ThatDan123:
CL: Removes network identity from matrixes

-Edits all scenes

#2463 is now partially done

~Just build testing now~

Tested fine with a headless server and a client moving through matrixes and flying a shuttle. But can someone else build test too since this is a big change.


ThatDan123:
This is ready to be fully tested by someone else

I tested headless with client, going between different matrixes and flying a shuttle. All seemed fine.

NoooneyDude:
Did you test with two clients?

ThatDan123:
Tested with two clients and headless. All seems good, ive fixed up all the warnings too

PR #6532, opened 2021-05-16
Might fix skeleton

ThatDan123:
CL: Adds scene checker to the net identity

-In the long term we'll need to try to remove the identity


corp-0:
how do we test this?

ThatDan123:
I think just starting a headless server, joining one client and then late joining another client will make the late joining see anyone who was already in the game as a skeleton

DTraitor:
> I think just starting a headless server, joining one client and then late joining another client will make the late joining see anyone who was already in the game as a skeleton

now you also see your own skull and skulls of people who joined after you did

NoooneyDude:
I couldn't see any skeletons when testing with two clients and disconnecting, restarting.

Note that this had a conflict (at least, locally) with #6516.

PR #6533, opened 2021-05-16
Fixes energy bow bullet prefab

ThatDan123:
CL: The energy bows now fire their correct bullets


corp-0:
go home codacy, you drunk

PR #6538, opened 2021-05-17
Fixes handcuffs and other items from being uninteractable.

### LINKED ISSUE #6523: Documenting the handcuff pickup bug
### Environment
- Game Version: upstream/develop (15-5-2021)
- OS Version: Win10

Handcuffs can no longer be picked up after uncuffing someone as reported in #5726
This issue's purpose is to document this nasty bug so others can have an either time fixing it. 

This bug only occurs on headless servers, it does not happen inside the editor unless you're joining a server through the editor.

### Steps to reproduce

1. Cuff someone
2. Uncuff them
3. Try picking up handcuffs

### Things that have been tested and observed.

- The game doesn't currently throw any errors when trying to pick them up as far as I'm aware
- There is no null checks in any of the restraint or interaction code that are silently causes this issue to appear without the game throwing any errors at us.
- Handcuffs green highlight missing?

### Things that need to be tested.

- Check if the game disables the pickupable component and forgets to re-enable it.
- Check if the game registers the click and if it executes any logic after it registers.
- Does the game change anything about handcuffs after they enter player inventory?

Aranclanos:
### Purpose
Fixes handcuffs and other items from being interactable.

If the clients received an update of removal from an itemslot that they were observing (even theirs), they would never set the pickupeable itemslot to null.
Because they believed the pickupeables had an itemslot, they would refuse to interact with it. 
It didnt happen when they were the ones dropping the object because the WillInteract() just believed that it was in their own inventory and returned true.

### DIFF
--- a/UnityProject/Assets/Scripts/Messages/Server/UpdateItemSlotMessage.cs
+++ b/UnityProject/Assets/Scripts/Messages/Server/UpdateItemSlotMessage.cs
@@ -38,19 +38,11 @@ namespace Messages.Server
 					slot = ItemSlot.GetIndexed(NetworkObjects[0].GetComponent<ItemStorage>(), msg.SlotIndex);
 				}
 
-				var previouslyInSlot = slot.ItemObject;
-				var pickupable = msg.Item == NetId.Invalid ? null : NetworkObjects[1].GetComponent<Pickupable>();
-				slot.ClientUpdate(pickupable);
-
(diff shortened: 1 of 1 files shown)


NoooneyDude:
YES! Thank you.

NoooneyDude:
This alone makes staging ready for release now in my opinion...

PR #6540, opened 2021-05-17
Fix Floating Effect

ThatDan123:
CL: Makes the floating effect use sync vars instead and is triggered by server now


NoooneyDude:
Hmm, some spaces for whitespace and not tabs, for your information.

MaxIsJoe:
The whole point of RpcLeanTween was to provide a library of networked lean tween functions so we don't have to do stuff like this. Are you sure that there isn't another solution?

PR #6543, opened 2021-05-17
Reflective Blob Now Reflects

### LINKED ISSUE #6535: Make reflective blob reflect lasers
-Reflectors which can reflect lasers are already in the game, just need to make a variant of the component or add the component (forgotten how i coded it :p) to the reflective blob structure

ThatDan123:
CL: Reflective blob can now reflect lasers, however is still damaged by them so will still eventually die to continuous lasers

-Also makes statues initially not pushable

fixes #6535

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/LaserReflectObject.cs
+++ b/UnityProject/Assets/Scripts/Objects/LaserReflectObject.cs
@@ -1,0 +1,42 @@
+﻿using Messages.Server;
+using ScriptableObjects.Gun;
+using UnityEngine;
+using Weapons.Projectiles;
+using Weapons.Projectiles.Behaviours;
+
+namespace Objects
+{
+	public class LaserReflectObject : MonoBehaviour, IOnHitDetect
+	{
+		[SerializeField]
+		//Use to check whether a bullet is a laser
+		private LayerMaskData laserData = null;
+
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Objects/LaserReflectObject.cs
```diff
@@ -0,0 +1,42 @@
...
+			//Only reflect lasers
+			if (data.BulletObject.TryGetComponent<Bullet>(out var bullet) == false || bullet.MaskData != laserData) return;
+
+			ShootAtDirection(GetNewDirection(data), data);
+		}
+
+		private Vector2 GetNewDirection(OnHitDetectData data)
+		{
+			var normal = data.BulletShootNormal;
+			var newDirection = data.BulletShootDirection - 2 * (data.BulletShootDirection * normal) * normal;
```
Bod9001, line 26: I assume this is copying from the code with the tile version of this?

ThatDan123: Uses a different parameter but yeah

ThatDan123: Though the one from the tile version use global variables as well

PR #6551, opened 2021-05-18
Matrix Sync Test

ThatDan123:
-Adds a test to make sure theres a matrix sync on all the correct matrixes

-Added more checks to the interaction messages to make sure it grabs correct netId

-Fixes ore gen error

I have tested by removing a matrix sync in the scene and running test in editor and it works.

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Ore Generation/OreGenerator.cs
+++ b/UnityProject/Assets/Scripts/Systems/Ore Generation/OreGenerator.cs
@@ -53,7 +53,7 @@ public class OreGenerator : MonoBehaviour
 		wallTilemap = metaTileMap.Layers[LayerType.Walls].GetComponent<Tilemap>();
 		tileChangeManager = GetComponent<TileChangeManager>();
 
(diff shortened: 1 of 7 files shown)


UnityProject/Assets/Scripts/Systems/Ore Generation/OreGenerator.cs
```diff
@@ -50,49 +53,59 @@ public void RunOreGenerator()
 		wallTilemap = metaTileMap.Layers[LayerType.Walls].GetComponent<Tilemap>();
 		tileChangeManager = GetComponent<TileChangeManager>();
 
-		if (CustomNetworkManager.IsServer)
+		if (CustomNetworkManager.IsServer) return;
```
NoooneyDude: Are you sure you want to return if it's the server?
```suggestion
		if (CustomNetworkManager.IsServer == false) return;
```

ThatDan123: Good catch

PR #6555, opened 2021-05-18
fixes highlight disappearing

Bod9001:
scale was getting set too small randomly somehow


Aranclanos:
I'd suggest finding the reason of the bug before slapping a fix

corp-0:
> I'd suggest finding the reason of the bug before slapping a fix

Same. That mysterious reason might be doing more than just making highlight small

Bod9001:
it's because of the lean tween animation squishing it down when you pick it up

corp-0:
oooh, so it doesn't go back to normal after. What is your change doing?

Bod9001:
just reset scale and, tells it not to bring scale over

PR #6578, opened 2021-05-19
Re-textures high-voltage cable

Bod9001:
since there was many issues with people incorrectly wiring high-voltage cables, outing to sprites to highlight how the cables connect
new
image
old 
image


PerfectTangent:
cute. you could add little tips of copper to the ends of the cable if you like, but that looks good. It communicates what it needs to very clearly.

PR #6587, opened 2021-05-20
Possible Rejoin Fix

ThatDan123:
-Might fix the odd occasion where you rejoin a game but end up on the new player join screen instead of rejoining body

-Caches lighting system state for admin tools

-Adds methods in player list to search for offline and online players as some methods should check both.

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/PlayerList.cs
+++ b/UnityProject/Assets/Scripts/Managers/PlayerList.cs
@@ -354,11 +354,11 @@ public partial class PlayerList : NetworkBehaviour
 	/// <returns></returns>
 	private ConnectedPlayer GetInternalLoggedOff(Func<ConnectedPlayer, bool> condition)
 	{
-		for (var i = 0; i < loggedIn.Count; i++)
+		for (var i = 0; i < loggedOff.Count; i++)
 		{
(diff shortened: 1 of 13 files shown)


DTraitor:
What happens if you join with the second client while the first client is on the server?

ThatDan123:
The new joining player should get kicked for release builds

UnityProject/Assets/Scripts/Managers/PlayerList.cs
```diff
@@ -302,7 +312,47 @@ public List<ConnectedPlayer> GetAllByUserID(string byUserID, bool includeOffline
...
+
+		return ConnectedPlayer.Invalid;
+	}
+
+	/// <summary>
+	/// Check logged off players
+	/// </summary>
+	/// <param name="condition"></param>
+	/// <returns></returns>
+	private ConnectedPlayer GetInternalLoggedOff(Func<ConnectedPlayer, bool> condition)
```
NoooneyDude, line 355: This is identical to `GetInternalLoggedIn()`?

ThatDan123: dont worry im an idiot

PR #6590, opened 2021-05-21
Adds a delay to disarm

### LINKED ISSUE #5719: Disarms can be spammed extremely quickly and need a cooldown.
### Environment
- Game Version: 5608
- OS Version: Windows 10

Right now, there doesn't appear to be a cooldown timer on disarms. This means that you can disarm a person 5 times in a second. This makes disarms more powerful than they should be due to how quickly you can throw them out. There should be a cooldown between disarms so you can't spam them, possibly a second as in /tg/.

### Steps to reproduce
1. Go to another person.
2. On disarm intent, click on them.
3. Disarm them several times in just a single second, much more in multiple seconds.

Bod9001:
fixes #5719

keep track of who disarmed and then if it's to recent block it


NoooneyDude:
Won't the `Cooldowns` helper class work here?

Bod9001:
will take a look

PR #6594, opened 2021-05-21
QuickBuild Tool

NoooneyDude:
- Adds a new QuickBuild tool:
quickbuild
Operates independently of the main build window.
Settings are persistent and won't be picked up by git (except `Quick Load`).

>The `Disable Scenes` tab is the old `DisableNonEssentialScenes` tool.


UnityProject/Assets/Scripts/Core/Editor/QuickBuild.cs
```diff
@@ -0,0 +1,449 @@
...
+			}
+		}
+
+		#endregion
+
+		#region Disable Scenes Tab
+
+		private static List<string> AdditionalScenesToRemove = new List<string>()
+		{
+			"Fallstation Centcom",
```
corp-0, line 254: I remember NaughtyAttributes had  an attribute to work with scenes in inspector while internally keep using strings.

corp-0: aaaah but I imagine naughtyattributes doesn't work with this custom window, right?

NoooneyDude: Yep, I'm using it. Ignore the Disable Scenes region - that's existing code that's just been namespaced.

PR #6596, opened 2021-05-21
Outline and underlay for chat

DTraitor:
### Purpose
When we moved to TMP in chat we've lost shadows (they don't work with TMP so I removed them from chat entry). Here is comprasion of what we have rn and how it'll look after this PR merged.

Nothing:

Underlay + Outline:

### Notes
Underlay was set using notepad. You cant set it bigger than 1 in Unity.
If someone add's second layer of TMP that is completely black and disable changes I did in this PR it'll look even better than it is.

### Changelog:
CL: Added outline and underlay to chat entry

### DIFF
--- a/UnityProject/Assets/Prefabs/Chat/ChatEntry.prefab
+++ b/UnityProject/Assets/Prefabs/Chat/ChatEntry.prefab
@@ -473,7 +473,6 @@ GameObject:
   m_Component:
   - component: {fileID: 1885229338361628804}
   - component: {fileID: 6359359618520375730}
-  - component: {fileID: 4478343802284241693}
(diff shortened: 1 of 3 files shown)


DTraitor:
It seems tests are failing because of something unrelated to the PR

corp-0:
can you please test locally?

DTraitor:
I can test tomorrow but how do I do that?

corp-0:
if you have the side bar with many icons, you can click in the unitystation icon and that will open the test runner and just run all tests.

Bod9001:
I think there is shadows/outline with tmp. It's just that they are global

PR #6589, opened 2021-05-21
Square and additional scenes updated to doors V2 + silo

PerfectTangent:
### Purpose

This PR updates Square station and all additional scenes except the Fall station syndicate scene to Doors V2. I also made a medium quantity of small other changes to Square and some of the additional scenes and added a tiny amount of prefab variants for SSUs. Makes progress on #6554.

### Notes:

- fixed _a lot_ of stuff that was fucked up with the Centcom scene.
- Gary the Guard borg has been confined to Centcom's attic for the safety of visiting personnel.
- removed the deathsquad hardsuits and disruptor cannons that could just be freely taken by ERT personnel in Centcom.
- added a couple new suit storage units for ERT hardsuits. Also mapped a few of them into Centcom and Fall station centcom where ERT hardsuits were previously stored.
- slightly increased the size of R&D in Square and replaced all of the science walls with rwalls.
- added a few APCs to Square's maintenance halls.
- added a mineral silo to Square's vault.
- fixed a couple weird bits of Lavaland's mining base layout.
- removed the old mining equipment vendors in Lavaland and replaced them with the new version.
- liberally added ore boxes to lavaland.
- reworked both syndicate Outpost and syndicate Pog a bit, and removed all of the old legacy weaponry vendors and lockers in them. Removed a total of about _30_ stalker SMGs from both scenes combined. They aren't needed since nukies have the majority of their intended equipment available now in the uplink, which they should be paying for. I did leave a small number of weapon lockers in the very back of the syndicate Outpost scene.
- slightly reoriented the operations console in the Syndicate Outpost scene to make it a bit more visibly apparent.
- some other shit, idk. When I close my eyes all I see are doors, now.

PR #6598, opened 2021-05-21
fixes light switches in Fall centcom and wizard asteroid

PerfectTangent:
### Purpose
Orphaned commit from #6589. Feexis sweetchis.


corp-0:
are you perhaps conflicting with yourself?

PerfectTangent:
I don't think I have enough experience to see why I would conflict with myself on the changes made in a commit that didn't actually make it into the previous PR.

```
UnityProject/Assets/Scenes/AdditionalScenes/Fallstation Centcom.unity
UnityProject/Assets/Scenes/AdditionalScenes/WizardAsteroid.unity
```

corp-0:
> I don't think I have enough experience to see why I would conflict with myself on the changes made in a commit that didn't actually make it into the previous PR.

It happens because both PRs change the same files and you have commits that don't exist in the develop branch as they were squashed into a single commit. It is an easy fix, though, just need to update from upstream/develop and solve the conflict by overwritting upstream/develop since your local has the updated version.

PR #6606, opened 2021-05-23
Better possible fixes for plasma fires

Semuca:
### Purpose
I realised the fix I had for plasma fires before was a pretty bad way for going about it, so I've done a new method to reduce some red unity errors that may be causing immortal plasma fires.

### Explanation:
An InvalidOperationError was being called because the lighting thread tried to create sparks which could light fires. This could overlap with the ForEachSafe function in TileList.cs, meaning two of the same function on the same object would try to be called at the same time. I've fixed this by changing the requirements for the ForEachSafe function to break, from `lockedposition != null && lockedposition != localposition` to `lockedposition != null`. I don't see why the `localposition` was checked as well, because even if it was the same as `lockedposition` the variable would still be changed, causing the InvalidOperationError. However, from my testing it hasn't broken anything.

Because of that, I've also had to lock the hotspotsToAdd and hotspotsToRemove during updates since other threads can now edit them while the update is running without an error being thrown and the code breaking.

### Notes:
I've realised my way of reproducing the immortal plasma fires doesn't work, and it was probably a fluke of the bug that caused me to think that spot was special. It seems the only way to find out if these fires are fixed is to wait and see if the bug occurs again. From my testing though, I haven't seen it reappear.

### Changelog:
CL: Fixes a possible cause of neverending plasma fires

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/ReactionManager.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/ReactionManager.cs
@@ -80,50 +80,56 @@ namespace Systems.Atmospherics
 			Profiler.BeginSample("HotspotModify");
 			//perform the actual logic that needs to happen for adding / removing hotspots that have been
 			//queued up to be added / removed
-			foreach (var addedHotspot in hotspotsToAdd)
+			lock (hotspotsToAdd)
 			{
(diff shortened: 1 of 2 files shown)


jamessimo:
I don’t suppose anyone else can think of a way to reproduce this bug?

Bod9001:
looks good

corp-0:
test merging since the bug is way too hard to reproduce locally and test.

PR #6608, opened 2021-05-23
Added information to help UI development

Akbadain29:
### Purpose
Added in file locations so people know how to start making a UI in the documentation. I'll try and add more instructions to link everything together later, but I wanted to add this in before I forget again.


corp-0:
shouldn't it be "a**n** user interface" ?

PerfectTangent:
The use of "an" is a bit weird with words that start with "u." Some varieties of english accents will use "a" instead of "an" for words like unicorn, university, useful, union, umbrella, etc.. It's just some weird shit that has become a standard in some accents. English is a unreliable, absolute trash brain language. Grammatically, you're right that it would be "an user interface." That would be the correct usage.

corp-0:
what would be the proper way in this case? "A user interface"?

PerfectTangent:
Either is fine for me. I personally would say "**a** user interface."

corp-0:
alrighty then

PR #6609, opened 2021-05-23
Shuttle engines now change brightness on thrust

### LINKED ISSUE #6067: Shuttle engines are way too bright
Shuttle engines project a really really bright blooming light, I guess this effect was supposed to be shown when the engines are active, also the escape shuttle's engines are backwards, but they correct themselves when the shuttle is launched, weird.

Unitystation_TmUV55c4gf

Unitystation_chohjaTWy6

Unitystation_v3j3SksdjM
MY EYES AAAAAAAAAA (notice that the engines are backwards)

GsiytTc5HF
After launch.

Bod9001:
Fixes : #6067

this makes the light, brighten and dim depending on speed of the shuttle/matrix ,and when off not visible at all

### DIFF
--- a/UnityProject/Assets/Scripts/Shuttles/ShipThruster.cs
+++ b/UnityProject/Assets/Scripts/Shuttles/ShipThruster.cs
@@ -6,9 +6,9 @@ using UnityEngine;
 public class ShipThruster : MonoBehaviour
 {
 
-	public MatrixMove shipMatrixMove; // ship matrix move
-	public ParticleSystem particleFX;
-	public LightSprite lightSprite;
+	private MatrixMove shipMatrixMove; // ship matrix move
+	private ParticleSystem particleFX;
+	private LightSprite lightSprite;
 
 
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Shuttles/ShipThruster.cs
```diff
@@ -1,20 +1,24 @@
...
 using System.Collections.Generic;
+using Light2D;
 using UnityEngine;
 
 public class ShipThruster : MonoBehaviour
 {
 
 	public MatrixMove shipMatrixMove; // ship matrix move
 	public ParticleSystem particleFX;
+	public LightSprite lightSprite;
```
corp-0: I know everything else is public in this class because this is one of the older classes where that was fine, but new fields should always be private unless we really need to set and get it somewhere else (and even then, I'd suggest using properties instead).

Bod9001: ok

PR #6615, opened 2021-05-25
This works? (test fix)

Bod9001:
wasn't able to grab packages for some reason


ThatDan123:
nope, still not working

Bod9001:
this technically doesn't count since, I removed Fire base

Bod9001:
let me check if it still compiles?

ThatDan123:
I think it might be working fine.

Bod9001:
doesn't compile,  must be importing project

Bod9001:
humm, this works buts add 100 mb file

PR #6616, opened 2021-05-25
This work? (Test fix, with remote packages)

Bod9001:


Bod9001:
don't think this is possible unless we find a new URL that hosts the packages

PR #6617, opened 2021-05-25
The server will now check for illegal character settings

### LINKED ISSUE #6478: Why is sus character creation not checked serverside? a.k.a 1upping issue #4292
EDIT:
So i messed a bit more and it turns out its all client side, good thing im not getting paid for spotting shit like this lmao
eitherway one can probably turn themselves blue and whatnot, yeah
image

image
image
image
image
image
you get the point

discord Jakesolly#0966

how to reproduce:
use dnspy or from source doesnt matter lol
write to the server in a way that bypasses the client side checks, i just modified the randomize button function so that it types whatever i want
image

 ps: thanks for changing invalid account names into 'nerd' >:(

MaxIsJoe:
### Purpose
fixes #6478

### Notes:
Needs testing from someone who can build the game because unity is being a pain with me right now and I can't build anything.

### Changelog:
CL: [Fix] - The server will now check for illegal character sheets and will kick players who try modifying their sheets to spawn in with settings that are not allowed by the game.


UnityProject/Assets/Scripts/Core/Input System/InteractionV2/Validations.cs
```diff
@@ -772,4 +772,83 @@ public static bool HasBothHands(GameObject player)
...
 		return health.HasBodyPart(BodyPartType.LeftArm, true) && health.HasBodyPart(BodyPartType.RightArm, true);
 	}
+
+
+	/// <summary>
+	/// Validates that the player's character name.
+	/// </summary>
+	/// <param name="characterName">CharacterSettings.Name</param>
+	/// <returns>True if illegal.</returns>
+	public static bool HasIllegalCharacterName(String characterName)
```
Bod9001: kind of funny place to put it since it's under interactions  "ore/Input System/InteractionV2/Validations.cs "

Bod9001:
it's only checking name age and skin Tone, 
I think there is some validation with the sprites and colour options, E.G  hair and such, not too sure though,
  I'm interested to why you think you would need to be kick surely just falling back to defaults/unset would be good enough

UnityProject/Assets/Scripts/Core/Input System/InteractionV2/Validations.cs
MaxIsJoe: Moved to server validations.

MaxIsJoe:
> it's only checking name age and skin Tone,
> I think there is some validation with the sprites and colour options, E.G hair and such, not too sure though,
> I'm interested to why you think you would need to be kick surely just falling back to defaults/unset would be good enough

We have agreed about this before, we kick players instead of playing with their settings.
Also there are no ways you can modify your character sheet to spawn in a "illegal" sprite. Illegal colored hair for lizards? Sure, but other than that you can't do anything else.

Bod9001:
ok cool, You could plug it up to the BodyPartCustomisationBase and what inherit from it to kick as well, E,g trying to set alpha values

Bod9001:
kicking seems a bit extreme, I'm thinking a case of someone joins from different server with different characters but I suppose we would have to server sided at some point the characters that is and the user settings

MaxIsJoe:
> kicking seems a bit extreme, I'm thinking a case of someone joins from different server with different characters but I suppose we would have to server sided at some point the characters that is and the user settings

I don't think kicking is extreme since it's not preventing them from playing the game or on the server.
This is a much better solution as well so in the future when character sheets become corrupted or a server has different settings players don't spawn with tampered character sheets and we won't have to wonder "why the hell are characters not spawning correctly?".

corp-0:
> CL: [Security] 

``Fix`` is alright. The less categories, the better.

MaxIsJoe:
> > CL: [Security]
> 
> `Fix` is alright. The less categories, the better.

You can edit them yourself as a maintainer btw

corp-0:
yep, but if I don't bring the attention to the change, you might keep creating categories.

MaxIsJoe:
> yep, but if I don't bring the attention to the change, you might keep creating categories.

image
I'm not creating new ones, I already said that these are the tags we can use for changelogs and security is fitting for an issue that's tagged with security.. oh well ¯\_(ツ)_/¯

edit : no idea why would anyone care about that as well tbh when it's just me and another person who label our changelogs unlike the rest of the contributors.

Bod9001:
what you think about hooking it up with BodyPartCustomisationBase ? and what inherits from it

MaxIsJoe:
> what you think about hooking it up with BodyPartCustomisationBase ? and what inherits from it

Here's a problem with this approach, this approach spawns the player in.
It leaves a body on the ground with the broken character sheet, and the more i think about it, this is terrible. What if we change skin colors to individual body parts mid-round and we re-attach a limb or we change the player's skin color for any reason that it doesn't match with their sheet's skin color? What about special species that don't share the same logic as lizards?
this approach is terrible and will cause problems in the future, it's better to leave it in PlayerSpawn where whenever we want to check for any illegal settings in a character sheet we just update that function with a function that checks for what we need to check.

Bod9001:
well, the logical setting up the character setting should only be called once on it , so it shouldn't do it mid round, 

it's only way you can validate individual body part settings

waffielz:
1984

PR #6620, opened 2021-05-26
Fixes a few staging errors

Bod9001:
try to clamp down on reagent mixe stuff clashes, 
Fixes error with minus catalogue, and removes unneeded error log on net identity set, 
fixes pipe not rendering properly, fixes health error

### DIFF
--- a/UnityProject/Assets/Mirror/Runtime/NetworkIdentity.cs
+++ b/UnityProject/Assets/Mirror/Runtime/NetworkIdentity.cs
@@ -351,11 +351,11 @@ namespace Mirror
                 }
 
                 // old not empty
-                if (!string.IsNullOrEmpty(oldAssetIdSrting))
-                {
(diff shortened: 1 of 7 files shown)


UnityProject/Assets/Mirror/Runtime/NetworkIdentity.cs
```diff
@@ -351,11 +351,11 @@ internal set
...
-                if (!string.IsNullOrEmpty(oldAssetIdSrting))
-                {
-                    Debug.LogError($"Can not Set AssetId on NetworkIdentity '{name}' because it already had an assetId, current assetId '{oldAssetIdSrting}', attempted new assetId '{newAssetIdString}'");
-                    return;
-                }
+                // if (!string.IsNullOrEmpty(oldAssetIdSrting))
+                // {
+                    // Debug.LogError($"Can not Set AssetId on NetworkIdentity '{name}' because it already had an assetId, current assetId '{oldAssetIdSrting}', attempted new assetId '{newAssetIdString}'");
+                    // return;
+                // }
```
corp-0: don't comment code, just remove it. We can retrieve it with git if we need to.

Bod9001: ok

Issue #6270, opened 2021-03-19
Combat Dismemberment and Disemboweling B:150


PerfectTangent:
## Description:

**This is a health-related bounty.** This bounty calls for the implementation of limb and organ loss due to weapon attacks and explosions.

In ss13, victims of particularly vicious attacks may lose their head, legs or arms. This can result from a few different situations including:

- explosions.
- attacks from very powerful weapons.
- attacks from particular or specialized weapons (the plasma cutter being the most well known one).
- extremely forceful impacts such as players gettting launched by catastrophic decompression or getting too close to the singularity
- airlocks slamming with safeties disengaged.

Rarely, organs can be directly ejected or cut out of a victim's chest in the middle of combat from certain attacks (attacks from the energy sword and the power fist being the most common cases). Organs and limbs can also rarely be completely pulverized and destroyed.

Obviously, any of these things happening to a crew member is generally bad news. If a limb is lost in such a way, it will usually cause serious amount of bleeding. Losing an organ vital to survival is also very bad, especially if it is the heart or brain. Patients missing a head will die in most cases as their body lacks its brain.

**Armor also has an effect on how likely any of the methods of dismemberment or disembowelment above are to happen. Highly protective armor should reduce the risk of such things occurring.**

## What is required to fulfill the bounty:

- [x] Add dismemberment due to explosions, violent weapon attacks, extreme blunt force, and blast doors/airlocks, which can be counteracted by high tier armors.
- [x] Add disemboweling due to explosions, extreme blunt force, and powerful attacks directed at the chest, head, or groin.
- [x] Add pulverization of organs and limbs.

With the new health system in game, the majority of the effects of organ and limb loss have been addressed. However, currently, mobs do not die when they have lost their head or brain. That is an important feature specifically for the topic this bounty covers, and should be included with this.

- [x] Death should occur instantly if a target of an attack is decapitated or loses their brain.

## Suggested implementation

- Add a new field to ItemAttributesV2.cs for whether the attacks of the weapon causes dismembering when used to attack and how likely the weapon is to dismember.
- Change the damage method parameters to send this ``canDismember`` field and also the ``dismemberChance`` of the attack, so different weapons/objects can have a different chances of dismembering.
- After armor reduction, check the damage and compare it to a damage threshold set in the body part. If the reduced damage passed the threshold, roll the dismembering chance.

PerfectTangent:
price reevaluated to 60.00 USD

MaxIsJoe:
[...]

PR #6625, opened 2021-05-27
Adds in bleeding for dismembered parts

MaxIsJoe:
### Purpose
Part of #6270

Characters will now bleed when they lose a limb, re-adjusted dismemberment chance values for each race so some races would be more fragile than others and added some missing documentation.

### Blood Loss
Players who lose a limb have about 5-6 minutes to go to a doctor before they say goodbye to life due to blood loss, this time gets shorter with each dismembered limb and with the addition of Disemboweling in the next few PRs, this can drastically change to seconds as your body turns into a fountain of blood.
Right now all races share the same blood loss values, but later each race will lose more or less blood depending on how fragile they are.

### Lizard stronk, Moth weak *spit*
Moths have the highest chance of dismemberment on the highest severe damage levels, nearly 90%!
Lizards on the other hand have a 70% chance to lose their limb on highest damage levels.
Humans and Catpeople almost share the same values, but the difference between them is that catpeople are more likely to lose their heads and torso compared to other humans. (87% for felines, 85% for humans)

If you don't want to lose your limb, start wearing armor!
Though armor values are still lacking on everything soooo, wait until next PRs.

### Notes:
Currently to stop bleeding I've made Brusie Packs have the ability to stop bleeding temporarily until we discuss how we should approach limb loss bleed healing.

### Changelog:
CL: [New] - Characters will now start bleeding when losing a limb.
CL: [Improvement] - Races no longer share the same dismemberment chance values.
CL: [Fix] - Fixed an error that would prevent damage to spread to all limbs by bombs.

### DIFF
--- a/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
@@ -352,7 +352,7 @@ namespace HealthV2
 			//This is so you can still hit for example the Second Head of a double-headed thing, can be changed if we find a better solution for aiming at Specific body parts
 			if (damageSplit || attackType == AttackType.Bomb || attackType == AttackType.Fire || attackType == AttackType.Rad)
 			{
-				foreach (var ContainsLimb in ContainsLimbs)
+				foreach (var ContainsLimb in ContainsLimbs.ToArray())
 				{
 					ContainsLimb.TakeDamage(
 						damagedBy,
--- a/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
@@ -352,9 +352,8 @@ namespace HealthV2
 			//This is so you can still hit for example the Second Head of a double-headed thing, can be changed if we find a better solution for aiming at Specific body parts
 			if (damageSplit || attackType == AttackType.Bomb || attackType == AttackType.Fire || attackType == AttackType.Rad)
 			{
-				//We don't use for loops to avoid errors when the list gets modifed.
-				int limbCount = 0;
-				while (ContainsLimbs.Count > limbCount)
+				//We don't use foreach to avoid errors when the list gets modifed.
+				for(int limbCount = 0; ContainsLimbs.Count > limbCount; limbCount++)
 				{
 					ContainsLimbs[limbCount].TakeDamage(damagedBy,
 						damage / ContainsLimbs.Count,
--- a/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
@@ -327,11 +327,6 @@ namespace HealthV2
 		public virtual void SubBodyPartAdded(BodyPart implant)
 		{
 			SetupSpritesNID(implant);
-			if(ContainsLimbs.Count != 0)
-			{
-				Logger.Log($"[BodyPartContainer/{name} - Limbs detected! No longer bleeding..]");
-				IsBleeding = false;
-			}
 		}
 
 		/// <summary>
--- a/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
@@ -352,16 +352,18 @@ namespace HealthV2
 			//This is so you can still hit for example the Second Head of a double-headed thing, can be changed if we find a better solution for aiming at Specific body parts
 			if (damageSplit || attackType == AttackType.Bomb || attackType == AttackType.Fire || attackType == AttackType.Rad)
 			{
-				foreach (var ContainsLimb in ContainsLimbs.ToArray())
+				//We don't use for loops to avoid errors when the list gets modifed.
+				int limbCount = 0;
+				while (ContainsLimbs.Count > limbCount)
 				{
-					ContainsLimb.TakeDamage(
-						damagedBy,
+					ContainsLimbs[limbCount].TakeDamage(damagedBy,
 						damage / ContainsLimbs.Count,
 						attackType,
 						damageType,
 						damageSplit,
 						armorPenetration: armorPenetration
-					);
+						);
+					limbCount++;
 				}
 			}
 			else
--- a/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
@@ -362,7 +362,6 @@ namespace HealthV2
 						damageSplit,
 						armorPenetration: armorPenetration
 						);
-					limbCount++;
 				}
 			}
 			else
--- a/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
@@ -302,6 +302,11 @@ namespace HealthV2
 			{
 				prop.ImplantPeriodicUpdate();
 			}
+			if(IsBleeding)
+			{
+				float TotalBloodLossFromLostLimbs = LimbLossBleedingValue;
+				healthMaster.CirculatorySystem.Bleed(TotalBloodLossFromLostLimbs);
+			}
 		}
 
 		/// <summary>
--- a/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
@@ -316,11 +316,7 @@ namespace HealthV2
 		public virtual void SubBodyPartRemoved(BodyPart implant)
 		{
 			RemoveSpritesNID(implant);
-			if(ContainsLimbs.Count == 0)
-			{
-				Logger.Log($"[BodyPartContainer/{name} - No Limbs detected! Bleeding..]");
-				IsBleeding = true;
-			}
+			IsBleeding = true;
 		}
 
 		/// <summary>
--- a/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
@@ -353,7 +353,7 @@ namespace HealthV2
 			if (damageSplit || attackType == AttackType.Bomb || attackType == AttackType.Fire || attackType == AttackType.Rad)
 			{
 				//We don't use foreach to avoid errors when the list gets modifed.
-				for(int limbCount = 0; ContainsLimbs.Count > limbCount; limbCount++)
+				for(int limbCount = ContainsLimbs.Count - 1; limbCount >= 0; limbCount--)
 				{
 					ContainsLimbs[limbCount].TakeDamage(damagedBy,
 						damage / ContainsLimbs.Count,
--- a/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
@@ -304,8 +304,7 @@ namespace HealthV2
 			}
 			if(IsBleeding)
 			{
-				float TotalBloodLossFromLostLimbs = LimbLossBleedingValue;
-				healthMaster.CirculatorySystem.Bleed(TotalBloodLossFromLostLimbs);
+				healthMaster.CirculatorySystem.Bleed(LimbLossBleedingValue);
 			}
 		}
 
--- a/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
@@ -73,7 +73,7 @@ namespace HealthV2
 		/// How much blood does the body lose when there is lost limbs in this container?
 		/// </summary>
 		[SerializeField, Tooltip("How much blood does the body lose when there is lost limbs in this container?")]
-		private float LimbLossBleedingValue = 35f;
+		private float limbLossBleedingValue = 35f;
 
 		public RootBodyPartController RootBodyPartController;
 
@@ -304,7 +304,7 @@ namespace HealthV2
 			}
 			if(IsBleeding)
 			{
-				healthMaster.CirculatorySystem.Bleed(LimbLossBleedingValue);
+				healthMaster.CirculatorySystem.Bleed(limbLossBleedingValue);
 			}
 		}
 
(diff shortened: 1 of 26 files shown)


UnityProject/Assets/Scripts/HealthV2/Living/CirculatorySystem/CirculatorySystemBase.cs
```diff
@@ -48,5 +48,20 @@ public void AddFreshBlood(ReagentMix bloodPool, float amount)
 			bloodToAdd.Add(CirculatedReagent, bloodType.GetGasCapacity(bloodToAdd));
 			bloodPool.Add(bloodToAdd);
 		}
+
+		public void Bleed(float amount)
+		{
+			var bloodToRemove = new ReagentMix(BloodType, amount);
+			bloodToRemove.Remove(CirculatedReagent, bloodType.GetGasCapacity(bloodToRemove));
+			ReadyBloodPool.Subtract(bloodToRemove);
```
Bod9001: you don't need to subtract the blood component, you can just remove chemicalswhatever u amount of chemicals

UnityProject/Assets/Scripts/Items/Implants/Organs/Heart.cs
```diff
@@ -96,13 +96,28 @@ public void DoHeartBeat(LivingHealthMasterBase healthMaster)
...
+		LivingHealthMasterBase HealthBase = RelatedPart.HealthMaster;
 		if (circulatorySystem)
 		{
 			//circulatorySystem.HeartBeat(heartStrength * TotalModified);
 			//TODOH Balance all this stuff, Currently takes an eternity to suffocate
 			// Logger.Log("heart Available  " + ReadyBloodPool);
 			// Logger.Log("heart pumpedReagent " + pumpedReagent);
 
+			//Find out if we have lost limbs and how much blood we'll lose.
+			float TotalBloodLossFromLostLimbs = 0f;
```
Bod9001: I feel like this would be better if each body part bled individually instead of the heart bleeding?

MaxIsJoe: The heart is not bleeding, the heart is simply a tool that pushes blood out of the body with each heartbeat.

corp-0: certainly. ``Bleeding`` is a state of body parts that makes them leak blood each tick. The state should be handled by the body part itself.

UnityProject/Assets/Scripts/HealthV2/Living/CirculatorySystem/CirculatorySystemBase.cs
MaxIsJoe: I don't understand.

UnityProject/Assets/Scripts/Items/Implants/Organs/Heart.cs
MaxIsJoe: done.

UnityProject/Assets/Scripts/HealthV2/Living/CirculatorySystem/CirculatorySystemBase.cs
MaxIsJoe: done.

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
```diff
@@ -302,6 +302,11 @@ public virtual void ImplantPeriodicUpdate()
 			{
 				prop.ImplantPeriodicUpdate();
 			}
+			if(IsBleeding)
+			{
+				float TotalBloodLossFromLostLimbs = LimbLossBleedingValue;
```
corp-0: what's  the need for this assignation if you're using it right in the next line without any alterations from ``LimbLossBleedingValue``?

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
```diff
@@ -67,6 +67,14 @@ public class RootBodyPartContainer : MonoBehaviour, IBodyPartDropDownOrgans, ISe
 		[Tooltip("Individual limbs contained within this")]
 		public List<BodyPart> ContainsLimbs = new List<BodyPart>();
 
+		[HideInInspector] public bool IsBleeding = false;
+		
+		/// <summary>
+		/// How much blood does the body lose when there is lost limbs in this container?
+		/// </summary>
+		[Tooltip("How much blood does the body lose when there is lost limbs in this container?")]
+		public float LimbLossBleedingValue = 35f;
```
corp-0: why public?

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
MaxIsJoe: good point. Done.

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
MaxIsJoe: This was public for a reason. But since it's no longer being accessed due to a change requested by you and Bod it's no longer needed to be public. Fixed.

MaxIsJoe:
Alright I think the code quality part is settled, time to merge.

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
```diff
@@ -304,6 +316,11 @@ public virtual void ImplantPeriodicUpdate()
 		public virtual void SubBodyPartRemoved(BodyPart implant)
 		{
 			RemoveSpritesNID(implant);
+			if(ContainsLimbs.Count == 0)
```
Bod9001: wouldn't this have issues if you have multiple right arms for example?

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
corp-0: sure. Mind the access level of all code you write/modify so the intetion on how to use that code is crystal clear. Give a look to the rest of your code and qustion if they actually need to be public, being it fields or methods.

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
MaxIsJoe: whoops, forget to change that. on my way to fix it.

MaxIsJoe: fixed.

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
```diff
@@ -352,7 +352,7 @@ public virtual void SubBodyPartAdded(BodyPart implant)
 			//This is so you can still hit for example the Second Head of a double-headed thing, can be changed if we find a better solution for aiming at Specific body parts
 			if (damageSplit || attackType == AttackType.Bomb || attackType == AttackType.Fire || attackType == AttackType.Rad)
 			{
-				foreach (var ContainsLimb in ContainsLimbs)
+				foreach (var ContainsLimb in ContainsLimbs.ToArray())
```
corp-0: what type were these enumrables before the casting to Array/List?

MaxIsJoe: I have no idea but all of these caused bombs to break when exploding near players so I had to do this to fix this issue.
Now they explode normally again without breaking.

pyrosshade: Line 68
`public List<BodyPart> ContainsLimbs = new List<BodyPart>();`

corp-0: foreach should work fine with either arrays or lists, any enumerable actually. I'm trying to think how this change fixed the problem or if it in fact did anything.

MaxIsJoe: it did do something. It works now and doesn't throw errors so it fixed it. Test it yourself.
https://stackoverflow.com/questions/604831/collection-was-modified-enumeration-operation-may-not-execute

Bod9001: the issue is that, the bomb would cause a limb to get removed but it's applying the damage still so it breaks the list it's iterating over

MaxIsJoe: Yes, and this is the fix for it.

corp-0: so the concern we have now is generating a considerable amount of garbage when you create a new list every time someone takes damage. We're discussing solutions at the moment.

Bod9001: the solution I would choose would have a list on the health master base that you had the limb/organ that you want to be cut off , then it loops through that once it's finished taking damage

MaxIsJoe: Yes I know I'm late on this, but you're wrong about how this works Bod.
We're trying to damage the limb, not cut it off in that for loop.

Creating another list to track a separate list is terrible as well, I've been looking for a solution for the past few hours and all what stackoverflow and the internet has told me is that the only way to avoid this error is to just use `ToArray()`
There is no way to escape this I believe.

corp-0: creating a separate list once vs creating a list every time someone takes damage. That's the thing.

MaxIsJoe: Except that this list will be useless 99% of the time and will be sitting there in memory for no reason + it has a very specific case that no other function or object will use so this is basically the same as the problem of creating a public variable that only one function uses it and has no place anywhere else.

Don't even get me started on having to maintain it and make sure that it's synced with the original list so the game doesn't throw a fit when it detects that a bodypart is no longer inside that list.

corp-0: so instead of making the list one time you think it is better to create new lists every time someone takes damage and when it is no longer used, make the GC dispose it at the cost of performance?

MaxIsJoe: No. That's why I'm looking for a solution right now that's actually good for performance.
Creating a list that's going to be hell to maintain and synced with the original list and is taking up memory all game is not good, nor is using this method right now.

MaxIsJoe: FOUND THE FIX
HAHA

corp-0:
Tested on server - client environment. Server spawned a synd bomb and client exploded it on their feet. The enumeration error is gone, but I did experience a freeze before seeing the result of the explosion.

corp-0:
can someone else please try reproduce my experience?

Bod9001:
with a quick test was good for me

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
Lizurt: Do I understand correctly that the problem was that when an explosion damage the limbs in `foreach()`, some of them can be removed from the list(limbs fly off), so due to the severing of the limbs, the iterations are interrupted and the damage stops being dealt?

I know I'm late with this question, but I would still like to know the answer to it, as I have some suspicions about the solution to this problem in recent commits.

MaxIsJoe: You're correct and yeah everyone is a lil sus about my solution because of index changes.
I'm trying other solutions now.

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartContainer.cs
```diff
@@ -339,16 +352,16 @@ public virtual void SubBodyPartAdded(BodyPart implant)
 			//This is so you can still hit for example the Second Head of a double-headed thing, can be changed if we find a better solution for aiming at Specific body parts
 			if (damageSplit || attackType == AttackType.Bomb || attackType == AttackType.Fire || attackType == AttackType.Rad)
 			{
-				foreach (var ContainsLimb in ContainsLimbs)
+				//We don't use foreach to avoid errors when the list gets modifed.
+				for(int limbCount = 0; ContainsLimbs.Count > limbCount; limbCount++)
```
Lizurt: Some items in the list may be skipped under the current implementation. 

If an item is removed from the list, `limbCount` already points to the next item in the list. But `limbCount` is also increased at the end of the loop. So we skip the item in the list.

The simplest solution is to go through the list from the end, for example, like this: `for (int limbCount = ContainsLimbs.Count - 1; limbsCount >= 0; limbCount--)`.

By the way, it is better to also make this list a LinkedList. Removing items from a middle of a regular list can make the garbage collector suffer. In this case, it is better to use your `foreach ()` implementation via `Enumerator`

MaxIsJoe: I've added your solution of looping the thing backwards, LinkedList doesn't seem to be working as intended so I'm going to leave it as that for now.

corp-0: > foreach () implementation via Enumerator

How would that look @Lizurt ?

Lizurt: > How would that look @Lizurt ?

I worked with `Enumerator` (Iterator) only in Java, so it looks like we can't use the `Enumerator` here, since it goes through the values, not the nodes. You will have to go down even one more level and use your own abstract `Enumerator`. Something like(a bit pseudocoded):
```
LinkedListNode<Limb> limbNode = list.First;
LinkedListNode<Limb> nextLimbNode;
while (limbNode != null)
{
    nextLimbNode = limbNode.Next;
    limbNode.Value.TakeDamage(args);
    limbNode = nextLimbNode;
}
```

You will also have to delete a limb in the list by node and not by its value, unless you want the complexity O(n) instead of O(1) when removing.

Lizurt: Remember that this is just a recommendation. Since the list is small and deletions are rare, it's not so critical to leave it as a simple list imo.

jamessimo: I have tested it and it works from a gameplay perspective, cant speak towards the looping method.

jamessimo:
Tested and work for me, cant speak towards the loop methods however.

corp-0:
Are you conflicting with your other Pr that got merged?

MaxIsJoe:
> Are you conflicting with your other Pr that got merged?

I excepted for this PR and disembowling to conflict at some point when one of them gets merged, but body tint? that's an odd on.

MaxIsJoe:
Merge conflicts solved.

PR #6633, opened 2021-05-28
makes it so on to spawn, register tiles remove themselves from locations hopefully fixing the null reference bug?

Bod9001:
don't worry the changes have already been merged showing them for some reason though they've already been merged,
also improved explosion code a bit

fixes NRE with table interactions while it's under construction if you don't have anything in your hand
so, mirror with  objects that were in the scene when it was safe when server requests to destroy them it just hides them instead resulting in inconsistent destruction path , (if it were responding then it gets destroyed normally if it was there already it would gethidden) so, I've made it so it destroys even if it was a scene object and tested with a few round restarts with editor hosting and build as client

also removes the annoying " scene needing to be received" thing that stops builds from proceeding,
 for testing purposes becauseit's particularly annoying when it gets half way through a build and stops

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterTile.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterTile.cs
@@ -305,6 +305,12 @@ public class RegisterTile : NetworkBehaviour, IServerDespawn
 
 	public virtual void OnDespawnServer(DespawnInfo info)
 	{
+		if (objectLayer)
+		{
+			objectLayer.ServerObjects.Remove(LocalPositionServer, this);
+			objectLayer.ClientObjects.Remove(LocalPositionClient, this);
+		}
+
 		//cancel all relationships
 		if (sameMatrixRelationships != null)
 		{
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterTile.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterTile.cs
@@ -398,8 +398,8 @@ public class RegisterTile : NetworkBehaviour, IServerDespawn
 		objectLayer?.ClientObjects.Remove(LocalPositionClient, this);
 		objectLayer?.ServerObjects.Remove(LocalPositionServer, this);
 
-		LocalPositionClient = Vector3Int.zero;
-		LocalPositionServer = Vector3Int.zero;
+		LocalPositionClient = TransformState.HiddenPos;
+		LocalPositionServer = TransformState.HiddenPos;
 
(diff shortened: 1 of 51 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterTile.cs
```diff
@@ -305,6 +305,12 @@ public void OnDestroy()
 
 	public virtual void OnDespawnServer(DespawnInfo info)
 	{
+		if (objectLayer)
+		{
+			objectLayer.ServerObjects.Remove(LocalPositionServer, this);
+			objectLayer.ClientObjects.Remove(LocalPositionClient, this);
+		}
+
```
ThatDan123, line 313: Only called on server, wont affect clients

ThatDan123:
Also are all those files in this PR wanted?

Bod9001:
yeah, should make it so walls absorbed more explosion damage

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterTile.cs
Bod9001: will look into more, dam I really hate that need to save scenes cancel build thing

Bod9001: isn't the problem server side?

UnityProject/Assets/Scripts/Tilemaps/Behaviours/TilemapDamage.cs
```diff
@@ -74,28 +74,44 @@ private float DealDamageAt(float damage, AttackType attackType, Vector3Int cellP
...
-			if(damage >= 1)
+			if(damageTaken >= 1)
 				SoundManager.PlayNetworkedAtPos(basicTile.SoundOnHit, worldPosition);
 		}
 
 		var totalDamageTaken = data.GetTileDamage(Layer.LayerType);
 
 		if (totalDamageTaken >= basicTile.MaxHealth)
 		{
+			float ExcessEnergy = basicTile.Armor.GetForce( totalDamageTaken - basicTile.MaxHealth, attackType);
```
corp-0: camelCase

corp-0:
I have no idea what this PR is for. Is it rebalancing explosions? Is it fixing a bug on RegisterTile like the tittle suggest? What bug is it fixing?

corp-0:
huh, so I guess I should read PR descriptions before reading code changes. Sorry for that, Bod.

corp-0:
Adding a DNM since Bod said he will look into it further.

Bod9001:
Merging so will Be in next build

PR #6634, opened 2021-05-28
More  locks!!

Bod9001:
if it still clashes after this then, time to redesign system

### DIFF
--- a/UnityProject/Assets/Scripts/Player/PlayerSprites.cs
+++ b/UnityProject/Assets/Scripts/Player/PlayerSprites.cs
@@ -405,10 +405,6 @@ public class PlayerSprites : MonoBehaviour
 		}
 	}
 
-	private void SetupSprite(SpriteDataSO spriteSheetAndData, string clothesDictKey, Color? color = null)
-	{
-		clothes[clothesDictKey].spriteHandler.SetSpriteSO(spriteSheetAndData, color.GetValueOrDefault(Color.white));
-	}
 
 	private void OnClientFireStacksChange(float newStacks)
 	{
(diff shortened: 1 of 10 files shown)


UnityProject/Assets/Scripts/Player/PlayerSprites.cs
```diff
@@ -633,13 +629,13 @@ private void ValidateHideFlags()
...
 		// Need to check all flags with their gameobject names...
 		// TODO: it should be done much easier
-		ValidateHideFlag(ClothingHideFlags.HIDE_GLOVES, "hands");
-		ValidateHideFlag(ClothingHideFlags.HIDE_JUMPSUIT, "uniform");
-		ValidateHideFlag(ClothingHideFlags.HIDE_SHOES, "feet");
-		ValidateHideFlag(ClothingHideFlags.HIDE_MASK, "mask");
-		ValidateHideFlag(ClothingHideFlags.HIDE_EARS, "ear");
-		ValidateHideFlag(ClothingHideFlags.HIDE_EYES, "eyes");
-		ValidateHideFlag(ClothingHideFlags.HIDE_NECK, "neck");
+		// ValidateHideFlag(ClothingHideFlags.HIDE_GLOVES, "hands");
```
corp-0, line 632: nani?

Bod9001: I can't see anything that uses them, from a quick look through it looks like it just goes through all the clothing and then if any have the flag then it hides it, so the names are used nowhere?

corp-0:
test merging. As Bod said, if this doesn't work we better find another way to do threaded reactions.

PR #6664, opened 2021-05-31
sweet extensions go.NetId() fix

pyrosshade:
go.NetId() can now retreive matrix netIdentities again
~~Removes unreachable return statement~~


UnityProject/Assets/Scripts/Player/PlayerSync.cs
```diff
@@ -473,10 +473,6 @@ private void UpdateMe()
 
 	private void Synchronize()
 	{
-		if (isLocalPlayer && GameData.IsHeadlessServer)
```
Bod9001: Technically this can trigger on the headless server,m idk how it handles its own playerr

pyrosshade: I'm under the impression that headless is server only? There shouldn't be a localplayer for something that is not a client.

corp-0:
changelog is meant for the final user, both of these entries doesn't mean much for them.

UnityProject/Assets/Scripts/Util/SweetExtensions.cs
```diff
@@ -82,7 +82,11 @@ public static T GetRandom<T>(this List<T> list)
 
 	public static uint NetId(this GameObject go)
 	{
-		return go ? go.GetComponent<NetworkIdentity>().netId : global::NetId.Invalid; //maxValue is invalid (see NetId.cs)
+		return go == null
```
corp-0: I like ternaries but in this case the nested ternaries might be hard to read for people with less experience. I'd suggest using a simple if statement in this case.

UnityProject/Assets/Scripts/Util/SweetExtensions.cs
```diff
@@ -82,7 +82,11 @@ public static T GetRandom<T>(this List<T> list)
...
+			? global::NetId.Invalid
+			: go.GetComponent<Matrix>()
+				? go.GetComponentInChildren<NetworkIdentity>().netId
```
ThatDan123: Splitting it out into If statements would be better like Gilles said and then do TryGetComponent<Matrix>(out var matrix)

and then returning matrix.NetworkedMatrix.MatrixSync.netId

UnityProject/Assets/Scripts/Player/PlayerSync.cs
ThatDan123: Yeah, i dont think this check was ever true. So removing it wouldnt change anything. But just making it a Headless check (if that was the original intension) could break stuff, so would need testing.

ThatDan123: image

i donno maybe we should keep it in. Might be some weird interaction with headless

pyrosshade:
> 
> 
> changelog is meant for the final user, both of these entries doesn't mean much for them.

gotcha, changed.

PR #6610, opened 2021-05-23
Adds in dismembering players + No more alive headless players

MaxIsJoe:
### Purpose
Part of #6270

Adds in the ability to dismember players using weapons, bomb explosions and more while giving armor a way to soften the chance of dismembering a player. The idea of my approach is that the system is flexible enough to allow for anything to dismember the player as long as the damage is great enough for it to dismember them.
Also kills off the player when they're missing a brain, head or torso.

### Notes:
ticks one box off #6195 

catpeople, moths and lizards currently share the same values humans. This will be changed later after some tests on staging.
Only the most common and popular armor have armor values right now, the rest don't until we get some tests going.

### What's left?
In the next few PRs i'll add in the ability for players to bleed to death after they have lose a limb + tweak and add more armor values to balance this new addition to the game + the ability for limbs to simply alt F4 when then they receive enough damage for them to completely get wiped off the game.

### Changelog:
CL: [New] - Players can now be dismembered by other players and explosives.
CL: [Fix] - Players can no longer walk around headless or without a brain or torso and stay alive.

PR #6670, opened 2021-06-02
Changing clothing armor to old coverage and values.

PetMudstone:
### Purpose
- Undoes the various changes to the armor stats and coverage of various clothing items made in #6319 to the old values from before that PR. **Does not change the underlying code.**
  - Attempts to maintain most of the dismemberment protection values made in #6610, but some values were altered based on inheritance of prefab variants.


PetMudstone:
This is a draft due to the feature freeze.

PerfectTangent:
if this is just prefab changes, then it is exempt from the feature freeze. That just covers code additions.

Issue #6613, opened 2021-05-25
Unpowered Door Visuals


ThatDan123:
Hard to tell if a door is unpowered can be fixed by:

-Having message play when trying to open unpowered doors eg. "The door control panel lights are off" or something like that. Or literally "The door is unpowered".

-Have powered doors have a animated sprite e.g blinking light on control panel of door, then the upowered version would have that off / grey coloured instead

Fogapod:
I don't like the idea of blinking lights.

Logical arguments:
- unpowered means no power. does door have internal battery for these lights?

Gameplay arguments:
- good for traitors for doing shady traitor things. you can stay uncaught until someone tries to interact with the door
- bad for engineers because it's annoying to fix? currently true, but could be easily solved by some engineering HUD for inspecting wiring and electronics. i'm pretty sure this is planned

Maybe blinking lights could be used if the wiring is messed up, but the door is still powered

ThatDan123:
> * unpowered means no power. does door have internal battery for these lights?

No power = no blinking lights

PR #6673, opened 2021-06-03
Some Client Update Changes

ThatDan123:
-Stops client subscribing to updates it doesnt need.

-Adds door unpowered message

Partially implements #6613


ThatDan123:
Codacy just complaining about namespace for playerscript

corp-0:
it looks good. Maybe @Bod9001 should take a look?

PR #6758, opened 2021-06-06
Fixes Wrench Harm Behaviour

Akbadain29:
### Purpose
Changes Table SOs to use TableDeconstruction MonoBehaviour for Tile Interaction to isolate behaviour from other construction methods allowing for help intent with wrench to be used.

### Changelog:
**[Fix]** Gives help intent wrench functionality back for Tile Interaction DeconstructWhenItemUsed.

### DIFF
--- a/UnityProject/Assets/ScriptableObjects/Interaction/TileInteraction/TableDeconstruction.asset
+++ b/UnityProject/Assets/ScriptableObjects/Interaction/TileInteraction/TableDeconstruction.asset
@@ -1,0 +1,21 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!114 &11400000
+MonoBehaviour:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 0}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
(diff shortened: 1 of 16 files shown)


Akbadain29:
Codacy doesn't like it how I've created another TileInteraction similar to the existing one. Shouldn't it be fine though?

pyrosshade:
to quote codacy
>Move 'TableDeconstruction' into a named namespace.

Akbadain29:
Should it be namespace Objects.Table, even though it's technically a Tile? I'll put it in that and see if it makes it better.

corp-0:
how do you leave the wrench on the table now?

Akbadain29:
Same as before, help intent and clicking on the table with the wrench in your hand. The regular DeconstructWhenItemUsed doesn’t have that check anymore. The new one does.

corp-0:
Wait so you leave the wrench on the table clicking it with help intent and also can deconstruct it clicking it with help intent?

Akbadain29:
No, you still can only deconstruct it with harm intent. Help intent lets you place it on the table.

AutumnNova:
I would assume this is related to fixing the inability to un-wrench girders and such unless you use harm intent

Akbadain29:
Yes that’s what the change should do.

Akbadain29:
Except I thought it was to do with windows not girders, but that makes more sense. I haven’t tested what happens with girders though. I should probably do that.

Akbadain29:
Yeah I can unwrench and rewrench using help or harm intent for these griders.

PR #6775, opened 2021-06-10
Fixes admin windows not disabling character input

### LINKED ISSUE #5303: Keystrokes while typing in admin windows still register in terms of player movement, bring up chat, and more. B:25
## Description
Right now when you type in admin windows, such as player chat, your keystrokes will sometimes still do things like move your character, drop things, or even open normal chat (which is **especially** annoying as everything you type afterwards will be in normal chat and not admin to player chat or whatever). This shouldn't be happening.

The key word is sometimes. Pressing Y will sometimes do nothing while typing in admin windows, while other times it will open up normal chat and do as described above.

### Steps to Reproduce
1.  Open up an admin window, such as admin to player chat.
2. Start typing in any text fields.
3.  Press random keys, see if they do anything. With some (mis)fortune the WASD keys will move your character around and the T, Y, and U keys will open up chat for example.

Aranclanos:
### Purpose
Fixes admin windows sometimes not disabling character keyboard input when using text fields.
Fixes #5303

### Notes:
The search input field was enabling character input on each frame if that specific input field wasn't being used. The search box already uses our InputFieldFocus so that's handled already.

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Systems/MainHUD/UI Bottom/AdminSearchBar.cs
+++ b/UnityProject/Assets/Scripts/UI/Systems/MainHUD/UI Bottom/AdminSearchBar.cs
@@ -12,29 +12,6 @@ public class AdminSearchBar : MonoBehaviour
 		Searchtext = GetComponent<InputField>();
 	}
 
-	void Update()
-	{
-		if (Searchtext.isFocused)
-		{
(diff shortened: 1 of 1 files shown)


Bod9001:
note, in the VV rework I've added a component for TMP input fields focus

PR #6776, opened 2021-06-10
disables skin tone validation

MaxIsJoe:
quick PR to fix an issue where people keep getting kicked for playing dark skinned characters.
This comments out code until we figure out why it isn't working then fix it.


waffielz:
jim crow would dislike this

PR #6788, opened 2021-06-12
VV Improvements

Bod9001:
made it so editor could load into any scene that was previously loaded

CL: **[New]** VV has hierarchy now

CL: **[Improvement]**  VV Is now way better to use


corp-0:
solve the codacy issues but the ones about PascalCase with the GUI word and we're good.

Bod9001:
still fixing a few things

Bod9001:
is good now

waffielz:
vv gaming

PR #6791, opened 2021-06-13
Fixes uplink bought items being at hiddenpos if not assigned an inventory slot

### LINKED ISSUE #6761: Buying uplink items without any inventory slots to hold them in causes them to not appear.
### Environment
- Game Version: Staging 21060607
- OS Version: Windows 10

When buying uplink items, if you have no empty inventory slots for the item you're buying to fit in then it will not appear at all. This means that buying items without space for them will waste TC. In the case of items not having space for them to fit in your inventory they should just drop on the ground.

### Steps to reproduce
1. Be a traitor or nukie.
2. Access the uplink on your PDA, go to "Stealthy Weapons".
3. Buy as many energy daggers as you can. Notice that after energy daggers fill up your hands and pockets, they won't appear anywhere when you buy them anymore.

pyrosshade:
Fixes #6761

CL: **[Fix]** Items bought in PDA uplink that can't find an ideal slot to go to will now spawn on the ground.

### DIFF
--- a/UnityProject/Assets/Scripts/Items/PDA/PDALogic.cs
+++ b/UnityProject/Assets/Scripts/Items/PDA/PDALogic.cs
@@ -436,7 +436,7 @@ namespace Items.PDA
 
 			if (cost > UplinkTC) return;
 
-			var result = Spawn.ServerPrefab(objectRequested);
+			var result = Spawn.ServerPrefab(objectRequested,GetComponent<Pickupable>().ItemSlot.Player.WorldPosition);
 			if (result.Successful)
 			{
 				UplinkTC -= cost;


UnityProject/Assets/Scripts/Items/PDA/PDALogic.cs
```diff
@@ -436,7 +436,7 @@ public void SpawnUplinkItem(GameObject objectRequested, int cost)
 
 			if (cost > UplinkTC) return;
 
-			var result = Spawn.ServerPrefab(objectRequested);
```
Bod9001, line 439: humm, was going to suggest replacement strategy but then realise it doesn't return any slots, so this is the best thing for now

PR #6794, opened 2021-06-14
Fixes pickaxes and other interactions on tiles not outside the station

Aranclanos:
### Purpose
Fixes pickaxes and other interactions on tiles not working in non-station matrixes.
Adds a MouseToWorld() function to MouseUtils that returns the world position of the mouse location making sure that Z is set to 0 to be able to use it on MetaTileMap systems.
Replaces all parts in the project that manually set their Z mousepos to 0 to make them use this new function.

### Notes
When clicking on an asteroid tile, the position used had the camera Z's -1. Due to this, the matrix returned was the station as it's the default when nothing can be found. When checking if the station's matrix has a mineable wall at the position clicked, there was nothing and it failed.


corp-0:
should this PR have closed an issue? I could bet it was linked to a particular issue.

Aranclanos:
there's an issue about not being able to mine *diagonally* but I wasnt able to reproduce it

PR #6806, opened 2021-06-16
Undo Checks

ThatDan123:
-Removes the server checks for the sub system managers


Bod9001:
can you do application is Editor?

Issue #6784, opened 2021-06-11
Non-mainstation shuttles are not showing proper coordinates


pyrosshade:
![image](https://user-images.githubusercontent.com/14188696/121740421-0f5f2e80-cacb-11eb-96d3-009f25d2a990.png)

note in the bottom left the coords from alt-clicking, then note the coords in the piloting console

Bod9001:
confirmed 100, 100 and editor for some reason

PerfectTangent:
https://user-images.githubusercontent.com/68730437/121898121-0f407800-ccf1-11eb-9dba-f47c5446b2e9.mp4

this doesn't appear to be only affecting the ancient station shuttle

ThatDan123:
Fixed by #6814

Issue #6779, opened 2021-06-11
VV fails to activate on matrixes


Pockets-byte:
Linux-staging

On staging if you try to activate VV on matrixes, it completely fails, leaving you with what was last activated
this is kinda a pain for testing various things, as well as controlling speeds on shuttles and other various things

Pockets-byte:
If I click into the shuttle matrix move from a shuttle computer, it crashes my VV

Bod9001:
"On staging if you try to activate VV on matrixes, it completely fails, leaving you with what was last activated
this is kinda a pain for testing various things, as well as controlling speeds on shuttles and other various things"

do you mean you verbals on matrices. The last option in the admin right click?

Bod9001:
couldn't replicate the second one on the editor,

Pockets-byte:
uhh?
the VV option in the right click menu acts as if it wasn't clicked

pyrosshade:
have you tried testing this again? this seems like something that i fixed with #6664

Pockets-byte:
testing it again now

Pockets-byte:
I can confirm this still doesn't work on staging

Pockets-byte:
https://user-images.githubusercontent.com/55299415/121761401-83b2c580-cafd-11eb-973d-165faf7e66b4.mov

I am clicking in this

pyrosshade:
what shows in f5 when you try to do so?

corp-0:
yeah, I also just verified it in the same session that video was captured. 
```
NullReferenceException: Object reference not set to an instance of an object
SweetExtensions.NetId (UnityEngine.GameObject go) (at <4029ab2637b6496a9cf4fd9ebd32c4f4>:0)
Messages.Client.VariableViewer.RequestBookshelfNetMessage.Send (UnityEngine.GameObject _TheObjectToView, System.String adminId, System.String adminToken) (at <4029ab2637b6496a9cf4fd9ebd32c4f4>:0)
RightClickManager+<>c__DisplayClass21_0.<Generate>b__0 () (at <4029ab2637b6496a9cf4fd9ebd32c4f4>:0)
UI.Core.RightClick.RightClickMenuController.OnClickAction (UnityEngine.EventSystems.PointerEventData eventData, UI.Core.RightClick.RightClickRadialButton button) (at <4029ab2637b6496a9cf4fd9ebd32c4f4>:0)
[...]

Issue #6764, opened 2021-06-07
felinid improper layering


Pockets-byte:
Linux staging 

on staging felinids have improper layering for their ears, and possibly tail
ears render behind hair, and tail renders behind clothes and hair
![Screenshot at 2021-06-07 16-07-15](https://user-images.githubusercontent.com/55299415/121081704-158f9b00-c7ab-11eb-9af6-1ab4da2fe765.png)
![Screenshot at 2021-06-07 16-07-20](https://user-images.githubusercontent.com/55299415/121081705-158f9b00-c7ab-11eb-9eeb-9fc64a70a5d5.png)
![Screenshot at 2021-06-07 16-07-24](https://user-images.githubusercontent.com/55299415/121081706-158f9b00-c7ab-11eb-9b63-2d1de932c444.png)
![Screenshot at 2021-06-07 16-08-03](https://user-images.githubusercontent.com/55299415/121081707-16283180-c7ab-11eb-9691-4cd2fba863e0.png)
![Screenshot at 2021-06-07 16-08-14](https://user-images.githubusercontent.com/55299415/121081709-16283180-c7ab-11eb-879f-65a4cbe4d1d7.png)

Bod9001:
checked in editor can confirm, also tale is missing animation

jamessimo:
Lizards tails also are in the wrong place and appear in front of the player

ThatDan123:
Fixed by #6814

Issue #6762, opened 2021-06-06
item stick to a matrix even if the tiles underneath have been destroyed


Bod9001:
items that stay still on matrix should check if moved in new matrix when tile underneath is destroyed

pyrosshade:
Title is clearer than description, but if i have this correctly, items are moving with matrices despite technically seeming like they should belong to space?

Bod9001:
yeah, it's because the registered tile doesn't get constantly updated with position

ThatDan123:
Fixed by #6814

PR #6814, opened 2021-06-17
A bunch of miscellaneous fixes

Bod9001:
CL: [Fix] Fixes shuttle Coordinates not being updated on UI #6784 
CL: [Fix] fixes table climbing While buckled  #6715
CL: [Fix] Fixes VV matriix  #6779
CL: [Fix] Fixes layering issues on cat Race #6764
CL: [Fix] Fixes characters Settings for lizards
CL: [Fix] Fixes Body parts And internal bleedding 
CL: [Fix] fixes item sticking to matrices #6762

### DIFF
--- a/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
@@ -306,7 +306,7 @@ namespace HealthV2
 		public virtual void RemoveFromBodyThis()
 		{
 			BodyPartRemovalChecks();
-			RootBodyPartContainer parent = this.GetParent();
+			dynamic parent = this.GetParent();
 			if (parent != null)
 			{
 				parent.RemoveSpecifiedFromThis(this.gameObject);
--- a/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
@@ -369,6 +369,7 @@ namespace HealthV2
 		/// </summary>
 		/// <returns>A body part that contains it OR a body part container that contains it OR null if it is not
 		/// contained in anything</returns>
+		///TODO change to some type of inheritance/Interface model
 		public dynamic GetParent()
 		{
 			if (ContainedIn != null)
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterTile.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterTile.cs
@@ -235,6 +235,8 @@ public class RegisterTile : NetworkBehaviour, IServerDespawn
 
 	private CustomNetTransform cnt;
 
+	public CustomNetTransform CNT => cnt;
+
 	//cached for fast fire exposure without gc
 	private IFireExposable[] fireExposables;
 	private bool hasCachedComponents = false;
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterTile.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterTile.cs
@@ -235,7 +235,7 @@ public class RegisterTile : NetworkBehaviour, IServerDespawn
 
 	private CustomNetTransform cnt;
 
-	public CustomNetTransform CNT => cnt;
+	public CustomNetTransform customNetTransform => cnt;
 
 	//cached for fast fire exposure without gc
 	private IFireExposable[] fireExposables;
--- a/UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
+++ b/UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
@@ -382,6 +382,22 @@ namespace IngameDebugConsole
 		}
 
 #if UNITY_EDITOR
+		[MenuItem("Networking/Give me some AA!")]
+#endif
+		private static void MakeAA()
+		{
+			if (CustomNetworkManager.Instance._isServer)
+			{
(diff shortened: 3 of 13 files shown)


UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
```diff
@@ -306,7 +306,7 @@ public virtual void AddBodyPartSlot(ItemSlot ItemSlot)
 		public virtual void RemoveFromBodyThis()
 		{
 			BodyPartRemovalChecks();
-			RootBodyPartContainer parent = this.GetParent();
+			dynamic parent = this.GetParent();
```
corp-0, line 309: what's ``dynamic``?

Bod9001: it's basically python flexibility

corp-0: why does it need to be type flexible?

Bod9001: the problem is, root body part container and body part both could contain body parts, so it could Either or

corp-0: aaah I see. So this GetParent could either return a RootBodyPartContainer or BodyPart. This is screaming bad implementation to me, why aren't those clases parented under the same abstract class so we make this depend on abstracts instead  of concretes than can be of different type?

corp-0: https://en.wikipedia.org/wiki/Dependency_inversion_principle

Bod9001: idk, just wanted to try dynamic it works pretty well

Bod9001: also say that having to write a bunch of code, because only came across the issue once it was all mostly written up

Bod9001: *also saved having to write a bunch of code

corp-0: yes, I understand and I guess it is good enough for now. Can you slap a TODO comment there so we can revisit it later? I frankly don't feel comfortable with it being like that.

MaxIsJoe:
I understand the change for `RemoveFromBodyThis()` but how did you fix internal bleeding exactly and what issues did it have?

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterTile.cs
```diff
@@ -235,6 +235,8 @@ private set
 
 	private CustomNetTransform cnt;
 
+	public CustomNetTransform CNT => cnt;
```
MaxIsJoe: Can you please call this variable "CustomNetTrans" instead of "CNT"? Much more understandable.
Same thing for the private name as well.

corp-0:
> I understand the change for `RemoveFromBodyThis()` but how did you fix internal bleeding exactly and what issues did it have?

read the discussion I'm having above. Problem was being hidden by the use of ``var``. Sometimes, the GetParent would return a ``BodyPart`` and sometimes a ``RootBodyPartContainer``, both classes are not parented under a common super class thus the script could fail when we tried to get a particular type, in this case ``RootBodyPartContainer``. By using ``dynamic`` as the type, we change the variable to be dynamically typed instead of strongly typed (like GDScript works) which is more of a hack in my opinion.

MaxIsJoe (commented):
> > I understand the change for `RemoveFromBodyThis()` but how did you fix internal bleeding exactly and what issues did it have?
> 
> read the discussion I'm having above. Problem was being hidden by the use of `var`. Sometimes, the GetParent would return a `BodyPart` and sometimes a `RootBodyPartContainer`, both classes are not parented under a common super class thus the script could fail when we tried to get a particular type, in this case `RootBodyPartContainer`. By using `dynamic` as the type, we change the variable to be dynamically typed instead of strongly typed (like GDScript works) which is more of a hack in my opinion.

That sounds more like a fix for limbs getting removed, not internal bleeding. Did I miss something in the code that was changed that uses the new `dynamic` keyword?

MaxIsJoe:
> That sounds more like a fix for limbs getting removed, not internal bleeding. Did I miss something in the code that was changed that uses the new `dynamic` keyword?

Nevermind I found what's going on. Internal bleeding was trying to use `GetParent()` which caused the check to return false and skip over the code to not run any errors. `BodyPart currentParent = ContainedIn;` should eliminate the issue and would be better for performance since it's not trying to look for a gameObject when it just grabs a reference.
Smart.

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterTile.cs
NoooneyDude: Nooooo ; (
; )

ThatDan123:
> > I understand the change for `RemoveFromBodyThis()` but how did you fix internal bleeding exactly and what issues did it have?
> 
> read the discussion I'm having above. Problem was being hidden by the use of `var`. Sometimes, the GetParent would return a `BodyPart` and sometimes a `RootBodyPartContainer`, both classes are not parented under a common super class thus the script could fail when we tried to get a particular type, in this case `RootBodyPartContainer`. By using `dynamic` as the type, we change the variable to be dynamically typed instead of strongly typed (like GDScript works) which is more of a hack in my opinion.

Not too relevant for now but if we ever wanted to build for IL2CCP instead of mono (which should in theory have better performance) we'd have to change it as it doesnt support dynamic as its runtime generated code.

Issue #6492, opened 2021-05-13
Character sprites buggy B:35


NoooneyDude:
Character sprite representation is glitchy when moving around.

![buggy character sprites](https://i.imgur.com/7u5VGcK.gif)

NoooneyDude:
I managed to reproduce this locally with two clients on headless: have one client pull the other in a tight circle.

![buggy sprite](https://i.imgur.com/BMic60T.gif)

Both clients saw the exact same thing at all times.

As you can see, you can get it into a buggy state without needing lag to reproduce - permanently stays that way until moved.

NoooneyDude:
The first client died to a fire and was lying prone. I then respawned that client, and the second client observed the corpse as such:

![standing corpse](https://user-images.githubusercontent.com/6926225/118478251-c46c2880-b763-11eb-9d4c-599bfbc94850.png)

Note the lack of visible player clothing, equipment, and the fire is behind the player. Does clothing burn up now? Though I think it is related to the bug above.

NoooneyDude:
Another interesting image

![underwear](https://user-images.githubusercontent.com/6926225/118479946-bd461a00-b765-11eb-8efe-9645edf494c9.png)

Bod9001:
Some issues could be from, the fact that clothing directions are done client side while body directions  sprites are done server side, 
the main issue is it's kind of a mix of the old and new systems so that needs to be sorted

PerfectTangent:
this issue is being promoted to bounty status. Evaluated at 25 USD.

pyrosshade:
This seems like a possible desync issue, should i exclude it from my writeup?

PerfectTangent:
no, reference this issue in that issue if you think they are connected.

PerfectTangent:
bounty bumped to 75 USD

Fogapod:
clothing/body rotation desync:

Only body rotates, clothes face same direction.
This happens when:
- you rotate yourself from radial menu
- you are pulling someone and rotate them from radial menu
- you are walking pulling someone and change direction (first tile in new direction only body rotates, 2nd tile clothes rotate as well)

Demonstration:

https://user-images.githubusercontent.com/22667809/120896484-793e8a80-c62a-11eb-8914-b11d30eb68f2.mp4

### Steps to reproduce
1. rotate yourself from radial menu
2. rotate someone from radial menu (you should be pulling them)
3. pull someone changing direction

PerfectTangent:
[...]

PR #6815, opened 2021-06-17
Fixes character body sprites being out of sync

Aranclanos:
### Purpose
Fixes character body sprites being out of sync with the clothes of the same character.
Fixes (partially? dunno if to include the comments into the original issue post) #6492

So how it works:
When you move, you decide on your own character's rotation and so, as a client you change your own sprites. But moments later after informing the server about your rotation, you receive messages from the sprites themselves wanting a sprite change that  was from a frame (plus ping) ago, making your character look weird with your body and clothes not being the same.
This is extremely intensified with high ping, otherwise its a 1 frame desync.

This PR disables server-side rotation from sending the sprite updates, the client will handle that on its own.

### Notes:
Client still rejects server messages of rotation on players that they're pulling, will resolve on a followup PR


Bod9001:
looks good

PR #6816, opened 2021-06-17
Mobs will hit windows and grills again, Mobs will also hit adjacent tiles if they cant move to target

ThatDan123:
CL: [Fix] Mobs will hit windows and grills again

CL: [Improvement] Mobs will also hit adjacent tiles if they cant move to target


corp-0:
very nice. Why is it important to find grills in particular?

ThatDan123:
image

Basically the red target is what the raycast has hit and is what the mob trys to attack but the yellow is where the mob is getting blocked from doing a move to, so now it'll attack the raycast target then hit the adjacent tiles around it if it can. So it checks for windows, grills and objects

corp-0:
Yes, thanks. I gave the code a second read with more calm and got it.

waffielz:
sick

PR #6817, opened 2021-06-17
Refactor the Mob Tile Hit

ThatDan123:
-Reread through the code and realised it wasnt actually working as it needed to fully so i rewrote it.


corp-0:
xSucces?

ThatDan123:
> xSucces?

X was successful

PR #6829, opened 2021-06-19
Makes it so VV will renders in front of chat

Bod9001:
yeah, pretty self-explanatory, though this will make every window render above chat


waffielz:
thank uuuuuuuuuuuuuuuuuuuuuu

PR #6830, opened 2021-06-20
More body stuff: liver is a clam edition.

pyrosshade:
### Purpose
One of the first steps I'm taking into filling out the metabolism functions is looking closer at what's happening with things being eaten. Things already went into the stomach, which it handles bits and pieces of nutrition stuff, but some of the reactions you might normally expect in ss13 don't happen. On top of that, one of the few reaction effects that we DO have wasn't occurring properly with respect to doing so in the body. With a look into explosions, if this effect gets to be strong enough, the hope of adding items being thrown is to make sure internal explosions bust out and create an interesting scene. That and throwing items caught in explosions seems to have been planned. So with further testing we can dial in on making sure explosions in general are more interesting.

### Notes
Did some folder organization with reagent and reactions in project management
Alcohols are not yet adapted to change drunk time, nor do they break down in the liver yet, this will be in the next PR
Liver does not yet take damage or check thresholds, this too will be in a future PR. This will likely need tweaking and some maths to get right.
Toxins are essentially the main thing removed outright from the body for now.
ChemExplosions now have a basic potency variable, acts as a multiplier of the final strength calculation before it's passed off to Explosion Manager.
Potassium+Water does not use a Generic ChemExplosion effect now,

### Changelog:
CL: **[New]** Standard reactions now happen within the stomach
CL: **[New]** Chemical Reactions that explode now do damage to your insides if the reaction takes place in your body
CL: **[New]** Explosions outside the body now attempt to throw items
CL: **[New]** Liver now has functionality to filter alcohol and toxins from the bloodstream
CL: **[Balance]** Potassium chemical explosions will be nerfed until their math or base explosion math can be made to be more nuanced.


UnityProject/Assets/Scripts/Systems/Explosions/ExplosionNode.cs
```diff
@@ -59,7 +61,23 @@ public void Process()
 
 			foreach (var integrity in matrix.Get<Integrity>(v3int, true))
 			{
+				Debug.Log($"ExplosionNode handling : {integrity}");
```
NoooneyDude:
```suggestion
```
Was just for testing, I presume?

UnityProject/Assets/Scripts/Systems/Explosions/ExplosionNode.cs
```diff
@@ -79,5 +97,22 @@ public void Process()
...
+				case 0:
+					return SpinMode.None;
+				case 1:
+					return SpinMode.Clockwise;
+				case 2:
+					return SpinMode.CounterClockwise;
+				default:
+					return SpinMode.Clockwise;
+			}
+		}
```
NoooneyDude: May I suggest moving this to some utility class? `RandomUtils` seems appropriate. I see exact same method is used in `MassDriver` and `Singularity`.

NoooneyDude (commented):
"Alcohol" is frequently misspelled including folders/files, would you mind fixing them? Just helps when searching for things.

UnityProject/Assets/Scripts/Systems/Explosions/ExplosionNode.cs
pyrosshade: ah, yes

UnityProject/Assets/Scripts/Systems/Explosions/ExplosionNode.cs
pyrosshade: alrighty

pyrosshade: resolved with commit 3170a93

pyrosshade:
just need to fix up misspelling alcohol and it should be good

PR #6833, opened 2021-06-20
fixes a few staging Errors, adds Grills layer to asteroids, fixes plasma generation initialisation

Bod9001:
adds Grills layer to asteroids

I added more logging with inventory interactions

fix an error with number dial

insured stuff removes itself from the update manager properly

fixes plasma generation initialisation

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Implants/Organs/Liver.cs
+++ b/UnityProject/Assets/Scripts/Items/Implants/Organs/Liver.cs
@@ -66,48 +66,62 @@ namespace HealthV2
 			float drawnAmount = 0;
 
 			//figure out how much we are going to process or remove
-			foreach (Reagent reagent in blood.CurrentReagentMix.reagents.Keys)
+			lock (blood.CurrentReagentMix.reagents)
 			{
-				if (Alcohols.AlcoholicReagents.Contains(reagent) || Toxins.Contains(reagent))
+				foreach (Reagent reagent in blood.CurrentReagentMix.reagents.Keys)
 				{
-					float amount = Mathf.Min(tickPullProcessingAmnt,RelatedPart.BloodContainer.CurrentReagentMix[reagent]);
-					amount = Mathf.Min(amount, (processingContainer.MaxCapacity - processingContainer.ReagentMixTotal)-drawnAmount);
-
-					tempArray.Add(new Tuple<Reagent, float>(reagent, amount));
-
-					if (processingContainer.IsFull)
+					if (Alcohols.AlcoholicReagents.Contains(reagent) || Toxins.Contains(reagent))
 					{
-						Logger.LogTrace("Liver is full, please try again. or don't.",Category.Health);
-						break;
-					}
+						float amount = Mathf.Min(tickPullProcessingAmnt,RelatedPart.BloodContainer.CurrentReagentMix[reagent]);
+						amount = Mathf.Min(amount, (processingContainer.MaxCapacity - processingContainer.ReagentMixTotal)-drawnAmount);
 
-					drawnAmount += amount;
-					tickPullProcessingAmnt -= amount;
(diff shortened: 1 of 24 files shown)


UnityProject/Assets/Scripts/Objects/Engineering/PowerGenerator.cs
```diff
@@ -52,14 +52,21 @@ private enum SpriteState
...
+
+		public void Initialise()
+		{
+			registerTile = GetComponent<RegisterTile>();
+			securable = GetComponent<WrenchSecurable>();
+			baseSpriteHandler = GetComponentInChildren<SpriteHandler>();
+			electricalNodeControl = GetComponent<ElectricalNodeControl>();
 			var itemStorage = GetComponent<ItemStorage>();
 			itemSlot = itemStorage.GetIndexedItemSlot(0);
 			securable.OnAnchoredChange.AddListener(OnSecuredChanged);
```
Aranclanos, line 61: you're adding two listeners to this now, it will be called twice

Aranclanos: I don't understand what's wrong with the way it is already, please explain further what you're trying to accomplish

Bod9001: On spawn server isn't called when it's already on the scene I think

Aranclanos: its called at roundstart

Bod9001: tho  there are errors on staging relating to it not be initialised properlybefore being used

Bod9001: TryToggleOn  errored  because itemSlot was null

NoooneyDude: Yeah, ItemSlot / ItemStorage stuff are snowflakes and can't be used in Awake. Start might be ok? Would have see how they work.

NoooneyDude: Turns out, that same storage method `GetIndexedItemSlot` is used elsewhere in other monobehaviour `Start()`s so presumably it is ok there.

UnityProject/Assets/Scripts/Items/Implants/Organs/Liver.cs
```diff
@@ -66,70 +66,98 @@ public override void ImplantPeriodicUpdate()
...
-					if (tickPullProcessingAmnt <= 0) break;
+						drawnAmount += amount;
+						tickPullProcessingAmnt -= amount;
+						if (tickPullProcessingAmnt <= 0) break;
+					}
 				}
+
 			}
 
+			List<Tuple<Reagent,float>> Toremove = new List<Tuple<Reagent,float>>();
```
pyrosshade, line 94: this and the extra foreachs you presented aren't needed, that's adding a whole other temporary data structure on top of a temporary data structure. If temp array is having issues being removed from in the foreach, then i'll just change it so the temparray is being cleared afterwards. A bit miffed i wasn't awake to see this pr before it was merged.

corp-0: wouldn't iterating backwards be good enough to solve the issue?

pyrosshade: it's not an issue of how it's iterating, just an issue of what temparray is used for. it's simple enough to just clear temparray after the foreachs that read through it, which i'm gonna pull up a pr to do just that.

PR #6837, opened 2021-06-21
Fixes client overlays not being removed

### LINKED ISSUE #6704: Pka overlays don't go away
found on staging

1. use pka
2. profit

image

Aranclanos:
### Purpose
Removes the unused 'RemoveAll' argument and variables in tile update changes code.
Removes RemoveOverlaysOfName() and GetOverlayPosByName(), all overlay removals will now use OverlayType instead.
All remove overlay server functions will now be thread safe.
Separates the update tile message and remove tile message in SpawnSafeThread into two.
Fixes clients not removing certain specific overlays.
Fixes one way to reproduce #6398 (with bombs, although the bug that I'm fixing seems to have been aded after that issue was created)
Fixes #6704

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/TileChangeManager.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/TileChangeManager.cs
@@ -378,25 +378,6 @@ public class TileChangeManager : MonoBehaviour
 
 }
 
-public enum OverlayType
-{
-	//none is used to say there is no overlay, add new category if you need a new type
-	None,
-	Gas,
-	Damage,
-	Cleanable,
-	Fire,
-	Mining,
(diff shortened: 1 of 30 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/TileChangeManager.cs
```diff
@@ -407,16 +376,25 @@ public void InternalRemoveTile(Vector3 position, LayerType layerType, bool remov
...
-		None,
-		Gas,
-		Damage,
-		Cleanable,
-		Fire,
-		Mining
-	}
+}
+
+public enum OverlayType
```
corp-0: Move the enum to its own file and give it a namespace

Aranclanos:
done

PR #6841, opened 2021-06-21
Hold the data structure please

pyrosshade:
We don't need an intermediate data structure for an intermediate data structure. That just makes our RAM frown.

Also, don't mind the extra commits, my develop is a bit dirty. Only the commits after the merges are the real changes.


MaxIsJoe:
you know you can create different branches for changes instead of sitting on develop all the time, right?

pyrosshade:
Yes, i know, this branch was based from my develop, i just need to rebase it.

PR #6843, opened 2021-06-22
fixes job ban list not removing elements, mag boots exploit, nan bug in reagent mix, grills not being destroyed by fire properly

Bod9001:
fixes job ban list not removing elements
fixes grills not being destroyed by fire properly
fixes potential nan scenario in reagent mix
fixes mag boots make you run faster when turned on

### DIFF
--- a/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
@@ -324,11 +324,11 @@ namespace HealthV2
 			{
 				CharacterSettings settings = HealthMaster.gameObject.Player().Script.characterSettings;
 				ColorUtility.TryParseHtmlString(settings.SkinTone, out Tone);
-				BodyPartItemSprite.SetColor(Tone);
+				BodyPartItemSprite.OrNull()?.SetColor(Tone);
 			}
 			if(currentBurnDamageLevel == BurnDamageLevels.CHARRED)
 			{
-				BodyPartItemSprite.SetColor(bodyPartColorWhenCharred);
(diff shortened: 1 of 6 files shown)


UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
```diff
@@ -324,11 +324,11 @@ private void BodyPartRemovalChecks()
 			{
 				CharacterSettings settings = HealthMaster.gameObject.Player().Script.characterSettings;
 				ColorUtility.TryParseHtmlString(settings.SkinTone, out Tone);
-				BodyPartItemSprite.SetColor(Tone);
+				BodyPartItemSprite.OrNull()?.SetColor(Tone);
```
Aranclanos, line 327: why can this be null?

UnityProject/Assets/Scripts/Systems/Chemistry/ReagentMix.cs
```diff
@@ -196,7 +196,10 @@ public void Add(ReagentMix b)
 			}
 			else
 			{
-				Temperature = (Temperature * Total + b.Temperature * b.Total) / (Total + b.Total); //TODO Change to use different formula Involving Heat capacity of each Reagent
+				if (Total + b.Total != 0)
```
Aranclanos: You're already checking if both are 0 this check, this isn't needed.

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
Bod9001: well not every body part Has a Sprite

PR #6858, opened 2021-06-25
Fixes mixer prefab and convert sprite handler  Message to use reader and writer

Bod9001:
the changes to the mix is because it has a value of 10000 and the net Dial doesn't like it since it only goes to 9999
 writing directly with bites and stuff should hopefully reduce GC from the messages and increasing the amount of information Sent per message


Bod9001:
tested with headless Server

PR #6859, opened 2021-06-26
Fixes admin storage items being invisible on the UI

### LINKED ISSUE #6136: Aghost inventory does not function B:40
Screenshot at 2021-02-26 19-45-58
linux - 4117
current aghost inventory does not completely work as intended. clicking on items just drops them rather than putting them into your hand. also cannot see items held in hand nor interact with them. the same goes for clothing under hardsuits

also to note. right click does not work in inventory viewer. I'm  guessing it doesn't work in aghost inv either however I have no way to verify it

additionally it would be nice to be able to pick up and pull items from the ground

Aranclanos:
### Purpose
Fixes admin storage items being invisible on the UI
Fixes #6136

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Lifecycle/PlayerSpawn.cs
+++ b/UnityProject/Assets/Scripts/Core/Lifecycle/PlayerSpawn.cs
@@ -369,6 +369,8 @@ public static class PlayerSpawn
 
 		if (PlayerList.Instance.IsAdmin(forMind.ghost.connectedPlayer))
 		{
+			var adminItemStorage = AdminManager.Instance.GetItemSlotStorage(forMind.ghost.connectedPlayer);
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Core/Lifecycle/PlayerSpawn.cs
```diff
@@ -369,6 +369,8 @@ public static void ServerSpawnGhost(Mind forMind)
 
 		if (PlayerList.Instance.IsAdmin(forMind.ghost.connectedPlayer))
 		{
+			var adminItemStorage = AdminManager.Instance.GetItemSlotStorage(forMind.ghost.connectedPlayer);
+			adminItemStorage.ServerAddObserverPlayer(ghost);
```
Bod9001, line 373: Does it need to change the UI item Slot GetItemStorage?

PR #6861, opened 2021-06-26
Scaled human blood

jamessimo:
### Purpose
We need to scale up blood as right now a human only has 5u of blood and should have closer to 500

### Notes:
I have made this a draft as I need help understanding what else needs to be scaled up. No matter what I do nothing really seems to change and not sure if this is enough


Bod9001:
need to change the code/SOs for metabolic reactions MetabolismReaction.cs BodyHealthEffect stuff like PercentageBloodOverdose
and increase heartbeat strength tho that needs to be balanced bit, 
increased lung strength maybe
BodyFat needs to be balanced
Stomach
change how much each body part stores in blood

Bod9001:
Those are all I can think of currently

Bod9001:
like, it would be nice to do the fat too, since the player gets slowed quite a lot by lack of nutrients and there's other stuff I've specified that hasn't been done, do you want me to merge then work on those?

jamessimo:
> , do you want me to merge then work on those?

If possible yes, I have tried many times to make fat remove the slow down but nothing I seem to do works.

UnityProject/Assets/Scripts/Core/Factories/EffectsFactory.cs
```diff
@@ -44,12 +44,12 @@ public static void BloodSplat(Vector3 worldPos, ReagentMix bloodReagents = defau
 		string sizeDesc;
 
 
-		if (bloodReagents.Total < 0.1f)
+		if (bloodReagents.Total < 5f)
```
vlandemart: These are magic numbers. Move them into separate public/[SerializedField] variable so they would be easier to change later on.

UnityProject/Assets/Scripts/Core/Factories/EffectsFactory.cs
```diff
@@ -44,12 +44,12 @@ public static void BloodSplat(Vector3 worldPos, ReagentMix bloodReagents = defau
...
 		{
 			chosenTile = smallBloodTile;
 			sizeDesc = "drop";
 		}
-		else if(bloodReagents.Total > 0.1f && bloodReagents.Total < 0.3f)
+		else if(bloodReagents.Total > 5f && bloodReagents.Total < 15f)
```
vlandemart: Also a magic number

UnityProject/Assets/Scripts/HealthV2/Living/CirculatorySystem/CirculatoryInfo/CirculatoryInfo.cs
```diff
@@ -7,23 +7,23 @@ public class CirculatoryInfo : ScriptableObject
 {
 	//General data that the circulatory system needs.
 	[Tooltip("Maximum amount of blood in the system.")]
-	public float BLOOD_MAX = 7000;
-	public float BLOOD_SLIME_SPLIT = 3500;
+	public float BLOOD_MAX = 700;
```
vlandemart, line 10: Can't those be changed in a prefab instead of a script?
If you are sure you need to change a script, please correct tooltips for those, right now "BLOOD_NORMAL = 500" but tooltip says "The normal amount of blood in the body. 5000mL is the human average"

UnityProject/Assets/Scripts/HealthV2/Living/CirculatorySystem/CirculatorySystemBase.cs
```diff
@@ -35,7 +35,7 @@ public void SetBloodType(BloodType inBloodType)
 		private void Awake()
 		{
 			healthMaster = GetComponent<LivingHealthMasterBase>();
-			AddFreshBlood(ReadyBloodPool, 5);
+			AddFreshBlood(ReadyBloodPool, 500);
```
vlandemart: This is a magic number. Please, add variable like "StartingBlood" and use it instead, so it would be easier to change later

UnityProject/Assets/Scripts/Items/Medical/HealthScanner.cs
```diff
@@ -47,8 +47,8 @@ public void ServerPerformInteraction(HandApply interaction)
 
 			var health = interaction.TargetObject.GetComponent<LivingHealthMasterBase>();
 			var totalPercent = Mathf.Round(100 * health.OverallHealth / health.MaxHealth);
-			var bloodTotal = Mathf.Round(health.GetTotalBlood() * 1000);
-			var bloodPercent = Mathf.Round(bloodTotal / health.CirculatorySystem.BloodInfo.BLOOD_NORMAL * 100);
+			var bloodTotal = Mathf.Round(health.GetTotalBlood());
+			var bloodPercent = Mathf.Round(bloodTotal / health.CirculatorySystem.BloodInfo.BLOOD_NORMAL * 1000);
```
vlandemart: Why "bloodTotal / bloodNormal * 1000" is a blood percent?
In ideal case, our bloodTotal == bloodNormal, so that "bloodTotal / bloodNormal" would be 1. If you multiply it by 100, you get percentage.

vlandemart (changes requested):
Changes are requested.
Some magic numbers (even though they were present before the PR), and few mistakes

UnityProject/Assets/Scripts/HealthV2/Living/CirculatorySystem/CirculatoryInfo/CirculatoryInfo.cs
jamessimo: Ok fixed it in my latest

PR #6864, opened 2021-06-27
Fixes self-hit when spamclicking

### LINKED ISSUE #5467: Sometimes you can punch yourself when clicking on something else while on harm intent
## Environment
- Game Version: Staging 5422
- OS Version: Windows 10 1803

Sometimes rapidly clicking on something on harm intent, you will end up punching yourself

(May be due to my bad internet, please close it if that's the case after testing)
### Steps to reproduce

1. Rapidly click on harm intent
2. Punch yourself
3. don't die

Aranclanos:
### Purpose
Moves hiticons out from player prefabs into SpriteManager.
Hit icons will now be pooled in the SpriteHandlerManager.
Fixes the hiticons being target of interactions, which was making players hit themselves when spamclicking a target.
Fixes #5467

### Notes:
It was happening more to some and less to others because the icon isnt predicted, so it arrives after your ping time with the chances of your click cooldown being off.


ThatDan123:
Tests failing as you need to add the prefab to the network manager and commit the changes

Aranclanos:
its not a networked prefab, its just an animation

ThatDan123:
> its not a networked prefab, its just an animation

Still needs to be in the list, as the list is also used to spawn client side stuff which doesnt have network identities, so it just searches and expected all stuff in the prefab folder to be in there

PR #6865, opened 2021-06-27
Surgery fixes

Bod9001:
also allows moths to have any Colour
resets window position when open
changes name of attach procedure
need some replication steps for surgery bugging out


PerfectTangent:
> also allows moths to have any Colour

We turned off the moth color selection because it didn't look good on the prexisting colored moths. Did you get greyscale moth sprites?

PR #6867, opened 2021-06-27
Atmos GasMix Refactor

ThatDan123:
-Refactors gasmix to have fast array of data for fast interation and a dictionary for fast look up for specific gases

-GasMix will now only store the gases in the mix and not all possible gases

-Allows for easy adding of new gases at runtime and for base gases, ~and it'll be easy to add the logic for runtime creation of gases if thats ever needed (change dictionary to use int instead of enum and then generate new int Ids from enum count onwards) Will be improving on this in PR, still got some stuff to do.~

-Base gases are now SO's

-I completely removed the custom editor script for the gas canisters as it was just annoying to try to modify for the new system, ive basically set the canisters to 20000 moles and the internals tanks to their max. Probably need to go through and reset the values to what they are supposed to be. This also resets all mapped gas canister stuff to the default values.

~-Still draft as ive still got stuff to change but PR is here so issues can be discussed~


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Utils/AtmosUtils.cs
```diff
@@ -54,11 +55,12 @@ public static bool IsPressureChanged(MetaDataNode node, out Vector2Int windDirec
...
-						for (int j = 0; j < Gas.Count; j++)
+						foreach (var gas in Gas.Gases)
 						{
-							float moles = node.GasMix.Gases[j];
-							float molesNeighbor = neighbor.GasMix.Gases[j];
+							float moles = node.GasMix.GasData.GetGasMoles(gas.Key);
+							float molesNeighbor = neighbor.GasMix.GasData.GetGasMoles(gas.Key);
 
 							if (Mathf.Abs(moles - molesNeighbor) > AtmosConstants.MinPressureDifference)
 							{
```
ThatDan123: Currently this is looping through all possible gases, but it might be better if i changed it to loop through all current node gases do the checks and then loop through all neighbour gases (as it might have gases which the current node doesnt have so would have been missed) and do the checks there too. Or could make a list containing the unique gas types between the two nodes but that would GC so might just be better to loop through them both separately

ThatDan123: Also just realised for an optimisation it should break after the setting to true when they are different on line 67 as that only needs to happen once.

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/Gas.cs
```diff
@@ -68,33 +73,97 @@ public static Gas[] All
...
+	[Serializable]
+	public class GasValues
+	{
+		public GasType GasType;
+
+		//Moles of this gas type
+		public float Moles;
+	}
+
+	public enum GasType
```
Bod9001: doesn't this mean you can't add new gases in runtime

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/Reactions/PlasmaFireReaction.cs
```diff
@@ -51,23 +51,21 @@ public void React(GasMix gasMix, Vector3 tilePos, Matrix matrix)
 				}
 
 				gasMix.RemoveGas(Gas.Plasma, MolesPlasmaBurnt);
-				if (gasMix.Gases[Gas.Plasma] < 0) gasMix.Gases[Gas.Plasma] = 0;
-
 				gasMix.RemoveGas(Gas.Oxygen, MolesPlasmaBurnt * 2);
-				if (gasMix.Gases[Gas.Oxygen] < 0) gasMix.Gases[Gas.Oxygen] = 0;
```
ThatDan123, line 57: Just noting that checks for < 0 happen internally now so arent needed

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/Gas.cs
```diff
@@ -1,48 +1,54 @@
...
+		public static readonly Gas BZ = new Gas(GasType.BZ,"BZ", 20, 44.01f, false, 0.4f, "NONE", OverlayType.None, 8);
+		public static readonly Gas Miasma = new Gas(GasType.Miasma,"Miasma", 20, 44.01f, true, 0.4f, "Miasma", OverlayType.Miasma, 0);
+		public static readonly Gas Nitryl = new Gas(GasType.Nitryl,"Nitryl", 20, 44.01f, true, 0.4f, "Nitryl", OverlayType.Nitryl, 16);
+		public static readonly Gas Tritium = new Gas(GasType.Tritium,"Tritium", 10, 44.01f, true, 0.4f, "Tritium", OverlayType.Tritium, 1);
+		public static readonly Gas HyperNoblium = new Gas(GasType.HyperNoblium,"Hyper-Noblium", 2000, 44.01f, false, 0.4f, "NONE", OverlayType.None, 0);
+		public static readonly Gas Stimulum = new Gas(GasType.Stimulum,"Stimulum", 5, 44.01f, false, 0.4f, "NONE", OverlayType.None, 7);
+		public static readonly Gas Pluoxium = new Gas(GasType.Pluoxium,"Pluoxium", 80, 44.01f, false, 0.4f, "NONE", OverlayType.None, -10);
+		public static readonly Gas Freon = new Gas(GasType.Freon,"Freon", 300, 44.01f, true, 0.4f, "Freon", OverlayType.Freon, -5);
```
ThatDan123: Will probably move all these to SO's in another PR or this one

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/Gas.cs
ThatDan123: Theres two ways to make it work, either add some spare enum values at the end but probably better to just internally use int instead of the actual enum in the dictionary and then just keep a count on how many new added gases + enum total count to get a new id. Either way wouldnt be hard to allow for runtime gas creation in another PR

ThatDan123: Will probably remove the enum and just making the int auto generated, sounds easier.

ThatDan123: Actually will still need the enum for gas canister setting values but that'll just be to link to the gas type not for storing the data

UnityProject/Assets/Scripts/Systems/Fluids/MixAndVolume.cs
```diff
@@ -130,12 +132,12 @@ public void Add(MixAndVolume mixAndVolume, bool changeVolume = true)
...
-			var removeNewOne = new float[gasMix.Gases.Length];
-			for (int i = 0; i < gasMix.Gases.Length; i++)
+			var newOne = new GasData();
+			var removeNewOne = new GasData();
+			foreach (var gasData in gasMix.GasesArray)
 			{
-				removeNewOne[i] = gasMix.Gases[i] * percentage;
-				newOne[i] = gasMix.Gases[i] * (1 - percentage);
+				removeNewOne.SetMoles(gasData.GasType, gasMix.GasData.GetGasMoles(gasData.GasType) * percentage);
+				newOne.SetMoles(gasData.GasType, gasMix.GasData.GetGasMoles(gasData.GasType) * (1 - percentage));
```
ThatDan123: TODO stick `gasMix.GasData.GetGasMoles(gasData.GasType)` in a var so it doesnt need to be gotten twice, or is that a premature optimisation :p

Bod9001:
conflicts

ThatDan123:
Fixed

UnityProject/Assets/Scripts/Objects/Pipes/Devices/Filter.cs
```diff
@@ -12,24 +14,7 @@ public class Filter : MonoPipe
...
-			{"BZ",Gas.BZ},
-			{"MIAS",Gas.Miasma},
-			{"NO2",Gas.Nitryl},
-			{"TRIT",Gas.Tritium},
-			{"HN",Gas.HyperNoblium},
-			{"STIM",Gas.Stimulum},
-			{"PLX",Gas.Pluoxium},
-			{"FRE",Gas.Freon},
-		};
+		public Dictionary<string, GasSO> CapableFiltering;
```
NoooneyDude: ~~Curious why we assign value in awake now?~~ Also can probably be static, right?

UnityProject/Assets/Scripts/ScriptableObjects/Atmospherics/GasMixesSO.cs
```diff
@@ -0,0 +1,29 @@
...
+using UnityEngine;
+
+namespace ScriptableObjects.Atmospherics
+{
+	[CreateAssetMenu(fileName = "GasMixesSO", menuName = "ScriptableObjects/Atmos/GasMixesSO")]
+	public class GasMixesSO : ScriptableObject
+	{
+		[SerializeField]
+		private GasMix gasMix = new GasMix();
+		public GasMix GasMix => gasMix;
```
Bod9001: Might need some stuff so it doesn't just use the gas mix on the SO accidentally, Maybe like
copymix

Bod9001:
looks good

UnityProject/Assets/Scripts/Objects/Pipes/Devices/Filter.cs
ThatDan123: Had to move it to awake due to constructor violations with getting the GasSO's via the singleton.

Suppose it could be made static

NoooneyDude: Yeah, should've read more changes first. Thanks. Yeah there's a few lists here that could be made static, if we care about saving an insignificant amount of memory.. Up to you.

PR #6884, opened 2021-06-30
Dynamic Item Storage Fixes

ThatDan123:
-Switches some stuff from Item Storage to use Dynamic Item Storage

-Prevent law change for dead Ai's


Bod9001:
looks good

PR #6932, opened 2021-07-04
First iteration for integrating the new clearance framework

corp-0:
### Purpose
This PR won't make any gameplay changes just yet, but introduces the first brick for transitioning from old access to the new clearances framework. I will keep working on it with subsequent PR's by adding automatic tools to successfuly migrate.


corp-0:
EDIT: NVM I figured it out.

Tests are failing because ``ItemStorage`` changed and is missing a method I was calling. I guess I should use ``DynamicItemStorage`` now, but I don't know how to use it. This is the snippet of code that needs changes:

```cs
// Try get object in ID slot
			// Try get object in ID slot
var idSlot = playerStorage.OrNull()?.GetNamedItemSlot(NamedSlot.id).ItemObject;
if (idSlot != null && idSlot.TryGetComponent<IClearanceProvider>(out var idObject))
{
	return HasClearance(idObject.GetClearance());
}

// Nothing worked, let's go with active hand
var activeHandObject = playerStorage.GetActiveHandSlot().ItemObject;
if (activeHandObject != null && activeHandObject.TryGetComponent<IClearanceProvider>(out var handObject))
{
	return HasClearance(handObject.GetClearance());
}
```

PR #6955, opened 2021-07-06
makes all cargo crates cost more than the export price by at least 10%

### LINKED ISSUE #6953: Able to get unlimited laser tag crates
### Environment
- Game Version: Staging 21070515
- OS Version: Windows 10

By shipping back the crate the laser tag weapons come in, you're able to practically nullify the price of 500 credits due to the nanotrasen recycling program, thus making them free and you are practically able to order as many as you want. (Unsure if this was intentional)

### Steps to reproduce

1. Order a laser tag crate from the cargo console
2.  Unload all of the laser tag weapons on the floor
3. Ship back the crate the weapons came in

ThatDan123:
Fixes #6953

-Makes sure that all cargo crates cost more than the export price by at least 10%

### DIFF
--- a/UnityProject/Assets/ScriptableObjects/Cargo/Orders/50 Plasteel Sheets.asset
+++ b/UnityProject/Assets/ScriptableObjects/Cargo/Orders/50 Plasteel Sheets.asset
@@ -13,7 +13,7 @@ MonoBehaviour:
   m_Name: 50 Plasteel Sheets
   m_EditorClassIdentifier: 
   OrderName: 50 Plasteel Sheets
-  CreditCost: 16500
+  CreditCost: 17800
   Crate: {fileID: 1191069438927925888, guid: 007d9f81d20c55d479d86a1d7d4343db, type: 3}
   Items:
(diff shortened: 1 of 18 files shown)


Bod9001:
why  10%?

ThatDan123:
> why 10%?

10% minimum sounded ok

Bod9001:
humm, is it to fix bugs where stuff has the same price  stuff as it was new ?

mmails:
> humm, is it to fix bugs where stuff has the same price  stuff as it was new ?

I believe it's to fix items that are 500 or lower, since selling the crate the item came back with gives you 500 credits

Bod9001:
surely the test would include the price of the create?

PerfectTangent:
ThatDaniel is making sure that you can't get _free_ stuff from purchasing crates that have an equal resell value to their purchase price.

corp-0:
> surely the test would include the price of the create?

I remember it did, I don't know how the exploit by passed the check.

PR #6957, opened 2021-07-06
Lavaland tiles are now diggable, sand amount tweaks, filling in dug holes.

PetMudstone:
### Purpose
- You can now dig the lavaland tiles to get volcanic ash, which is identical to sand.
- Digging up sand or lavaland tiles now gives five sand or ash, up from three.
- Using five sand on dug-up sand tiles or five ash on dug-up lavaland tiles now turns them back into undug tiles.
- Renamed all the volcanic tiles to have similar names to their sprites. Reorganized volcanic tiles in palette.
- Changed the sprite of one volcanic tile which looked identical to another one. Had to change Wizard asteroid scene to replace the tile in question.

### Changelog:

CL: [New] Lavaland tiles are now diggable and yield volcanic ash which is identical in properties to sand.

CL: [New] You can use five sand on sand tiles you've dug up in order to return them to normal.

CL: [Balance] Diggable tiles now yield five sand.


waffielz:
ash glass gaming

PR #6958, opened 2021-07-06
Mining satchel is smaller and now fits in backpacks.

PetMudstone:
### Purpose
- The mining satchel is now medium sized, which means it fits in backpacks and some other storage containers.

### Changelog:

CL: [Balance] The mining satchel now fits in backpacks and other storage containers that can hold medium sized items.


waffielz:
ngl i never knew you made this change lmao, i just stopped putting it in my backpack out of habit

PR #6966, opened 2021-07-07
Liver, stop bleeding like a whoopie cushion [hotfix]

pyrosshade:
### Purpose
fixes errant reagentcontainer choice making the liver psuedo bleed

### Changelog:
CL: [Fix] Liver is no longer a source of buggy bloodloss.


UnityProject/Assets/Scripts/Items/Implants/Organs/Liver.cs
```diff
@@ -14,7 +14,7 @@ public class Liver : BodyPartModification
 		/// <summary>
 		///ReagentContainer which the liver uses to hold reagents it will process. Reagents like alcohol will be broken down into their ethanol reagent via ReactionSet
 		/// </summary>
-		private ReagentContainer processingContainer;
+		[SerializeField] private ReagentContainer processingContainer;
```
corp-0: how was this assigned before?

pyrosshade:
```
public override void SetUpSystems()
		{
			base.SetUpSystems();
			processingContainer = GetComponent<ReagentContainer>();
		}
```

like so

corp-0: I don't quite like having to manually assing the component. whats happening here? There are many of the same component in the same gameobject?

pyrosshade: I don't like it either, but yes. ReagentContainerBody is first along the list and the container intended for processing is below that in the components list. The stomach also has it's stomachContents reagentContainer manually set as well.

UnityProject/Assets/Scripts/Items/Implants/Organs/Liver.cs
```diff
@@ -14,7 +14,7 @@ public class Liver : BodyPartModification
 		/// <summary>
 		///ReagentContainer which the liver uses to hold reagents it will process. Reagents like alcohol will be broken down into their ethanol reagent via ReactionSet
 		/// </summary>
-		private ReagentContainer processingContainer;
+		[HideInInspector] public ReagentContainer processingContainer;
```
NoooneyDude: I would say using `NonSerialized` is more appropriate here (similar situation for other instances in this PR), because we don't need the value to persist on disk - just a small thing to keep prefab assets slightly smaller..

waffielz:
lmao

PR #6972, opened 2021-07-07
UI is now robust hopefully

Bod9001:
yeah,  made stuff more robust now shouldn't bug out

some inventory accessing Should work a little bit better, 
~~I think I still need to fix one edge case, But lunch got in the way ~~ is fixed
### Changelog:

CL: [Fix] another battle  won in a never-ending war against the UI bugs


Aranclanos:
image
no UI after restarting with domain reload off

backpack movement
cant open containers easily

scroll-click stopped working for hand swap


I haven't made a build to test further, this was on editor only

Bod9001:
I don't know what to do for the customer interactions with backpacks and stuff?

PR #6994, opened 2021-07-08
Server channel validation on chat message

### LINKED ISSUE #6561: How to prank call the entire station [honk]
image
image
image
[very based]

Deciding to which channels to send is on the clients side, thus you have no choice but to listen to me (:<<<

didnt check if i can chat in admin chat, probably can, also didnt check whether i can also see all chats, probably can

discord: Jakesolly#0966

Aranclanos:
### Purpose
The server will now validate the channels used when a player sends a chat message.
Fixes #6561


Bod9001:
surprisingly easy, thought it would be hard

PR #7040, opened 2021-07-13
Reduces CPU usage related to objects moving on centcomm

Aranclanos:
Updates the max transfer moles amount from the -still used- depreciated vent and scrubber prefabs from 100 to 5, stopping centcomm from consuming cpu by moving objects due to all the air movement.

image
the bright blue is the CPU consumption difference

note: the usage is from objects moving, not gas transfers. Take this as a hotfix, movement shouldn't be as expensive.


Bod9001:
yeah, big part of explosion lag to, it also should make shuttles less buggy

PR #7042, opened 2021-07-13
Rounds end limbo state fix

Aranclanos:
### Purpose
The round end report will now happen after the coroutine for the round restart ends triggers, avoiding the round getting stuck in a limbo state in the case of a runtime happening inside the round end report.

Note: I'm not sure why would a runtime happen yet, and it did, but this will make sure the round actually starts and the server doesn't get stuck in a limbo

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/GameModes/GameMode.cs
+++ b/UnityProject/Assets/Scripts/Systems/GameModes/GameMode.cs
@@ -386,7 +386,7 @@ namespace GameModes
 		/// <summary>
 		/// End the round and display any relevant reports
 		/// </summary>
-		public virtual void EndRound()
+		public void EndRoundReport()
 		{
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Systems/GameModes/GameMode.cs
```diff
@@ -386,7 +386,7 @@ public virtual void CheckEndCondition()
 		/// <summary>
 		/// End the round and display any relevant reports
 		/// </summary>
-		public virtual void EndRound()
+		public void EndRoundReport()
```
corp-0, line 389: why the removal of virtual?

Aranclanos: unused

PR #7044, opened 2021-07-13
Health gradient

Bod9001:
### notes:
so, I've tested with a build and connecting to it, when lava land is initialising there are a few net errors They don't seem to have any impact, on client
### Changelog:

CL: [Improvement] Critical now has a gradient and improved the health icon state

CL: [Fix] Fixes you  accidentally eating your own organs when trying to implant them

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Systems/Alerts/UI_HeartMonitor.cs
+++ b/UnityProject/Assets/Scripts/UI/Systems/Alerts/UI_HeartMonitor.cs
@@ -139,20 +139,19 @@ public class UI_HeartMonitor : TooltipMonoBehaviour
 			if (HealthPercentage > -0.66f)
 			{
 				CurrentSpriteSet = 3;
-				overlayCrits.SetState(HealthPercentage);
 			}
 			else if (HealthPercentage > -1)
 			{
 				// crit state has 2 sprite sets (blinking)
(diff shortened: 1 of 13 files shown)


UnityProject/Assets/Scripts/UI/Systems/Alerts/UI_HeartMonitor.cs
```diff
@@ -106,48 +106,57 @@ private void CheckHealth()
...
-			overlayCrits.SetState(OverlayState.crit);
-		}
-		else if (overallHealthCache <= -100)
-		{
-			CurrentSpriteSet = 6;
-			overlayCrits.SetState(OverlayState.death);
+			if (HealthPercentage >= 1)
+			{
+				CurrentSpriteSet = 0;
+				overlayCrits.SetState(HealthPercentage);
```
jamessimo, line 115: Should this function just be called at the end of the if/else chain since it seems to get called no matter what the output is

jamessimo: Also, this whole block may be better as a switch statement to make it more manageable

Bod9001: yeah, good point

PR #7045, opened 2021-07-13
Hardsuit speed adjustments and energy sword stuff.

PetMudstone:
### Purpose
- Adjusted hardsuit speeds to hopefully be more reasonable. They were too fast before and then too slow.
- Energy sword was nukie exclusive, for some reason. It is no longer nukie exclusive.


corp-0:
meh, I thought it was fine as 4. It felt byond like. But aight, if people are angy, not much we can do.

Issue #7047, opened 2021-07-13
Fires should damage and destroy windows based on temperature.


PetMudstone:
### Description
In /tg/, fires don't damage windows on contact. Instead windows all have a temperature threshold. If temperature around the window goes above that threshold then the window will start taking damage and will eventually break. This lets windows (especially reinforced ones) effectively block most fires while still succumbing to very hot ones.

PR #7055, opened 2021-07-15
Squarestation and Pogstation changes.

PetMudstone:
### Purpose
- Added a double-barrel shotgun to Squarestation's bartender's room.
- Pogstation tweaks.
  - Reduced the number of windows to stop fires from spreading as easily.
  - Firelocks have been added to the main halls, a number of civilian rooms, and cargo.
  - Added furniture to the outside of medical in a new nook of some sort. 
  - Whatever that small enclosed area is in arrivals now has sofas and a comfy chair.
- Removes the old Punpun and Gaggle mobs as they don't exist in scenes anymore.
- Used up cigars and joints now use proper sprites. 

### Changelog:

CL: [New] Pogstation now has multiple firelocks in its main halls and a few of its misc. rooms.


PerfectTangent:
can I say, I think that while Pog could definitely use some window reduction, I don't think the solution should be to reduce windows to avert fires from spreading easily. That doesn't address the root problem which is that windows should be taking temperature damage, not fire damage.

PetMudstone:
That is true, the absence of the feature described in #7047 definitely influenced my decision to remove a number of windows.

PR #6530, opened 2021-05-16
Body part slots v3 

Bod9001:
currently it's just to check what's been changed and fix any merge conflicts

### Notes:
still need to do handcuffs

fixes #6066

CL: UI slots are now determined by what clothing and body parts you have

PR #7066, opened 2021-07-16
Added pockets to trenchcoats, expanded storage and movement penalty to duffel bags, expanded webbing/belt whitelists to contain more items.

PerfectTangent:
### Purpose

This PR adds 2 pocket slots to trenchcoats and variants. This included the Curator's coat, which gives them a nice little something to make them special, and the HoS's trenchcoat, which is a buff they don't necessarily need. We can try it for a while and see whether it's overpowered.

This PR also rebalances and adds a significant number of traits that are missing from belts and webbing so they are more usable and can contain more equipment that is generally expected of them, either from looking at ss13 or from a common sense perspective. Webbing now functions like the chest rig.

Lastly, this PR adds the staple difference between backpacks and duffel bags from ss13; Duffel bags have a movement penalty, but can contain a larger quantity of items.

### Notes:

- The trenchcoat and bomber jacket and all variants have gained 2 extra pockets in addition to suit storage
- added grenades to the whitelist for the janibelt and sec belt (for cleaning grenades and flashbangs)
- added cigarettes to the whitelist for most belts
- added power cells to the whitelist for toolbelts
- added radios to the whitelist for most belts
- added pill bottles to the whitelist for medical belts
- added speedloaders to the whitelist for the detective's holster
- made security and explorer's webbing function like the chest rig
- Duffel bags have increased storage, but come with a movement penalty.

I am not against removing the pockets from the armored trenchcoat outright or keeping the extra pockets and nerfing the armored trenchcoat in other ways down the line, such as bringing it's armor protection more in line with the rest of the security staff. With #6530, we have the ability to make much more flexible clothing, and we should try to make a standard format for it. 

A point about the webbing; It doesn't make much sense intuitively from a player's perspective that there is a different whitelist for all different forms of webbing. Funtime Dudeski made a good argument about this. They look like reflavours of the same item, but they can't hold any of the same stuff. We previously consolidated the whitelist for suit storage for the same reason, so I would argue that we should get rid of the whitelists and make them all function like the chest rig, where they just serve as flexible small item storage. In the case that someone feels we should revert this change, then as a compromise, I have also consolidated the different forms of webbing storage capacities into one.

All forms of webbing are meant to be upgrades over other typical belt wearable storage containers, and not having whitelists at all reflects thats well. What is the point of paying for the item (with labor points as miner or with TC as traitor) if it is at best a sidegrade? You feel me?

Changes tested in editor.

### Changelog:

CL: [Improvement] Several trenchcoats and jackets come with additional pockets.
CL: [Balance] Many belts can contain a larger selection of items, including radios, cigarettes, and specific items for specific belts (such as pill bottles for the medibelt).
CL: [Balance] The detective's holster can now store revolver speedloaders.
CL: [Improvement] Duffel bags now function substantially different from the other roundstart bag choices. They come with a small movement penalty, but they can carry a larger quantity of items.
CL: [Balance] Security and explorer's webbing have had their whitelists removed, making them function identically to the chest rig. Death Squad Officers now come with security webbing instead of chest rigs.


waffielz:
movement penalty to duffel bags = 1984

PR #7087, opened 2021-07-18
Tile Conductivity

ThatDan123:
-Adds tile temperature conductivity which triggers at high temperature

-Fixes welder heating room issue


ThatDan123:
Dont ask about the git commit history, am bad at git. The changed files show the correct ones

corp-0:
> Dont ask about the git commit history, am bad at git. The changed files show the correct ones

that's what happens with merges

Looks good, though it makes me unconfortable to add yet another property to BasicTile

Bod9001:
"Looks good, though it makes me unconfortable to add yet another property to BasicTile" why so?

corp-0:
Because BasicTile is a mess. At least this property makes sense to be there, but we have floors with radiation penetration, walls with footstep sounds, cables that change the footsteps. Instead of specializing the tiles into their own classes, we just added everything in BasicTile and slapped bools.

PR #7093, opened 2021-07-20
Change Reaction.cs to not rely on the serilization of nullable floats. Change ReactionEditor.cs to be compatible. 

SkyHearn:
### Purpose
There was a bug that caused every reaction have no temperature requirements. Upon looking in to it, it was discovered that Unity wasn't serializing the temperature values because they were stored in nullable floats. Also, edits could not be made to reactions because the custom editor would not mark the changes as dirty and unity would not save them along with the project.
The changes I made to fix these problems are as follows:
Change Reaction.cs to not rely on the serialization of nullable floats. Change ReactionEditor.cs to be compatible with the changes to Reaction.cs. Use the new system on Ash.asset. Add a button to mark changes you make to chemical reactions in the reaction editor as dirty.
### Notes:
This **DOES NOT** fix the importer for reagents and reactions and does not modify any of the reactions (besides ash) to have minimum or maximum temperature requirements. 

### Changelog:
CL: [Fix] Fixes the bug where oil immediately turns to ash.


Bod9001:
waiting on redline to do review

PR #7094, opened 2021-07-20
Fixes players being unable to move

### LINKED ISSUE #7081: Clients being unable to move
some steps I found that reproduced it in the editor, 
having the movement go to 0 this usually happens from oxygen starvation,
you're able to move one it's futile but then after that, your position and so the server position of desynchronised,
though on the builds it does persist after a re-log

Aranclanos:
### Purpose
Players will no longer start a movement with 0 speed, getting stuck in the process.
Run, walk and crawl syncvars will now only be updated once in the same frame.
Fixes #7081

Note: I only focused on onyloss related movement slowdown, if there's more ways to lose the control of your character it hasn't been taken into consideration on this PR

### DIFF
--- a/UnityProject/Assets/Scripts/Player/PlayerMove.cs
+++ b/UnityProject/Assets/Scripts/Player/PlayerMove.cs
@@ -469,19 +469,30 @@ public class PlayerMove : NetworkBehaviour, IRightClickable, IServerSpawn, IActi
 
 	public void UpdateSpeeds()
 	{
-		RunSpeed = 0;
-		WalkSpeed = 0;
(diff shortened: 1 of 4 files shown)


corp-0:
so from your findings, the problem wasn't movement speed being zero but being a negative number? Do we ever want it to be zero? I'd insist in clamping it to crawlspeed, to be honest.

Aranclanos:
no, it was 0, just checking that way so codacy doesnt complain
But later in the day I noticed that your running speed could be 0 before your walk speed was 0, so depending on your damage level you could start walking faster once you took more dmg

PR #7095, opened 2021-07-20
Fixes PDA TC count not updating after using a telecrystal

### LINKED ISSUE #7053: Uplink TC count doesn't update in the PDA when adding TC until you close the uplink and reopen.

The counter for how much TC is left in your uplink does not update when you add more TC to the uplink until you close the uplink and reopen it. Then it updates to the amount you put in. It should update immediately after putting in extra TC.

Aranclanos:
### Purpose
Fixes PDAs not updating its TC counter UI once a telecrystal is used on them.
Fixes #7053

### DIFF
--- a/UnityProject/Assets/Scripts/Items/PDA/PDALogic.cs
+++ b/UnityProject/Assets/Scripts/Items/PDA/PDALogic.cs
@@ -8,6 +8,7 @@ using NaughtyAttributes;
 using Random = UnityEngine.Random;
 using UI.Action;
 using AddressableReferences;
+using UI.Items.PDA;
 
 namespace Items.PDA
 {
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Items/PDA/PDALogic.cs
```diff
@@ -356,6 +359,7 @@ private void ServerInsertItem(GameObject item, ItemSlot fromSlot, GameObject pla
 						$"After a moment it disappears, your Telecrystal counter ticks up a second later";
 
 					Chat.AddExamineMsgFromServer(player, uplinkMessage);
+					PDAGui.uplinkPage.UpdateTCCounter();
```
ThatDan123: Could this null if you spawn a PDA in and then use a telecrystal without first opening the UI?

Aranclanos: no, its set in the prefab of the tab

ThatDan123: But the tab wont be created until it has been opened the first time

Aranclanos: yeah whoops

PR #7100, opened 2021-07-21
[Draft] Breakable bodyParts and Blunt Force Trauma.

MaxIsJoe:
still WiP


MaxIsJoe:
Request : Please do not do any major changes to any of the BodyPart files while I'm working on this bounty, catching up with everything once I'm back tomorrow is going to be a headache.

(Yes I know that a lot of the health code was disorganized and ugly but I already was planning on cleaning it at some point + I've seen an intended behavior get removed during Arch's cleanup. At least ask me about what you're going to remove/adjust before touching any Trauma code)

MaxIsJoe:
Re-doing this in another PR after #7299 gets merged because this PR is so old

PR #7103, opened 2021-07-21
Prevents Drinks From Playing Global Sounds

ThatDan123:
-Drinks can no longer play global sounds


ThatDan123:
wdym compile errors

ThatDan123:
wdym compile errors

waffielz:
what if i wanna hear a centcom officer drink a soda from the next server over? smh

PR #7114, opened 2021-07-22
Cleanup and SingletonManager

darkvergus:
### Changelog:
Implemented SingletonManager and NetworkSingletonManager
General Cleanup
Normalized a bit of the code


UnityProject/Assets/Scripts/Util/SingletonManager.cs
```diff
@@ -0,0 +1,38 @@
+﻿using UnityEngine;
+
+public class SingletonManager<T> : MonoBehaviour where T : MonoBehaviour
```
corp-0, line 3: @ThatDan123 I remember you said something about the pattern new singletons should follow, can you take a look at this?

darkvergus: i have to work on this for a while since with the changes needed it requires alot of additions ill open it at a later time

corp-0: you could have ommitted the namespaces changes, at least for classes that weren't new. Anyways, don't be afraid to make smaller PRs and progresively implement your changes the next time!

darkvergus: sure thing i mostly just want to implement the singletonManager since there are quite a few managers which require it

darkvergus: also as far as I can see the project itself requires a lot of restructure

corp-0: honestly, yeah. Any help on that regard would make us happy.

UnityProject/Assets/Scripts/Util/NetworkSingletonManager.cs
```diff
@@ -0,0 +1,39 @@
...
+        get
+        {
+            if (shuttingDown)
+                return null;
+
+            lock (@lock)
+            {
+                if (instance == null)
+                {
+                    instance = (T)FindObjectOfType(typeof(T));
```
ThatDan123, line 21: FindObjectOfType is really bad for loading times, thats why most of the managers use the awake to set the instance

corp-0: Yeah, pretty sure darkvergus used that pattern because it is what he saw in most of your singletons. Can you provide an example of the right way to do it? So we can update our old singletons in the process by simply making them inherit this base singleton class darkvergus is doing.

ThatDan123:
```cs
public class SingletonManager<T> : MonoBehaviour where T : MonoBehaviour
{
	private static T instance;
	public static T Instance => instance;

	public virtual void Awake()
	{
		if (instance == null)
		{
			instance = this as T;
		}
		else
		{
			Destroy(gameObject);
		}
	}
}
```
  


Just need to call the base Awake in all the singleton scripts if they are already using Awake for something else

PR #7125, opened 2021-07-25
Small matrixmanager raycast performance fix

Aranclanos:
### Purpose
Small performance fix on the matrix manager raycast, on how it searches if the line of the raycast goes through a matrix's bounds

image
this exact thing inside the raycast is now around 4 times faster


Aranclanos:
DNM spotted something

PR #7138, opened 2021-07-26
Hotfixes handswap not working after a restart

Aranclanos:
### Purpose
Hotfixes handswap not working after a restart


corp-0:
I thought the agreed pattern was setting the instance in Awake

PR #7157, opened 2021-07-28
Make camera assemblies craftable.

PetMudstone:
### Purpose
Title. Camera assemblies can now be made using five metal sheets.


ThatDan123:
I might have made it so they can be constructed in the autolathe can you check please.

PetMudstone:
> 
> 
> I might have made it so they can be constructed in the autolathe can you check please.

I checked and camera assemblies are not constructible in the autolathe.

PR #7166, opened 2021-07-28
fixes item slots sending interaction twice

Bod9001:
Apparently sending an interaction twice is not a good thing, 

With the Item slot changes
itemStorage should always be the main component , there shouldn't be a time when it's a subobject

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 ---
 
 
+* 2021/07/27: **[New]** added tests for duplicate cables pipes by [Bod9001](https://github.com/Bod9001) in PR #[7147](https://github.com/unitystation/unitystation/pull/7147)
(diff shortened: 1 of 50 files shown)


ThatDan123:
Have you checked to make sure there isnt a storage on a child somewhere?

Bod9001:
well the problem is it would break every other item storage though

ThatDan123:
oh sorry didnt read the class as Item Slot, thought you were changing item storage stuff

ThatDan123:
Have you check that all the cases of item slot have the net identity on them?

ThatDan123:
What i mean is have you checked that all the item slot storages have their network identity on them? You've changed the hand prefab, but is that the only case?

Bod9001:
humm, hand prefab was fixing the "Apparently sending an interaction twice is not a good thing," thing

I don't believe anywhere there's a item with a sub game object that does the item storage on it well all item storage being parented to an item

PR #7175, opened 2021-07-30
Bunch of fixes

Bod9001:
just been looking through staging errors and fixing them


UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -192,6 +192,8 @@ public float MaxHealth
 
 		protected GameObject LastDamagedBy;
 
+		public List<BodyPart> DismemberingBodyParts = new List<BodyPart>();
```
corp-0: make it [NonSerialized]

UnityProject/Assets/Scripts/Items/Implants/Organs/Stomach.cs
```diff
@@ -9,7 +9,7 @@ namespace HealthV2
 {
 	public class Stomach : BodyPartModification
 	{
-		[NonSerialized] public ReagentContainer StomachContents;
+		public ReagentContainer StomachContents;
```
corp-0, line 12: why do we need to serialize the content?

corp-0: aaah nvm, it is the same thing as the liver container

corp-0:
does this close any recent issue?

Bod9001:
humm, not too sure

corp-0:
would be better to do individual PRs so we can keep track of what it fixes

Bod9001:
this would cover about 6 PRs then

corp-0:
6 PRs is fine. Another option would be at least stating in the PR description what this fixes, I can only guess by looking at the changes.

PR #7181, opened 2021-07-31
Fixes Facehugger not changing sprite after impregnation (#7132)

### LINKED ISSUE #7132: Facehuggers don't change sprites after they successfully infect someone and die.
### Environment
- Game Version: Production 4120
- OS Version: Windows 10

Title. Facehuggers don't change sprites after they latch onto someone's face die after successfully impregnating them. This can result in terror and confusion when a dead facehugger continues to wiggle about as if still alive.

### Steps to reproduce
1. Get impregnated by a facehugger.
2. After you get back up after being knocked down, examine the facehugger to see its dead description.
3. Notice that the dead facehugger is still wiggling.

AtnerNT:
### Purpose
Fixes  #7132 causing facehuggers to once again change sprites after they impregnate a host, the code was updating the clothing data but not the sprite itself

### Changelog:
CL: [Fix] Fixes a bug where facehuggers will still have their alive animation even after dying from impregnating a host.

### DIFF
--- a/UnityProject/Assets/Scripts/Clothing/FacehuggerImpregnation.cs
+++ b/UnityProject/Assets/Scripts/Clothing/FacehuggerImpregnation.cs
@@ -27,17 +27,16 @@ namespace Clothing
 		[Tooltip("Reference to larvae gameObject so we can spawn it")] [SerializeField]
 		private GameObject larvae = null;
 
-		[Tooltip("The SpriteHandler the facehugger should use once it successfully impregnates a host")] [SerializeField]
-		private SpriteHandler spriteHandler = default;
-
 		private bool isAlive = true;
 		private ClothingV2 clothingV2;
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Clothing/FacehuggerImpregnation.cs
```diff
@@ -27,6 +27,9 @@ public class FacehuggerImpregnation : NetworkBehaviour, IServerInventoryMove, IC
 		[Tooltip("Reference to larvae gameObject so we can spawn it")] [SerializeField]
 		private GameObject larvae = null;
 
+		[Tooltip("The SpriteHandler the facehugger should use once it successfully impregnates a host")] [SerializeField]
```
corp-0: all items should only use one sprite handler except for layered sprites. Could you get the sprite handler on awake instead?

PR #7182, opened 2021-07-31
fixes Atmospheric Tiles randomly becoming empty for like a millisecond

Bod9001:
this was done by not modifying the tile data until it was fully Calculated, 
also reduces GC on atmospheric slightly


Aranclanos:
looks fine, although dont manually type 2.7f there's a definition for that I believe

NoooneyDude:
This is still happening.

PR #7188, opened 2021-08-01
Adds Emag text feedback in chat

### LINKED ISSUE #6687: Emag text feedback is missing
Linux- staging

Loud emag interaction messages and so on don't work anymore
emagging lockers, doors and anything else makes it pretty clear

PerfectTangent:
### Purpose

This PR adds text feedback for when an Emag is used; closes #6687. In addition, more objects have flavor text for when they are emagged.

### Changelog:

CL: [Improvement] Emags give recognizable chat alerts to those nearby when used.

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Drawers/Morgue.cs
+++ b/UnityProject/Assets/Scripts/Objects/Drawers/Morgue.cs
@@ -2,6 +2,7 @@
 using UnityEngine;
 using AddressableReferences;
 using HealthV2;
+using Items;
 
 namespace Objects.Drawers
 {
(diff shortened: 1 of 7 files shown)


UnityProject/Assets/Scripts/Objects/Drawers/Morgue.cs
```diff
@@ -131,9 +131,8 @@ private void UseEmag(HandApply interaction)
 			alarmBroken = true;
 
 			Chat.AddActionMsgToChat(interaction,
-					$"You wave the {interaction.HandObject.name.ToLower()} over the {name.ToLower()}'s electrical panel. " +
 					"The status panel flickers and the buzzer makes sickly popping noises. You can smell smoke...",
```
NoooneyDude, line 140: Looks like the morgue hasn't been updated to use up a charge, so the removed text here won't be displayed.
You'll want get the emag component and check that it has charges on line 53 via `TryGetComponent<>()`, then you can pass this variable into the `Morgue.UseEmag()` method, here, to call the `UseCharge()` method on the component.

PerfectTangent: alright, cool. Done.

PR #7196, opened 2021-08-02
Patching missing tiles in Dark Jungle.

PetMudstone:
### Purpose
- I noticed the Dark Jungle scene had a lot of missing tiles so I replaced them. Easy as.


PetMudstone:
Can we rerun tests on this? Local tests succeeded, I sucessfully made a build, and the artifact only has a .log  file for some reason.

PR #7199, opened 2021-08-03
Forces an account being logged only once at the same time

### LINKED ISSUE #4031: Login Bug Exploit - Single User can create multiple characters on the same server with the same account
### Description

Was able to make 2 or more versions of the same character(s); i.e. clones having the same role **assistant**. 

### Detailed Bug Report

Before the game started, I accidentally started two game sessions whilst logged in the same account and thereby made 2 duplicate or clone characters. 

I then tried to replicate this bizarre occurrence a third time by trying to create another character. All three of the characters have the same profession, same features and the same character name _Jeffrey Steele_. To my surprise, the characters were successfully created. However upon creating the third copy of the character that specific game window crashed leaving me with a fully working third copy but no way to control the character. 

_A critical point to note is that upon trying to create a **fourth** and **fifth** iteration of the same character...my system's memory couldn't handle the load since I also had some other background processes running and those respective game copies crashed. _

This bug would be a critical issue if left unfixed as players could avoid death in a game session by just controlling their cloned characters. And or perhaps overflooding the server with clones. 

The server I was running this on was Unitystation - USA MRP; build - 4091

### Steps to Reproduce

Steps to reproduce the bug or behavior:  

1.  Upon starting the game, either click multiple times on _PLAY_  in the Station Hub and or drag the options on the game icon with the left button of the mouse. As such -->
image

2.  Proceed through the character creation and profession selection menus for each desired copy or clone. 

3.  You should end up with all three cloned characters starting in the same loadout area if you choose the same profession that is. 

UnityStation-Login-Glitch-Exploi

**UPDATE 04/25**

As requested by @DooblyNoobly  recreated the issue successfully on LRP Spess Adventures. 

In addition to recreating this, we tried to test this issue out with the same character but using different professions. We observed the following using the given info: 

-The same name _Jeffery Steele_ 
-The first profession selected was Botanist. Second was selected as Botanist but was assigned clown by the game or system. 

unitystation 4_24_2020 11_47_53 PM

Results 

- I noticed that this time however the system assigned the clone with a character name of "Unknown" 
- But in the chat name assigned to the clone was "Jefferey Steele2" 
- Despite assigning the profession as a botanist the system assigned the clone the profession of "clown" 
- I could interact with both original and clone players at the same time
- I could chat with other players at the same time. 

** Additionally I tasked @rishabhyata with performing a similar test with same character and clone but different profession (Assistant & Security Officer); he successfully replicated the bugs as well. His screenshots are added in comments

### LINKED ISSUE #7054: It is possible to rejoin the round as a new character by simply switching to one in character customization.
### Environment
- OS Version: Windows 10

It is possible to join a round again as a new character by switching to them. This allows players to multikey and get themselves out of various situations such as being in perma or being dead. I have not been able to replicate this myself but I have witnessed or heard reports of multiple players who exploited this bug on Solidarity.

### Steps to reproduce
1. Join a server as a character.
2. Log out.
3. Switch to another character.
4. Rejoin the server, you might freshly spawn as the new character you chose.

Aranclanos:
### Purpose
Upon client login, the server will now disconnect another client if its using the same account.
Fixes relogging fast not giving you back your character.
Fixes #7054
Fixes #4031

### Notes:
There's a special check for the server player, so testing with the same account as host & client can be done without issues


PerfectTangent:
> There's a special check for the server player, so testing with the same account as host & client can be done without issues

Nice, I think this is a really good compromise to not impede devving.

NoooneyDude:
Does this mean we can only connect at most one other client locally?

Edit: yeah. You could change the check to see if at least one player is the host, then allow any amount of extra clients.

When testing builds, I usually spin up one headless server and two clients.

PR #7223, opened 2021-08-09
Health - body part and organ code separation

### LINKED ISSUE #6948: Torso inside torso invincibility surgery
### Environment
- Game Version: Staging (21070515)
- OS Version: Arch Linux

You are able to attach a loose torso from a corpse inside your own chest, leaving you first with any health analysis broken and second with some kind of invincibility unable to die, here's a screenshot of me without a hard-suit 10 minutes in the void of space.
robot

### Steps to reproduce

1. Find loose chest from a corpse
2. start surgery and open torso
3. attach loose chest inside you own torso
4. close and enjoy invincibility

Extra (you are even able to open the chest inside the torso inside your chest)
(possible infinite torso inside torso)

Aranclanos:
### Purpose
Cleans up a big chunk of health code, making it simpler and easier to understand without losing any features

Removes RootBodyPartContainer
Removes the usage of dynamic classes
Starts the separation of body parts and organs, with the idea of them being different from eachother. With the health master holding body parts and the body parts holding organs.
Removes organs components from body parts prefabs.
Cleans up a bit gibbing code.
Adds a couple of TODOs to continue the separation and clean-up, as for now organ prefabs still have a body part component.
Fixes #6948

CL:[fix] you can no longer become unkillable by surgically attaching a torso within another torso.


MaxIsJoe:
woohoo, finally.

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
```diff
@@ -15,52 +16,48 @@ namespace HealthV2
...
+		[NonSerialized]
+		public List<Organ> OrganList = new List<Organ>();
+
+		/// <summary>
+		/// Player sprites for rendering equipment and clothing on the body part container
+		/// </summary>
+		[Tooltip("Player sprites for rendering equipment and clothing on this")]
+		public PlayerSprites playerSprites;
+
+		[HideInInspector] public bool IsBleeding = false;
```
NoooneyDude, line 39:
```suggestion
		[NonSerialized]
		public bool IsBleeding = false;
```
Guessing we could make this  `[NonSerialized]`? Few others marked `[HideInInspector]` that could probably be treated the same, though no biggie.

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -1260,6 +1147,100 @@ protected virtual void LethalElectrocution(Electrocution electrocution, float sh
...
+				implant.RelatedPresentSprites.Add(newSprite);
+
+				var newOrder = new SpriteOrder(sprites.Item1);
+				newOrder.Add(i);
+				newSprite.UpdateSpritesForImplant(implant, implant.ClothingHide, Sprite, newOrder);
+				InternalNetIDs.Add(newSprite.GetComponent<NetworkIdentity>().netId);
+
+				i++;
+				i++;
+				i++;
```
NoooneyDude:
```suggestion
				i += 3;
```
Increment thrice? If that's what you want, perhaps `i += 3` with a quick comment would look less like a mistake.

UnityProject/Assets/Scripts/HealthV2/Living/Surgery/Procedures/ImplantProcedure.cs
```diff
@@ -21,14 +21,14 @@ public class ImplantProcedure : SurgeryProcedureBase
 			{
 				if (OnBodyPart != null)
 				{
-					OnBodyPart.AddBodyPartSlot(interaction.HandSlot);
+					OnBodyPart.OrganStorage.ServerTryTransferFrom(interaction.HandSlot);
 				}
 				else
 				{
-					PresentProcedure.ISon.GetComponent<LivingHealthMasterBase>().GetRootBodyPartInZone(interaction.TargetBodyPart).AddBodyPartSlot(interaction.HandSlot);
+					PresentProcedure.ISon.GetComponent<LivingHealthMasterBase>().BodyPartStorage.ServerTryTransferFrom(interaction.HandSlot);
```
NoooneyDude, line 28: `ISon`? Oh, the target body I take it.

I like the `Finnish surgery procedure`, too. 😄

UnityProject/Assets/Scripts/Items/Medical/Gauze.cs
```diff
@@ -31,47 +31,41 @@ public override void ServerPerformInteraction(HandApply interaction)
...
 			{
-				if(container.BodyPartType == interaction.TargetBodyPart)
+				if(bodyPart.BodyPartType == interaction.TargetBodyPart)
 				{
-					foreach(BodyPart limb in container.ContainsLimbs)
+					if(bodyPart.IsBleedingExternally)
 					{
-						if(limb.IsBleedingExternally)
+						bodyPart.StopExternalBleeding();
+						if(interaction.Performer.Player().GameObject == interaction.TargetObject.Player().GameObject)
```
NoooneyDude, line 43:
```suggestion
						if (interaction.Performer == interaction.TargetObject)
```

UnityProject/Assets/Scripts/Items/Medical/Gauze.cs
```diff
@@ -31,47 +31,41 @@ public override void ServerPerformInteraction(HandApply interaction)
...
-							}
-							else
-							{
-								Chat.AddActionMsgToChat(interaction.Performer.gameObject,
-								$"You stopped {interaction.TargetObject.Player().Script.visibleName}'s bleeding.",
-								$"{interaction.PerformerPlayerScript.visibleName} stopped {interaction.TargetObject.Player().Script.visibleName}'s bleeding.");
-							}
+							Chat.AddActionMsgToChat(interaction.Performer.gameObject,
+							$"You stopped your {interaction.TargetObject.Player().Script.visibleName}'s bleeding.",
+							$"{interaction.PerformerPlayerScript.visibleName} stopped their own bleeding from their {interaction.TargetObject.ExpensiveName()}.");
```
NoooneyDude, line 47: "You stopped your Clarke Hanford's bleeding." 🤔
"Clarke Hanford stopped their own bleeding from their Clarke Hanford." 😆 May as well fix this while it is in the light.
```suggestion
							Chat.AddActionMsgToChat(interaction.Performer,
							$"You stopped the bleeding on your {bodyPart}.",
							$"{interaction.PerformerPlayerScript.visibleName} stopped the bleeding on their {bodyPart}.");
```
Though `bodyPart` probably doesn't print nicely. Does it have an attributes component for `bodyPart.ExpensiveName()`?

Should probably check this before applying suggestion. Same goes for the next.

UnityProject/Assets/Scripts/Items/Medical/Gauze.cs
```diff
@@ -31,47 +31,41 @@ public override void ServerPerformInteraction(HandApply interaction)
...
+						}
+						else
+						{
+							Chat.AddActionMsgToChat(interaction.Performer.gameObject,
+							$"You stopped {interaction.TargetObject.Player().Script.visibleName}'s bleeding.",
+							$"{interaction.PerformerPlayerScript.visibleName} stopped {interaction.TargetObject.Player().Script.visibleName}'s bleeding.");
```
NoooneyDude, line 53:
```suggestion
							Chat.AddActionMsgToChat(interaction.Performer,
							$"You stopped the bleeding on {interaction.TargetObject.Player().Script.visibleName}'s {bodyPart.ExpensiveName()}.",
							$"{interaction.PerformerPlayerScript.visibleName} stopped the bleeding 
 on {interaction.TargetObject.Player().Script.visibleName}'s {bodyPart.ExpensiveName()}."); // Might read a lil weird for the recipient.
```

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -128,13 +121,16 @@ public float MaxHealth
 		/// <summary>
 		/// A list of all body parts of the creature
 		/// </summary>
-		public HashSet<BodyPart> BodyPartList = new HashSet<BodyPart>();
+		public List<BodyPart> BodyPartList = new List<BodyPart>();
```
NoooneyDude, line 123: Just curious, is there something about `HashSet` that means we should avoid it, or were the collection's advantages just not used in this case?

NoooneyDude (dismissed):

Bod9001:
"Starts the separation of body parts and organs, with the idea of them being different from eachother. With the health master holding body parts and the body parts holding organs.
Removes organs components from body parts prefabs."

I told you We are not doing that,  because the system requires that each Body part (includes organs) Is a game object that makes up the player

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartController.cs
```diff
@@ -14,49 +15,33 @@ public class RootBodyPartController : NetworkBehaviour
 	[SyncVar(hook = nameof(UpdateChildren))]
 	private string SerialisedNetIDs = "";
 
-	public List<RootBodyPartContainer> RootBodyParts = new List<RootBodyPartContainer>();
+	public List<uint> ToSerialised = new List<uint>();
```
Bod9001: this stuff is changed in the Unity update

UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartDamage.cs
```diff
@@ -273,11 +273,6 @@ public float TotalDamage
 			}
 		}
 
-		public void DamageInitialisation()
-		{
-			this.AddModifier(DamageModifier);
```
Bod9001, line 278: where has this gone?

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
Bod9001: There were occasions where double body parts ended up getting added so I used HashSet to just prevent that

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -128,13 +121,16 @@ public float MaxHealth
...
 
 		/// <summary>
-		/// A list of all body part containers of the creature.
-		/// A body part container is a grouping of like body parts (legs, arms, eyes, etc)
+		/// The storage container for the body parts
 		/// </summary>
-		public List<RootBodyPartContainer> RootBodyPartContainers = new List<RootBodyPartContainer>();
+		[HideInInspector]
+		public ItemStorage BodyPartStorage;
```
Bod9001, line 129: the player shouldn't have item storage, All they have is dynamic item storage

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -536,7 +512,6 @@ private void SetConsciousState(ConsciousState NewConsciousState)
 					healthStateController.SetConsciousState(NewConsciousState);
 					OnConsciousStateChangeServer.Invoke(oldState, NewConsciousState);
 				}
-				ConsciousState = NewConsciousState;
```
Bod9001, line 538: Doesn't this need to be set?

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -554,20 +529,13 @@ private void SetConsciousState(ConsciousState NewConsciousState)
...
 			{
-				float bodyParts = RootBodyPartContainers.Sum(Container => Container.ContainsLimbs.Count());
+				float bodyParts = BodyPartList.Count;
 				damage /= bodyParts;
 			}
 
-			foreach (var Container in RootBodyPartContainers)
+			foreach (var bodyPart in BodyPartList)
 			{
-				if (damageSplit)
```
Bod9001, line 562: what happened to damage Split?

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -801,41 +722,20 @@ public RootBodyPartContainer GetRootBodyPartInZone(BodyPartType bodyPartAim)
...
-		/// <summary>
-		/// Gibs the creature if possible, not implemented
-		/// </summary>
-		private void TryGibbing(float damage)
-		{
-			//idk
-			//TODO: Reimplement
-		}
-
-		public void CheckDismemberBody()
```
Bod9001, line 829: Where is this handled?

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -1260,6 +1147,100 @@ protected virtual void LethalElectrocution(Electrocution electrocution, float sh
...
+		/// </summary>
+		/// <param name="implant">Body Part to display</param>
+		public void SetupSpritesNetId(BodyPart implant)
+		{
+			int i = 0;
+			bool isSurfaceSprite = implant.IsSurface;
+			var sprites = implant.GetBodyTypeSprites(playerSprites.ThisCharacter.BodyType);
+			foreach (var Sprite in sprites.Item2)
+			{
+				var newSprite = Spawn.ServerPrefab(implant.SpritePrefab.gameObject, Vector3.zero, playerSprites.BodySprites.transform).GameObject.GetComponent<BodyPartSprites>();
```
Bod9001, line 1102: note Might have to adjust some sprite rendering orders Since if it'll been thrown on the same object misses separate objects before it might change the rendering order

UnityProject/Assets/Scripts/Items/Implants/Organs/BodyPartUISlots.cs
```diff
@@ -24,25 +24,6 @@ public class BodyPartUISlots : Organ, IDynamicItemSlotS
 		[SerializeField] [FormerlySerializedAs("Storage")]
 		private List<BodyPartUISlots.StorageCharacteristics> storage;
 
-		public override void RemovedFromBody(LivingHealthMasterBase livingHealthMasterBase)
```
Bod9001, line 27: where is this stuff handled now?

UnityProject/Assets/Scripts/Items/Implants/Organs/Stomach.cs
```diff
@@ -47,22 +47,15 @@ public override void ImplantPeriodicUpdate()
 
 			if (AllFat)
 			{
-				StartCoroutine(DelayAddFat());
+				var Added = Spawn.ServerPrefab(BodyFatToInstantiate.gameObject).GameObject.GetComponent<BodyFat>();
+				BodyFats.Add(Added);
+				RelatedPart.OrganStorage.ServerTryAdd(Added.gameObject);
```
Bod9001, line 52: the problem here is it causes the loop to get an extra element while looping through

UnityProject/Assets/Scripts/Items/Weapons/Projectiles/Behaviours/ProjectileIncendiary.cs
```diff
@@ -65,15 +65,9 @@ public bool OnHit(MatrixManager.CustomPhysicsHit hit)
 
 			if (hit.CollisionHit.GameObject == null) return true;
 
-			//TODO REMOVE AFTER MOBS ARE MOVED TO NEW HEALTH
-			if (hit.CollisionHit.GameObject.TryGetComponent(out LivingHealthMasterBase health))
```
Bod9001, line 69: have mobs been moved over to new health?

UnityProject/Assets/Scripts/Player/PlayerSprites.cs
```diff
@@ -182,68 +174,55 @@ public void SubSetBodyPart(BodyPart Body_Part, string path)
 
 		if (customisationStorage != null)
 		{
-			bool pass = true;
+			var data = customisationStorage.Data.Replace("@£", "\"");
```
Bod9001, line 177: a good point here is that you need customisation for organs to

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -705,12 +673,13 @@ BodyPartType bodyPartAim
...
 			LastDamagedBy = damagedBy;
 
-			foreach (var bodyPartContainer in RootBodyPartContainers)
+			foreach (var bodyPart in BodyPartList)
 			{
-				if (bodyPartContainer.BodyPartType == bodyPartAim)
+				if (bodyPart.BodyPartType == bodyPartAim)
 				{
 					//Assuming is only going to be one otherwise damage will be duplicated across them
-					bodyPartContainer.TakeDamage(damagedBy, damage, attackType, damageType, armorPenetration);
```
Bod9001, line 712: so,  Since this was expecting root body containers, 
the extra logic for having multiple body parts in an area would be handled by them

so, this has to be changed to take account of how many body parts are there, 
so for Bomb , Fire , Rad it splits the damage over all body parts in target area

For everything else it picks a random body part and applies the damage to that

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -554,20 +529,13 @@ private void SetConsciousState(ConsciousState NewConsciousState)
 		{
 			if (damageSplit)
 			{
-				float bodyParts = RootBodyPartContainers.Sum(Container => Container.ContainsLimbs.Count());
+				float bodyParts = BodyPartList.Count;
 				damage /= bodyParts;
 			}
 
-			foreach (var Container in RootBodyPartContainers)
+			foreach (var bodyPart in BodyPartList)
```
Bod9001: You can't use BodyPartList since that would include organs as well, you need to do damage to the root body parts to propagate the damage down properly through the different armour of the body parts

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
MaxIsJoe: that should be a private bool, with a `IsBleeding => isBleeding`

UnityProject/Assets/Scripts/Items/Implants/Organs/Stomach.cs
MaxIsJoe: I think we need a comment to make sure people are aware of that

UnityProject/Assets/Scripts/Items/Weapons/Projectiles/Behaviours/ProjectileIncendiary.cs
MaxIsJoe: not yet, but I've been working on making damage work between all mobs again. not just the player.

Aranclanos: this was checking for the new class again a few lines below but yeah I guess I'll add it back with the old class

UnityProject/Assets/Scripts/Items/Implants/Organs/Stomach.cs
Aranclanos: the preriodic update of organs now loops backwards to avoid that

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
Aranclanos: thats for another pr focused on player sprites, this particular function was just moved

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
Aranclanos: this doesn't seem to happen anymore, I'm unaware of the cause of the bug

UnityProject/Assets/Scripts/Items/Medical/Gauze.cs
Aranclanos: the target object seems to be a body part, not the person

UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartDamage.cs
Aranclanos: was called in awake so moved it there

UnityProject/Assets/Scripts/Items/Medical/Gauze.cs
MaxIsJoe: Like I said on discord, better get the bodyPart's name from the BodyPartDamageDiscreptor script as it's designer defined and constant.

UnityProject/Assets/Scripts/Items/Implants/Organs/Stomach.cs
Bod9001: well what about when you remove element?

UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartDamage.cs
Bod9001: can't see in awake in BodyPartDamage.cs

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
Aranclanos: since there aren't body parts inside of body parts, the damagesplit dmg value is calculated a few lines above and then applied to all

UnityProject/Assets/Scripts/Items/Implants/Organs/Stomach.cs
Aranclanos: it would runtime, I'll make a reverse for loop

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
Aranclanos: BodyPartList doesnt include organs anymore

Aranclanos:
damage is the weakest part of this PR, I applied some changes to somewhat patch it up to later handler in another PR, but at the moment they are untested until later today

corp-0:
the crafting PR added PlayerCrafting to the player prefab, so to solve the conflict just use yours and add the crafting component.

UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartDamage.cs
Aranclanos: partial class

Aranclanos:
Updated with everything talked above solved, although the issues with the strings of gauzes remain, they are just out of the scope for this pr

UnityProject/Assets/Scripts/Items/Medical/Gauze.cs
NoooneyDude: Not sure I understand your comment. `interaction.Performer` is the performing player's gameobject, and the `Player()` extension returns `ConnectedPlayer`. The `GameObject` member is a reference to the `ConnectedPlayer`'s gameobject, same as previous, so it could just be shortened to `interaction.Performer`.

If you want to call it as outside the scope of the PR as it was already like that, that's fine, but it's a pretty quick change.

Aranclanos: targetobject is a bodypart, and thats a different network object, so player() doesnt get the connection if you use it, I guess that's the reason of why using gauzes on yourself doesnt trigger the proper message

NoooneyDude: Ooh. I getcha. Thanks.

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
Bod9001: how?

UnityProject/Assets/Scripts/Items/Implants/Organs/Stomach.cs
Bod9001: Wouldn't it still miss an element out?

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
Bod9001: there could still be multiple in the same area so for example implanting multiple arms

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
Aranclanos: I'm just not adding organs to this list anymore

UnityProject/Assets/Scripts/Items/Implants/Organs/BodyPartUISlots.cs
Aranclanos: SetHealthMaster() in BodyPart

UnityProject/Assets/Scripts/Items/Implants/Organs/Stomach.cs
Aranclanos: no, its the point of the reverse loop -- then again this is handled a bit weirdly already by a disbowel call so it isnt needed

UnityProject/Assets/Scripts/Player/PlayerSprites.cs
Aranclanos: there arent many changes on how sprites are handled, mostly its just moving stuff around, like in this case. If the concern is customization for organs (beards, hair, eyes, etc) its working as previously

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/RootBodyPartController.cs
Aranclanos: understandable, however this list was only updated to discard the RootBodyParts

Aranclanos:
Damage targeted to one body part type will now spread across all of the same type instead of picking the first one.

aka, if you aim towards a right arms for 20 damage and your target has 2 right arms, you will deal 10 to each one

Aranclanos:
note: this is a change requested in this PR based on inconsistence - example: if a body part was inside a body part, the damage would spread, but if the same body part was in a root body container it would not spread and pick the first. It depended on where the person clicked when using the surgery menu.

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
Aranclanos: solved on last commit

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
Aranclanos: solved on last commit

Aranclanos:
fixed conflicts

UnityProject/Assets/Scripts/Items/Implants/Organs/BodyPartUISlots.cs
Bod9001: why Separate it out?

Aranclanos:
updated

Bod9001:
just saying that the admin Heal loops for everybody part and that would have to include organs to so you don't instantly drop dead from having a damaged liver or heart

Bod9001 (approved):
apart from the few questions it looks good

corp-0 (approved):
after this approval, only one unresolved conversation left to merge.

UnityProject/Assets/Scripts/Items/Implants/Organs/BodyPartUISlots.cs
Aranclanos: more organized that way, plus this is no longer an override since its not longer an organ but a monobehaviour

PR #7227, opened 2021-08-09
apparently this was never merged?

Bod9001:
general bug fixes


Bod9001:
nah was looking at wrong branch

PR #7232, opened 2021-08-09
Remove Gib Check

### LINKED ISSUE #7204: Cant interact with corpses properly
### Environment
- Game Version:  Build 21080118
- OS Version: windoes 10

On the staging branch it is not possible to butcher animals or loot bodies

### Steps to reproduce
1.die
2.get cloned
3.try and loot your old self
4.anger

1. kill goat
2. click goat with knife
3. starve

ThatDan123:
Fixes #7204

-Players dont use old health so dont need to check for allowed gibbing

-Note: new health doesnt have gibbing check but it hasnt been reimplemented yet


corp-0:
uhm, probably the right move would be reimplementing gib check in new health and keep the configuration

corp-0:
ah wait, this is only removing from LivingHealthBehaviour, my bad.

PR #7233, opened 2021-08-09
Fix Emotes

### LINKED ISSUE #7155: Emotes do not work as headless clients.
### Environment
- Game Version: Staging 21080118 and Production 4121
- OS Version: Windows 10

When you try to emote (for example, typing "*dance" sans quotes in chat) as a client connected to a headless server, the special message associated with the emote as well as any additional actions (such as sprite movement) will not appear. Emotes still work for hosts and clients connected to localhosts.

### Steps to reproduce
1. Join a server off the hub.
2. Once in-game as your character, type in "*backflip" in chat.
3. Watch as nothing happens.

ThatDan123:
Fixes #7155

-Haven't built tested but still works in editor so at least wont be a another regression, but the logic should work now 

Tested in editor with both *dance and /me dance

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Chat/Chat.Process.cs
+++ b/UnityProject/Assets/Scripts/Core/Chat/Chat.Process.cs
@@ -48,8 +48,6 @@ public partial class Chat
 	public Color blobColor;
 	public Color defaultColor;
 
-	private static GameObject playerGameObject;
-
 	private static bool playedSound;
 
 	/// <summary>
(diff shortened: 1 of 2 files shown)


corp-0:
original problem was it not working on clients. I will build this later today if not tomorrow. If anyone can build and test before me, that'd be great as well.

PetMudstone:
I will investigate.

PetMudstone:
So I have good news after testing on headless server, the *dance, *tremble, and *twitch emotes make your character move as they should. Bad news, the chat message associated with each emote doesn't appear. This is problematic since most emotes are text only at the moment. Additionally, the *backflip emote doesn't rotate your character as it should and the *surrender emote is also non-functional and doesn't make your character lie down.

I also have been unable to test using multiple clients on a headless server since the first client gets DC'd when a second client joins, though that isn't the fault of anything in this PR I'm pretty sure.

NoooneyDude:
> So I have good news after testing on headless server, the *dance, *tremble, and *twitch emotes make your character move as they should. Bad news, the chat message associated with each emote doesn't appear. This is problematic since most emotes are text only at the moment. Additionally, the *backflip emote doesn't rotate your character as it should and the *surrender emote is also non-functional and doesn't make your character lie down.
> 
> I also have been unable to test using multiple clients on a headless server since the first client gets DC'd when a second client joins, though that isn't the fault of anything in this PR I'm pretty sure.

Yep, was a result of #7199. Annoying. Thanks for testing.

Edit: oh, is the lack of emote message a regression or is it already like that in develop?

ThatDan123:
ok, moved the logic around again. Should now do the message and action correctly

PR #7235, opened 2021-08-10
Examine text will now show how many items there is now in a stack

### LINKED ISSUE #7225: Examining stacks of items should tell you how many items are in each stack.
### Description
Title. The message in chat when you examine a stack of items should tell you how many items are in that stack.

MaxIsJoe:
title. closes #7225 

### Changelog:
CL: [New] - Stackable items' examine text now shows how many items they're holding.

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Stackable.cs
+++ b/UnityProject/Assets/Scripts/Items/Stackable.cs
@@ -346,7 +346,7 @@ public class Stackable : NetworkBehaviour, IServerLifecycle, ICheckedInteractabl
 	{
 		string displayName = gameObject.ExpensiveName();
 
-		string str = "This " + displayName + " contains (" + Amount + ") stacks.";
+		return $"This {gameObject.ExpensiveName()} contains {Amount} stacks.";
 
 		return str;
 	}
--- a/UnityProject/Assets/Scripts/Items/Stackable.cs
+++ b/UnityProject/Assets/Scripts/Items/Stackable.cs
@@ -8,7 +8,7 @@ using UnityEngine;
 /// <summary>
 /// Allows an item to be stacked, occupying a single inventory slot.
 /// </summary>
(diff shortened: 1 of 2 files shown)


corp-0:
use the stackable component itself instead

MaxIsJoe:
> use the stackable component itself instead

but.. that's what I'm doing?

corp-0:
I meant, instead of adding a reference to Stackable and a check in Attributes component, implement IExaminable directly in Stackable component. Then you need no reference and no check. It is also cleaner because that string is only relevant for objects that are stackable.

MaxIsJoe:
Done.

UnityProject/Assets/Scripts/Items/Stackable.cs
```diff
@@ -341,4 +341,13 @@ public void ServerPerformInteraction(HandApply interaction)
 	{
 		ServerCombine(interaction.TargetObject.GetComponent<Stackable>());
 	}
+
+	public string Examine(Vector3 worldPos)
+	{
+		string displayName = gameObject.ExpensiveName();
+
+		string str = "This " + displayName + " contains (" + Amount + ") stacks.";
```
corp-0:
```suggestion
		return $"This {gameObject.ExpensiveName()} contains {Amount} stacks.";
```
I think it is fine without the variables.

MaxIsJoe: added

MaxIsJoe:
fixed

UnityProject/Assets/Scripts/Items/Stackable.cs
```diff
@@ -341,4 +341,9 @@ public void ServerPerformInteraction(HandApply interaction)
 	{
 		ServerCombine(interaction.TargetObject.GetComponent<Stackable>());
 	}
+
+	public string Examine(Vector3 worldPos)
+	{
+		return $"This {gameObject.ExpensiveName()} contains {Amount} stacks.";
```
NoooneyDude, line 347: "This `metal sheet` contains 20 `stacks`."? "It contains 20 items." would be better.

PR #7238, opened 2021-08-10
[Draft] Loud Headsets for Command.

### LINKED ISSUE #7119: Loud headsets for Heads of Staff
### Description

I am unsure of their name on /tg/, but they were headsets that made your radio chatter 2 times as bigger, thus making your messages more noticeable. (And more annoying)

MaxIsJoe:
WIP. closes #7119

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Headset.cs
+++ b/UnityProject/Assets/Scripts/Items/Headset.cs
@@ -4,13 +4,37 @@ using Mirror;
 /// <summary>
 ///     Headset properties
 /// </summary>
-public class Headset : NetworkBehaviour
+public class Headset : NetworkBehaviour, ICheckedInteractable<HandApply>
 {
 	[SyncVar] public EncryptionKeyType EncryptionKey;
+	public bool IsLoudHeadset = false;
+	[SyncVar] public bool LoudHeadsetEnabled = false;
 
 	public void init()
 	{
 		getEncryptionTypeFromHier();
+	}
+
+	public bool WillInteract(HandApply interaction, NetworkSide side)
+	{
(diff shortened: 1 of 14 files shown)


UnityProject/Assets/Scripts/Items/Headset.cs
```diff
@@ -4,15 +4,39 @@
 /// <summary>
 ///     Headset properties
 /// </summary>
-public class Headset : NetworkBehaviour
+public class Headset : NetworkBehaviour, ICheckedInteractable<HandApply>
```
ThatDan123, line 7: Think you want HandActivate?

corp-0:
I'm not convinced using speech modifiers is the way to go. This should only affect how text is displayed in chat.

MaxIsJoe:
> I'm not convinced using speech modifiers is the way to go. This should only affect how text is displayed in chat.

Any suggestions on how to approach this then?

UnityProject/Assets/Scripts/Items/Headset.cs
MaxIsJoe: What's the difference? HandApply already seems to work

ThatDan123: Well hand apply is using it on something, hand activate is clicking on it in hand.

corp-0:
> Any suggestions on how to approach this then?

I would look for the code that sends text to chat and try to make it flexible so we can pass font size and style as parameters. This would also help to make chat more dynamic, like in ss13, where people screaming appear bigger in chat.

MaxIsJoe:
Loudspeak PR is too old that it still uses the old version of Unity, starting from scratch.

PR #7241, opened 2021-08-10
Some step-by-step guides for the new crafting system

Lizurt:
### Purpose

Adds the small step-by-step guides for the new crafting system.

### Notes

Wasn't tested because I don't even know how to open this `.md` file in a browser.

Please check my grammar on the page. English is definitely not my native language.

### DIFF
--- a/docs/development/Crafting-System.md
+++ b/docs/development/Crafting-System.md
@@ -14,7 +14,7 @@ This page will help you to create/remove recipes/categories. Any detailed inform
 6. Press the "Add to the singleton if necessary" button.
 7. Edit your new recipe freely(tooltips will help you).
 8. Add the opportunity for a player to learn this recipe.
-  * You can add the recipe to the Player's default known recipes list in the PlayerCrafting component.
+  * You can add the recipe to the default known recipes list in the PlayerCrafting component of the Player_V4 prefab.
   * You can add the recipe to a recipe book.
 
 ### How to delete a crafting recipe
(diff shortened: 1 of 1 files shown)


docs/development/Crafting-System.md
```diff
@@ -0,0 +1,90 @@
...
+
+1. Open the Unity Editor 
+2. Follow the path: /Assets/ScriptableObjects/Crafting/Recipes/
+3. Find a necessarry category to put a new recipe into.
+4. RMB -> Create -> ScriptableObjects -> Crafting -> CraftingRecipe
+5. Give the new recipe a name. 
+6. Press the "Add to the singleton if necessary" button.
+7. Edit your new recipe freely(tooltips will help you).
+8. Add the opportunity for a player to learn this recipe.
+  * You can add the recipe to the Player's default known recipes list in the PlayerCrafting component.
```
NoooneyDude: Might want to specify where the `PlayerCrafting` component is.

Though, discussion in Discord suggests moving this into an SO for creating recipe collections, and then having a list of these recipe collections defined on the occupation SOs, at some point. Something like that.

Lizurt: > Might want to specify where the PlayerCrafting component is.

Well it's already specified or something like that:

> ... to the Player's ...

Do you mean it's better to write `... to the Player_V4's ...`?

NoooneyDude: Yeah, that it is on the player prefab.

Lizurt: > Though, discussion in Discord suggests moving this into an SO for creating recipe collections, and then having a list of these recipe collections defined on the occupation SOs, at some point. Something like that.

I think we may edit the documentation when we implement this.

NoooneyDude: Yep, outside the scope of this one.

Issue #6826, opened 2021-06-19
Overlap Pipe Error


ThatDan123:
-Turn on the pipes warning setting in the logger

-Play on square and you'll get warnings about overlapping pipes at these coords

![image](https://user-images.githubusercontent.com/56727168/122643001-1efef880-d105-11eb-81b8-939c670278be.png)

-Delete pipe and place again using tile palette, and the warning still happens.

-If you turn off the auto delete duplicate pipe in UnderFloorLayer line 60 (by the log) and go in game and unwrench the pipes at the coords you'll see theres a duplicate pipe which you cant interact with.

-It is not a direct error with placing pipes using the brush, I put logs there and only one pipe was place when testing (using the box tool, the paint tool can duplicate).

AtnerNT:
We have a test for this now and all cases of overlapping pipes/wires have been removed

MaxIsJoe:
It's a bit annoying that Unity still lets you paint overlapping pipes, but at least we prevent PRs from being merged when they have them.

PR #7251, opened 2021-08-13
Fixes pipes sharing the same tile affecting server performance

Aranclanos:
### Purpose
Fixes pipes sharing the same tile affecting performance and leaving a ghost pipe sprite once removed.

-what was wrong with the old check?
it saved and used a SO to check for directions, which were always the same


ThatDan123:
Do you think this has fixed #6826 ?

PR #7255, opened 2021-08-13
Titanium wall, window, floor, and airlock resprites by Onule.

PetMudstone:
### Purpose
- Resprites all the titanium tiles and airlocks as determined by popular vote in a poll. Sprites taken from [this PR](https://github.com/tgstation/tgstation/pull/50454) by @Onule.
  - Older styled titanium floor tiles are still around for both mappers and players to use (see below). Their colors have been adjusted to match the new titanium floor tiles.
- Titanium floor tiles can now be used in-hand to convert them to other kinds of titanium tiles, including older styled ones.
  - Plastitanium tiles can also be used in-hand to convert them to red tiles. 
- Titanium and plastitanium floors now give tile items with the same color as them.
image
image

#### Credits
@Onule: New sprites for titanium.


ThatDan123:
Network manager stuff needs commiting

PR #7256, opened 2021-08-13
Crafting system hotfixes

### LINKED ISSUE #7252: Crafting recipe doesn't show recipes for clients.
### Environment
- Game Version: Staging 21081300
- OS Version: Windows 10

Title. When connected to a server as a client the crafting menu doens't show any recipes, making it impossible in practice to perform any recipe that can't be done by just using items on other items.

### Steps to reproduce
1. Connect to a server.
2. Open the crafting menu upon spawning in.
3. Notice that none of the categories have recipes available.

### LINKED ISSUE #7253: The game thinks you're trying to perform simple crafting recipes when using ingredients on items and objects unrelated to recipes.
### Environment
- Game Version: Staging 21081300
- OS Version: Windows 10

Ingredients used in simple recipes think you're trying to make recipes when you use them on items or objects unrelated to recipes. For context, a simple recipe is a recipe which can be done by simply using an item on another item. An example would be the burger recipe, where you can use a steak on a bun to make a burger without opening the crafting menu.

This is bad because the recipe check seemingly overrides other interactions, such as trying to put plasma sheets in the plasma generator.
image
image

### Steps to reproduce

1. Use a clown mask on a random item or object.
2. You receive a message saying that you can't  make a "Clown Burger" because there aren't enough ingredients.

Lizurt:
### Purpose

- fixes https://github.com/unitystation/unitystation/issues/7252
   - Some recipes in the crafting singleton had a wrong index.
- fixes https://github.com/unitystation/unitystation/issues/7253
   - If we can't craft something, the interaction cooldown updates anyway, so we skip other interactions. Now we try to craft something if we are sure that we can craft something in the `WillInteract()` method.
- The recipe info placeholder(a xeno maid) now hides correctly.
   - There was a category without a category button. That's why we had NullPointerException, so the recipe info placeholder was still active.

### Changelog:

CL: [Fix] Now a xeno maid "recipe" shouldn't appear.
CL: [Fix] Ingredients of a simple recipe no longer affect objects that they shouldn't.
CL: [Fix] The crafting menu now should show recipes correctly.


UnityProject/Assets/Scripts/Player/PlayerCrafting.cs
```diff
@@ -102,7 +105,16 @@ public List<CraftingRecipe> GetKnownRecipesInCategory(RecipeCategory recipeCateg
...
 		public bool KnowsRecipe(CraftingRecipe recipe)
 		{
-			return GetKnownRecipesInCategory(recipe.Category).Contains(recipe);
+			try
+			{
+				return GetKnownRecipesInCategory(recipe.Category).Contains(recipe);
+			}
+			catch(Exception e)
+			{
+				Console.WriteLine(e.StackTrace);
```
corp-0: log it instead

Lizurt: Sorry, it's removed now. I was using this for debugging

UnityProject/Assets/Scripts/Player/PlayerCrafting.cs
```diff
@@ -480,67 +481,74 @@ public void TryToStartCrafting(CraftingRecipe recipe, NetworkSide networkSide)
...
+		/// <returns>True if a crafting action has started, false otherwise.</returns>
 		[Server]
 		public bool TryToStartCrafting(
 			CraftingRecipe recipe,
 			List<CraftingIngredient> possibleIngredients,
 			List<ItemAttributesV2> possibleTools,
 			List<ReagentContainer> reagentContainers,
-			bool ignoreIngredientsAndTools = false
+			bool ignoreIngredientsAndTools = false,
+			bool shouldGiveFeedback = true
```
corp-0: https://refactoring.guru/es/smells/long-parameter-list

Lizurt: Alright, I replaced sove parameters with a new class.

PR #7274, opened 2021-08-15
fixes players getting kicked for bad mapping

Bod9001:
yeah 
image
was causing people to be kicked

CL: [fix] fixed an issue with a lightswitch in the ancient station that was causing players to get kicked.

### DIFF
--- a/UnityProject/Assets/Mirror/Runtime/NetworkIdentity.cs
+++ b/UnityProject/Assets/Mirror/Runtime/NetworkIdentity.cs
@@ -705,6 +705,11 @@ namespace Mirror
             }
 
             // Debug.Log("OnStartClient " + gameObject + " netId:" + netId);
+            if (NetworkBehaviours == null)
+            {
+	            Debug.LogError("hey NetworkBehaviours is null on " + gameObject + "at " + GetGameObjectPath(gameObject) + " at "  + gameObject.transform.localPosition  );
+	            return;
+            }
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Mirror/Runtime/NetworkIdentity.cs
```diff
@@ -723,6 +728,16 @@ internal void OnStartClient()
                 }
             }
         }
```
corp-0, line 730: This could probably go somewhere else, a GameObjectUtils class? Or maybe as an extension of GameObject

Bod9001: it has namespace issues since unity is in its own namespace

corp-0: I have no idea what line I commented, I’m on my phone, but I meant the path method

Bod9001: mirror*

Bod9001: *assembly definition

ThatDan123:
Be a good idea to make a scene test to detect child networked objects of networked objects since you've removed the net identity on the player sprites now (though that wouldnt be picked up on a scene test)?

PR #7282, opened 2021-08-16
fixes mirror not launching properly when launching a scene directly

Bod9001:
yeah, due to the statics it was thinking it was loading a map while it wasn't so, it would never receive the spawn player message so thought there were no players


Bod9001:
let's see how long this one line change takes to get merged

MaxIsJoe:
> let's see how long this one line change takes to get merged

PR merge speedrun %any

UnityProject/Assets/Mirror/Runtime/NetworkServer.cs
```diff
@@ -70,6 +70,10 @@ static void Initialize()
             if (initialized)
                 return;
 
+            //CUSTOM UNITYSTATION CODE//
+            //So it can handle Not having domain reloading on
+            isLoadingScene = false;
```
corp-0, line 75: so we force the scene to be flagged as not loading when initializing the server?

Bod9001: it should be set to true since this should run before that

corp-0: ELI5 how does it affect scene loading?

Bod9001: It doesn't

ThatDan123 (approved):

corp-0 (approved):

PR #7293, opened 2021-08-18
map serialization first iteration

Bod9001:
first step to saving tile maps


UnityProject/Assets/Scripts/MapSaver/MapSaver.cs
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	public static TileMapData SaveTileMap(MetaTileMap MetaTileMap)
+	{
+		//# Matrix4x4
+		//§ TileID
+		//◉ Colour
+
+		//☰ Layer
+		//@ location
```
Aranclanos: ENUMS PLEASE

corp-0: uhm, enums are named numbers so it wouldn't work to use instead of the characters Bod is using to serliaze the map. A class to define the constants would work.

```cs
public static class MapSerializationConstants
{
    public static string Matrix = "#";
    public static string TileID = "§";
    ...
}
```

Bod9001: oh yeah, characters are used for improving the readability

corp-0: use constants so it is easier to swap them in the future if we need

UnityProject/Assets/Scripts/MapSaver/MapSaver.cs
```diff
@@ -0,0 +1,182 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Text;
+using TileManagement;
+using UnityEngine;
+using System.Linq;
+
+public static class MapSaver
+{
+	public readonly static char Matrix4x4Char = '#' ;
```
corp-0: now that they are in the same class and this class does serialization and deserialization, I don't expect these to be needed elsewhere, right? If that's the case, make then private, otherwise say and I resolve the conversation.

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
```diff
@@ -215,6 +216,11 @@ public void Update()
 			}
 		}
 
+		public Dictionary<Layer, Dictionary<Vector3Int, TileLocation>> GetPresentTiles()
```
corp-0: getters are for java nerds, we use properties instead

UnityProject/Assets/Scripts/MapSaver/TestMapSaverScript.cs
```diff
@@ -0,0 +1,17 @@
...
+
+public class TestMapSaverScript : MonoBehaviour
+{
+	public MetaTileMap MetaTileMap;
+
+
+	[NaughtyAttributes.Button()]
+	public void SaveMap()
+	{
+		Logger.Log(JsonConvert.SerializeObject(MapSaver.SaveTileMap(MetaTileMap)));
```
corp-0: loggin the whole map is intended or was it for debugging?

Bod9001: is debugging

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
Bod9001: main reason is because, it's high-risk item no random should be accessing it couldn't think of a better  geter

UnityProject/Assets/Scripts/MapSaver/MapSaver.cs
Bod9001: it might be a separate class called map loader for the loading

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
corp-0:
```cs
public Dictionary<Layer, Dictionary<Vector3Int, TileLocation>> GetPreesentTiles => PresentTiles;
```
How is the property less secure?

corp-0: Normally I wouldn't call it ``GetPresentTiles`` but ``PresentTiles`` is already taken.

Bod9001: ok

corp-0 (approved):

corp-0 (changes requested):
Can you solve the codacy issues?

corp-0:
the test is missing namespace and should be moved to the tests folder.  I imagine the empty foreach that saves objects is a TODO.

corp-0:
try to don't add more to this pr so it is easier to review. I had already given my green tick

Bod9001:
"the test is missing namespace and should be moved to the tests folder. I imagine the empty foreach that saves objects is a TODO." 
is a temporary script will be removed once finished

corp-0:
alrigthy

UnityProject/Assets/Scripts/MapSaver/MapSaver.cs
```diff
@@ -434,22 +434,28 @@ public static ObjectMapData SaveObjects(MetaTileMap MetaTileMap)
...
 
+				if (Field.FieldType.IsGenericType && Field.FieldType.GetGenericTypeDefinition() == typeof(Dictionary<,>)) continue; //skipping all dictionaries For now
+
 				//if Field is a class and is not related to unity engine.object Serialise it
 				var PrefabDefault = Field.GetValue(PrefabInstance);
 				var MonoSet = Field.GetValue(SpawnedInstance);
 
 				var selfValueComparer = PrefabDefault as IComparable;
-				bool result;
+				bool AreSame;
```
corp-0:
```suggestion
				bool areSame;
```

UnityProject/Assets/Scripts/MapSaver/MapSaver.cs
```diff
@@ -424,7 +424,7 @@ public static ObjectMapData SaveObjects(MetaTileMap MetaTileMap)
 					var APrefabDefault = Field.GetValue(PrefabInstance);
 					var AMonoSet = Field.GetValue(SpawnedInstance);
 
-					if ((Field.FieldType.IsSubclassOf(typeof(UnityEngine.Object)) == false)) continue; //Handle this with custom stuff
+					if ((Field.FieldType.IsSubclassOf(typeof(UnityEngine.Object)))) continue; //Handle this with custom stuff
```
corp-0: what's going on with the parenthesis?

UnityProject/Assets/Scripts/MapSaver/MapSaver.cs
```diff
@@ -434,22 +434,28 @@ public static ObjectMapData SaveObjects(MetaTileMap MetaTileMap)
...
+				if (Field.FieldType.IsGenericType && Field.FieldType.GetGenericTypeDefinition() == typeof(Dictionary<,>)) continue; //skipping all dictionaries For now
+
 				//if Field is a class and is not related to unity engine.object Serialise it
 				var PrefabDefault = Field.GetValue(PrefabInstance);
 				var MonoSet = Field.GetValue(SpawnedInstance);
 
 				var selfValueComparer = PrefabDefault as IComparable;
-				bool result;
+				bool AreSame;
 				if (PrefabDefault == null && MonoSet == null)
```
corp-0, line 445: guidelines say to use ``{`` for if statements, even if the body is one line.

Bod9001: "if (Field.FieldType.IsGenericType && Field.FieldType.GetGenericTypeDefinition() == typeof(Dictionary<,>)) continue;"
that seems to be fine though, and is used quite commonly out throughout the code base

corp-0: yeah, we have a lot of ``continue`` and ``return``' but in this particular case you are even mixing both styles in the same function

UnityProject/Assets/Scripts/Managers/NetworkManagement/CustomNetworkManager.cs
```diff
@@ -66,6 +68,28 @@ public override void Awake()
...
+	public void Update()
+	{
+		if (allSpawnablePrefabs.Count > CurrentLocation)
+		{
+			for (int i = 0; i < 50; i++)
+			{
+				if (allSpawnablePrefabs.Count > CurrentLocation + i)
+				{
+					if (allSpawnablePrefabs[CurrentLocation + i] == null) continue;
+					var PrefabTracker = allSpawnablePrefabs[CurrentLocation + i].GetComponent<PrefabTracker>();
```
corp-0, line 82: is this really alright? A loop with GetComponent in an ``Update`` method.

corp-0: also you can merge the getting component assignation and null checking with a ``TryGetComponent``

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/ObjectLayer.cs
```diff
@@ -16,6 +16,19 @@ public class ObjectLayer : Layer
 	public TileList ServerObjects => serverObjects ?? (serverObjects = new TileList());
 	public TileList ClientObjects => clientObjects ?? (clientObjects = new TileList());
 
+	public TileList GetTileList(bool isServer)
```
corp-0, line 19: use properties for getters

corp-0: ah wait, this has a parameter. Nevermind.

UnityProject/Assets/Scripts/MapSaver/MapSaver.cs
NoooneyDude: Yeah I think the unspoken rule is to allow jump statements on the same line, anything else should get its own block.

UnityProject/Assets/Scripts/Managers/NetworkManagement/CustomNetworkManager.cs
Bod9001: well it only runs for 50 times an update, it would do it all if I put it in start (a bigger Stutter)
and once it's finished it shouldn't use it again

corp-0 (approved):
I'm approving considering the system is a wip and should be cleaner in following iterations.

ThatDan123 (approved):

PR #7295, opened 2021-08-19
Fix == operator on MatrixInfo

Copern:
### Purpose
MatrixInfo == operator is not behaving like an equality operator (a null MatrixInfo == null is false, excuse me what?). The == operator now behaves as it should.

Edit: Slightly more verbose at Gilles request.


Bod9001 (approved):

ThatDan123 (approved):

Aranclanos:
out of curiosity, did this have any effect in the game?

corp-0:
Unknown effects. Depends on how much was the operator used and how many times we were in a situation where left was null.

PR #7306, opened 2021-08-21
Null check dynamic item storage

ThatDan123:
-Catch null objects for dynamic item storage


corp-0:
When could the object be null?

Bod9001 (approved):

corp-0 (approved):

PR #7317, opened 2021-08-22
Fix mob speeds

NoooneyDude:
- CL: [Fix] Fixes default mob speeds. Individual mobs will likely still require tuning.

### DIFF
--- a/UnityProject/Assets/Scripts/NPC/AI/MobPathFinder.cs
+++ b/UnityProject/Assets/Scripts/NPC/AI/MobPathFinder.cs
@@ -323,7 +323,7 @@ namespace Systems.MobAIs
 
 				if (!movingToTile)
 				{
-					if (TickDelay != 0f)
+					if (TickDelay.Approx(0) == false)
 					{
 						yield return WaitFor.Seconds(TickDelay);
 					}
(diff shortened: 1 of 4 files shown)


corp-0:
does this only affect movement? Is that the only decision they take?

NoooneyDude:
> does this only affect movement? Is that the only decision they take?

To be honest, I'm not entirely up to speed with mob code. I just found that playing with that variable in particular achieved the desired result, and I suspect the original values were lost during the mob prefab variant rebasing.

UnityProject/Assets/Scripts/NPC/AI/MobPathFinder.cs
```diff
@@ -319,9 +323,9 @@ IEnumerator PerformFollowPath(List<Node> path)
 
 				if (!movingToTile)
 				{
-					if (tickRate != 0f)
+					if (TickDelay != 0f)
```
NoooneyDude:
```suggestion
					if (TickDelay.Approx(0) == false)
```

ThatDan123 (approved):

corp-0 (approved):

PR #7318, opened 2021-08-23
Traumacodefix

MaxIsJoe:
CL: [Fix] - Fixes bleeding not working.
CL: [Balance] - Re-adjusts base trauma values to allow for them to happen again more frequently for lower damage weapons.

### DIFF
--- a/UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartTraumaDamage.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartTraumaDamage.cs
@@ -8,7 +8,7 @@ namespace HealthV2
 {
 	public partial class BodyPart
 	{
-		
+
 		private bool isBleedingInternally = false;
 
 		private bool isBleedingExternally = false;
@@ -92,7 +92,7 @@ namespace HealthV2
 		/// How much damage can this body part last before it breaks/gibs/Disembowles?
 		/// <summary>
 		public float DamageThreshold = 18f;
-		
+
 
 		[SerializeField] private bool gibsEntireBodyOnRemoval = false;
 
@@ -170,26 +170,25 @@ namespace HealthV2
 		private void CheckCutSize()
 		{
 			CheckTraumaDamageLevels();
-			if(currentSlashCutDamage <= 0)
+			switch (currentSlashCutDamage)
 			{
-				currentCutSize = BodyPartCutSize.NONE;
-			}
-			else if(currentSlashCutDamage > 75)
-			{
-				currentCutSize = BodyPartCutSize.LARGE;
-			}
-			else if(currentSlashCutDamage > 50)
-			{
-				currentCutSize = BodyPartCutSize.MEDIUM;
-			}
-			else if(currentSlashCutDamage > 25)
-			{
-				currentCutSize = BodyPartCutSize.SMALL;
-			}
-
-			if(currentCutSize >= BodyPartSlashLogicOnCutSize && CanBleedExternally)
-			{
-				StartCoroutine(ExternalBleedingLogic());
+				case float n when n.IsBetween(0, 25):
+					currentSlashDamageLevel = SlashDamageLevel.NONE;
+					break;
+				case float n when n.IsBetween(26, 50):
+					currentSlashDamageLevel = SlashDamageLevel.SMALL;
+					break;
+				case float n when n.IsBetween(26, 50):
+					currentSlashDamageLevel = SlashDamageLevel.MEDIUM;
+					break;
+				case float n when n > 50:
+					currentSlashDamageLevel = SlashDamageLevel.LARGE;
+					break;
+				default:
+					currentSlashDamageLevel = SlashDamageLevel.NONE;
+					Logger.LogError(
+						$"Unexpected slash cut damage on: {gameObject}, {currentSlashDamageLevel}");
+					break;
 			}
 		}
 
@@ -449,7 +448,7 @@ namespace HealthV2
 			}
 		}
 
-		
+
 		/// <summary>
 		/// Checks if the bodypart is damaged to a point where it can be gibbed from the body
 		/// </summary>
--- a/UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartTraumaDamage.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartTraumaDamage.cs
@@ -24,7 +24,7 @@ namespace HealthV2
 		public float InternalBleedingBloodLoss = 12;
 		public float ExternalBleedingBloodLoss = 6;
 
-		[SerializeField, Range(0.2f, 4.25f)] private float baseTraumaDamageMultiplier = 0.25f;
+		[SerializeField, Range(2.0f, 12.0f)] private float baseTraumaDamageMultiplier = 2.0f;
 
 		public float MaximumInternalBleedDamage => maximumInternalBleedDamage;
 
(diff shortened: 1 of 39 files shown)


UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartTraumaDamage.cs
```diff
@@ -192,6 +193,69 @@ private void CheckCutSize()
...
+			//Slash
+			if(currentSlashCutDamage <= 0)
+			{
+				currentSlashDamageLevel = SlashDamageLevel.NONE;
+			}
+			if(currentSlashCutDamage >= 25)
+			{
+				currentSlashDamageLevel = SlashDamageLevel.SMALL;
+			}
+			if(currentSlashCutDamage <= 50)
```
NoooneyDude:
```suggestion
			if(currentSlashCutDamage >= 50)
```

NoooneyDude: Do these enums have the same value for each shared level? If so, consider moving all of this duplicate code into a method, which each damage type can call.

NoooneyDude: In fact, it might be better to just stick with one enum for the different damage types, with burn code handling the higher 75 level a little differently than the others, though I haven't delved too much into damage code.

corp-0: To handle the ranges, a switch case with IsBetween would be a lot cleaner

corp-0:
```cs
switch (currentSlashCutDamage)
{
    case float n when n.IsBetween(0, 25):
        currentSlashDamageLevel = DamageLevel.NONE;
        break;
    case float n when n.IsBetween(26, 50):
        currentSlashDamageLevel = DamageLevel.SMALL;
        break;
    case float n when n.IsBetween(26, 50):
        currentSlashDamageLevel = DamageLevel.MEDIUM;
        break;
    case float n when n > 50:
        currentSlashDamageLevel = DamageLevel.LARGE;
    default:
        currentSlashDamageLevel = DamageLevel.NONE;
        Logger.LogError($"Unexpected slash cut damage on: {healthMaster.gameObject}, {currentSlashDamageLevel}");
}
```
It would also allow you to handle out of range damage, like if for whatever reason you got negative damage.

MaxIsJoe: I wanted to try using switch but it was annoying me a bit, this makes code much cleaner. Thank you Gillies!

UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartTraumaDamage.cs
```diff
@@ -169,26 +169,89 @@ private void DEBUG_ApplyTestPierce()
...
+			CheckTraumaDamageLevels();
+			switch (currentSlashCutDamage)
+			{
+				case float n when n.IsBetween(0, 25):
+					currentSlashDamageLevel = SlashDamageLevel.NONE;
+					break;
+				case float n when n.IsBetween(26, 50):
+					currentSlashDamageLevel = SlashDamageLevel.SMALL;
+					break;
+				case float n when n.IsBetween(26, 50):
```
corp-0: I just realized I wrongly set the ranges in the code I gave you. Small and Medium use the same values, so that needs fixing. 

Also consider the option of merging wound severity into a single enum like Nooney suggested.

MaxIsJoe: I fixed that

UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartTraumaDamage.cs
MaxIsJoe: did as Nooney asked and merged enums into one

NoooneyDude: Hmm, my screen is still full of `if` `else`'s. You can still make this more concise by reusing that same switch statement.


```cs
private TraumaDamageLevel GetDamageLevel(float damage)
{
    // That fancy switch statement in CheckCutSize() here
    // case n when xyz:
    //     return DamageLevel.MEDIUM;
    // ...
}
```
Then you just need to:
```cs
currentSlashDamageLevel = GetDamageLevel(currentSlashDamage);
currentPierceDamageLevel = GetDamageLevel(currentPierceDamage);
// ...
// For the special burn damage check, you can still do it, just outside the function
```

MaxIsJoe: Done 👍

Bod9001 (dismissed):

UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartTraumaDamage.cs
```diff
@@ -166,29 +166,74 @@ private void DEBUG_ApplyTestPierce()
...
+					break;
+				case float n when n > 76:
+					currentSlashDamageLevel = TraumaDamageLevel.CRITICAL;
+					break;
+				default:
+					currentSlashDamageLevel = TraumaDamageLevel.NONE;
+					Logger.LogError(
+						$"Unexpected slash cut damage on: {gameObject}, {currentSlashDamageLevel}");
+					break;
 			}
```
NoooneyDude: Now that we have the method, we don't need this switch here now, right? Also, consider the case when damage is `75.5` - think it'll drop to the switch's default case. And what about `25.5`? I think it'd be safe to just use the same number for one case's upper value with the next case's lower value, though I'm not sure how `IsBetween()` works when the value is right on the boundary. You'd have to look into it.

```suggestion
```

MaxIsJoe: I'd rather not use decimal numbers at all right now as all damage numbers are whole at the moment, only `health` goes near `.5` territories and stuff like that.

"But why not just use an int then?" because damage numbers are floats, if we were working with separate trauma damage numbers like we used to before the rework then yeah it would have made sense to use integers but we're no longer doing that.
Also I don't want to deal with more number bugs, this PR was intended to fix them; not create more.

corp-0: from what I remember, IsBetween had a third optional parameter to make it not end inclusive, which would fix the issue without much changes. You would just check a range like ``IsBetween(0, 25, true)`` and the next one ``IsBetween(25, 50, true)``

corp-0: it was the other way around, you use false to make it non-inclusive

corp-0: https://rextester.com/PVZR19760 ye it works

MaxIsJoe:
Okay I think I've covered everything so far, can we finally merge this already so I can return to working on other things?

Bod9001 (approved):

corp-0 (approved):

PR #7332, opened 2021-08-27
Fixes the dissapearing bell bug

### LINKED ISSUE #5659: Picking up the service bell with right click menu causes it to stop existing.
### Environment
- Game Version: 5608
- OS Version: Windows 10

Since you can't pick up service bells by left clicking on them (that dings them instead), you might try to pick them up using the right click menu instead. However, trying to pick up a service bell in this way cause the service bell to stop existing, for some reason.
### Steps to reproduce

1. Right click on service bell.
2. Go to pick up option.
3. Try to pick up service bell, watch it disappear.

Lizurt:
### Purpose

Fixes https://github.com/unitystation/unitystation/issues/5659

### Notes:

Gotcha! I found the cause of the bug on the third attempt.

### Changelog:

CL: [Fix] Fixed the disappearing service bell bug.


corp-0:
So the problem with this implementation of the bell is that it extends ``Pickupable`` which isn't the intended way. ``ServiceBell`` should just extend ``MonoBehaviour`` and the prefab have a ``Pickupable`` component like any other item.

Lizurt:
> So the problem with this implementation of the bell is that it extends `Pickupable` which isn't the intended way. `ServiceBell` should just extend `MonoBehaviour` and the prefab have a `Pickupable` component like any other item.

I also considered this as a solution, but then I thought: what if someone or something will change an order of components? Then the `Pickupable` component will block the bell's logic and it won't be obvious to the one who will debug. The only way I see is the complicating the `Pickupable`'s logic, adding something like this:
```cs
// ... the Pickupable class and its methods ...
foreach (Action action in actions)
{
    if (action.Exectute()) 
    {
        // we have ringed and we don't need to pick up the bell
        return;
    }
}
// ... the picking up logic ...
```

So in this approach we don't create a new `MonoBehaviour`, but complicate the `Pickupable` logic. I don't see any other example of these `actions`. So maybe the current approach isn't that bad?

corp-0:
Not sure if this prefab inherits from the base item prefab. If it does, it would already have a pickupable component near the top (last in execution order) and that order isn’t expected to ever change, at least not so dramatically so it interrupts execution before the new component. If it doesn’t, it should.

Lizurt:
Alright, I made the `Pickupable` and the `ServiceBell` components separated. Also I changed the base of the service bell. It's `Item` now.

UnityProject/Assets/Scripts/Items/ServiceBell.cs
```diff
@@ -17,15 +15,13 @@ public class ServiceBell : Pickupable, IServerSpawn
 
 		[SerializeField] private SpriteHandler BellSpriteRenderer;
 
+		public bool WillInteract(HandApply interaction, NetworkSide side)
+		{
+			return interaction.Intent != Intent.Grab && interaction.TargetObject == gameObject;
```
corp-0: test for empty hand too and not harm intent

UnityProject/Assets/Scripts/Systems/Inventory/Pickupable.cs
```diff
@@ -133,7 +133,7 @@ public virtual bool WillInteract(HandApply interaction, NetworkSide side)
 		return true;
 	}
 
-	public void ClientPredictInteraction(HandApply interaction)
+	public virtual void ClientPredictInteraction(HandApply interaction)
```
corp-0: this doesn't need to be virtual anymore

corp-0:
to fix  the test, just run the test locally and commit the changes it generates

corp-0 (approved):

Bod9001 (approved):

UnityProject/Assets/Scripts/Items/ServiceBell.cs
```diff
@@ -17,15 +15,16 @@ public class ServiceBell : Pickupable, IServerSpawn
 
 		[SerializeField] private SpriteHandler BellSpriteRenderer;
 
+		public bool WillInteract(HandApply interaction, NetworkSide side)
+		{
+			return interaction.Intent != Intent.Grab
+			       && interaction.Intent != Intent.Harm
+			       && interaction.TargetObject == gameObject
+			       && interaction.HandObject == null;
+		}
```
ThatDan123, line 24: Can you add `if (DefaultWillInteract.Default(interaction, side) == false) return false;` here, or else I think you'll be able to use the bell from any distance

ThatDan123: nvm, i'll do it in another PR

PR #7334, opened 2021-08-28
Add prefab-mode-only and scene-mode-only attributes

NoooneyDude:
- Adds prefab-editing mode and scene-editing only attributes.
Reduces inspector complexity significantly for mapping.
- Applies the former attribute to a bunch of common components.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Attributes.meta
+++ b/UnityProject/Assets/Scripts/Core/Attributes.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: afac2c9f530e04242bdace68abcf1ce2
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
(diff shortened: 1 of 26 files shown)


Bod9001:
one good thing to think about is how much should you be able to change Through an item on the map?

NoooneyDude:
> one good thing to think about is how much should you be able to change Through an item on the map?

I don't think there's much advantage in any of the variables I've added the attribute to being changed on the map, but point out any that you disagree on.

corp-0:
It is fine to merge, I think. I don't know how close are we to change mapping but we won't be editing scenes anymore in the future, right?

Bod9001:
with the mapping tools you could do it through the editor or in game the plan is currently

Bod9001:
all I say add an override somewhere that you can toggle

corp-0 (dismissed):

Bod9001 (approved):

corp-0 (approved):

Issue #7356, opened 2021-09-04
You can see and hear mobs from other matrices as a client.


PetMudstone:
### Environment
- Game Version: Staging 21090215
- OS Version: Windows 10

### Detailed description of bug
As a client, you can see mobs from other matrices when you spawn on the station. You are also able to hear them in the case of mobs that make noises like xenomorphs. Notably, you can't actually interact with these mobs and these mobs can't interact with you. Similarly, the mobs will appear to phase through walls and objects on the station.

### Steps to reproduce
1. Join a server.
2. Walk around for a bit until you see carp, xenomorphs, hivebots, etc. walking around the station.

Aranclanos:
the sound position gets overriden by the transform.position of the mob if the client receives a netID so i believe this was fixed in the linked PR

PR #7369, opened 2021-09-05
Fixes hivebot ghosts wandering around the station

Aranclanos:
### Purpose
CL: [Fix] Fixed mobs in movable matrixes being in the incorrect position for clients

### Notes:
That function fails because the movable matrix values aren't ready and the WorldToLocal() fails, but we dont need to fetch the matrixinfo that way since we already have the matrix.
This of course doesnt mean things are good now, matrix initialization need a lot of work


PetMudstone:
Does this fix #7356?

ThatDan123 (approved):

Bod9001 (approved):

PR #7378, opened 2021-09-07
Fix changelogs, maybe

corp-0:
### Purpose
Since we adopted branch protection rules, changelog action broke. This little change would make the changelog push the changes with ``--force`` which *should* circumvent the review requirement.


ThatDan123 (approved):

Bod9001 (approved):

corp-0:
it didn't work :c

Issue #7389, opened 2021-09-08
Fix codacy issues


corp-0:
### Description
Thank to Paul from SS14 for pointing out how we aren't retroactively fixing issues reported by Codacy.

There are enough issues for everyone, from all levels of complexity and importance. So please, look [here](https://app.codacy.com/gh/unitystation/unitystation/issues/index) for issues we had in the repo from long before we added codacy and retroactively fix them on your own pace.

PR #7390, opened 2021-09-09
Little clean up in movement related code

corp-0:
### Purpose
I was solving some simple codacy issues and ended up making small refactorization of movement related code. This is mostly just adding namespace, moving an inner interface to its own file, null checking and solving potential logic issues where we were using ``&`` instead of ``&&``

### Notes:
The issues I solved here could have eventually ended in unexpected behaviour, but I wouldn't bet any of it fixed any bug.


corp-0:
related to #7389

ThatDan123 (approved):

Bod9001 (approved):

PR #7394, opened 2021-09-10
Revert "Add prefab-mode-only and scene-mode-only attributes"

NoooneyDude:
Reverts unitystation/unitystation#7334

Breaks inspectors.


Bod9001 (approved):

ThatDan123:
Could it not be fixed?

NoooneyDude:
> Could it not be fixed?

I had a quick tinker but not entirely sure how to solve the problem of lists not rendering right. Might be a NaughtyAttributes conflict or somethting.

PR #7400, opened 2021-09-11
Fix atmos devices

NoooneyDude:
- CL: [Fix] Fixes atmospheric devices not rotating properly.
Removes the directional components.
- Updates packages.


Bod9001 (approved):

Bod9001:
does it work by removing the directional? how was the fixed done?

NoooneyDude:
Yeah by removing the directional so it doesn't change transform at onspawnserver

Bod9001:
just noting some of the stuff might be mapped with directional so it might lose the direction, when loading up the maps,

PR #7401, opened 2021-09-11
Pop in Chat Bubble Text

ThatDan123:
Chat bubble text now pops in!

CL: [Improvement] Chat bubble text now pops in!

-Clown is now even more annoying

-Settings menu contains some customisation over character pop in speed and time. Can also make it instant like the old chat

-Adds a singleton manager, will convert over all old managers in another PR

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/SettingsManager/DisplaySettings.cs
+++ b/UnityProject/Assets/Scripts/Managers/SettingsManager/DisplaySettings.cs
@@ -79,16 +79,16 @@ namespace Managers.SettingsManager
 			}
 			private bool chatBubbleSizeChanged = false;
 
-			public bool ChatButtleInstantChanged
+			public bool ChatBubbleInstantChanged
 			{
-				get => chatButtleInstantChanged;
+				get => chatBubbleInstantChanged;
 				set
 				{
-					chatButtleInstantChanged = value;
+					chatBubbleInstantChanged = value;
 					changed |= value;
 				}
 			}
-			private bool chatButtleInstantChanged = false;
+			private bool chatBubbleInstantChanged = false;
 
 			public bool ChatBubblePopInSpeedChanged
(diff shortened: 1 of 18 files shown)


corp-0:
I love the colourful clown text and me myself have no trouble reading it but I'm afraid it might hinder accessibility for some people. Could we maybe add a checkbox on the settings to enable and disable the effect?

UnityProject/Assets/Scripts/Managers/SettingsManager/DisplaySettings.cs
```diff
@@ -77,6 +78,39 @@ public bool ChatBubbleSizeChanged
 				}
 			}
 			private bool chatBubbleSizeChanged = false;
+
+			public bool ChatButtleInstantChanged
+			{
+				get => chatButtleInstantChanged;
```
NoooneyDude: `chatButtle`? 🙂

NoooneyDude (dismissed):

NoooneyDude:
> I love the colourful clown text and me myself have no trouble reading it but I'm afraid it might hinder accessibility for some people. Could we maybe add a checkbox on the settings to enable and disable the effect?

Yep, but perhaps a generic "Accessibility" toggle, or similar, instead of something for this specific instance would be a good idea.

ThatDan123:
> > I love the colourful clown text and me myself have no trouble reading it but I'm afraid it might hinder accessibility for some people. Could we maybe add a checkbox on the settings to enable and disable the effect?
> 
> Yep, but perhaps a generic "Accessibility" toggle, or similar, instead of something for this specific instance would be a good idea.

I've just done a setting for just the clown colour, think its best we have lots of options instead of just a generic accessibility toggle imo

Bod9001 (approved):

corp-0 (approved):

PR #7404, opened 2021-09-14
Fix osx builds for staging

JustinTrotter:
### Purpose
Fix build so that required libraries for macOS are properly generated.
 <br>

### Notes:

Sound and music for macOS builds on staging still do not work. I believe this is due to a pathing issue with the addressables for sound and music.

### Changelog:

CL: [Fix] Staging should now be playable on macOS


NoooneyDude:
Huh, I just see meta files. does this really fix it?

NoooneyDude (approved):

Bod9001 (approved):

PR #7407, opened 2021-09-14
Fixes disposal outlets not working after only one release

### LINKED ISSUE #7365: Disposal outlets become unable to eject their contents after releasing items.
### Environment
- Game Version: Staging 21090215
- OS Version: Windows 10

When a disposal outlet releases an item coming from disposals, it goes into a animation loop where it repeatedly opens and closes. During this loop it will no longer eject anything else that enters the outlet. This includes both other items and players. 
Additionally, items are released onto the same tile the outlet is on instead of the tile in front of it. Outlets otherwise release players properly unless broken by this bug.

### Steps to reproduce
1. Find the disposal sorting area in cargo.
2. Place an item into a nearby disposal bin and return to the disposal sorting area.
3. Witness the outlet release the item onto the same tile as it and enter an animation loop.
4. Place another item into a nearby disposal bin and return to the disposal sorting area again.
5. Watch as the outlet doesn't release the second item while in its animation loop.

Aranclanos:
### Purpose
CL: [Fix] Fixed disposal outlets not working after their first trash release.
Fixes #7365

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Disposals/DisposalVirtualContainer.cs
+++ b/UnityProject/Assets/Scripts/Objects/Disposals/DisposalVirtualContainer.cs
@@ -15,7 +15,6 @@ namespace Objects.Disposals
 		[SerializeField]
 		private AddressableAudioSource ClangSound;
 
-		private Matrix Matrix;
 		private ObjectBehaviour ContainerBehaviour;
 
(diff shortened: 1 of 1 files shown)


ThatDan123 (approved):

NoooneyDude (approved):
Why on earth did I make the station the thrower :p

PR #7410, opened 2021-09-15
fixes loops with door hacking and added mmanual test for loops

Bod9001:
yeah, you have to be careful about loops with the new hacking code but, hopefully the manual test I made should get all cases of loops


Bod9001 (commented):
I wonder if this is coded properly

Bod9001:
humm, I wonder if that actually counts to the required Reviews, would be hilarious if so xD

ThatDan123 (approved):

corp-0 (approved):

PR #7418, opened 2021-09-16
Fixes the health scanner showing the wrong value for oxy damage

### LINKED ISSUE #7344: Oxygen damage doesn't seem to put you in crit or kill you properly.
### Environment
- Game Version: Staging 21083022
- OS Version: Windows 10

Oxygen damage behaves strangely in regards to crit and death states. For some reason having 100 oxygen damage will not cause you to become unconscious and having 200 oxygen damage will not kill you. Reaching these levels is quite negligible for your character, in fact.
Reaching 600 oxygen damage causes you to go into crit, but reaching 1,200 oxygen damage and above doesn't seem to kill you oddly enough. Dealing other kinds of damage on top of the 1,200+ oxygen damage will immediately kill your character, however.
In regards to the "Overall status" of someone's health viewable via the health analyzer, oxygen damage seems to adjust the percentage less than other types of damage.
image
image

### Steps to reproduce
1. Go out into space while wearing a spacesuit and with a health analyzer. Keep your oxygen turned off.
2. Wait as you suffocate for a minute or so.
3. Check yourself with the health analyzer. Despite you most likely having around 200 oxygen damage by this point you will not have fallen unconscious let alone be dead.

Aranclanos:
### Purpose
CL: [Fix] Fixed the health scanner showing the wrong value for oxy damage
CL: [Fix] Fixed "catperson" string appearing in the health scanner output
Fixes #7344

players overall health is only affected by the brain oxy dmg, and now for the different body parts, but now this raises the question: Should health analyzers show oxy damage from the different body parts? If so, how?
I'm not sure if oxygen damage does anything outside of the brain organ

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Medical/HealthScanner.cs
+++ b/UnityProject/Assets/Scripts/Items/Medical/HealthScanner.cs
@@ -66,8 +66,12 @@ namespace Items.Medical
 				{
 					fullDamage[i] += bodypart.Damages[i];
 				}
-
 				partMessages.AppendLine(GetBodypartMessage(bodypart));
+			}
+
+			if (health.brain)
+			{
+				fullDamage[(int) DamageType.Oxy] = health.brain.RelatedPart.Oxy;
 			}
 
 			if ((int)totalPercent != 100)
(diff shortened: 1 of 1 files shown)


corp-0:
I don't think we gain much by tracking oxy damage on each body part

NoooneyDude (approved):

Bod9001:
oxygen damage on each body part affects the functionality of that body part, for example if you take oxygen damage on legs your movement speed will be slower

Bod9001 (approved):

PR #7428, opened 2021-09-18
Cannot Be Null Attribute and Test

ThatDan123:
-Adds a new attribute: [CannotBeNull]

-Adds a utility for recursive action on child game objects

Put that on a field which you know will never be null e.g a GameObject field on a script on an prefab (e.g spawning item in item spot) and then a test will run and if there's a prefab or a scene object which has that field null then it'll fail.

Only works on MonoBehaviours

Means we can reduce the number of null checks for those specific fields are we know it'll always have a value on every prefab/scene object

Just have this issue: https://github.com/unitystation/unitystation/pull/7428#issuecomment-929509662


Bod9001:
could reuse the naughty attributes one?
https://github.com/dbrizov/NaughtyAttributes#required

corp-0:
what's the state of this one?

ThatDan123:
Its working fine for SO's and components for when the variable its checking is on those classes directly.

Having issues for when it needs to check a list of a class and those individual variables e.g:

image

image

corp-0:
uhm, for a list we could enforce assigning an empty list instead of null

ThatDan123:
> uhm, for a list we could enforce assigning an empty list instead of null

The things inside the list not the list itself, see `List<PoolData>` and the CannotBeNull attribute in `PoolData`

Bod9001:
could try using some reflection?

vlandemart (approved):
Looks fine to me

PR #7447, opened 2021-09-24
Fixes a lot of server affecting bugs

DooblyNoobly:
## Mapping:
CL: [Fix] Fallstation Map: Fixes Security Camera Already Instantiated Bug on Fall Station
CL: [Fix] Relicta Map: Fixes Fire Closet being parented to Tool Closet

##NREs:
- Lots of NRE's found in HackingBase:
- One found and fixed in Addressables:
- Attempted to clamp Arithmetic operation to stop overflow for the Reactor Chamber temp calculations
- Fixes missing null check on HealthMaster on BodyPart UI Object validation
- Fixed wrong State condition check in BoltsModule for Doors
- Continue if object not found for body part UI slot Item
- Fixes GetOxy null NRE

### DIFF
--- a/UnityProject/Assets/Scenes/AwaySites/Relicta.unity
+++ b/UnityProject/Assets/Scenes/AwaySites/Relicta.unity
@@ -1545,6 +1545,7 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   MatrixSync: {fileID: 0}
+  matrix: {fileID: 0}
 --- !u!114 &60906935
 MonoBehaviour:
   m_ObjectHideFlags: 0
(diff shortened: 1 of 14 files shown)


Bod9001 (dismissed):

MaxIsJoe (dismissed):
Everything seems good, i would be a little worried about the null checks instead of just ensuring that things get assigned correctly but its good for now :thumbsup:

NoooneyDude:
> Everything seems good, i would be a little worried about the null checks instead of just ensuring that things get assigned correctly but its good for now 👍

smash it

Bod9001:
the clamp on the reactor stuff, needs to be looked into some time but I suppose it stops it exploding randomly

ThatDan123 (approved):

Bod9001 (approved):

PR #7448, opened 2021-09-24
Adds Rshub compatibility

DooblyNoobly:
Pretty simple, if you are already logged into firebase and pass the server and port argument via the commandline then you will auto connect to a server just like you would with an "authed" StationHub.


NoooneyDude:
Want to rebase on latest develop?

DooblyNoobly:
Yeah I tried but its saying all of develop is different files compared to my branch
I might just start again

PR #7449, opened 2021-09-24
RShub Compatibility 02

DooblyNoobly:
Pretty simple, if you are already logged into firebase and pass the server and port argument via the commandline then you will auto connect to a server just like you would with an "authed" StationHub.


Bod9001 (approved):

corp-0:
looks good

corp-0 (approved):

PR #7458, opened 2021-09-25
Ignore user editor settings

NoooneyDude:
My recently-opened scenes are not your recently-opened scenes.


corp-0:
just adding to gitignore isn't enough, you have to purge it from history as well. Take a read here: https://stackoverflow.com/questions/1274057/how-can-i-make-git-forget-about-a-file-that-was-tracked-but-is-now-in-gitign

NoooneyDude:
It's alright, was just to stop recently-opened scenes from resetting when I reset to develop.

Ah, right.

ThatDan123 (approved):

Bod9001 (approved):

PR #7478, opened 2021-09-28
Fixes gateway trap

### LINKED ISSUE #7392: Gateway portal is a trap
### Environment
- Game Version: Staging 21090800

when you get teleported out of the gateway you get stuck on the tile it places you

easiest place to replicate is in the relicta away site

https://user-images.githubusercontent.com/80334192/132609717-75870d3b-d368-497f-8e4e-e1a5f51ecff7.mp4

### Steps to reproduce

1.  start round making sure the away site is relicta
2. step into the gateway portal

MaxIsJoe:
fixes #7392 

`UpdateMe()` was acting a little bit weird and has often called `DetectPlayer()` more than once which caused weird issues where if two players walked in less than a second they'd both be teleported back to teleport cords even if players did not get trapped and one of them has moved a couple of tiles away from the teleporter.
This is without the mentioning of the bug where for some reason the player's matrix would get stuck on the teleporter if we don't halt `DetectPlayer()` which was the whole reason I looked into this.

Has been tested locally and on a client and could no longer reproduce this bug after introducing this fix.

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Gateway/StationGateway.cs
+++ b/UnityProject/Assets/Scripts/Objects/Gateway/StationGateway.cs
@@ -1,4 +1,5 @@
-﻿using Mirror;
+﻿using System.Collections;
+using Mirror;
 using UnityEngine;
 using System.Linq;
 using UnityEditor;
@@ -31,6 +32,8 @@ namespace Objects
 		private Sprite[] Offline = null;
 		[SerializeField]
 		private Sprite[] PowerOff = null;
+
+		private bool isOnCooldown = false;
 
 		private WorldGateway selectedWorld;// The world from the list that was chosen
 
@@ -118,7 +121,8 @@ namespace Objects
 
 			if (loadNormally)
 			{
-				WaitTimeBeforeActivation = Random.Range(RandomCountBegining, RandomCountEnd);
+				//WaitTimeBeforeActivation = Random.Range(RandomCountBegining, RandomCountEnd);
+				WaitTimeBeforeActivation = 30f;
 			}
 
 			Invoke(nameof(ConnectToWorld), WaitTimeBeforeActivation);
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Objects/Gateway/StationGateway.cs
```diff
@@ -146,14 +149,14 @@ protected virtual void UpdateMe()
...
 				{
 					DetectPlayer();
 					timeElapsedServer = 0;
 				}
 
 				timeElapsedServerSound += Time.deltaTime;
-				if (timeElapsedServerSound > SoundLength && isOn)
+				if (timeElapsedServerSound > SoundLength && isOn && isOnCooldown == false)
 				{
 					DetectPlayer();
```
ThatDan123: Think the actual issue is this.

Not sure why I made it do a detect on sound play time as well as detection time, think it was the second thing I coded for the project :p

MaxIsJoe: Removed it but my fix is still required for the other issue mentioned.

MaxIsJoe:
pls merge

ThatDan123 (approved):

corp-0:
is this the same issue #7489 fixed?

MaxIsJoe:
> is this the same issue #7489 fixed?

Nope, different fix for different issues.

Aranclanos:
what is this trying to fix?

MaxIsJoe:
> what is this trying to fix?

1. Sound event causes the game to try to detect the player it shouldn't
2. An odd issue where if a player walked in then another; both would get set back to teleportCords even if the first player has walked away for a couple of tiles.

I'm sure the second and third has been fixed at it's root thanks to your PR, but the the sound event shouldn't be trying to detect players.

Aranclanos:
take a different approach, this class already has some sort of cooldown, adding it to the cooldown of UpdateMe you'll end up with 3 different cooldowns

Aranclanos:
As in, simplify it rather to add more into the mess
its true that the gateway code, while short, looks a bit overcomplicated if not nonsensical at points

PR #7479, opened 2021-09-28
Gives AI back its power. Fixes 7471

AtnerNT:
### Purpose
Fixes a bug where the APCPoweredDevice ConnectToClosest function would always return false.

### Notes:
The colliders for the APCs are remaining disabled during runtime, so Physics2D overlaps don't work. Gonna try find out why this is happening so this is a draft for now. The Pr does work in its current state, but is not ideal

### Changelog:
CL: [Fix] AI's now connect to the APC on spawn again


ThatDan123:
FindObjectsOfType is terrible, we need to fix the issue of the colliders being off not this

ThatDan123:
Think ive fixed the colliders now in #7485

AtnerNT:
Oh easy- if that PR works, Ill just close this one

PR #7481, opened 2021-09-28
Adds the ability to climb crates

### LINKED ISSUE #4878: You should be able to climb on top of crates.
## Description
Title. In /tg/ you can climb on top of crates the exact same way you can climb onto tables. Notably, it takes just a second to climb onto a crate vs. a table. If you could climb onto crates, it would become impossible to trap people using them (except if you put people into locked ones, I guess). The script for climbing crates should also be usable for other objects, like surgical tables.

MaxIsJoe:
closes #4878

Adds in a script that's just plug-n-play for gameObjects that should be climbable (like crates)
Climb time is adjustable in the inspector

CL: [improvement] added the ability to climb crates.

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Climbable.cs
+++ b/UnityProject/Assets/Scripts/Objects/Climbable.cs
@@ -1,0 +1,72 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class Climbable : MonoBehaviour, ICheckedInteractable<MouseDrop>
+{
+
+	public float ClimbTime = 1.0f;
+
+
+	public bool WillInteract(MouseDrop interaction, NetworkSide side)
+	{
+		if (!DefaultWillInteract.Default(interaction, side)) return false;
+
+		PlayerSync playerSync;
+		CustomNetTransform netTransform;
+		if(interaction.UsedObject.TryGetComponent(out playerSync))
+		{
+			if (playerSync.IsMoving || playerSync.playerMove.IsBuckled)
+			{
+				return false;
+			}
+
+			// Do a sanity check to make sure someone isn't dropping the shadow from like 9000 tiles away.
+			var mag = (interaction.TargetObject.transform.position - playerSync.ServerPosition).magnitude;
+			if (mag > PlayerScript.interactionDistance)
+			{
+				return false;
+			}
+		}
+		else if(interaction.UsedObject.TryGetComponent(out netTransform)) // Do the same check but for mouse draggable objects this time.
+		{
+			var mag = (interaction.TargetObject.transform.position - playerSync.ServerPosition).magnitude;
+			if (mag > PlayerScript.interactionDistance)
+			{
+				return false;
+			}
+		}
+		else // Not sure what this object is so assume that we can't interact with it at all.
+		{
+			return false;
+		}
+
+		return true;
+	}
+
+	public void ServerPerformInteraction(MouseDrop interaction)
+	{
+		StandardProgressActionConfig cfg =
+			new StandardProgressActionConfig(StandardProgressActionType.Construction, false, false, false);
+		var x = StandardProgressAction.Create(cfg, () =>
+		{
+			PlayerScript playerScript;
+			if (interaction.UsedObject.TryGetComponent(out playerScript))
+			{
+				playerScript.PlayerSync.SetPosition(this.gameObject.WorldPosServer());
+			}
+			else
+			{
+				var transformComp = interaction.UsedObject.GetComponent<CustomNetTransform>();
+				if (transformComp != null)
+				{
+					transformComp.AppearAtPositionServer(this.gameObject.WorldPosServer());
+				}
+			}
+		}).ServerStartProgress(interaction.UsedObject.RegisterTile(), ClimbTime, interaction.Performer);
+
+		Chat.AddActionMsgToChat(interaction.Performer,
+			$"You begin climbing onto the {interaction.TargetObject.gameObject.ExpensiveName()}...",
+			$"{interaction.Performer.ExpensiveName()} begins climbing onto the {interaction.TargetObject.gameObject.ExpensiveName()}...");
+	}
+}
--- a/UnityProject/Assets/Scripts/Objects/Climbable.cs
+++ b/UnityProject/Assets/Scripts/Objects/Climbable.cs
@@ -1,4 +1,4 @@
-using UnityEngine;
+-using UnityEngine;
 
 namespace Objects
 {
@@ -15,12 +15,12 @@ namespace Objects
 				if (playerSync.IsMoving || playerSync.playerMove.IsBuckled) return false;
 
 				// Do a sanity check to make sure someone isn't dropping the shadow from like 9000 tiles away.
-				float mag = (interaction.TargetObject.transform.position - playerSync.ServerPosition).magnitude;
+				float mag = (interaction.TargetObject.transform.position - playerSync.ClientPosition).magnitude;
 				if (mag > PlayerScript.interactionDistance) return false;
 			}
 			else if (interaction.UsedObject.TryGetComponent(out netTransform)) // Do the same check but for mouse draggable objects this time.
 			{
-				float mag = (interaction.TargetObject.transform.position - playerSync.ServerPosition).magnitude;
+				float mag = (interaction.TargetObject.transform.position - netTransform.ServerPosition).magnitude;
 				if (mag > PlayerScript.interactionDistance) return false;
 			}
 			else // Not sure what this object is so assume that we can't interact with it at all.
--- a/UnityProject/Assets/Scripts/Objects/Climbable.cs
+++ b/UnityProject/Assets/Scripts/Objects/Climbable.cs
@@ -8,8 +8,6 @@ namespace Objects
 
 		public bool WillInteract(MouseDrop interaction, NetworkSide side)
 		{
-			if (DefaultWillInteract.Default(interaction, side) == false) return false;
-			
 			PlayerSync playerSync;
 			CustomNetTransform netTransform;
 			if (interaction.UsedObject.TryGetComponent(out playerSync))
--- a/UnityProject/Assets/Scripts/Objects/Climbable.cs
+++ b/UnityProject/Assets/Scripts/Objects/Climbable.cs
@@ -1,4 +1,4 @@
--using UnityEngine;
+using UnityEngine;
 
 namespace Objects
 {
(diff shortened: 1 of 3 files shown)


Bod9001 (dismissed):

UnityProject/Assets/Scripts/Objects/Climbable.cs
```diff
@@ -0,0 +1,72 @@
...
+		if(interaction.UsedObject.TryGetComponent(out playerSync))
+		{
+			if (playerSync.IsMoving || playerSync.playerMove.IsBuckled)
+			{
+				return false;
+			}
+
+			// Do a sanity check to make sure someone isn't dropping the shadow from like 9000 tiles away.
+			var mag = (interaction.TargetObject.transform.position - playerSync.ServerPosition).magnitude;
+			if (mag > PlayerScript.interactionDistance)
```
corp-0: move the magnitud check to a private method so you can reuse it.

UnityProject/Assets/Scripts/Objects/Climbable.cs
```diff
@@ -0,0 +1,72 @@
...
+
+public class Climbable : MonoBehaviour, ICheckedInteractable<MouseDrop>
+{
+
+	public float ClimbTime = 1.0f;
+
+
+	public bool WillInteract(MouseDrop interaction, NetworkSide side)
+	{
+		if (!DefaultWillInteract.Default(interaction, side)) return false;
```
corp-0:
```suggestion
		if (DefaultWillInteract.Default(interaction, side) == false) return false;
```

UnityProject/Assets/Scripts/Objects/Climbable.cs
```diff
@@ -0,0 +1,72 @@
...
+	public void ServerPerformInteraction(MouseDrop interaction)
+	{
+		StandardProgressActionConfig cfg =
+			new StandardProgressActionConfig(StandardProgressActionType.Construction, false, false, false);
+		var x = StandardProgressAction.Create(cfg, () =>
+		{
+			PlayerScript playerScript;
+			if (interaction.UsedObject.TryGetComponent(out playerScript))
+			{
+				playerScript.PlayerSync.SetPosition(this.gameObject.WorldPosServer());
```
corp-0: shouldn't this have a small offset in y?

corp-0 (changes requested):
solve the codacy issues and my review, then it should be good to go.

UnityProject/Assets/Scripts/Objects/Climbable.cs
ThatDan123: Pretty sure `if (!DefaultWillInteract.Default(interaction, side)) return false;` will check the distance, though double check

corp-0: yeah, that makes sense. Maybe Max found that check in our current implementation of climbing tables? I know we had some issues with distance before.

corp-0:
if we want to reuse this component for say, printers, we need to also hold a reference to whatever has climbed on top of the thing

UnityProject/Assets/Scripts/Objects/Climbable.cs
MaxIsJoe: wdym? why would we offset this?
If you mean like an offset where it visually appears as if the player is ontop of the crate; I'd rather that we add pixel shifting first instead of messing with the player's position that will probably get set back to an Int due to matrixes and what not.

UnityProject/Assets/Scripts/Objects/Climbable.cs
MaxIsJoe: No, it doesn't work.

MaxIsJoe:
> if we want to reuse this component for say, printers, we need to also hold a reference to whatever has climbed on top of the thing

pretty sure you can already just get that from the matrix
no idea why you'd want a reference of what's ontop of the object as well and sounds like it won't be used by anything currently.

UnityProject/Assets/Scripts/Objects/Climbable.cs
```diff
@@ -0,0 +1,60 @@
...
+		public bool WillInteract(MouseDrop interaction, NetworkSide side)
+		{
+			PlayerSync playerSync;
+			CustomNetTransform netTransform;
+			if (interaction.UsedObject.TryGetComponent(out playerSync))
+			{
+				if (playerSync.IsMoving || playerSync.playerMove.IsBuckled) return false;
+
+				// Do a sanity check to make sure someone isn't dropping the shadow from like 9000 tiles away.
+				float mag = (interaction.TargetObject.transform.position - playerSync.ServerPosition).magnitude;
```
ThatDan123: `playerSync.ServerPosition` will need to be client position when the side is the client

MaxIsJoe: wouldn't that allow for players to cheat? or is that variable stored on the server and not the client itself?

ThatDan123: `WillInteract` is called by both client and server to validate the interaction. `playerSync.ServerPosition` is only valid server side so only use it when the side (using the parameter) is server then use the client position for client

UnityProject/Assets/Scripts/Objects/Climbable.cs
corp-0: true

UnityProject/Assets/Scripts/Objects/Climbable.cs
MaxIsJoe: I think I've solved what you wanted

Bod9001 (dismissed):
just waiting on that Dan

ThatDan123:
Tests are failing

MaxIsJoe:
> Tests are failing

found out why and fixed it

UnityProject/Assets/Scripts/Objects/Climbable.cs
```diff
@@ -0,0 +1,60 @@
...
+			}
+
+			return true;
+		}
+
+		public void ServerPerformInteraction(MouseDrop interaction)
+		{
+			StandardProgressActionConfig cfg =
+				new StandardProgressActionConfig(StandardProgressActionType.Construction, false, false, false);
+			var x = StandardProgressAction.Create(cfg, () =>
```
NoooneyDude:
```suggestion
			StandardProgressAction.Create(cfg, () =>
```

NoooneyDude (dismissed):

corp-0 (dismissed):

MaxIsJoe:
was about to fix that, thank

corp-0:
What did I do to dismiss the review?

MaxIsJoe:
> What did I do to dismiss the review?

I think you dismissed Nooney's suggestion before he resolved it by approving the changes.

also did I do what Dan wanted correctly?

corp-0:
No, Mooney did a suggestion to commit changes from it. 

I'll wait for Daniel to resolve the conversation

NoooneyDude:
> What did I do to dismiss the review?

Haha, that's pretty weird 🤔 surely the label didn't upset it?

MaxIsJoe:
> > What did I do to dismiss the review?
> 
> Haha, that's pretty weird 🤔 surely the label didn't upset it?

Say sorry to it >:(

UnityProject/Assets/Scripts/Objects/Climbable.cs
```diff
@@ -0,0 +1,60 @@
...
+				float mag = (interaction.TargetObject.transform.position - netTransform.ServerPosition).magnitude;
+				if (mag > PlayerScript.interactionDistance) return false;
+			}
+			else // Not sure what this object is so assume that we can't interact with it at all.
+			{
+				return false;
+			}
+
+			return true;
+		}
```
ThatDan123:
```suggestion
		public bool WillInteract(MouseDrop interaction, NetworkSide side)
		{
			if (interaction.TargetObject == null) return false;
			if (interaction.UsedObject == null) return false;

			var targetObjectPos = side == NetworkSide.Server ?
				interaction.TargetObject.WorldPosServer() : interaction.TargetObject.WorldPosClient();

			if (interaction.UsedObject.TryGetComponent<PlayerSync>(out var playerSync))
			{
				if (playerSync.IsMoving || playerSync.playerMove.IsBuckled) return false;

				// Do a sanity check to make sure someone isn't dropping the shadow from like 9000 tiles away.
				float mag = (targetObjectPos - (side == NetworkSide.Server ?
					             playerSync.ServerPosition : playerSync.ClientPosition)).magnitude;

				return mag <= PlayerScript.interactionDistance;
			}

			// Do the same check but for mouse draggable objects this time.
			if (interaction.UsedObject.TryGetComponent<CustomNetTransform>(out var netTransform))
			{
				if (netTransform.PushPull.IsNotPushable) return false;

				float mag = (targetObjectPos - (side == NetworkSide.Server ?
					             netTransform.ServerPosition : netTransform.ClientPosition)).magnitude;

				return mag <= PlayerScript.interactionDistance;
			}

			return false;
		}
```

UnityProject/Assets/Scripts/Objects/Climbable.cs
```diff
@@ -0,0 +1,60 @@
...
+
+		public void ServerPerformInteraction(MouseDrop interaction)
+		{
+			StandardProgressActionConfig cfg =
+				new StandardProgressActionConfig(StandardProgressActionType.Construction, false, false, false);
+			StandardProgressAction.Create(cfg, () =>
+			{
+				PlayerScript playerScript;
+				if (interaction.UsedObject.TryGetComponent(out playerScript))
```
ThatDan123:
```suggestion
				if (interaction.UsedObject.TryGetComponent<PlayerScript>(out var playerScript))
```

UnityProject/Assets/Scripts/Objects/Climbable.cs
```diff
@@ -0,0 +1,60 @@
...
+				{
+					playerScript.PlayerSync.SetPosition(this.gameObject.WorldPosServer());
+				}
+				else
+				{
+					var transformComp = interaction.UsedObject.GetComponent<CustomNetTransform>();
+					if (transformComp != null)
+					{
+						transformComp.AppearAtPositionServer(this.gameObject.WorldPosServer());
+					}
```
ThatDan123:
```suggestion
					if (interaction.UsedObject.TryGetComponent<CustomNetTransform>(our var transformComp))
					{
						transformComp.AppearAtPositionServer(this.gameObject.WorldPosServer());
					}
```

UnityProject/Assets/Scripts/Objects/Climbable.cs
ThatDan123: I've made a suggestion for it

MaxIsJoe:
Done.

ThatDan123 (approved):

corp-0 (approved):

PR #7485, opened 2021-09-28
Fix Server Colliders

### LINKED ISSUE #7471: AI broken on staging
### Environment
- Game Version: Staging Latest
- OS Version: Arch Linux

AI camera tracking doesnt work, it cannot even track mobs
lifeforms

also when you try to camera more as AI is says that you cannot because there is no power, but there is power when i was testing it
AI2

### Steps to reproduce

1.join the game as AI or respawn as AI

ThatDan123:
Note: is untested, but previously the function to turn off colliders was only called from OnStartClient, this now prevents the server running it.

Fixes #7471

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Transform/CustomNetTransform.cs
+++ b/UnityProject/Assets/Scripts/Core/Transform/CustomNetTransform.cs
@@ -187,11 +187,13 @@ public partial class CustomNetTransform : NetworkBehaviour, IPushable
 		{
 			InitServerState();
 		}
-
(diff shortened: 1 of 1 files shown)


corp-0:
how should it be tested?

ThatDan123:
Think theres a bug currently when spawning as AI having no power as the overlap cant see the collider of the APC. So to test just spawn as AI and see if you have power

NoooneyDude (approved):

Bod9001 (approved):

PR #7502, opened 2021-10-02
Removes an expensive null check in matrix get from a fixed runtime

Aranclanos:
### Purpose
Removes an expensive null check in matrix get from a fixed runtime.
Fixed in #7489


ThatDan123 (approved):

ThatDan123:
You could add a TryCatch

corp-0:
We might need to investigate nullables, as suggested by the ss14 boys and use that instead

NoooneyDude (approved):

PR #7503, opened 2021-10-02
Removes an expensive trycatch in PlayerScript IsGhost for a runtime already fixed

Aranclanos:
### Purpose
Removes an expensive trycatch in PlayerScript IsGhost for a runtime already fixed
Fixed in #7489


ThatDan123:
Keep the Try Catch but remove the `gameObject == null` check. The Try Catch has no performance cost until it is catching something

The null check is the drain here

NoooneyDude (approved):

ThatDan123 (approved):

PR #7504, opened 2021-10-03
Fix VS false errors

NoooneyDude:
- Fix VS false errors due to duplicate dlls.


corp-0:
vs false errors?

NoooneyDude:
> vs false errors?

Yeah, all these and more:

errors

corp-0:
Can you explain what those dlls were?

NoooneyDude:
> Can you explain what those dlls were?

Not entirely sure, but you can find the new location here https://github.com/unitystation/unitystation/tree/develop/UnityProject/Assets/Parse/Plugins/dotNet45

corp-0 (approved):

Bod9001 (approved):

PR #7515, opened 2021-10-04
Removes a trycatch and a null check in bodypart ashing

Aranclanos:
### Purpose
Removes the CannotBeAshed bool, ashing will now check for fire resistance.
Organizes bodyPart damage, organ damage and trauma calls.
Removes unneeded an unneeded try catch in body part ashing.
Adds DamageOrgans() to differentiate, separate and organize the damage inbetween a body part and an organ.

### DIFF
--- a/UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartDamage.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartDamage.cs
@@ -3,6 +3,8 @@ using System.Collections;
 using System.Collections.Generic;
 using NaughtyAttributes;
 using UnityEngine;
+using System.Linq;
+using Random = System.Random;
 
 namespace HealthV2
 {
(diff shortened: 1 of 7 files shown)


UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartDamage.cs
```diff
@@ -221,52 +224,67 @@ public void AffectDamage(float damage, int damageType)
...
+				}
 			}
 
-			if(Severity <= DamageSeverity.LightModerate) return;
-			if (damageType == DamageType.Burn || attackType == AttackType.Fire
-			                                  || attackType == AttackType.Laser || attackType == AttackType.Energy)
+			if (damageType == DamageType.Burn || attackType == AttackType.Fire ||
+			    attackType == AttackType.Laser || attackType == AttackType.Energy)
 			{
-				ApplyTraumaDamage(TraumaticDamageTypes.BURN, true);
```
MaxIsJoe, line 266: why was this pass removed?

Aranclanos: it never did anything due to the `if(Severity <= DamageSeverity.LightModerate) return;` line that I also removed a couple lines before
so now the lightmoderate check will still exist and used, but inside ApplyTraumaDamage

MaxIsJoe: Everything seems to be good, tested as well and no issues creeping up.

MaxIsJoe (approved):
Everything seems to be good, tested as well and no issues creeping up.

NoooneyDude (approved):

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -287,13 +284,13 @@ private void PeriodicUpdate()
 		{
 			if (IsDead == false)
 			{
-				for (int i = BodyPartList.Count - 1; i >= 0; i--)
+				foreach (var bodyPart in BodyPartList)
 				{
-					BodyPartList[i].ImplantPeriodicUpdate();
+					bodyPart.ImplantPeriodicUpdate();
 				}
```
ThatDan123, line 290: Are you sure these for loops arent needed due to changing the BodyPartList and so needs to be looped backwards?

Aranclanos: yes

ThatDan123 (approved):

PR #7531, opened 2021-10-07
Fixes a whole lot of food bugs.

VernaSaurus:
CL: [Fix] Adds icons for all missing crafting tabs
CL: [Fix] allows some previously broken foods to be microwaved again
CL: [Fix] makes the food processor no longer vanish into the infinite abyss!


UnityProject/Assets/Scripts/Systems/CraftingV2/CraftingActionParameters.cs
```diff
@@ -16,13 +16,24 @@ public class CraftingActionParameters
 		/// </summary>
 		public bool IgnoreToolsAndIngredients => ignoreToolsAndIngredients;
 
+		private bool shouldProclaimSuccess;
+
+		/// <summary>
+		/// 	Should we send feedback about a SUCCESSFUL crafting status to the player?
+		/// </summary>
+		public bool ShouldProclaimSuccess => shouldProclaimSuccess;
```
Lizurt: Maybe it would be better to replace these 2 bools with 1 enum or a bit mask?

I mean `bool shouldGiveFeedback` and `bool shouldProclaimSuccess`

UnityProject/Assets/Scripts/Systems/CraftingV2/CraftingIngredient.cs
```diff
@@ -72,9 +72,21 @@ public bool WillInteract(HandApply interaction, NetworkSide side)
...
+					interaction.PerformerPlayerScript.PlayerCrafting.CanClientCraft(
+						relatedRecipe.Recipe,
+						possibleIngredients,
+						possibleTools
+					) == CraftingStatus.AllGood
+				)
+				{
+					return true;
+				}
+				}
```
Lizurt, line 86: Usually I don't judge other programmers' code styles, but sadly can't ignore the lack of tabulation here since it affects readability a lot in my opinion

UnityProject/Assets/Scripts/Systems/CraftingV2/CraftingIngredient.cs
```diff
@@ -85,6 +97,7 @@ public bool WillInteract(HandApply interaction, NetworkSide side)
 				{
 					return true;
 				}
+				}
```
Lizurt, line 99: Same thing with tabulation

UnityProject/Assets/Scripts/Systems/CraftingV2/CraftingIngredient.cs
NoooneyDude: Maybe we could look into https://github.com/marketplace/actions/dotnet-format

VernaSaurus:
alright, i'll fix the indentation

VernaSaurus:
changed from bools to an enum, as per request

VernaSaurus:
I keep realizing i made a bunch of small mistakes oops

VernaSaurus:
you know what while i'm at it i'll also fix the stack deletion bug

VernaSaurus:
microwaves are now fixed

UnityProject/Assets/Scripts/Objects/Machines/Microwave.cs
```diff
@@ -228,7 +228,21 @@ private bool TryTransferToMicrowave(PositionalHandApply interaction)
...
+			else
+			{
+				var item = stack.ServerRemoveOne(); // might delete the item, i'm not sure
+				added = Inventory.ServerAdd(item, storageSlot);
+				if(!added)
+				{
+					stack.ServerIncrease(1); // adds back the lost amount to prevent microwaves from eating stackable items
+				}
+			}
+			if (storageSlot == null || added == false)
```
NoooneyDude: Do we want to allow full stacks to be cooked at once? ~~Cramming 50 cutlets into the microwave might be a little excessive~~ (edit: actually, looks like the max. stack size for cutlets is 6, which is just fine). Would it be simpler and make more sense to only allow adding one item per microwave slot? Then again, sure, why not: I can just imagine an SS13 chef cramming them all in, Mr. Bean style. 

You might also be interested in `TryGetComponent`:
```csharp
ItemObject.TryGetComponent<Stackable>(out var stackable)
{
    // do something with stackable
}
else
{
    // didn't have stackable
}
```

Just a tiny bit nicer than null checking a `GetComponent`.

Anyway, I think it makes more sense to fiddle around with the stackable at the moment of item transformation, somewhere in `CheckCooked()` method. See example:
```csharp
// spawn cooked item (cookedItem)
if (cookable.TryGetComponent<Stackable>(out var inStackable)
        && cookedItem.TryGetComponent<Stackable>(out var outStackable))
{
    outStackable.ServerSetAmount(inStackable.Amount);
}
// despawn old item (cookable)
// add cooked item to microwave inventory (ready for further cooking if possible)
```
This'd be a lot more concise addition. Though, by this point, I imagine `CheckCooked()`'d be a pretty chunky method, so might want to move the body of `cookable.AddCookingTime()` to a new method, `CookItem(cookable)`, just to keep complexity down.

I try not to impede too much on people's style but I think I've suggested some good improvements here.

NoooneyDude (commented):
`{`, `}`, should be on their own lines - it's part of the style guide for consistency.

UnityProject/Assets/Scripts/Objects/Machines/Microwave.cs
VernaSaurus: I tried the second part earlier, and it would constantly replace the value with 1 regardless of what I set it to. I have no idea why this is, but it was frustrating.

VernaSaurus: The first part though, now that I can do.

VernaSaurus:
did the thing, but i'm going to leave putting food into microwaves as-is since cutlet creation is likely going to be switched to the crafting system and cutlets will have their stackable component removed

UnityProject/Assets/Scripts/Systems/CraftingV2/CraftingIngredient.cs
```diff
@@ -72,18 +72,33 @@ public bool WillInteract(HandApply interaction, NetworkSide side)
...
 				}
-
-				if
-				(
-					interaction.PerformerPlayerScript.PlayerCrafting.CanCraft(
+				if(side == NetworkSide.Client)
+				{
+					if 
+					(
+						interaction.PerformerPlayerScript.PlayerCrafting.CanClientCraft(
```
NoooneyDude:
```suggestion
				if (side == NetworkSide.Client)
				{
					if (interaction.PerformerPlayerScript.PlayerCrafting.CanClientCraft(
```

UnityProject/Assets/Scripts/Systems/CraftingV2/CraftingIngredient.cs
```diff
@@ -72,18 +72,33 @@ public bool WillInteract(HandApply interaction, NetworkSide side)
...
+						) == CraftingStatus.AllGood
+					)
+					{
+						return true;
+					}
+				}
+				else if(side == NetworkSide.Server){
+					if 
+					(
+						interaction.PerformerPlayerScript.PlayerCrafting.CanCraft(
```
NoooneyDude:
```suggestion
				else if (side == NetworkSide.Server)
				{
					if (interaction.PerformerPlayerScript.PlayerCrafting.CanCraft(
```

NoooneyDude (approved):

Bod9001 (approved):

UnityProject/Assets/Scripts/Player/PlayerCrafting.cs
```diff
@@ -452,7 +452,12 @@ CraftingActionParameters craftingActionParameters
...
-				if (craftingActionParameters.ShouldGiveFeedback && craftingStatus != CraftingStatus.AllGood)
+				if (craftingActionParameters.Feedback == FeedbackType.GiveAllFeedback)
+				{
+					GiveClientSidedFeedback(craftingStatus, recipe, false);
+					return;
+				}
+				else if (craftingActionParameters.Feedback == FeedbackType.GiveOnlySuccess && craftingStatus == CraftingStatus.AllGood)
 				{
 					GiveClientSidedFeedback(craftingStatus, recipe, false);
 					return;
```
Lizurt, line 463: These returns block crafting logic below, so we cannot longer craft

PR #7543, opened 2021-10-08
Remove extra NetworkIdentity component on lighters

NoooneyDude:
- Removes extra NetworkIdentity component from lighters. Fixes error on start.

### DIFF
--- a/UnityProject/Assets/Mirror/Runtime/NetworkIdentity.cs
+++ b/UnityProject/Assets/Mirror/Runtime/NetworkIdentity.cs
@@ -321,7 +321,7 @@ namespace Mirror
             if (hasSpawned)
             {
 				// Unitystation edit: add more information
-                Debug.LogError($"{name} in {transform.parent.OrNull()?.name} at position {transform.position} has already spawned. " +
+                Debug.LogError($"{name} in {(transform.parent == null ? "(no parent)" : transform.parent.name)} at position {transform.position} has already spawned. " +
 						$"Don't call Instantiate for NetworkIdentities that were in the scene since the beginning (aka scene objects). " +
 						$"Otherwise the client won't know which object to use for a SpawnSceneObject message.");
                 SpawnedFromInstantiate = true;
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Mirror/Runtime/NetworkIdentity.cs
```diff
@@ -320,7 +320,10 @@ internal void Awake()
 
             if (hasSpawned)
             {
-                Debug.LogError($"{name} has already spawned. Don't call Instantiate for NetworkIdentities that were in the scene since the beginning (aka scene objects).  Otherwise the client won't know which object to use for a SpawnSceneObject message.");
+				// Unitystation edit: add more information
+                Debug.LogError($"{name} in {transform.parent.name} at position {transform.position} has already spawned. " +
```
ThatDan123:
```suggestion
                Debug.LogError($"{name} in {transform.parent.OrNull()?.name} at position {transform.position} has already spawned. " +
```

Could be null if theres something wrong.

UnityProject/Assets/Mirror/Runtime/NetworkIdentity.cs
```diff
@@ -320,7 +320,10 @@ internal void Awake()
 
             if (hasSpawned)
             {
-                Debug.LogError($"{name} has already spawned. Don't call Instantiate for NetworkIdentities that were in the scene since the beginning (aka scene objects).  Otherwise the client won't know which object to use for a SpawnSceneObject message.");
+				// Unitystation edit: add more information
+                Debug.LogError($"{name} in {transform.parent.OrNull()?.name} at position {transform.position} has already spawned. " +
+						$"Don't call Instantiate for NetworkIdentities that were in the scene since the beginning (aka scene objects). " +
+						$"Otherwise the client won't know which object to use for a SpawnSceneObject message.");
```
NoooneyDude:
```suggestion
                Debug.LogError($"{name} in {(transform.parent == null ? "(no parent)" : transform.parent.name)} at position {transform.position} has already spawned. " +
						$"Don't call Instantiate for NetworkIdentities that were in the scene since the beginning (aka scene objects). " +
						$"Otherwise the client won't know which object to use for a SpawnSceneObject message.");
```

Bod9001 (approved):

corp-0 (approved):

PR #7544, opened 2021-10-08
Failure log parameter to admin check

ThatDan123:
Should only do the log for admin only actions not shared actions

### DIFF
--- a/UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
+++ b/UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
@@ -534,7 +534,7 @@ public partial class PlayerNetworkActions : NetworkBehaviour
 	[Command]
 	public void CmdRespawnPlayer()
 	{
-		if (AdminCommandsManager.IsAdmin(connectionToClient, out _) || GameManager.Instance.RespawnCurrentlyAllowed)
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
```diff
@@ -534,7 +534,7 @@ public void CmdCommitSuicide()
 	[Command]
 	public void CmdRespawnPlayer()
 	{
-		if (AdminCommandsManager.IsAdmin(connectionToClient, out _) || GameManager.Instance.RespawnCurrentlyAllowed)
+		if (AdminCommandsManager.IsAdmin(connectionToClient, out _, false) || GameManager.Instance.RespawnCurrentlyAllowed)
```
Bod9001, line 537: when would be true?

ThatDan123: It the parameter default is true, and is used else where like in the admin commands manager. The respawn player command is used for the ghost UI, so normal players can click it and they will fail the admin check which sends the log and can be seen on the admin webhook.

NoooneyDude (approved):

Bod9001 (approved):

PR #7553, opened 2021-10-10
Update TileChangesNewClientSync.cs

NoooneyDude:
- Reduces tile change batch size back to original 20.


corp-0:
lol next one will be 15

corp-0 (approved):

Bod9001 (approved):
We should optimise this message at some point

PR #7557, opened 2021-10-11
Fixes players missing tile updates if pipes were modified 

Aranclanos:
### Purpose
The server will now use an enum to dictate the orientation of a tile instead of sending an entire matrix4x4
Clients will now ask for the tile updates after the matrix are initialized.
Increases the max chunk size of a new player tile updates from 25 to 200.


Bod9001 (changes requested):
like, it hardly makes sense to plug-in an enum for something that is only used by pipes, and then have it be everywhere in the tile change system,
and especially in the future pipes are going to have layer offsets so, you can have multiple pipes on one tile without them interfering with each other, 
and the plan is just to use an offset on the matrix4x4 in the suggested changes that would require adding another field  to the change manager, imo it is not particularly hard checking if matrix4x4 = identity and not sending it if so and then making a more compact serialiser for it if it is different, since the current issue is that json.net includes properties so, each matrix4x4 has like five references to identity matrix that basically itself again

Aranclanos:
This issue was handled by #7561

PR #7558, opened 2021-10-11
Re-add files to DMI splitter tool that are necessary for it to function.

PetMudstone:
### Purpose
- Re-adds files to "DMI_splitter_and_meta_maker" that are needed for the DMI extraction and splitter tool to function. They got removed at some point and broke the tool.


Bod9001 (approved):

corp-0:
why are the meta needed? Isn't the tool an external tool?

PetMudstone:
Frankly I don't know why they're needed code wise. However, from testing the tool with and without the .meta files having them absent will only make a folder for a single .dmi in the "example" folder without filling said folder with any extracted .pngs or other files.

corp-0 (approved):

PR #7561, opened 2021-10-12
Compact tiles, tile changes are now using network reader and writer

Bod9001:
still need to calculate the upper limit but 1000 has been tested with no colour and matrix changes so, 350 I think is a good estimation if there were colour and make exchanges

Did a build, with editor as client
 tested as billed as client with reactor explosion and wall welding, worked


Aranclanos:
image
this is with like 9 pipe tiles in a single chunk, I'm not sure how many exactly because I noticed the changeList is increasing by like 100 each couple seconds with effects tile changes of atmos

Bod9001:
ok cool

Bod9001:
still a few bugs with how, stuff is initialised for example it doesn't like tiles been moved from underfloor, because in the fluorescent initialised by the time you get the messages

Bod9001:
should be good now

Aranclanos:
a few points:
- you're sending a matrix4x4 and wasting bandwidth for a coordinal position (norht, south, east, west)
- you're adding a new custom serialization/deserialization with magic numbers to replace a third party custom one. I don't think this is a good idea in the long run, maintaining it will be a headache
- the ore generator already has an isserver check
- changing IsSpaceAt accepting effects, tables and underfloor will results in unexpected errors and inaccuracies
- clients manually call their own MetaDataSystems Initialize, checking for IsServer defeats the purpose of it

if the problem is to ask for the tileupdates too early, you can follow the example of my pr, which fixes all points I mentioned and doesn't add further complications, but simplifies everything, looking forwards towards the longterm health of the project

Bod9001:
"changing IsSpaceAt accepting effects, tables and underfloor will results in unexpected errors and inaccuracies"
I think it's agreed-upon that tables, effects and underfloor  shouldn't affect whether or not something is space,
I think it's actually causing the inverse since when I made it so that it initialised when underfloor was initialised, that caused some parts of the station to be depressurised,

so the mere fact that underfloor isn't initialised by the time initialisation is done stop the station from being depressurised,
unexpected behaviour indeed


"you're sending a matrix4x4 and wasting bandwidth for a coordinal position (norht, south, east, west)"
the reader writer stream is I daresay 10 times more efficient than the  json serialisation even more efficient for data per packet

"you're adding a new custom serialization/deserialization with magic numbers to replace a third party custom one. I don't think this is a good idea in the long run, maintaining it will be a headache"

like, you could just replace the 1 , 2 and 255 with enum casted to bytes   that could be something that can be done, 

"clients manually call their own MetaDataSystems Initialize, checking for IsServer defeats the purpose of it" 
the reason it is there is because the client is too slow to initialise it before it starts receiving underfloor net messages, from the server, so it needs to be initialised before receiving those messages

imo I could porting over the code that I was working on to ensure client has loaded all scenes before requesting observing ( receiving updates about their data and stuff ) but I still had a bug with still adding observing to the online scene

corp-0:
change initialization to make Aran happy and we are good to go

Bod9001:
what would be specified changes?

Bod9001:
ok, will work on tomorrow

corp-0 (approved):

NoooneyDude (approved):

PR #7568, opened 2021-10-13
Fixes radiation not dealing damage

### LINKED ISSUE #7391: Radiation doesnt cause damage
### Environment
- Game Version: Staging 21090800
- OS Version: Arch Linux

Now radiation doesn't cause any kind of damage to the players on game.

example from being on top of the nuclear reactor (highest source of radiation possible):
reactor

also if you step on top of the fully started reactor core the health doll will do this.

https://user-images.githubusercontent.com/80334192/132604404-2811e6ad-2be0-42ce-a48c-d7c0073c55aa.mp4

### Steps to reproduce

1. take health analyzer and Geiger counter 
2. start the reactor until you get the biggest amount of radiation possible
3. step on top of the core and check health

Aranclanos:
### Purpose
Fixes radiation not dealing damage.
Removes radiation damage from bodyparts.
Adds radiation stacks to LivingHealthMasterBase.
Radiation will deal a maximum of 1 toxin damage per tick as long as there are radiation stacks.
Fixes #7391

### DIFF
--- a/UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
@@ -132,6 +132,11 @@ namespace HealthV2
 		[HideInInspector]
 		public PlayerSprites playerSprites;
 
+		/// <summary>
+		/// Amount of radiation sustained. Not actually 'stacks' but rather a float.
+		/// </summary>
+		public float RadiationStacks;
+
 		// FireStacks note: It's called "stacks" but it's really just a floating point value that
 		// can go up or down based on possible sources of being on fire. Max seems to be 20 in tg.
 		private float fireStacks => healthStateController.FireStacks;
@@ -241,7 +246,7 @@ namespace HealthV2
 		{
 			if (CustomNetworkManager.IsServer == false)
 				return;
-			UpdateManager.Add(PeriodicUpdate, 1f);
+			UpdateManager.Add(PeriodicUpdate, tickRate);
 		}
 
 		void OnDisable()
@@ -304,11 +309,27 @@ namespace HealthV2
 		[Server]
 		public void CalculateRadiationDamage()
 		{
-			var radLevel = (RegisterTile.Matrix.GetRadiationLevel(RegisterTile.LocalPosition) * (tickRate / 5f) / 6);
+			var radLevel = RegisterTile.Matrix.GetRadiationLevel(RegisterTile.LocalPosition) * tickRate * 0.001f;
 
-			if (radLevel == 0) return;
+			if (radLevel > 0)
+			{
+				foreach (var bodyPart in BodyPartList)
+				{
+					var damage = bodyPart.ArmorDamageReduction(radLevel, AttackType.Rad, 0);
+					RadiationStacks += damage;
+				}
+			}
 
-			ApplyDamageAll(null, radLevel * 0.001f, AttackType.Rad, DamageType.Radiation);
+			if (RadiationStacks > 0)
+			{
+				//the damage is sloooow
+				var toxDamage = Mathf.Min(RadiationStacks, 1);
+
+				ApplyDamageAll(null, toxDamage, AttackType.Rad, DamageType.Tox, armorPenetration: 100);
+				RadiationStacks -= toxDamage;
+
+				//maybe add burn damage or cool things like catching on fire if the stacks are super high here?
+			}
 		}
 
 		/// <summary>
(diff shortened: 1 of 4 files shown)


NoooneyDude (approved):

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -304,11 +309,27 @@ private void PeriodicUpdate()
...
 		{
-			var radLevel = (RegisterTile.Matrix.GetRadiationLevel(RegisterTile.LocalPosition) * (tickRate / 5f) / 6);
+			var radLevel = RegisterTile.Matrix.GetRadiationLevel(RegisterTile.LocalPosition) * tickRate * 0.001f;
+
+			if (radLevel > 0)
+			{
+				foreach (var bodyPart in BodyPartList)
+				{
+					var damage = bodyPart.ArmorDamageReduction(radLevel, AttackType.Rad, 0);
+					RadiationStacks += damage;
```
Bod9001, line 319: radiation should be localised to each body part, not to everybody part

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -304,11 +309,27 @@ private void PeriodicUpdate()
...
+				}
```
Bod9001, line 320: Health with these changes just push it towards a more messed up system with no unified structure

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
Bod9001: So, it has to go back to what radiation should do?

Radiation should be a penalty for standing in x area for too long, that usually has engine reactor that type of thing,
so, it's fair enough to have it as a slow thing, that deals damage over time,
the stacks works well, that slowly get transitioned and toxin, toxin because it's something more interesting than burn or brute damage,
 even though IRL radiation would be burns, 

The question is what you get from per body part radiation stacks?,
For example if you had a exposed arm you could amputate it to stop the effects of the radiation, very much like IRL,
if you had chest only protection it was stop your internal organs from taking damage, however stuff like arms and legs would still take damage, maybe this could be the downside of a ghetto radiation suit

Is there any downsides?
I can't think of any

Bod9001: Feel free to suggest any downsides

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
Aranclanos: nah this is miles better than having random `if ((int) DamageType.Radiation == i) continue;` checks

Bod9001: to be fair there's a bunch of them,  for excluding oxygen damage from overall health

Bod9001: except for the brain's oxygen damage

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
corp-0: I don't get how in this game you can get one part exposed and not another was it randomly?

Bod9001: no, if you have armour that only covers the chest, it means that stuff like your arms would get exposed to radiation

corp-0:
So what should we do with this pr?

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
WarlockD: I think it needs to be stacked on how many radiation is absorbed.  The yearly exposure for a person is 100 milli seivet gives you a clear increase in cancer.  Anything at 1 Sv causes throwing up, organ damage and the like.  Doesn't need to be burning though.  Anything after 10 should cause light burning and increased chance of tumors on the expanded area.

At 100 Sv, that's when the fun begins.  Almost guaranteed to amputate something or get surgery to remove skin tumors.  Skin grafts, etc (burn damage).  If you get that accumulated dose at your head or chest, the chance of nervous system shutdown is high.  Higher after 200.

Just some ideas:)  I mean basically, toxin damage ad medium dose's, and you get cooked like a microwave if your just sitting next to the thing.  Or picking it up in the middle of the woods where 20% normal nuclear accidents happen:P

VernaSaurus:
This can probably be closed thanks to #7843

NoooneyDude:
Did #7843 fix #7391?

PR #7573, opened 2021-10-14
Fix declaring war not counting down, and fix syndies being unable to shoot their guns

### LINKED ISSUE #6186: Nuke Ops cant use their guns
Apparently nuke ops cant use the guns they spawn with.

~To reproduce respawn as nuke ops try to fire guns~

This problem only happens when you spawn as a nuke ops naturally, this can be reproduced by setting the next gamemode as ops then you should spawn in after you ready up for the next round.

### LINKED ISSUE #7532: syndicate base is broken pls fix <3
### Environment
- Game Version: Staging 21100617
- OS Version: Arch Linux

MANY issues are present on the syndicate base where the nuclear operatives spawn.

(Edit) Summary:

- [ ] Syndicate base starts unpowered and there is no way to power it back (Electrical)
- [x] There are no buttons to get out of the base (Fixed on #7537)
- [x] Syndicate guns dont shoot under strange circumstances (Weapons) (Fixed on #7573)
- [x] War console once war is declared the countdown doesnt start (Fixed on #7573)
- [x] Syndicate shuttle is duplicated underneath, possibly the entire base is duplicated (Fixed on #7539)

## Power issues

first off when the nukies spawn on their rooms they are unable to exit out of them by normal means due to the base being unpowered and the doors being closed shut.

in this video you can see how connecting the plasma gen doesnt work to power the station, also all wires on the server get broken the moment the syndicate base is spawned:

might be related to issue #7333 

https://user-images.githubusercontent.com/80334192/136317326-a9463385-992e-4a0f-bc09-13fdaa71c881.mp4

## Buttons

the base lacks buttons to exit it and to access the syndicate shuttle

button3
button2
button

## Syndicate Weapons

while testing it i tried shooting the syndicate guns on staging as a nukie and nothing happened. This error showed off in the console each time i tried to shoot a syndicate gun.

[error.txt](https://github.com/unitystation/unitystation/files/7298606/error.txt)

NOTE: i restarted the round and the first thing i did was respawning myself as a nukie, i immediately tried the guns and they shot without problems, however after i declared war in the war console they broke again somehow with the same error.

## War console countdown

when you declare war the countdown on the console doesnt start, however the TCs get deposited in the TC withdrawal consoles.

https://user-images.githubusercontent.com/80334192/136318295-f806eaf5-a318-45ea-82a9-481f54c88503.mp4

## Syndicate shuttle

i dont know what the heck is happening here, just respawn as a nukie on staging and try the same things done here.

might be related to #7493 

https://user-images.githubusercontent.com/80334192/136318479-4bb60d64-731e-4ff0-8a19-f0a3e3562cd0.mp4

NOTE2: this was all tested on the staging server squarestation map.

### Steps to reproduce

god pls send help too many bugs

VernaSaurus:
partially closes #7532 and fully closes #6186 
IN DETAIL: declaring war now has the timer actually count down instead of remaining 20:00, and syndicates can fire ID locked weapons without causing an NRE when their ID is in a PDA.

CL: [Fix] Fixed declaring war not counting down
CL: [Fix] Fixed syndies being unable to shoot their guns

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Clearance/ClearanceCheckable.cs
+++ b/UnityProject/Assets/Scripts/Systems/Clearance/ClearanceCheckable.cs
@@ -30,6 +30,12 @@ namespace Systems.Clearance
 			{
 				return true;
 			}
+			// If the player has null or no defined access, access is denied
+			if (requesterClearance == null ||
+			    requesterClearance.Any() == false)
+			{
+				return false;
+			}
 
 			switch (type)
 			{
--- a/UnityProject/Assets/Scripts/Systems/Clearance/ClearanceCheckable.cs
+++ b/UnityProject/Assets/Scripts/Systems/Clearance/ClearanceCheckable.cs
@@ -30,9 +30,8 @@ namespace Systems.Clearance
 			{
 				return true;
 			}
-			// If the player has null or no defined access, access is denied
-			if (requesterClearance == null ||
-			    requesterClearance.Any() == false)
+			// If the player has null access, access is denied
+			if (requesterClearance == null)
 			{
 				return false;
 			}
--- a/UnityProject/Assets/Scripts/Objects/Consoles/SyndicateOpConsole.cs
+++ b/UnityProject/Assets/Scripts/Objects/Consoles/SyndicateOpConsole.cs
@@ -38,6 +38,7 @@ namespace SyndicateOps
 		private void OnDisable()
 		{
 			UpdateManager.Remove(CallbackType.PERIODIC_UPDATE, RewardTelecrystals);
+			UpdateManager.Remove(CallbackType.PERIODIC_UPDATE, CountDown);
 		}
 
 		public void AnnounceWar(string declarationMessage)
@@ -51,6 +52,7 @@ namespace SyndicateOps
 				$"Attention all crew! An open message from the syndicate has been picked up on local radiowaves! Message Reads:\n" +
 				$"{declarationMessage}" ,CentComm.UpdateSound.Alert);
 				UpdateManager.Add(RewardTelecrystals, 60);
+				UpdateManager.Add(CountDown, 1);
 			}
 		}
 
@@ -63,7 +65,14 @@ namespace SyndicateOps
 			{
 				UpdateManager.Remove(CallbackType.PERIODIC_UPDATE, RewardTelecrystals);
 			}
-
+		}
+		public void CountDown()
+		{
+			timer -= 1;
+			if(timer == 0)
+			{
+				UpdateManager.Remove(CallbackType.PERIODIC_UPDATE, CountDown);
+			}
 		}
 	}
 }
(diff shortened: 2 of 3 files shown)


Bod9001 (dismissed):

Aranclanos:
the timer already has a countdown in GUI_SyndicateOpConsole

VernaSaurus:
> the timer already has a countdown in GUI_SyndicateOpConsole

this just... isn't true? I know because I checked everywhere, and there isn't a countdown in that file.

VernaSaurus:
> the timer already has a countdown in GUI_SyndicateOpConsole

I can see why you thought it was. What happens is that there isn't actually a countdown in GUI_SyndicateOpConsole, there's only an update to the timer. The timer only counts down as the timer decreases in SyndicateOpConsole.

UnityProject/Assets/Scripts/Systems/Clearance/ClearanceCheckable.cs
```diff
@@ -30,6 +30,12 @@ public bool HasClearance(IEnumerable<Clearance> requesterClearance)
 			{
 				return true;
 			}
+			// If the player has null or no defined access, access is denied
+			if (requesterClearance == null ||
+			    requesterClearance.Any() == false)
+			{
+				return false;
```
corp-0, line 36: this makes no sense, it just returned true above using exactly the same condition

VernaSaurus: no? this checks if the _requester_ is null or has no clearance, whereas the one before it checks if _the pin_ is null or requires no clearance

corp-0: Right, my bad. Why is the check needed? Was it erroring? (Btw don't resolve conversations, leave them until the reviewer resolve them themselves)

VernaSaurus: the check is needed because for some reason occasionally the requester is null, and there was no null check for it so I added one (yes, it was erroring, preventing you from firing an ID locked gun under certain situations and resulting in an NRE

corp-0: Add a log then, might be worth checking why is the requester sometimes null in the near future. I think the second condition you added to the if isn't needed there if it is just a null check.

VernaSaurus: fair enough, should I remove it?

corp-0: Can you test without it? The system should already account for empty clearance

VernaSaurus: alright, i'll test it in a bit

VernaSaurus: ok I tested it

VernaSaurus: it works without it, so that should be good to go now

Bod9001 (approved):

corp-0 (approved):

PR #7574, opened 2021-10-14
Removes unneeded  Cool down ( fixes plasma cutter )

### LINKED ISSUE #6907: You can't put plasma cutters into backpacks by clicking on backpacks with them.
### Environment
- OS Version: Windows 10

When clicking on a backpack with a plasma cutter, the plasma cutter will not be put inside the backpack. This happens even if the backpack has free slots. Opening the backpack's inventory by clicking on it with an empty hand lets you put the plasma cutter in its slots in the UI, so it's clearly not because the plasma cutter cannot fit.

### Steps to reproduce
1. Obtain a plasma cutter.
2. Try to put the plasma cutter in your backpack by clicking on it with the plasma cutter. Fail.
3. Open the backpack's inventory with an empty hand, place the plasma cutter in one of its slots.

Bod9001:
why remove cool down?

I could have tried fixing it but the way it is implemented that since it's on the will interact side it will always trigger to a point where it will block random interaction,  the equivalent for server the form interaction is already implemented before the interaction takes place on an object, so I don't see much use for this since intrinsically will just cause issues down the line

closes #6907

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Input System/InteractionV2/Validations.cs
+++ b/UnityProject/Assets/Scripts/Core/Input System/InteractionV2/Validations.cs
@@ -19,11 +19,6 @@ using Objects.Wallmounts;
 /// </summary>
 public static class Validations
 {
-	//Monitors the time between interactions and limits it by the min cool down time
-	private static Dictionary<GameObject, DateTime> playerCoolDown = new Dictionary<GameObject, DateTime>();
-	private static Dictionary<GameObject, int> playersMaxClick = new Dictionary<GameObject, int>();
-	private static double minCoolDown = 0.1f;
-	private static int maxClicks = 5;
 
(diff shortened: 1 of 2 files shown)


corp-0:
are you certain this has no repercusions on any other part of the game? What was the reason for us to have a cooldown in interaction validations to begin with?

Bod9001:
looks like  it was used as a client side limit to the number of interactions,

I could code a better solution for, the client actually sending off for interactions and not just checking for interactions

corp-0:
sounds reasonable to move the cooldown where it actually belongs, since it is a validation, instead of getting rid of it

Bod9001:
ok, will work on tomorrow

corp-0 (approved):

NoooneyDude (approved):

Issue #5276, opened 2020-10-10
Friendly mobs (Ian, Runtime, most pets) should stop trying to move when you pull/grab them.


PetMudstone:
## Description
Title. In /tg/, friendly mobs such as Ian, Runtime, and most pets (parrots and monkeys are excluded since they're jerks) don't move when you pull/grab them. This is so you can easily move them to elsewhere without the threat of them breaking free of your grasp and running off. It would be helpful if friendly mobs didn't break free of your grasp in Unitystation as well.

Mute-Man:
Making a note that this definitely applies to Simplebots, such as the Floorbot, Medibot, and Cleanbot.

PR #7585, opened 2021-10-16
Closet refactor

### LINKED ISSUE #2870: Refactor DNAscanner
## Description
DNAscanner component is a bit of a mess. In particular I'd like to remove the need for ClosetStatus.ClosedWithOccupant and the bespoke logic in ClosetHandler that's added just to support DNA scanner. Remove that enum value and move occupancy / non-occupancy status to a bool syncvar in DNAscanner.

Once that's removed, also modify the ServerSetIsClosed method so it calls ServerHandleContentsOnStatusChange AFTER updating the syncvar, and remove the "force" parameter in ServerAddInternalItemInternal as it was added only to work around this issue

Also consider whether something other than inheritance can be used for DNAscanner, as inheritance just makes DNAscanner more hard to make sense of.

### LINKED ISSUE #3408: Red locker light init can be erratic
See how red light is only present on one locker. Others have no red lights, but do show green if toggled
img

### LINKED ISSUE #5367: Emagged Lockers only showing serverside
## Description
Emagging lockers does not show correctly for clients
### Detailed Bug Report
Server
Server host
Client
Client

### Steps to Reproduce

1.  Open 2 instances of the game
2.  Host a server on one and join on the other
3.  Spawn an Emag and Emag a locker from either game
4. View black dot on server, and green on client

NoooneyDude:
- Adds a new component `ObjectContainer`. Intended to handle object storage for things like closets. The component is still fairly barebones. Only closets (and similar) have been updated with the new component for now.
- New interface `IEscapable`. Intended to allow mobs (inc. players) to escape from containers. Again, only for closets for now.
- Refactors closets to rely on `ObjectContainer`. Fixes #3408 (if it was still an issue), fixes #5367.
- Untangles DNA scanners from closets. Now has a much more healthier relationship. Closes #2870.
- CL: [Improvement] Mobs can now be trapped in closets and will attempt to escape. Alleviates #5276.
mouse trap
- CL: [Improvement] Lockers and DNA scanners now have dynamic examination text.

> Unable to actually test cloning as I can no longer connect multiple clients locally.


Bod9001 (approved):

corp-0:
Brilliant work! I need to test cloning before approving otherwise the code looks very good.

PerfectTangent:
perhaps we could get an option to configure which mobs can break out of closets and which can't? I see applications with pet cages.

corp-0 (approved):

PR #7621, opened 2021-10-19
CNT optimisation and other optimisations

Bod9001:
CL: [Fix] fixed a bug with space walking

a good portion of focus on optimising CNT 

reuses matrix gets so there is multiple in CNT

### DIFF
--- a/UnityProject/Assets/Scripts/Player/PlayerSync.cs
+++ b/UnityProject/Assets/Scripts/Player/PlayerSync.cs
@@ -496,10 +496,8 @@ public partial class PlayerSync : NetworkBehaviour, IPushable, IPlayerControllab
 			{
 				CheckMovementServer();
 			}
-			else
-			{
-				CheckMovementClient();
-			}
+
+			CheckMovementClient(); //needed for space floating movement to update for the Server if it is not headless
 
 			if (!ClientPositionReady)
 			{
--- a/UnityProject/Assets/Scripts/Player/PlayerSync.cs
+++ b/UnityProject/Assets/Scripts/Player/PlayerSync.cs
@@ -496,8 +496,10 @@ public partial class PlayerSync : NetworkBehaviour, IPushable, IPlayerControllab
 			{
 				CheckMovementServer();
 			}
-
-			CheckMovementClient(); //needed for space floating movement to update for the Server if it is not headless
+			else if (CustomNetworkManager.IsHeadless == false)
+			{
+				CheckMovementClient(); //needed for space floating movement to update for the Server if it is not headless
+			}
 
 			if (!ClientPositionReady)
 			{
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Matrix.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Matrix.cs
@@ -311,7 +311,7 @@ public class Matrix : MonoBehaviour
 		return (isServer ? ServerObjects : ClientObjects).Get(localPosition);
 	}
 
-	public IEnumerable<T> GetAs<T>(Vector3Int localPosition, bool isServer) where T : class
+	public IEnumerable<T> GetAs<T>(Vector3Int localPosition, bool isServer) where T : RegisterTile
 	{
 		if (!(isServer ? ServerObjects : ClientObjects).HasObjects(localPosition))
 		{
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Matrix.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Matrix.cs
@@ -212,29 +212,6 @@ public class Matrix : MonoBehaviour
 	}
 
 
-	public bool IsPassableLeaveTileCardinalMatrix(Vector3Int Localorigin, Vector3Int Localposition, bool isServer,
-		CollisionType collisionType = CollisionType.Player, bool includingPlayers = true, GameObject context = null,
-		List<LayerType> excludeLayers = null, List<TileType> excludeTiles = null, bool ignoreObjects = false,
-		bool isReach = false, bool onlyExcludeLayerOnDestination = false)
-	{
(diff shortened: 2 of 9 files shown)


Bod9001:
needs more work

corp-0:
> needs more work

does this mean it isn't ready for review?

Bod9001:
yeah

Bod9001:
good to go

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Matrix.cs
```diff
@@ -311,6 +311,27 @@ public IEnumerable<RegisterTile> GetRegisterTile(Vector3Int localPosition, bool
...
+		if (!(isServer ? ServerObjects : ClientObjects).HasObjects(localPosition))
+		{
+			return Enumerable.Empty<T>(); //?
+		}
+
+		var filtered = new List<T>();
+		foreach (RegisterTile t in (isServer ? ServerObjects : ClientObjects).Get(localPosition))
+		{
+			T x = t as T;
+			if (x != null)
```
ThatDan123, line 326:
```suggestion
			if (t is T x)
```
Does that work and avoid the null?

UnityProject/Assets/Scripts/Player/PlayerSync.cs
```diff
@@ -496,10 +496,8 @@ private void Synchronize()
 			{
 				CheckMovementServer();
 			}
-			else
-			{
-				CheckMovementClient();
-			}
+
+			CheckMovementClient(); //needed for space floating movement to update for the Server if it is not headless
```
ThatDan123:
```suggestion

                        else if (CustomNetworkManager.IsHeadless == false)
			{
			    CheckMovementClient(); //needed for space floating movement to update for the Server if it is not headless
			}
```

Think thats the headless check, or is this needed for headless too bit confused by what you mean by movement

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Matrix.cs
Bod9001: I don't know if it technically casts twice there though that's the thing?

UnityProject/Assets/Scripts/Player/PlayerSync.cs
Bod9001: shall I run some headless tests?

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Matrix.cs
```diff
@@ -311,6 +311,27 @@ public IEnumerable<RegisterTile> GetRegisterTile(Vector3Int localPosition, bool
 		return (isServer ? ServerObjects : ClientObjects).Get(localPosition);
 	}
 
+	public IEnumerable<T> GetAs<T>(Vector3Int localPosition, bool isServer) where T : class
```
ThatDan123, line 315: Could you also comment this to say GetAs is only to be used if T is a class which inherit from RegisterTile

ThatDan123:
```suggestion
	public IEnumerable<T> GetAs<T>(Vector3Int localPosition, bool isServer) where T : RegisterTile
```

Actually would this work instead, but add a comment anyway?

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Matrix.cs
ThatDan123: Dont think it would

UnityProject/Assets/Scripts/Player/PlayerSync.cs
ThatDan123: yeah, would be for the best

ThatDan123:
Please profile for headless too, be good to check for the improvements

UnityProject/Assets/Scripts/Player/PlayerSync.cs
Bod9001: done

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Matrix.cs
Bod9001: done

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Matrix.cs
Bod9001: odd IDE isn't recommending the change anymore

Bod9001:
"Please profile for headless too, be good to check for the improvements"
it's hard to check since CNT is also mobs so, they be do different stuff on each profile

ThatDan123:
> "Please profile for headless too, be good to check for the improvements" it's hard to check since CNT is also mobs so, they be do different stuff on each profile

If you limit the build to just one map scene(remember to update the map SO's) then it should be consistent?

Bod9001:
I'll try that

corp-0:
seems like it isn't compiling

Bod9001:
even with the same scene it's still different the number of CNT 
7 vs 4 

0.16ms without optimisation for 4 = 0.04‬
0.25~ms Optimised for 7 =0.035

Bod9001:
oh yeah..
the optimisation is for items that are moving humm, didn't test that

ThatDan123 (approved):

corp-0:
If it has to inherit from RegisterTile wouldn't it be where T: RegisterTile ?

Bod9001:
I had to revert since it complained there was no cast between T to register tile

corp-0:
Was it just a warning or an actual error?

Bod9001:
actual error

corp-0 (approved):

PR #7631, opened 2021-10-21
Clients will no longer use the subsystem manager

Aranclanos:
### Purpose
Clients will no longer use the subsystem manager to have underfloor interactables active.
Adds a TODO with a tiny description of how expensive and unnecesary InitialiseUnderFloorUtilities() is for a client

Basically nothing changes, instead of using the subsystem manager to call InitialiseUnderFloorUtilities() and slap a bunch of if(isserver) just call it from here

Stole the idea from bod's pr
I believe this way its easier to read and understand, I'm already doubting of why we even have a subsystem manager that only sets up the pipes and environmental atmos managers

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/AtmosSystem.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/AtmosSystem.cs
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Generic;
 using ScriptableObjects.Atmospherics;
 using UnityEngine;
+using Mirror;
 
 namespace Systems.Atmospherics
 {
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/AtmosSystem.cs
```diff
@@ -13,11 +14,9 @@ public class AtmosSystem : SubsystemBehaviour
 
 		private Dictionary<int, RoomGasSetter> toSet = new Dictionary<int, RoomGasSetter>();
 
+		[Server]
```
Bod9001, line 17: I think this doesn't actually block it just posts a warning if called on client?

Bod9001 (approved):

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/AtmosSystem.cs
Aranclanos: it both stops it and sends a warning

ThatDan123: If you dont want the warning you can use [ServerCallback]

ThatDan123 (approved):

Issue #7638, opened 2021-10-22
Refactor Old Managers to use SingletonManager


ThatDan123:
https://github.com/unitystation/unitystation/blob/develop/UnityProject/Assets/Scripts/Managers/SingletonManager.cs

-Cant be done with networked managers

PR #7640, opened 2021-10-23
Update disposals to ObjectContainer

### LINKED ISSUE #4654: Disposals Issue and features B:150

NoooneyDude:
- Updates disposals to use `ObjectContainer`.
May or may not fix #4654 part 4.
Likely fixes part 5 of #4654.
Fixes part of #4654, though there seems to be an issue where the dropped item is ejected invisible.
- CL: [Fix] Fixes air ejection hiss sound not playing when a disposal traversal instance reaches an open-ended disposal pipe.
- Converts `DisposalsManager` to `SingletonManager` as per #7638.

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Disposals/DisposalVirtualContainer.cs
+++ b/UnityProject/Assets/Scripts/Objects/Disposals/DisposalVirtualContainer.cs
@@ -61,9 +61,6 @@ namespace Objects.Disposals
 
 			foreach (var obj in objects)
 			{
-				// may have despawned while in storage (e.g. grenade)
-				if (obj == null) continue;
-
 				if (obj.TryGetComponent<IPushable>(out var pushable) == false) continue;
 
 				if (obj.TryGetComponent<RegisterObject>(out _) == false && obj.TryGetComponent<CustomNetTransform>(out var cnt))
--- a/UnityProject/Assets/Scripts/Objects/Disposals/DisposalVirtualContainer.cs
+++ b/UnityProject/Assets/Scripts/Objects/Disposals/DisposalVirtualContainer.cs
@@ -1,7 +1,7 @@
 using System.Collections.Generic;
-using AddressableReferences;
-using Messages.Server;
+using System.Linq;
 using UnityEngine;
+using AddressableReferences;
 
 namespace Objects.Disposals
 {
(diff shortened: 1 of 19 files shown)


UnityProject/Assets/Scripts/Objects/Disposals/DisposalVirtualContainer.cs
```diff
@@ -1,287 +1,96 @@
...
+			container.RetrieveObjects();
 
-			containedItems.Clear();
-			containedObjects.Clear();
-			containedPlayers.Clear();
-		}
+			foreach (var obj in objects)
+			{
+				// may have despawned while in storage (e.g. grenade)
+				if (obj == null) continue;
```
NoooneyDude:
```suggestion
```

ThatDan123 (approved):

Bod9001 (approved):

PR #7646, opened 2021-10-23
Fix jump to mob when target is hidden

### LINKED ISSUE #7004: Players who are in lockers or disposals are treated as if they are at 0,0 by the ghost teleport menu.
### Environment
- OS Version: Windows 10

When a player is in a locker or in disposal pipes, then the ghost teleport menu will show them as being at 0,0. Trying to teleport to them will bring you to 0,0 instead of their actual position.
image
image

### Steps to reproduce
1. Get into a locker.
2. Hit F5, type in "suicide". You will die. Move to become a ghost.
3. Find yourself in the ghost teleportation menu. You will be shown as being at 0,0.

NoooneyDude:
- CL: [Fix] Fixes jump to mob teleporting you to space if the target is hidden. Fixes #7004.

### DIFF
--- a/UnityProject/Assets/Scripts/Player/TeleportUtils.cs
+++ b/UnityProject/Assets/Scripts/Player/TeleportUtils.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using Systems.Ai;
@@ -61,10 +61,7 @@ namespace Systems.Teleport
 					status = "(Cant tell if Dead/Alive or Ghost)";
 				}
 
-				//Gets Position of Player
-				var tile = player.gameObject.GetComponent<RegisterTile>();
-
-				var teleportInfo = new TeleportInfo(nameOfObject + "\n" + status, tile.WorldPositionClient, player.gameObject);
+				var teleportInfo = new TeleportInfo(nameOfObject + "\n" + status, player.AssumedWorldPos, player.gameObject);
 
 				yield return teleportInfo;
 			}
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Player/TeleportUtils.cs
```diff
@@ -181,7 +178,9 @@ public static IEnumerable<TeleportInfo> GetCameraTrackPlayerDestinations()
 
 		public static void TeleportLocalGhostTo(TeleportInfo teleportInfo)
 		{
-			var latestPosition = teleportInfo.gameObject.transform.position;
+			var latestPosition = teleportInfo.gameObject.TryGetComponent<ObjectBehaviour>(out var behaviour)
+					? behaviour.AssumedWorldPositionServer()
```
ThatDan123, line 182: This is client side right? So this wont work as there isnt a client version of `behaviour.AssumedWorldPositionServer()`

NoooneyDude: Are you sure this method is client-side?

ThatDan123: line 189 calls TeleportLocalGhostTo which is line 192 and that is a command

NoooneyDude: You might be right as I've only tested locally.

ThatDan123: All the teleport stuff is based on client side knowledge. 

We'd need to refactor the system so that it asks the server for locations to fill the list instead

NoooneyDude: You're right.

ThatDan123: or as a work around, just send the gameobject in the command then the server can just send you to the assumed object pos instead, definitely easier. But the coord shown in the teleport list would be wrong still

Issue #7591, opened 2021-10-16
Matrix rotation is buggered


NoooneyDude:
![buggered](https://i.imgur.com/hBkIzra.png)

AtnerNT:
Is this still an issue this matrix reworK?

MaxIsJoe:
This has been fixed some time ago.

PR #7647, opened 2021-10-23
Fixes clients having lighttubes and other sprites on the wrong direction

### LINKED ISSUE #7406: Camera/light tube orientations wrong/duplicated
### Environment
- Game Version: UnityStationDevelop-21090800

I'm surprised nobody reported this yet (at least i couldnt find it).
image
image
These are all over the place.
Note that light tube sprites get duplicated while camera is rotated without duplicated sprite.
Not all objects are like this, It looks like

Aranclanos:
### Purpose
Organizes the SpriteHandler initialization.
New clients will now request the spritehandler updates after all matrix are set up instead of counting down from 3 seconds.
Fixes #7406
CL: [Fix] fixed clients seeing light tubes and other sprites with wrong rotations.
Tested connecting to staging and following the order of sprites values being set.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Sprite Handler/SpriteHandler.cs
+++ b/UnityProject/Assets/Scripts/Core/Sprite Handler/SpriteHandler.cs
@@ -65,8 +65,6 @@ public class SpriteHandler : MonoBehaviour
 	/// </summary>
 	private bool isPaletteSet = false;
 
-	private bool initialised = false;
-
 	private NetworkIdentity networkIdentity;
 
 	private bool isSubCatalogueChanged = false;
@@ -292,13 +290,8 @@ public class SpriteHandler : MonoBehaviour
 
 	public void SetColor(Color value, bool networked = true)
 	{
-		if (initialised == false) TryInit();
 		if (setColour == value) return;
 		setColour = value;
-		if (HasImageComponent() == false)
-		{
-			GetImageComponent();
-		}
 
 		SetImageColor(value);
 		if (networked)
@@ -340,7 +333,6 @@ public class SpriteHandler : MonoBehaviour
 
 	public void PushClear(bool networked = true)
 	{
-		if (initialised == false) TryInit();
 		if (HasSpriteInImageComponent() == false) return;
 
 		SetImageSprite(null);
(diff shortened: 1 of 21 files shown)


PerfectTangent:
Screenshot 2021-10-23 190227

nice.

corp-0:
nice indeed

UnityProject/Assets/Scripts/Core/Sprite Handler/SpriteHandler.cs
```diff
@@ -386,7 +378,6 @@ public void SetPaletteOfCurrentSprite(List<Color> newPalette, bool networked = t
 
 	public void PushTexture(bool networked = true)
 	{
-		if (initialised == false) TryInit();
```
Bod9001, line 389: I think the problem is there's a few places Where Sprite hand function is called  in awake in a different script, potentially leading to scenario where It's called before being initialised

UnityProject/Assets/Scripts/Core/Sprite Handler/SpriteHandler.cs
```diff
@@ -585,15 +576,32 @@ void Awake()
 	{
 		if (Application.isPlaying)
 		{
-			TryInit();
-		}
-	}
+			spriteRenderer = GetComponent<SpriteRenderer>();
```
Bod9001, line 579: also awake  is called when object is enabled,  this might cause unintended/unexpected behaviour  so idk

NoooneyDude:
Don't suppose this fixes #7591?

corp-0:
I really really really want to merge this

UnityProject/Assets/Scripts/Core/Sprite Handler/SpriteHandler.cs
NoooneyDude: What's the outcome if this occurs? An error? Or fails silently / unexpected results?

Bod9001: An error /unexpected results
since it was called before it would then override the Sprite with the one default set
 and it possibly could NRE on the spriteRenderer reference

NoooneyDude: Alright. I'd be happy to merge if it threw errors as they can be easily found, but it sounds like it'll create issues that might take a while to notice.

Aranclanos:
updated

Bod9001:
What you think about the awake when being enabled?

PetMudstone:
Tested and works fine as a headless client.

corp-0 (approved):

Bod9001 (approved):

PR #7658, opened 2021-10-25
Fixes clients being stuck on preparing

Aranclanos:
### Purpose
CL: [Fix] Fixes clients being stuck on preparing

### Notes:
If a client disconnected while the scenes were being loaded on the server, the objects would still set the connection ID to its observer list due to this scene observer list (two different lists, in this case im refering to the one im changing in the commit) never clearing on player disconnect.
Afterwards, if a player connected and received the same connectionID, the SubSceneManager networkbehaivour would never start as the spawn message was stopped by the server once it checked that the connectionID is in the observer list of said networkidentity

did I make it sound too confusing?


corp-0:
Beautiful

VernaSaurus:
based pr

~~speedmerge~~

ThatDan123 (approved):

Bod9001 (approved):

PR #7661, opened 2021-10-25
Shorter traitor uplink code

### LINKED ISSUE #7660: Shorter traitor uplink codes or why sometimes less is better
Right now when you spawn as a traitor you receive your code for the uplink as a series of 3 words and 3 numbers in succession leading to something like this:

- Zulu-X-Ray-Charlie-123

at first glance it might work good enough for anyone to type on the PDA but not only you have to type it exactly right but its just too long to keep read of it and remember mid game, i have received multiple miss reports of the traitor up link code not working for them due to them having forgotten a word or by pure chance miss-typed it on their notes.

also for it being soo long its a lot harder to:

1. share it with another traitor if you gave him your PDA
2. remember it as a changeling if you absorbed a traitor
3. its takes to long to share it mid round

my suggestion is to make the code shorter, something like

- Alpha-123

not only is good enough to remember but like who is even gonna try to crack by chance 999 combinations of codes not even counting the different words?

some problems could come with a shorter code but:

1. if the words are too few it can also be good to increase the pool of words from where to choose
1. maybe add an extra number so its really hard to guess the code

Daxam12:
### Purpose

Title
Suggested by #7660 

CL: [Improvement] traitor uplink code shortened to one word+###


Bod9001 (dismissed):

corp-0:
if it is only one word + a number, the for isn't needed anymore

corp-0 (approved):

NoooneyDude (approved):
All good, though you might be interested in our sweet `PickRandom()` extension, so e.g. `return codeList.PickRandom() + nums`.

NoooneyDude:
What's your Discord name?

PR #7664, opened 2021-10-26
Fix secured objects considered as cargo, entering closets with magboots

### LINKED ISSUE #7617: players with active magboots cant go inside lockers
### Environment
- Game Version: Staging 21101813
- OS Version: Arch Linux

if you have mag boots active your character doesn't go inside lockers

https://user-images.githubusercontent.com/80334192/137836660-0a09675f-bbf8-45fb-8346-ccbd122736a6.mp4

### Steps to reproduce

1. get the drip
2. activate drip
3. your drip is too strong you cannot go inside lockers now

NoooneyDude:
CL: [Fix] Secured objects, such as conveyors, are no longer considered cargo and so won't be sold.
CL: [Fix] Fixes being unable to enter closets with magboots enabled (closes #7617).

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Attributes.cs
+++ b/UnityProject/Assets/Scripts/Items/Attributes.cs
@@ -3,7 +3,7 @@ using UnityEngine;
 using Mirror;
 using Core.Editor.Attributes;
 using Messages.Client.Interaction;
-
+using NaughtyAttributes;
 
 [RequireComponent(typeof(Integrity))]
 [RequireComponent(typeof(CustomNetTransform))]
@@ -34,7 +34,7 @@ public class Attributes : NetworkBehaviour, IRightClickable, IExaminable
 	private bool willHighlight = true;
 
 	[Tooltip("How much does one of these sell for when shipped on the cargo shuttle?")]
-	[SerializeField, PrefabModeOnly]
+	[SerializeField, BoxGroup("Cargo"), PrefabModeOnly]
 	private int exportCost = 0;
 
 	public int ExportCost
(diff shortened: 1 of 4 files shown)


PerfectTangent:
so, I was talking with @corp-0 about the cargo selling thing. checking if an object is secured solves a lot of the existing problems with cargo, but it may be desirable to make whether something is sold a toggleable attribute so we have maximum control over what is and isn't sold by the shuttle.

NoooneyDude:
> so, I was talking with @corp-0 about the cargo selling thing. checking if an object i secured solves a lot of the existing problems with cargo, but it may be desireable to make whether something is sold a toggleable attribute so we have maximum control over what is and isn't sold by the shuttle.

Can you think of something that'll use it - perhaps in the future - ? Else I'd rather just add that function when required, if I'm honest.

PerfectTangent:
well, I don't think it's optimal to sell just about anything that doesn't have an export value.

corp-0:
This won't sell anything that's secured to the floor, regardless of export value

NoooneyDude:
> This won't sell anything that's secured to the floor, regardless of export value

That's right. Do you think that is an issue?

corp-0:
> > This won't sell anything that's secured to the floor, regardless of export value
> 
> That's right. Do you think that is an issue?

No, I think I misread Jack. Will you change this to don't sell stuff without selling value?

Bod9001 (dismissed):

NoooneyDude:
Cargo code is nightmare fuel.

Bod9001 (approved):

ThatDan123:
Codacy wants you to use a string builder

corp-0:
in codacy we trust

NoooneyDude:
I think refactoring it to use SB would require touching more files. Can we let it slide? This PR was originally small ; )

Bod9001 (approved):

Bod9001:
Didn't see who the original approvals was and wondered why It didn't work xD

ThatDan123 (approved):

PR #7666, opened 2021-10-27
Fixes AI camera tracking & door electrocution

Aranclanos:
### Purpose
Fixes #7494
CL: [Fix] Fixes AIs not being able to track living forms on clients.
Removes the disabling of Collider2D on clients.
CL: [Fix] Fixes electrocution toggle on doors not working.

PR #7677, opened 2021-10-28
fixes blob

Bod9001:
reasons why blob was broken
A it was set to non where it would not set the randomized zone properly so doing no damage
B it would spawn extra blob structures underneath the parent for some reason not the object layer
CL: [Fix] fixed blob spawning blob structures improperly.
CL: [Fix] fixed blob randomized damage actually causing no damage.
also tightened up blob spawning, meaning it has to be on station to be valid

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Antagonists/Antags/Blob/BlobStarter.cs
+++ b/UnityProject/Assets/Scripts/Systems/Antagonists/Antags/Blob/BlobStarter.cs
@@ -270,7 +270,7 @@ namespace Blob
 				gameObject.GetComponent<PlayerSync>().SetPosition(position, true);
 			}
 
-			var spawnResult = Spawn.ServerPrefab(AntagManager.Instance.blobPlayerViewer, gameObject.GetComponent<PlayerSync>().ServerPosition, gameObject.transform.parent);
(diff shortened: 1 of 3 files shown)


ThatDan123 (dismissed):

UnityProject/Assets/Scripts/Systems/Antagonists/Antags/Blob/BlobStarter.cs
```diff
@@ -254,9 +254,12 @@ private void FormBlob()
 
 			var bound = MatrixManager.MainStationMatrix.Bounds;
 
+			//To ensure that it has to be station matrix
+			var on = MatrixManager.AtPoint(gameObject.AssumedWorldPosServer().RoundToInt(), true);
```
Aranclanos: use the registertile matrix instead

Bod9001:
Bullets still travel through on client but, removing client prediction on bullets is a thing to do

ThatDan123 (approved):

ThatDan123:
> Bullets still travel through on client but, removing client prediction on bullets is a thing to do

Shouldve been fixed by #7666

corp-0 (approved):

Issue #7261, opened 2021-08-13
Station bounced radios B: 75


PerfectTangent:
### Description

This bounty calls for the implementation of station bounced radios. Station bounced radio will function similarly to a standard radio headset if you put it in your pocket, belt-slot or in your hand.

Radios differ from headsets in a few key ways:

- Can have their frequency tuned, this changes which channel is used for sending and receiving messages by default.
  - Radios can be tuned to a minimum of 120.1 KHz and a maximum of 159.9 KHz. The default radios have a constricted range of a minimum of 144.1 KHz and to a maximum of 148.9 KHz, meaning these radios can only be tuned to the common and AI channels. 
- radios cannot have their encryption keys removed. (why, though? perhaps we want to allow it.)
- in addition to talking into the radio while equipped in a fashion like headsets, radios can also be switched to broadcast whatever they hear so long as they are in broadcast mode. Station bounced radios will pick up anything said in this mode within a 3 tile range.
- Pressing Z with a radio selected will bring up a UI that allows you to modify all the variables mentioned above.
- since unitystation has toggleable radio channel buttons and characters can have multiple radios available to use, each radio should get it's own channel buttons when equipped.
- radios can be combined with remote signallers to allow the signaller to remotely trigger or disable the broadcasting mode.

## What is needed to fulfill the bounty:

- [ ] add station bounced radios.

MaxIsJoe:
if #7678 gets merged I'd like to work on this

PR #7678, opened 2021-10-29
Signals - The Start - Part one

MaxIsJoe:
Adds a new signal manager which will help us easily add things like telecom servers, remote bomb detonators, station bounced radios, etc.

Most code is documented or self-documented and right now there is a proof of concept item that only admins can spawn that lets them test this with door buttons. This item is like an item that exists on /tg/ but is incredibly debuggy and uses place holder graphics + can't be crafted.

You only need to extend the SignalEmitter and SignalReceiver components with your own logic and it will work without a hassle. 
SignalReceived, Respond and SignalFailed are all the functions you need to go crazy with whatever you want to do.

Lays the groundwork for #7261

### NOTES 

- The debug receiver item can't be paired while it's in your hand for some reason, just drop it on the ground then pair it with the emitter.
- This is still the beginning, I plan on moving all items and radio chat to this new system later in the next PRs.
- This system is designed to be "hassle free" and easier to understand as in it acts as a high level API for objects that need to communicate between each other and any item or system should be able to adopt it.


NoooneyDude:
We have a radio manager. What are you going to do with this existing system?

UnityProject/Assets/Scripts/Managers/SignalsManager.cs
```diff
@@ -0,0 +1,143 @@
...
+				if (receiver.SignalTypeToReceive == SignalType.BOUNCED && AreOnTheSameFrequancy(receiver, emitter))
+				{
+					SignalStrengthHandler(receiver, emitter, signalDataSo);
+				}
+			}
+		}
+
+		private bool AreOnTheSameFrequancy(SignalReceiver receiver , SignalEmitter emitter)
+		{
+			return receiver.Frequency == emitter.Frequancy;
```
Lizurt: I don't recommend you to use `==` with floats, it's not reliable

Bod9001: it works okay if you implicitly set it but if it's something that's calculated then it's dodgy

MaxIsJoe:
> We have a radio manager. What are you going to do with this existing system?

Slowly start Deprecating the old one and move existing items to use the new signal manager, this system is meant to offer an easier way to implement items that communicate between each other especially for telecommunications.
I plan on adding servers and relays, signal jamming, station bounced radios, encryption to salt messages in private radios, and most importantly just a simple system that just works and acts as a high level API thats easy to understand and use for future features added to the game.

UnityProject/Assets/Scripts/Managers/SignalsManager.cs
MaxIsJoe: What would be the recommended function to check this? Right now there doesn't seem to be any issues because I'm not fully utilizing frequencies yet but I do see this now as a problem if the client sets the freq to something like 144.2 but the host decides to read it as 144.2000003 or something like that

MaxIsJoe: I say this because we had problems like this before with things like tweens and the food processor

Lizurt: > What would be the recommended function to check this?

IIRC we have a `float.Approx(float)` method.

Anyway, why do we have to work with kilohertz and not just hertz? I mean we could use an int field with values like 144900 instead of 144.9

MaxIsJoe: I haven't seen anything use things other than kilohertz but imo it just looks better when its four to five digits instead of 6

MaxIsJoe: done

UnityProject/Assets/Scripts/Communications/SignalEmitter.cs
```diff
@@ -0,0 +1,31 @@
...
+using ScriptableObjects.Communications;
+using UnityEngine;
+using Mirror;
+
+namespace Communications
+{
+	public class SignalEmitter : NetworkBehaviour
+	{
+		[SerializeField] private SignalDataSO signalData;
+		[SyncVar] public float Frequancy = 122f;
```
ThatDan123:
```suggestion
		[SyncVar] public float Frequency = 122f;
```

ThatDan123: Also make this private and have a public getter

UnityProject/Assets/Scripts/Communications/SignalReceiver.cs
```diff
@@ -0,0 +1,42 @@
...
+using UnityEngine;
+using ScriptableObjects.Communications;
+
+namespace Communications
+{
+	public class SignalReceiver : NetworkBehaviour
+	{
+		public SignalType SignalTypeToReceive = SignalType.PING;
+		[SyncVar] public float Frequency = 122F;
+		[SyncVar] public SignalEmitter Emitter;
```
ThatDan123: Unlikely you'll need to sync this, also wont work. Actually might for networkbehaviours, but it shouldnt be needed to anyway.

UnityProject/Assets/Scripts/Communications/SignalReceiver.cs
```diff
@@ -0,0 +1,42 @@
...
+		public float DelayTime = 3f; //How many seconds of delay before the SignalRecieve logic happens for weak signals
+
+
+		private void Awake()
+		{
+			if(isClient) return;
```
ThatDan123: This doesnt do what you think it does, use `CustomNetworkManager.Instance.IsServer == false`, but that cannot be used in awake, doesnt use awake to add the receiver

UnityProject/Assets/Scripts/Communications/SignalReceiver.cs
```diff
@@ -0,0 +1,42 @@
...
+		private void OnDestroy()
+		{
+			if(isClient) return;
+			SignalsManager.Instance.Recivers.Remove(this);
+		}
+
+		/// <summary>
+		/// Logic to do when
+		/// </summary>
+		public virtual void RecieveSignal(SignalStrength strength) { }
```
ThatDan123:
```suggestion
		public virtual void ReceiveSignal(SignalStrength strength) { }
```

UnityProject/Assets/Scripts/Communications/SignalReceiver.cs
```diff
@@ -0,0 +1,42 @@
...
+		{
+			if(isClient) return;
+			SignalsManager.Instance.Recivers.Add(this);
+		}
+
+		private void OnDestroy()
+		{
+			if(isClient) return;
+			SignalsManager.Instance.Recivers.Remove(this);
+		}
```
ThatDan123: Use OnEnable and OnDisable instead to add/remove stuff

UnityProject/Assets/Scripts/Items/Devices/ButtonSignalReceiver.cs
```diff
@@ -0,0 +1,52 @@
...
+using Mirror;
+using Objects.Wallmounts;
+
+///NOTE : THIS ITEM IS CURRENTLY HERE FOR TESTING REASONS ONLY
+///PLEASE REMOVE THIS (OR POLISH IT AND ADD IT TO THE LIST OF THINGS SCIENCE CAN MAKE) ONCE WE MAKE THE MOVE TO FULLY USE THE SIGNAL MANAGER
+namespace Items
+{
+	public class ButtonSignalReceiver : SignalReceiver, ICheckedInteractable<HandApply>
+	{
+		private DoorSwitch _switch;
```
ThatDan123:
```suggestion
		private DoorSwitch switch;
```

UnityProject/Assets/Scripts/Items/Devices/ButtonSignalReceiver.cs
```diff
@@ -0,0 +1,52 @@
...
+
+		public override void RecieveSignal(SignalStrength strength)
```
ThatDan123:
```suggestion
		public override void ReceiveSignal(SignalStrength strength)
```

UnityProject/Assets/Scripts/Items/Devices/ButtonSignalReceiver.cs
MaxIsJoe: plain `switch` causes compiler errors

UnityProject/Assets/Scripts/Managers/SignalsManager.cs
```diff
@@ -0,0 +1,145 @@
...
+			{
+				if (!signalsManager)
+				{
+					signalsManager = FindObjectOfType<SignalsManager>();
+				}
+
+				return signalsManager;
+			}
+		}
+
```
ThatDan123:
```suggestion
	public class SignalsManager : SingletonManager<SignalsManager>
	{

```
Don't network managers, and use SingletonManager

UnityProject/Assets/Scripts/Managers/SignalsManager.cs
```diff
@@ -0,0 +1,145 @@
...
+		public List<SignalReceiver> Recivers = new List<SignalReceiver>();
```
ThatDan123:
```suggestion
		public List<SignalReceiver> Receivers = new List<SignalReceiver>();
```

UnityProject/Assets/Scripts/Managers/SignalsManager.cs
```diff
@@ -0,0 +1,145 @@
...
+		/// Loops through all receivers and sends the signal if they match the signal type and/or frequancy
+		/// </summary>
+		[Server]
+		public void SendSignal(SignalEmitter emitter, SignalType type, SignalDataSO signalDataSo)
+		{
+			foreach (SignalReceiver receiver in Recivers)
+			{
+				if (receiver.SignalTypeToReceive != type) return;
+
+				if (!receiver.Frequency.IsBetween(signalDataSo.MinMaxFrequancy.x, signalDataSo.MinMaxFrequancy.y)) return;
```
ThatDan123:
```suggestion
				if (receiver.Frequency.IsBetween(signalDataSo.MinMaxFrequancy.x, signalDataSo.MinMaxFrequancy.y) == false) return;
```

UnityProject/Assets/Scripts/Managers/SignalsManager.cs
```diff
@@ -0,0 +1,145 @@
...
+				{
+					StartCoroutine(DelayedSignalRecevie(receiver.DelayTime, receiver, strength));
+				}
+				emitter.SignalFailed();
+			}
+		}
+
+		private IEnumerator DelayedSignalRecevie(float waitTime, SignalReceiver receiver, SignalStrength strength)
+		{
+			yield return new WaitForSeconds(waitTime);
```
ThatDan123:
```suggestion
			yield return new WaitForSeconds(waitTime);
```

Use WaitFor (think thats whats its called) Class instead

UnityProject/Assets/Scripts/Items/Devices/ButtonSignalReceiver.cs
corp-0: yeah, because it is a reserved word. Use ``doorSwitch`` instead.

UnityProject/Assets/Scripts/Communications/SignalReceiver.cs
ThatDan123: Tbh, doubt any of this will need syncvaring, clients will only need it for UI so nettabs will be fine.

MaxIsJoe: will add syncvars later if they seem to be needed.

MaxIsJoe:
Alright, done.

Bod9001 (dismissed):

UnityProject/Assets/Scripts/Communications/SignalEmitter.cs
```diff
@@ -0,0 +1,37 @@
+using System.Collections;
+using System.Collections.Generic;
+using Managers;
+using ScriptableObjects.Communications;
+using UnityEngine;
+using Mirror;
+
+namespace Communications
+{
+	public class SignalEmitter : NetworkBehaviour
```
corp-0: I don't think this class is meant to ever be used directly, right? Make it ``abstract`` so we can make sure it is never instantiated.

UnityProject/Assets/Scripts/Communications/SignalEmitter.cs
```diff
@@ -0,0 +1,37 @@
...
+		public float Frequency
+		{
+			get => frequency;
+			set => frequency = value;
+		}
+
+		/// <summary>
+		/// Tells the SignalManager to send a signal to a receiver
+		/// </summary>
+		public virtual void SendSignal()
```
corp-0: I normally dislike methods were you have to remember to do call ``Base`` before your own logic. Could we instead make the two methods children of this class are supposed to implement ``abstract`` and interface with the manager using another method that will call the abstract method.

UnityProject/Assets/Scripts/ScriptableObjects/SignalDataSO.cs
```diff
@@ -0,0 +1,21 @@
...
+using Managers;
+
+namespace ScriptableObjects.Communications
+{
+	[CreateAssetMenu(fileName = "SignalData", menuName = "ScriptableObjects/SignalData")]
+	public class SignalDataSO : ScriptableObject
+	{
+		[Tooltip("Is this signal global or do we want to check how far it is from a receiver?")]
+		public bool UsesRange = true;
+		[Tooltip("Measured in tiles")]
```
corp-0: you can add here the NaughtyAttributes.ShowIf attribute for more neatness.

MaxIsJoe:
Done.

UnityProject/Assets/Scripts/Communications/SignalEmitter.cs
```diff
@@ -0,0 +1,47 @@
...
+				return;
+			}
+			SignalFailed();
+		}
+
+		/// <summary>
+		/// All the checks and logic that needs to be implemented to send a signal to a receiver.
+		/// if false, a signal won't be sent.
+		/// </summary>
+		protected abstract bool SendSignalLogic();
```
corp-0, line 38: since it now includes the checks and its return value determines if we send the signal or call the failing signal logic, I would rename it to ``TrySendSignal``.

UnityProject/Assets/Scripts/Communications/SignalReceiver.cs
```diff
@@ -0,0 +1,42 @@
...
+		private void OnDisable()
+		{
+			if(CustomNetworkManager.Instance._isServer == false) return;
+			SignalsManager.Instance.Receivers.Remove(this);
+		}
+
+		/// <summary>
+		/// Logic to do when
+		/// </summary>
+		public abstract void ReceiveSignal(SignalStrength strength);
```
corp-0, line 34: maybe this could be called ``OnSignalReceived``

MaxIsJoe: Signal Manager uses ReceiveSignal so calling it "OnSignalReceive" won't make sense.

MaxIsJoe: image

corp-0: it sounds better in my head but I am not making it a demand, just a suggestion. You can resolve this conversation if you don't  feel like changing it.

UnityProject/Assets/Scripts/Items/Devices/ButtonSignalReceiver.cs
```diff
@@ -0,0 +1,52 @@
...
+				Respond(Emitter);
+				return;
+			}
+			Emitter.SignalFailed();
+		}
+
+
+		public override void Respond(SignalEmitter signalEmitter)
+		{
+			Chat.AddLocalMsgToChat("Signal sent!", signalEmitter.gameObject);
```
corp-0:
```suggestion
			Chat.AddLocalMsgToChat("Signal received!", signalEmitter.gameObject);
```

corp-0 (dismissed):

corp-0 (approved):

Bod9001 (approved):

PR #7683, opened 2021-10-30
Add tool to remove missing scripts.

NoooneyDude:
- Add tool to remove missing scripts from prefabs.
- Removes missing scripts from a few objects.


Bod9001 (approved):

Aranclanos:
it could be a test too if you want

ThatDan123:
I swear we have a test for missing scripts, could be broken?

ThatDan123 (approved):

PR #7685, opened 2021-10-30
Matrix at point optimisation

Bod9001:
matrix local bounds are cached and Recalculated if tile is placed outside bounds
matrix get at point is 10* faster,
also improves Custom raycast solution

CL: [Improvement] matrix get at point optimized (90% improvement)

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
@@ -1,7 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using Messages.Client.SpriteMessages;
+using System.Diagnostics;
 using System.Linq;
 using Chemistry;
 using Doors;
@@ -40,6 +40,9 @@ public partial class MatrixManager : MonoBehaviour
 	private static LayerMask tileDmgMask;
 
 	public Dictionary<int, MatrixInfo> ActiveMatrices { get; private set; } = new Dictionary<int, MatrixInfo>();
+
+	public List<MatrixInfo> ActiveMatricesList  { get; private set; } = new List<MatrixInfo>();
+
 	public List<MatrixInfo> MovableMatrices { get; private set; } = new List<MatrixInfo>();
 
 	public static bool IsInitialized;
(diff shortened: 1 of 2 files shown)


Aranclanos:
ActiveMatricesList is not being cleared
also why do we need it if we have the dictionary ActiveMatrices

Aranclanos:
~~is CashedBoundsInt updating when a matrix moves?~~ nvm its local position

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
```diff
@@ -251,6 +260,15 @@ public void Update()
 		}
 
 
+		public static bool IsWithinBounds(BoundsInt boundsInt, Vector3 Point)
+		{
+			if (Point.x > boundsInt.xMin && Point.x < boundsInt.xMax  &&
+			    Point.y > boundsInt.yMin && Point.y < boundsInt.yMax)
+				return true;
+
+			return false;
```
Aranclanos: there's a BoundsInt.Contains(Vector3Int position) you can use instead of creating this

UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
```diff
@@ -97,6 +100,7 @@ void ResetMatrixManager()
 		Instance.mainStationMatrix = null;
 		MovableMatrices.Clear();
 		ActiveMatrices.Clear();
+		ActiveMatricesList.Clear();
```
Bod9001, line 103: ActiveMatricesList  yeah it is
reason why it is so you can do for i Loop that's better before performance

NoooneyDude: There's only half a dozen or so entries in the dictionary right? so I don't think a list would do much. Seems like the difference of iterating dictionary values vs list is quite marginal?

Aranclanos: I dont think ActiveMatrices and ActiveMatricesList should exist at the same time, its the one or the other

Bod9001: though remember that it is called like 50 times Frame roughly

Bod9001: I think it's fine to have both

Aranclanos (approved):

ThatDan123 (approved):

PR #7693, opened 2021-10-31
No more client prediction on guns

Bod9001:
why?
Because of the problem of ping, you could have the client shoot and predict the shot but there's no way to guarantee the shot will be same and server, without making expensive state engine, that stores every single state for the last 2 seconds, then using that to reference when the client shot, 
so, with the current implementation The client could shoot at someone and hit them on their screen but not actually hit them, 
I believe this is a big frustration for players,

so, This changes it so that the server processes all the Relations of hitting and moving, 
the downside to this is that the clients shots will be delayed and will be from a different position that the client shot from
 due to the movement prediction, however the shots will be exactly as it is on server And will hit the exact same targets as on server, making the experience more reliable


Bod9001:
oops accidentally included revert

Bod9001:
It's okay I reverted the revert

Daxam12 (dismissed):

ThatDan123:
Can you check that bullet pooling still works and if the bullets are appearing at the correct place before they become visable.

Bod9001:
I can confirm bullets are pooled on Server at least

Bod9001:
can't fix the momentary appear on clients without reconfiguring client spawn code

Aranclanos:
do players want this?

Bod9001:
well, let's have a poll

PerfectTangent:
I don't think the vast majority of players could accurately weigh the pros and cons of this PR, especially since both ways have negative consequences. This is more up to us to decide whether we think it's best for the game or not.

PerfectTangent:
We may be able to set up a demonstration by simulating an acceptably high amount of ping and have one character strafe and fire at another character and record the results. If it feels gross at a level of ping players will often deal with, then we may not want to do this, but if it feels alright, then we continue with it.

Bod9001:
true

corp-0:
so what's the current state of this pr?

Bod9001:
idk Needs a debate on it

corp-0:
seems like this isn't compiling

Bod9001:
fixed

NoooneyDude (dismissed):

NoooneyDude:
ah, tests

NoooneyDude:
tests are still in pain

Bod9001:
ok

corp-0:
I want to merge this for next playtest, can you take a look at codacy's suggestions?

Bod9001:
image
image
Codacy  being dumb again

Bod9001:
yeah, All the code that Codacy  is talking about has been removed

corp-0:
Let me rerun in it in that case

corp-0:
Suggestions changed

Bod9001:
image
The last one is this since its direct Assigning should be fine

corp-0 (approved):

PR #7700, opened 2021-11-01
Fixes built objects spawn position

Aranclanos:
### Purpose
CL: [Fix] Fixes build girders and other constructable objects spawning incorrectly if two different players were building at the same time


Bod9001 (approved):

ThatDan123:
Which line fixes it?

EDIT: Just for reference the issue is the class var SentByPlayer is being changed by the next message, so having a local var fixes it.

Aranclanos:
no longer using SentByPlayer after the wait time is over

ThatDan123 (approved):

PR #7709, opened 2021-11-02
Increases performance in playersync

Aranclanos:
### Purpose
Increases playersync idle performance by not searching the matrix where the human is and using the one in registerTile when doing gravity checks

image
with 50 humans: before and after

CL: [Fix] large performance increase for player movement when player is *not* actively moving (92% improvement).


ThatDan123 (approved):

ThatDan123:
And no allocation too, always good

corp-0 (approved):

corp-0:
Poggers

PR #7717, opened 2021-11-02
Generral fixes

Bod9001:
fixes lungs being funny
fixes right click radial menu


UnityProject/Assets/Scripts/Util/ScriptsUtilMonoBehaviourSingleton.cs
```diff
@@ -14,6 +14,11 @@ public virtual void Awake()
 	}
 
 
+	public virtual void Start()
+	{
+		Instance = this as T;
+	}
```
ThatDan123, line 20: isnt it already set at Awake?

Bod9001: yeah, but that reference doesn't stick for some reason and just becomes null

ThatDan123 (approved):

corp-0 (approved):

PR #7718, opened 2021-11-02
 radial fix  ( please read )

Bod9001:
so, on my installation of the repository
some references are Missing for the SVGs for the radial menu, causes errors but you can still use it,
I went in and fix the references manually however, it is only on my branch it seems like so, someone else to have to try it out and see if it causes errors Spam when they open radial menu or not


ThatDan123:
Did you try reimporting the resources (think some rightclick stuff is in there still), prefab and scripts folders?

NoooneyDude:
This fixed it for me too. Thanks.

NoooneyDude (approved):

ThatDan123 (approved):

PR #7725, opened 2021-11-03
Use GasContainer's gasmix for ambient air when inside

NoooneyDude:
CL: [Fix] Fixes the pressure for one atmosphere being slightly off.
CL: [Improvement] Enterable objects now store their own gas mix which the player interacts with. This means you can hide in a morgue to escape a raging fire (though be careful you don't suffocate) or use a welded locker as a ghetto escape pod, provided you can figure out a means to shove off. Fixes issue 13 and feature H of #4654 (though you now take some high-pressure damage, and disposal intakes will be in vacuum).
- Adds a couple component presets for GasContainer.


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterTile.cs
```diff
@@ -746,6 +746,12 @@ public virtual bool IsAtmosPassable(Vector3Int enteringFrom, bool isServer)
 		return true;
 	}
 
+	/// <summary>Fetches the metadata of the tile.</summary>
+	public MetaDataNode GetMetaData(bool createIfNotExists = true)
+	{
+		return Matrix.MetaDataLayer.Get(LocalPosition, createIfNotExists);
+	}
+
```
NoooneyDude:
```suggestion
```

UnityProject/Assets/Scripts/HealthV2/Living/Respiration/RespiratorySystemBase.cs
```diff
@@ -68,31 +62,30 @@ void UpdateMe()
...
+			GasMix ambientGasMix;
+			if (objectBehaviour.parentContainer != null &&
+					objectBehaviour.parentContainer.TryGetComponent<GasContainer>(out var gasContainer))
+			{
+				ambientGasMix = gasContainer.GasMix;
 			}
+			else
+			{
+				Vector3Int position = objectBehaviour.AssumedWorldPositionServer();
+				ambientGasMix = MatrixManager.GetMetaDataAt(position).GasMix;
```
ThatDan123: Can you get it from the register tile matrix of the player, as i think this will loop through all matrixes (expensive?) to get the necessary tile?

Bod9001 (dismissed):

UnityProject/Assets/Scripts/HealthV2/Living/Respiration/RespiratorySystemBase.cs
NoooneyDude: Thanks. Missed this one!

Bod9001 (approved):

ThatDan123 (approved):

PR #7732, opened 2021-11-04
Fix caramel reaction and RPG reloading.

### LINKED ISSUE #7672: you cant reload the nuke ops RPG
### Environment
- Game Version: Staging 21102523
- OS Version: Arch Linux

when you shoot a the syndicate RPG it says that i has 0 ammo left, but when you try to put a new rocket inside it it shows a message that the RPG is already loaded

### Steps to reproduce

1.  spawn/buy RPG
2. shoot RPG
3. cant reload The aR Pee Gee

PetMudstone:
### Purpose
- Gives the caramel reactions minimum temperatures and properly adds them to the default reaction set so they can occur. Now sugar can be properly turned into caramel by heating it over 413.5 K.
- Makes rockets into clips so you can reload the RPGL with them. Fixes #7672

### Changelog:

CL: [Fix] Heating sugar above 413.5 K will now actually create caramel as intended.
CL: [Fix] Rockets can now be used to reload the RPGL.

### DIFF
--- a/UnityProject/Assets/Prefabs/Items/Weapons/Projectiles/AmmoCasings/CaselessAmmo/Ammo84mmRocketPM9HE.prefab
+++ b/UnityProject/Assets/Prefabs/Items/Weapons/Projectiles/AmmoCasings/CaselessAmmo/Ammo84mmRocketPM9HE.prefab
@@ -1,25 +1,5 @@
 %YAML 1.1
 %TAG !u! tag:unity3d.com,2011:
---- !u!114 &-2625148932084804424
-MonoBehaviour:
-  m_ObjectHideFlags: 0
-  m_CorrespondingSourceObject: {fileID: 0}
-  m_PrefabInstance: {fileID: 0}
-  m_PrefabAsset: {fileID: 0}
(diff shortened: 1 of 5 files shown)


PerfectTangent:
Not a fan of the solution for the RPG, if I understand it correctly. Ammunition should be able to be loaded directly into weapons that have internal mags. Not being able to do that is a bug/unimplemented feature, and so the issue should stay open. Do shotgun shells count as magazines or something in the current implementation?

PetMudstone:
> Not a fan of the solution for the RPG, if I understand it correctly. Ammunition should be able to be loaded directly into weapons that have internal mags. Not being able to do that is a bug/unimplemented feature, and so the issue should stay open. Do shotgun shells count as magazines or something in the current implementation?

Shotgun shells also use the "Magazine Behavior" script and are considered clips. This is how they can be loaded into shotguns. I do think this sort of behavior is to put it informally, jank, and would rather see an alternative way to reload guns and magazines with cartridges and other individual pieces of ammo. Regardless, this does solve the RPGL reloading issue for now.

NoooneyDude (approved):

ThatDan123 (approved):

PR #7753, opened 2021-11-06
Use SO manager for SO references

### LINKED ISSUE #5126: ScriptableObject Singletons causing slight frame hitch when it tries to find the instances
## Description
SingletonScriptableObject probably needs to be rid of resources loading code (especially LoadAll) as it actually loads the whole asset database and then searches it. See lines 19 to 31 in SingletonScriptableObject.cs.

Doobly believes its best to make the singletons managers, put them in the managers trees and they then reference the Scriptable object.

NoooneyDude:
Turns out we had a MonoBehaviour stuffed full of unused SO singleton instances.

- Converts the above into a singleton SO manager.
- Modifies SingletonScriptableObject to attempt to find the reference in the SO manager first. Closes #5126.
CL: [Fix] Fixes a few small lag spikes felt by the client right after spawning and reduces total load time by ~0.5 seconds (for me).

> There's not too much point to SingletonScriptableObject now as we can just ask `SOs` for the SO reference (like `SOs.Instance.CommonSounds`, but I didn't want to update all the current uses. There also seems to be two SOs related to atmospherics that are used before the SO manager awakes.


UnityProject/Assets/Scripts/Managers/SOs.cs
```diff
@@ -0,0 +1,90 @@
...
+				{ typeof(PoolConfig), PoolConfig },
+				{ typeof(RaceSOSingleton), RaceSOSingleton },
+				{ typeof(SOAdminJobsList), AdminJobsList },
+				{ typeof(SpeechModManager), SpeechModManager },
+				{ typeof(SpriteCatalogue), SpriteCatalogue },
+				{ typeof(SurgeryProcedureBaseSingleton), SurgeryProcedureBaseSingleton },
+				{ typeof(UIActionSOSingleton), UIActionSOSingleton },
+				{ typeof(UplinkCategoryList), UplinkCategoryList },
+				{ typeof(UplinkPasswordList), UplinkPasswordList },
+			};
```
ThatDan123, line 82:
```suggestion
			typeSOMap = new Dictionary<Type, ScriptableObject>()
			
			BindingFlags bindingFlags = BindingFlags.Public;

			foreach (FieldInfo field in typeof(SOs).GetFields(bindingFlags))
			{
				typeSOMap.Add(field.GetType(), field.GetValue(this));
			}
```

Think this'll work, needs testing though

NoooneyDude: I was trying to avoid reflection. Dunno why I felt compelled as so.

ThatDan123: > I was trying to avoid reflection. Dunno why I felt compelled as so.

Yeah, understandable. Not a big deal to keep it as it currently is, just thought i'd point it out.

NoooneyDude: It does look nicer I suppose. But yeah, can just get the reference directly from the manager in the future, skipping all this. Would be nice to not have to update this manager when adding new SOs.

Bod9001 (approved):

ThatDan123 (approved):

PR #7787, opened 2021-11-10
added coordinates to objects that fail PowerLightsRelatedTests

### LINKED ISSUE #7092: PowerLightsRelatedTests could use cordinates for violating objects.
### Description
image
All the "LightTubeFixture" instances share the same name and are all in a scene with dozens of other instances with the same names. They would be easier to find and identify if they had their transform values appended the messages mentioning them.

SluggyWuggies:
Adds the (x , y) local transform values of lights and switches to the console.
fixes #7092
hoorah?


Bod9001 (approved):
Looks good, Though There are a few formatting  Misalignments you can use auto format to fix for you But you don't need to fix them, just remember For Next time

corp-0 (approved):

PR #7789, opened 2021-11-11
add fermenting

VernaSaurus:
CL: [New] added fermenting of produce. A number of plants can now be fermented into various reagents
CL: [New] fermenting barrels can be made with 8 wood planks
CL: [Improvement] adds a use sound to the grinder so you can actually tell that it's done something and makes it so you can't put a beaker inside if there's already one in it


UnityProject/Assets/Scripts/Objects/Machines/AIOGrinder.cs
```diff
@@ -23,17 +23,20 @@ public class AIOGrinder : NetworkBehaviour
...
 			? itemSlot.ItemObject.GetComponent<ReagentContainer>()
 			: null;
-
+		[SerializeField] private AddressableAudioSource grindSound = null;
 		private ItemSlot itemSlot;
 		private ItemStorage itemStorage;
 		public bool GrindOrJuice => grindOrJuice;
 		private bool grindOrJuice = true;
+		private RegisterTile registerTile;
+		public Vector3Int WorldPosition => registerTile.WorldPosition;
```
NoooneyDude:
```suggestion
		private Vector3Int WorldPosition => registerTile.WorldPosition;
```

UnityProject/Assets/Scripts/Objects/Kitchen/InteractableBarrel.cs
```diff
@@ -0,0 +1,88 @@
...
+	{
+		private FermentingBarrel barrel;
+
+		private void Start()
+		{
+			barrel = GetComponent<FermentingBarrel>();
+		}
+		public string Examine(Vector3 worldPos = default)
+		{
+			return $"It is currently {(barrel.Closed ? "closed, meaning the items inside are fermenting." : "open, meaning you can put things inside.")}";
```
NoooneyDude: What if there're no items inside? Might give a false impression.

VernaSaurus: true
i'll add a message to show if there's something in it or not

VernaSaurus: while i'm at it, should i add the bubbling sound it makes on /tg/ when you ferment something?

UnityProject/Assets/Scripts/Objects/Machines/FermentingBarrel.cs
```diff
@@ -0,0 +1,122 @@
...
+				barrelSpriteHandler.ChangeSprite(1);
+			}
+			closed = !closed;
+		}
+
+		public void Update()
+		{
+			foreach (var slot in itemStorage.GetItemSlots())
+			{
+				if (slot.IsEmpty == true) break;
```
NoooneyDude: Can there only be one slot in this item storage?
```suggestion
				if (slot.IsEmpty == true) continue;
```

UnityProject/Assets/Scripts/Objects/Machines/FermentingBarrel.cs
```diff
@@ -0,0 +1,122 @@
...
+
+				if(closed){
```
NoooneyDude: This can be simplified by removing the closed check, as the update method won't be called while it is open, right?

But more importantly, `GetComponent()` is somewhat expensive, so instead of getting the components we need on the stored items each update, please cache these by adding to a list when the item is added to the barrel. The microwave should be doing this.

I'd also recommend using `TryGetComponent<MyComponent>(out var component)`.

NoooneyDude (changes requested):

UnityProject/Assets/Scripts/Objects/Kitchen/InteractableBarrel.cs
```diff
@@ -0,0 +1,89 @@
...
+	/// Allows Microwave to be interacted with. Player can put food in the microwave to cook it.
+	/// The microwave can be interacted with to, for example, check the remaining time.
+	/// </summary>
+	[RequireComponent(typeof(FermentingBarrel))]
+	public class InteractableBarrel : MonoBehaviour, IExaminable, ICheckedInteractable<HandApply>,
+			IRightClickable, ICheckedInteractable<ContextMenuApply>
+	{
+		private FermentingBarrel barrel;
+
+		private void Start()
```
ThatDan123:
```suggestion
		private void Awake()
```

UnityProject/Assets/Scripts/Objects/Machines/FermentingBarrel.cs
```diff
@@ -0,0 +1,130 @@
...
+using Chemistry;
+using Chemistry.Components;
+
+namespace Objects.Kitchen
+{
+	/// <summary>
+	/// A barrel into which players can insert certain food items.
+	/// Upon being inserted, they will be fermented into certain reagents.
+	/// </summary>
+	public class FermentingBarrel : NetworkBehaviour
```
ThatDan123:
```suggestion
	public class FermentingBarrel : MonoBehaviour
```

Cant see why its needs to be a networkbehaviour

ThatDan123 (changes requested):

ThatDan123 (approved):

NoooneyDude (approved):

PR #7813, opened 2021-11-14
Lockageddon add Locks to atmospherics

Bod9001:
This should prevent enumeration errors, since it is locked for the duration of the enumeration

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Implants/Organs/Lungs.cs
+++ b/UnityProject/Assets/Scripts/Items/Implants/Organs/Lungs.cs
@@ -132,7 +132,10 @@ public class Lungs : BodyPartFunctionality
 		{
 			efficiency = 1;
 		}
-		ReagentMix AvailableBlood = RelatedPart.HealthMaster.CirculatorySystem.UsedBloodPool.Take((RelatedPart.HealthMaster.CirculatorySystem.UsedBloodPool.Total * efficiency) / 2f);
+
+		ReagentMix AvailableBlood =
+			RelatedPart.HealthMaster.CirculatorySystem.UsedBloodPool.Take(
+				(RelatedPart.HealthMaster.CirculatorySystem.UsedBloodPool.Total * efficiency) / 2f);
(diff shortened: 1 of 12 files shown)


ThatDan123:
These errors have only started to happen recently are you sure this is gonna fix it?

Can't we just loop backwards?

Bod9001:
I'm sure this will fix it

Aranclanos:
Have you reproduced the errors before trying to fix them?

Bod9001:
100% thread related

ThatDan123:
might be able to use this instead since we don't care about the list order:

https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentbag-1?view=net-5.0

Though it looks like theres no `.Clear()` method in our unity version( need Unity 2021.2), needs .NET Standard 2.1, so you'll need to manually clear it / make a new instance of it.

[EDIT] Nvm, cant really use ConcurrentBag as you cant do a simple remove

ThatDan123:
Theres also a direct Thread Safe list alternative:

https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.synchronizedcollection-1?redirectedfrom=MSDN&view=dotnet-plat-ext-5.0

But even though its older than ConcurrentBag we dont have access to it. Might be able to add it in an external dll

See https://stackoverflow.com/a/4655236

ThatDan123:
Actually nvm, forgot this all needs to be serializable.

[EDIT] can actually have a serialise list for editor and then the bag as well See #7835

corp-0:
so which one do we want? This or the bag?

Bod9001:
bag Has a tiny bit of inherent gc with removing gases, 
but At some point we need to properly lock gases so, Threads can work properly together

corp-0 (approved):

NoooneyDude (approved):
I'll merge this, because the current issues in develop are pretty bad.

I'd still much prefer we keep the locks internal to the atmos classes and just expose a threadsafe method or property or somesuch.

PR #7816, opened 2021-11-15
Blocks mimes from force opening closed doors

### LINKED ISSUE #7286: Mime Invisible Wall can force open any airlock
### Environment
- Game Version: 4121
- OS Version: MacOS

Mimes are able to force open any airlock without access simply by using Invisible Wall on top of the airlock and then running into it repeatedly. The door will open but the invisible wall blocks your path. However, as soon as the invisible wall despawns there's enough time to pass through the airlock.

### Steps to reproduce

1.  Be Mime
2. Find Airlock you can't open and face it directly
3. Use Invisible Wall ability
4. Repeatedly move into or click on the airlock
5. Watch it magically open
6. Wait for Invisible Wall to time out
7. Gain Access

MaxIsJoe:
fixes #7286

CL: [Fix] fixed an exploit that allowed mimes to force open closed doors with the invisible wall spell.

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Spells/Spell.cs
+++ b/UnityProject/Assets/Scripts/Systems/Spells/Spell.cs
@@ -169,6 +169,7 @@ namespace Systems.Spells
 
 					if (matrixInfo.Matrix.Get<DoorMasterController>(localPos, true).Any(door => door.IsClosed))
 					{
+						//This stops tile based spells from being cast ontop of closed doors
 						Chat.AddExamineMsg(caster.GameObject, "You cannot cast this spell while a door is in the way.");
 						return false;
 					}
--- a/UnityProject/Assets/Scripts/Systems/Spells/Spell.cs
+++ b/UnityProject/Assets/Scripts/Systems/Spells/Spell.cs
@@ -1,6 +1,8 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
+using Doors;
 using UnityEngine;
 using Mirror;
 using ScriptableObjects.Systems.Spells;
@@ -165,6 +167,10 @@ namespace Systems.Spells
 					var matrixInfo = MatrixManager.AtPoint(castPosition, true);
 					var localPos = MatrixManager.WorldToLocalInt(castPosition, matrixInfo);
 
+					if (matrixInfo.Matrix.Get<DoorMasterController>(localPos, true).Any())
+					{
+						return false;
+					}
 					if (matrixInfo.MetaTileMap.HasTile(localPos, tileToSummon.LayerType)
 					&& !SpellData.ReplaceExisting)
 					{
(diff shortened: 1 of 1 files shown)


Bod9001:
Wouldn't this cause issues with stuff like firing a fire bolt at a door?

MaxIsJoe:
> Wouldn't this cause issues with stuff like firing a fire bolt at a door?

From what i can tell via the code; no. 
It should only block spells that involve manipulating tiles.

Bod9001:
surely a better check would be checking if the  position is blocked from interaction?

corp-0:
Yeah, I'm not sure if I like this solution, I find it worse than the issue.

pyrosshade:
This solution would likely have to be undone in the future. If not by US, then by those who intend to branch from us and have spells that work specifically on doors (Knock Knock).

MaxIsJoe:
the solution only affects one spell we have right now as far as im aware and like i said on discord this fix is for a non issue and I don't think we should argue about "whats the correct way to do this" for something that is already an intentional design.

NoooneyDude:
This means a mime that attempts to spawn a wall in the doorway they've just passed through (obvious strategy to escape pursuers) will get confused as the spell didn't work for some reason, or worse.

Nope, you need to figure out _why_ the door is opening. Slap a breakpoint in the door code where it opens and read the stack trace.

corp-0:
> this fix is for a non issue

If you think it is a non issue then don't fix it

MaxIsJoe:
> > this fix is for a non issue
> 
> If you think it is a non issue then don't fix it

Listen Gillies, I've been spending a lot of time trying to fix this issue and pulling my hair out over matrix code for the past few weeks and after asking for help multiple times on discord to fix this issue with failed answers I finally found the "best" way to personally handle this based on my capabilities.
I'm fixing this issue because it's marked as an issue and despite me saying it's not one it was still left as one so I thought my opinion was just *one* voice in a sea of crowd. 

Don't tell me to not fix issues otherwise I would have stopped fixing issues completely.

MaxIsJoe:
review pls

corp-0:
I guess not allowing closed doors is good enough **for now**. No idea if we ever want to make a spell that needs a closed door in the future.

corp-0 (dismissed):

UnityProject/Assets/Scripts/Systems/Spells/Spell.cs
```diff
@@ -165,6 +167,11 @@ IEnumerator DespawnAfterDelay()
 					var matrixInfo = MatrixManager.AtPoint(castPosition, true);
 					var localPos = MatrixManager.WorldToLocalInt(castPosition, matrixInfo);
 
+					if (matrixInfo.Matrix.Get<DoorMasterController>(localPos, true).Any(door => door.IsClosed))
```
Bod9001, line 170: just add big old note next to it so if anyone looking for the code sees it instantly

MaxIsJoe: Forgot to commit it, anyways merge pls.

corp-0 (approved):

NoooneyDude (approved):

PR #7818, opened 2021-11-15
Glass tables break upon climbing on them

### LINKED ISSUE #6699: Climbing on glass tables should cause them to break.
### Description
In /tg/, you can climb on glass tables like any other tables. However, it turns out that glass isn't particularly well suited for holding up spacemen and any glass table that is climbed on will break under the weight of the person that climbs it. This creates a glass shard which will hit the person who broke the table once in a random body part and will also stun the climber for 10 seconds (which honestly seems excessive to me).
image

MaxIsJoe:
closes #6699 (nice nice)

CL: [New] Glass tables have a 90% chance to break upon climbing on them


Bod9001 (approved):

NoooneyDude (approved):

Q4ik:
Ok good stuff. Thanks

PR #7819, opened 2021-11-15
Tile removal code clean-up

Aranclanos:
### Purpose
The removal of tiles code will now be way more linear and cleaner.
Cleans different tile and layer related files from unused functions and variables.
Removes the tile removal queue in SpawnSafeThread, the queue in MetaTileMap will be the onlyone used now.
The MetaTileMap queue will now only trigger if the thread is not the main thread.
Removes the TryDoNotDoneTiles functions from remove tile message and update tile message.
Removes tile message will now be rejected by the server.

### Notes:
I noticed that the medbay floor now takes 4 bombs to open a hole to space, when before it was only 2.


UnityProject/Assets/Scripts/Messages/Server/UpdateTileMessage.cs
```diff
@@ -151,93 +112,4 @@ public static void SendTo(GameObject managerSubject, NetworkConnection recipient
 			return msg;
 		}
 	}
-
-	public static class UpdateTileMessageReaderWriters
```
Bod9001: UpdateTileMessageReaderWriters?

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
```diff
@@ -1835,69 +1685,20 @@ public void ClearPreview()
...
-					// + "\n" + "\n"
-					// + " direction.x = " +  direction.x + " calculationFloat " + calculationFloat
-					// + "\n" + " xSteps " + xSteps + " gridOffsetx " + gridOffsetx  +"  vexinvX " + vexinvX);
-
-
-					// }
-				}
-				else
-				{
-					// float Testy = ((vecHit.y - 0.5f) - Mathf.Floor(vecHit.y));
```
Bod9001, line 1871: this is debug stuff, useful

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
```diff
@@ -1835,69 +1685,20 @@ public void ClearPreview()
...
-						// Debug.DrawLine(wold + (Vector3.up * 4f), wold + (Vector3.down * 4), Color.blue, 30);
-						// }
-						// else
-						// {
-						// Debug.DrawLine(wold + (Vector3.right * 4), wold + (Vector3.left * 4), Color.blue, 30);
-						// }
-
-						// ColorUtility.TryParseHtmlString("#ea9335", out var Orange);
-						// var map = ((Vector3) vec).ToWorld(PresentMatrix);
-						// Debug.DrawLine(map + (Vector3.right * 0.09f), map + (Vector3.left * 0.09f), Orange, 30);
```
Bod9001, line 1909: this too

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
```diff
@@ -1909,17 +1710,12 @@ public void ClearPreview()
...
 							}
+							Vector3 AdjustedNormal = ((Vector3) normal).ToWorld(matrix);
+							AdjustedNormal = AdjustedNormal - (Vector3.zero.ToWorld(matrix));
 
-
-							Vector3 AdjustedNormal = ((Vector3) normal).ToWorld(presentMatrix);
-							AdjustedNormal = AdjustedNormal - (Vector3.zero.ToWorld(presentMatrix));
-
-
-							// Debug.DrawLine(wold, wold + AdjustedNormal, Color.cyan, 30);
```
Bod9001, line 1933: yeah this too

UnityProject/Assets/Scripts/Messages/Server/UpdateTileMessage.cs
Bod9001: the reader writer massively improves the performance of tile change message,  since it blocks the inclusion of colour and matrix4x4 when appropriate

Bod9001:
" The MetaTileMap queue will now only trigger if the thread is not the main thread." 

 the main reason it was done for the main thread to was performance reasons,  
 
 since you could stagger tile changes so the client/Server wouldn't lag out
 
 I think this is important functionality

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
corp-0: leaving commented out code is a bad practice. If we ever need it again, we use git.

NoooneyDude:
you must be proud.

UnityProject/Assets/Scripts/Messages/Server/UpdateTileMessage.cs
Aranclanos: its unused

Bod9001: how so? mirror automatically uses it, because it has the 
public static UpdateTileMessage.NetMessage Deserialize(this NetworkReader reader)
and uses it

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
Bod9001: but am too dumb to use git,  change it to a debug bool?

UnityProject/Assets/Scripts/Messages/Server/UpdateTileMessage.cs
ThatDan123: Yeah, mirror uses code generation to use it so will appear unused.

corp-0:
let's discuss what to do with the comments so we merge this soon

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
WarlockD: Why not a preprocessor, c# supports it.  Like #ifdef TILE_DEBUG #endif

Bod9001:
Having a dubug bool, you pass through seems like a good solution?

corp-0:
preprocessor? Yeah, why not

Bod9001:
cool, the question is should that be done in this PR? , I don't mind making my own PR to add it

corp-0:
if you don't mind then we can merge this as is and let Aran do his stuff

Bod9001:
still have the discussion about 
"" The MetaTileMap queue will now only trigger if the thread is not the main thread."

the main reason it was done for the main thread to was performance reasons,

since you could stagger tile changes so the client/Server wouldn't lag out

I think this is important functionality"

Aranclanos:
which interaction in the game removes so many tiles at the same time that we have to consider queueing it up to the next frame?

ThatDan123:
> which interaction in the game removes so many tiles at the same time that we have to consider queueing it up to the next frame?

The lavaland random generation changes a lot of tiles when it loads, might be a frame issue there

Bod9001:
and explosions, gas spreading to, matrix collisions

Aranclanos:
I just dont think any of those are relevant enough to add a queue, specially when the present queue just makes it wait for the extra frame which in reality won't make a difference for player-experience if the server is struggling or not

Bod9001:
wasn't there a check to make sure it didn't take too long?

			while (stopwatch.ElapsedMilliseconds < TargetMSpreFrame)
			{
			yeah

Bod9001:
what shall we do for the tile stagger? imo I feel like it's good thing to have

Aranclanos:
updated

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
```diff
@@ -204,7 +204,12 @@ public void Update()
 
 		private void ApplyTileChange(TileLocation tileLocation)
 		{
-			if (mainThread.Equals(Thread.CurrentThread))
+			if (QueuedChanges.Contains(tileLocation))
+			{
+				return;
+			}
```
ThatDan123, line 210: What happens if a new change comes in before the old one is processed? Does this just stop the new one?

ThatDan123: actually nvm, the TileLocation would be different right? This would only match if it's exactly the same

Aranclanos: itd be the same, its a weird system

ThatDan123 (approved):

Bod9001 (approved):

PR #7824, opened 2021-11-16
Electrical thread try catch

ThatDan123:
Might as well add it here too

-Also removed some unused atmos stuff

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Electricity/Electrical processes/ElectricalManager.cs
+++ b/UnityProject/Assets/Scripts/Systems/Electricity/Electrical processes/ElectricalManager.cs
@@ -65,7 +65,7 @@ namespace Systems.Electricity
 						}
 						catch (Exception e)
 						{
-							Logger.LogError($"Electrical MainThreadProcess Error! \n{e}", Category.Electrical);
+							Logger.LogError($"Electrical MainThreadProcess Error! \n{e.GetStack()}", Category.Electrical);
 						}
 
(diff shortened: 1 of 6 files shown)


UnityProject/Assets/Scripts/Systems/Electricity/Electrical processes/ElectricalManager.cs
```diff
@@ -59,7 +59,15 @@ private void UpdateMe()
...
 					lock (ElectricalLock)
 					{
-						electricalSync.PowerNetworkUpdate();
+						try
+						{
+							electricalSync.PowerNetworkUpdate();
+						}
+						catch (Exception e)
+						{
+							Logger.LogError($"Electrical MainThreadProcess Error! \n{e}", Category.Electrical);
```
corp-0: include the trace

ThatDan123: apparently you can just put `e` or `e.ToString()` and it'll put the whole message with stacktrace

ThatDan123: looked into the unity code for it and gagged, so ive made an extension instead

corp-0 (approved):

Bod9001:
Might get a bit Spamy with errors but Should catch those Weird instances where it is not initialised properly

Bod9001 (approved):

PR #7826, opened 2021-11-17
Cached world bounds

Aranclanos:
### Purpose
World bounds will now be cached.
Creates a new bounds contains function to resolve it failing when the Z max and min point was equal.
AtPoint will now check for world bounds, avoiding the WorldToLocal call.
CL: [Fix] Fixed matrixes world bounds from being incorrect.

image
performance difference with humans floating in space and in the station

CL: [Fix] world bounds will now be caches. Improves performance for player movement between matrixes by approximately 75%

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Others/NukeDiskScript.cs
+++ b/UnityProject/Assets/Scripts/Items/Others/NukeDiskScript.cs
@@ -41,7 +41,7 @@ namespace Items.Command
 
 		public void OnSpawnServer(SpawnInfo info)
 		{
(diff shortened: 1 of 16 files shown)


corp-0:
beautiful

Aranclanos:
DNM, movable matrixes not updating the bounds on movement

Aranclanos:
Well thats taken care, but shuttle rotations have a possibility of being super wacky

NoooneyDude (approved):

ThatDan123 (approved):

PR #7833, opened 2021-11-17
gibbing  looks better now and Drops contained items on body parts

Bod9001:
pretty much says what it does on the tin, 

CL: [Improvement] gibbing has been visually improved. Organs now scatter from the body and unequip associated clothing.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Transform/CustomNetTransform.Move.cs
+++ b/UnityProject/Assets/Scripts/Core/Transform/CustomNetTransform.Move.cs
@@ -366,7 +366,7 @@ public partial class CustomNetTransform
 		//add player momentum
 		float playerMomentum = 0f;
 		//If throwing nearby, do so at 1/2 speed (looks clunky otherwise)
-		float speedMultiplier = Mathf.Clamp(correctedInfo.WorldTrajectory.magnitude / (throwRange <= 0 ? 1 : throwRange), 0.6f, 1f);
+		float speedMultiplier = Mathf.Clamp(correctedInfo.WorldTrajectory.magnitude / (throwRange <= 0 ? 1 : throwRange), 0.2f, 1f);
 		serverState.Speed = (Random.Range(-0.2f, 0.2f) + throwSpeed + playerMomentum) * speedMultiplier;
 		correctedInfo.InitialSpeed = serverState.Speed;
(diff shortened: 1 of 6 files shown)


UnityProject/Assets/Scripts/Core/Transform/CustomNetTransform.Move.cs
```diff
@@ -366,7 +366,7 @@ public void Throw(ThrowInfo info)
 		//add player momentum
 		float playerMomentum = 0f;
 		//If throwing nearby, do so at 1/2 speed (looks clunky otherwise)
-		float speedMultiplier = Mathf.Clamp(correctedInfo.WorldTrajectory.magnitude / (throwRange <= 0 ? 1 : throwRange), 0.6f, 1f);
+		float speedMultiplier = Mathf.Clamp(correctedInfo.WorldTrajectory.magnitude / (throwRange <= 0 ? 1 : throwRange), 0.2f, 1f);
```
ThatDan123, line 369: how is this change with normal throwing?

Bod9001: humm, only difference I noticed is that stuff like ID cards throw really slowly if you throw them close, but I just imagine that as them sliding along the floor for ages

Bod9001: if I don't have this it snaps beyond the target position too fast, and instantly stops, make it look like it just appears there  , if the target position is close to the origin of the body

corp-0: Should we make throw force a parameter?

pyrosshade: > 
> 
> Should we make throw force a parameter?

A bit odd that we don't already.

Bod9001: I think one of the issues is that it expects a target position, to slow down in such not too sure though the code is a bit hard to read

NoooneyDude (approved):

corp-0 (approved):

PR #7834, opened 2021-11-17
Game runn tests

Bod9001:
so, this runs with the play mode option of the test runner, 
and using a custom-made SO template for running each test, 
currently the implemented actions are 
Waite  for x seconds
SpawnX, spawn a prefab for X number at position
KeyInput input a key into the system, like the player is pressing a key
PrefabAt, tests if the prefab is present at the specified location
RespawnPlayer, response the play was specified occupation

more will be added


pyrosshade:
would we be able to simulate mouse input for menuing, like for testing certain consoles?

Bod9001:
yeah, tho would prefer a more reflection based approached like searching each game object and then picking that game object out and calling the component on it

PR #7835, opened 2021-11-17
Concurrent Bag GasData

ThatDan123:
Testing Concurrent Bag for GasData.

https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentbag-1?view=net-5.0

Alternative to #7813.

Pro's: 
-No need to put locks everywhere

Con's: 
-Can't change the values of a gasmix in the editor inspector during runtime easily 
(can add a function to do so though, just means extra steps)

-Cant use For loops only Foreach

-Concurrent Bag are unordered, each iteration could be a different order
(can't think of a reason that this would affect its current use but is worth noting)


UnityProject/Assets/Scripts/Util/SweetExtensions.cs
```diff
@@ -537,6 +538,33 @@ public static int RoundToLargestInt(this float source)
...
+			if (y.Equals(toRemove)) continue;
+
+			temp.Add(y);
+		}
+
+		foreach (var item in temp)
+		{
+			source.Add(item);
+		}
+	}
```
ThatDan123, line 566: Downside this GC's but is only called when there is no amount of a gas in the data so the entry is removed, so doesnt happen often.

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/Gas.cs
```diff
@@ -36,33 +37,45 @@ public static class Gas
...
-			for (int i = 0; i < GasesArray.Count; i++)
+			foreach (var gasData in Gases)
 			{
-				GasesArray[i].Pool();
+				gasData.Pool();
 			}
 
-			GasesArray.Clear();
+			//TODO Unity 2021.2 has .NET standard 2.1 which has a .Clear() method
+			Gases = new ConcurrentBag<GasValues>();
```
ThatDan123: This also GC's see comment, could make a pool in the mean time? or can do which will remove all entries:

`
while (Gases.IsEmpty == false)
			{
				Gases.TryTake(out _);
			}
`

[EDIT] using ^ now so should be no GC unless it inherently does

UnityProject/Assets/Scripts/Util/SweetExtensions.cs
Bod9001: problem is the transfer method, removes all the gas from a gas mix

ThatDan123: oh yeah that too

ThatDan123: Might be able to make a custom function for transfer so it doesnt need to loop through all the gases and removing them individually

Bod9001: you going to have to use it everywhere though

ThatDan123: > problem is the transfer method, removes all the gas from a gas mix

Which transfer method, i cant really find one which'll do a lot of removing

Bod9001: in CalcMeanGasMix 
meanGasMix.Clear();
and 
	public static GasData CopyTo(this GasData oldData, GasData CopyTo)
	CopyTo.Clear();

Bod9001: just to name a few

ThatDan123: CalcMeanGasMix => doesnt remove the moles so wont trigger the remove all the time (will only trigger if the mean moles gets close to 0)

meanGasMix.Clear(); => doesnt trigger the remove code

public static GasData CopyTo => doesnt trigger the remove code (unless moles is very small but should've already been cleaned up when it was first changed)

CopyTo.Clear(); => doesnt trigger the remove code

The remove code is only triggered here:

image

Which only happens if the moles is tiny (0.0001)

Bod9001: ConcurrentBag surely would have to be cleared if you're clearing the entire mix 
otherwise you would have loads of nitrogen = zero 
oxygen = zero

ThatDan123: If you are clearing the whole mix then this is called:

image

`public static void Remove<T>(this ConcurrentBag<T> source, T toRemove)` is only used to remove specific entries not to clear the whole thing

Bod9001: oh ok, so, it's the Remove x  That generates gc ok

Bod9001:
what you think about that in future we need to actually lock the gasmix while redoing calculation is to it?
so scrubber interacts with node and locks it while it is interacting, then releases when it's finished

Since currently my lock one only handles when you're accessing the gas mix directly

Aranclanos:
instead of using bags, could I recommend ThreadSafeList? its a class I made that pipes use

corp-0:
should we close this PR now that the other one got merged?

Issue #7841, opened 2021-11-18
TODO fill out reagent data on plants


Bod9001:
so, the only issue currently is that, there needs to be a way of populating the initial contents if you spawn for example a banana directly, 
imo it should check the seed and then assume a potency of 40 , then set the appropriate reagent values, 

should move over all the Initial contents to plant data, then when the plant grows crops the reagent container is populated

PetMudstone:
Reagent amounts in /tg/ plant produce seems to follow the formula of:
reagents in produce = "maximum volume of produce" * ("potency"/1000) rounded to the nearest whole number with a minimum count of at least 1reagent each.

PetMudstone:
Whoops this isn't actually related to this issue at all, retracting.

PR #7843, opened 2021-11-18
Fixes radiation hopefully and metabolism

Bod9001:
CL: [New] plants now can produce reagents and put it into the reagent container of the food, #7841 is opened so all content is converted over to plant data instead of initial contents,
hopefully makes the radiation thread more robust,
and balances metabolic reactions, stuff like toxin was damaging the stomach causing it not being able to digest the rest of the toxin

CL: [Fix] fixed a case of radiation thread failing.


UnityProject/Assets/Scripts/HealthV2/Living/MedicalChemistry/BodyHealthEffect.cs
```diff
@@ -56,14 +56,14 @@ public override void PossibleReaction(BodyPart sender, ReagentMix reagentMix, fl
 				{
 					foreach (var Effect in Effects)
 					{
-						sender.TakeDamage(null, Effect.EffectPerOne * 3 * LimitedreactionAmount * -OverdoseDamageMultiplier, Effect.AttackType,
+						sender.TakeDamage(null, Effect.EffectPerOne * 6 * LimitedreactionAmount * -OverdoseDamageMultiplier, Effect.AttackType,
```
ThatDan123: Might be better to make this number a constant?

Bod9001: ok

corp-0: what is that number?

Bod9001: // This balance is about right with 1 u ingested 1 * effect it about does one damage

NoooneyDude (dismissed):

NoooneyDude (approved):

ThatDan123 (approved):

PR #7844, opened 2021-11-18
Reagent code linq removal

Aranclanos:
### Purpose
Removes some linq usage from reaction and metabolism code to increase performance and reduce GC.
Removes some copypasted code with reagent application.

image
the difference with 21 humans heartbeats on this frame, 1.1mb of GC gone and performance increased by 20%

image
this happens twice since thats how hmans work, so, another example with 27 humans but this time the old way was removed entirely. Just to get a rough idea of the difference

CL: [Fix] substantial improvement to how player bodies handle internal reagents. 20% performance gain and it consumes less memory

### DIFF
--- a/UnityProject/Assets/Scripts/ChemistryComponents/ReagentContainer.cs
+++ b/UnityProject/Assets/Scripts/ChemistryComponents/ReagentContainer.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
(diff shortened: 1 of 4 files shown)


corp-0:
Chemistry Reaction Tests are failing, you will have to check if it is because code changes and it is trying to call methods that don't exist anymore of it is not getting the result it expects, then we can say your code changes introduced regression.

NoooneyDude (approved):

Bod9001 (approved):

PR #7849, opened 2021-11-19
Move chat UI to new UI canvas

NoooneyDude:
Part one of an attempt to consolidate the canvases and refactor the `UIManager` prefab. The finished refactor will:
- make scaling and resolution options uniform for all UIs.
- simplify adding more client UI.
- unify the client UI lifecycle.

Final structure will look something like
- `UI.prefab`
 \- holds the main canvas. Every other UI element will be a child so that scaling and resolutions apply to all.
 \- resides in OnlineScene and isn't destroyed on scene change.
     - `ChatUI.prefab` - all UI elements regarding the chat system.
     - `xHud.prefab` - each antagonist / occupation should specify the appropriate prefab in their SO to be instantiated and added (or perhaps just enabled, we'll see) to the UI prefab at runtime.
     - `Windows object` - this will be empty, with any windows lazy-loaded (and then cached) at runtime from window prefabs.
     - Any other main UI elements at this level (like top-right windows)
     - `NetUI container` to hold active, cached networked windows

This PR:
- Moves chat UI into this new UI prefab, separating the UI from system (prefab-wise).
- Defines standard text sizes so that they can be easily modified (player feedback welcome for defaults).
- Somewhat gusses up the chat UI.
- Fixes debug tool to spam chat.
- Reduces the amount of gameobjects per chat entry.
- Namespaces a few classes.


Bod9001:
I think the Airlock construction and deconstruction pr Has added a UI window so, Maybe merge that one first?

Bod9001:
looks good, just got to fix tests

Bod9001:
nah, I'm just being dumb they added a netUI is good to go

NoooneyDude:
> looks good, just got to fix tests

Running tests locally now, but do you know why the artifact isn't generated correctly?

corp-0:
If it only has one text file, that usually means compile errors

NoooneyDude:
Tests are fine locally. I'll rerun them here.

corp-0 (approved):

Bod9001 (approved):

PR #7850, opened 2021-11-19
Refactor all uses of serializable dicts to make use of generics #7111

davodico:
### Purpose
 <br>

Closes #7111, all inheritors usage are replaced with direct usage of SerializableDictionary<TKey, TValue> from SerializableDictionary.cs. Unused classes are removed.
### Notes:

No functionality was added it was just refactor.

PR #7854, opened 2021-11-20
Update ReagentContainerTest.cs

davodico:
Replaced DictionaryReagentFloat with direct usage of SerializableDictionary<Reagent, float>.

### Purpose
 <br>

In order to direct usage of SerializableDictionary<Reagent, float> passing the test it is needed to change it.

### DIFF
--- a/UnityProject/Assets/Tests/Chemistry/ReagentContainerTests.cs
+++ b/UnityProject/Assets/Tests/Chemistry/ReagentContainerTests.cs
@@ -54,28 +54,28 @@ namespace Tests.Chemistry
 				50,
 				new ReagentMix(a, 10),
 				new ReagentMix(b, 10),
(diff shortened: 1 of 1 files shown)


Bod9001:
surely this would better done in #7850?

corp-0:
It is fine, we just need to merge this one first

corp-0 (approved):

corp-0:
Wait, tests for this pr are also failing

davodico:
Yea, the same issue as for another but from another side,  cannot convert parameter DictionaryReagentFloat of DictionaryReagentFloat to SerializableDictionary<Chemistry.Reagent, float>.  What do you think is the best way now? 
Because even if I will add a fixed test to another pull request tests will be still failing.  Not sure how to make it right.

corp-0:
Modify everything in a single PR

PR #7863, opened 2021-11-21
Different null comparison since unity might be funny about it

Bod9001:
yeah, not too sure but I know that nulls are a bit funny in unity

### DIFF
--- a/UnityProject/Assets/Scripts/ScriptableObjects/SingletonScriptableObject.cs
+++ b/UnityProject/Assets/Scripts/ScriptableObjects/SingletonScriptableObject.cs
@@ -18,21 +18,21 @@ namespace ScriptableObjects
 		{
 			get
 			{
-				if (!_instance && SOs.Instance != null)
+				if (_instance == null && (SOs.Instance == null) == false)
 				{
 					_instance = SOs.Instance.GetEntry<T>();
 				}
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/ScriptableObjects/SingletonScriptableObject.cs
```diff
@@ -18,21 +18,21 @@ public static T Instance
 		{
 			get
 			{
-				if (!_instance && SOs.Instance != null)
+				if (_instance == null && (SOs.Instance == null) == false)
 				{
```
corp-0, line 22: != Was fine here, I think. If anything the == false makes it harder to read than !=

Bod9001: is unity okay with that?

corp-0: With != ? I don't see why not, it should be using the same equality comparison

Bod9001: ok cool

ThatDan123 (approved):

corp-0 (approved):

PR #7871, opened 2021-11-24
Fixes office chairs not moving the buckled player properly

### LINKED ISSUE #7694: Remote Controled People Bug
### Environment
- Game Version: Arch Linux 64-bit
- OS Version: Build 21102918

You can essentialy remote control people by strapping them into chairs and pushing that chair, you can keep doing this forever. no matter how far away they appear, you (and they) see them whereever you pushed them, the player cannot be harmed this way and cannot pickup items, but they can shoot from that position. and also their camera is there. when they unbuckle, they teleport to the spot where they REALLY are.

### Steps to reproduce

1. Build an Office Chair
2. Strap someone else in
3. push them

Aranclanos:
### Purpose
CL: [Fix] Fixes office chairs not moving the buckled player properly
Fixes #7694

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Transform/CustomNetTransform.cs
+++ b/UnityProject/Assets/Scripts/Core/Transform/CustomNetTransform.cs
@@ -47,6 +47,7 @@ public partial class CustomNetTransform : NetworkBehaviour, IPushable
 	public bool IsFixedMatrix = false;
 
 	private BuckleInteract buckleInteract = null;
+	private bool hasBuckleInteract;
 
 	/// <summary>
 	/// If it has ItemAttributes, get size from it (default to tiny).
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/Core/Transform/CustomNetTransform.cs
```diff
@@ -803,14 +803,11 @@ public void NotifyPlayer(NetworkConnection playerGameObject)
 	[Server]
 	private void UpdateOccupant()
 	{
-		if (occupiableDirectionalSprite != null && occupiableDirectionalSprite.HasOccupant)
+		if (buckleInteract != null && buckleInteract.OccupantPlayerScript != null)
```
ThatDan123:
```suggestion
		if (hasBuckleInteract && buckleInteract != null && buckleInteract.OccupantPlayerScript != null)
```
New bool `hasBuckleInteract`, set to true in awake if the component was found.

Just a small performance fix as it prevents a null check being constantly run even when theres no buckleInteract

(Could also unnest it just for looks :p )

ThatDan123 (approved):

corp-0 (approved):

PR #7877, opened 2021-11-25
Fixes and improve sounds for food processors

### LINKED ISSUE #5906: Add sound to food processors.

MaxIsJoe:
closes #5906
CL: [Fix] fixes and improves sounds for food processors


corp-0:
I think this is fine but not quite sure, Daniel knows better how to loop sounds. We could use with a method from the sound manager in the future that easily loops sounds so we don't have to repeat all of these steps. I will approve and wait for Daniel to confirm.

corp-0 (dismissed):

UnityProject/Assets/Scripts/Objects/Machines/FoodProcessor.cs
```diff
@@ -253,17 +258,22 @@ private void RpcShake(float duration, float distance, float delayBetweenShakes)
 			shaker.StartShake(duration, distance, delayBetweenShakes);
 		}
 
-		private void OnSyncPlayAudioLoop(bool oldState, bool newState)
+		private async void HandleAudio(float duration)
 		{
-			if (newState)
+			if(CustomNetworkManager.IsServer == false) return;
```
ThatDan123:
```suggestion
```

Not really need as it is already checked at line 235

ThatDan123 (approved):

corp-0 (approved):

PR #7883, opened 2021-11-26
why do I have to use this method ( test do not merge )

Bod9001:
Will this work

### DIFF
--- a/.github/workflows/codacy-analysis.yml
+++ b/.github/workflows/codacy-analysis.yml
@@ -1,0 +1,19 @@
+
+name: Codacy Analysis CLI
+
+on: ["push"]
+
+jobs:
+  codacy-analysis-cli:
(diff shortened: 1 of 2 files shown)


corp-0:
the workflows that PR runs are the ones in our repository so this won't ever trigger unless merged. Let's wait for the support team to answer, we shouldn't need to make a workflow for codacy.

corp-0:
well never mind I see the thing running in the list of checks in progress

Bod9001:
apparently didn't find anything wrong with it xD 
humm I wait for support

corp-0:
probably because of this ``max-allowed-issues: 2147483647``

corp-0:
but still, I would rather have the codacy.app panel and all

Bod9001:
yeah, apparently supposed to report back but doesn't seem to be working

PR #7910, opened 2021-11-29
Infinite loop tracker

Aranclanos:
### Purpose
Adds an infinite loop tracker which will log frames that exceed one minute.
The UpdateManager will now save the last invoked action in case it can be used as a checkpoint for the tracker.

The tracker is a manager who will run a thread checking for the main thread frame duration.


ThatDan123:
Thats pretty cool

ThatDan123 (approved):

Bod9001 (approved):

Bod9001:
what you think about the codacy analysis?

PR #7915, opened 2021-11-29
delete restyled

corp-0:
fucker never obeyed our style


Bod9001 (approved):
Press F to pay respects

ThatDan123 (approved):

PR #7918, opened 2021-11-30
Raycast distance check

Bod9001:
there is some stuff that abuses the Raycast  causing mega lag, 
this puts a hard limit on it, so it can't be abused and logs in error,
found out that sounds also have a daggy range check, so optimiseed that as well

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
@@ -340,9 +340,9 @@ public partial class MatrixManager : SingletonManager<MatrixManager>
 			WorldTo = Worldorigin + (Vector3) (direction.normalized * distance);
 		}
 
-		if ((WorldTo - Worldorigin).Value.magnitude > 35)
+		if ((WorldTo - Worldorigin).Value.magnitude > 25)
 		{
-			Logger.LogError(" Limit exceeded on raycast, Look at stack trace for What caused it ");
+			Logger.LogError(" Limit exceeded on raycast, Look at stack trace for What caused it "); //Meant to catch up stuff that's been naughty and doing stuff like 900 tile Ray casts
 			return new CustomPhysicsHit();
 		}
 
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
```diff
@@ -340,6 +340,13 @@ private static bool IsInMatrix(Vector3Int worldPos, bool isServer, MatrixInfo ma
 			WorldTo = Worldorigin + (Vector3) (direction.normalized * distance);
 		}
 
+		if ((WorldTo - Worldorigin).Value.magnitude > 25)
+		{
+			Logger.LogError(" Limit exceeded on raycast, Look at stack trace for What caused it "); //Meant to catch up stuff that's been naughty and doing stuff like 900 tile Ray casts
+			return new CustomPhysicsHit();
+		}
```
ThatDan123: Isnt the distance parameter also used to control the range? Also 25 seems maybe too small

Bod9001: 25 is about off screen and little bit more, on the edges up and below it plenty extra

ThatDan123: What about the distance parameter?

Bod9001: Distance parameter is just used to , automatically convert it into a Line  cast from a Raycast

NoooneyDude: Could probably allow it to go through but just log the magnitude?

NoooneyDude (dismissed):

Bod9001:
but lag tho?

UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
Bod9001: but lag tho?

ThatDan123:
Might be good to stick the magnitude value into the log

corp-0:
I also would let it pass and log the abuse, it will save someone in the future not knowing why their features that requires a longer ray doesn't work

Aranclanos:
Dont let it pass, its lagging the server at times making it uplayable. In fact i'd suggest on merging this and make the log better in a different PR so we can get this fix in a build faster.

corp-0 (approved):

NoooneyDude (approved):

PR #7930, opened 2021-12-05
Wielded items

### LINKED ISSUE #7254: Add two-handed wielding of certain items.
### Description
In /tg/, certain items such as fire axes and spears can be wielded with both hands to increase their damage. Other items such as potted plants, the cultist bastard sword, gibtonite, and the (non-Chaplain) chainsaw _require_ two hands to even hold.

There are a few variables that are relevant to the wielding component.
- wielded: Boolean, whether or not the item in question is currently being wielded.
- force_multiplier: Multiplies the damage of the item when it is wielded, doesn't work with the below two variables.
- force_wielded: Damage of the item when it is wielded.
- force_unwielded: Damage of the item when it is unwielded.
- require_twohands: If the item needs two-hands to even pick up and hold. Used for the chainsaw, potted plants, gibtonite, etc.
- icon_wielded: Sprite used for the item while it is being wielded.

When a wieldable item is used in-hand it runs a "wield" method which sets "wielded" to true unless you have an item in your other hand or if you only have one hand because your other hand got removed. Items with "require_twohands" set to true run the "wield" method upon being picked up and if wielding fails, the item is dropped.
Upon an item being wielded, its damage is adjusted either by being set to "force_wielded" or by being multiplied by "force_multiplier". Its sprite also gets updated to whatever is set in "icon_wielded". The in-hand also updates to show the wielding character holding the item with both hands.
Below, a character holding a bone spear one-handed
image
Below, a character wielding a bone spear two-handed
image
Additionally, the item gets "(Wielded)" appended to the end of its name and prevents you from swapping hands. 
Bone spear with "(Wielded)" appended to its name.
image
Message upon trying to swap hands.
image

When a wielded item is used in-hand or moved out of the hand slot (either by dropping it or moving it to another inventory slot) it is in it runs an "unwield" method which sets "wielded" to false. Unwielding an item that requires two hands to use drops it in the case of potted plants and gibtonite but has alternate functions for the bastard sword and chainsaw which I will not elaborate on here.
Upon an item being unwielded it either removes the "force_multiplier" multiplier or sets damage to "force_unwielded". It also sets the icon and in-hand sprites back to normal.

ilinm1:
### Purpose
Adds ability to carry some items (like spear) with both hands, changing its damage. Fixes #7254

### Notes:
Only glass and bone spears are currently wieldable cause Im lazy and some items like ratvarian spear are missing wielded sprites.

Also git tries to merge commits from deleted branch for some reason.

### Changelog:
CL:[New] Some items are wieldable now


ilinm1:
Ill make new PR since this branch is hopeless

PR #7935, opened 2021-12-06
Singularity "Warp" Shader

jamessimo:
### Purpose
This adds a "warp" effect around the singularity using a grabpass shader on the prefab. This is to make the singularity look like its really beanding space time and has the nice effect of objects getting sucked in! 
image

### Notes:
The Singularity has 2 shaders on it. One applies to the "lit" layers and the other only applies on the water or stars layer.
Added a check to mouse utils to make sure the texture can be read. 
Added some other WIP shaders under shaders/experimental. These are not used anywhere but are being worked on. 
Effects may need to be tweaked as we get some user feedback.

### Changelog:
CL: [Improvement] Singularity now looks like its bending space time using shaders!

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Singularity.cs
+++ b/UnityProject/Assets/Scripts/Objects/Singularity.cs
@@ -58,7 +58,13 @@ namespace Objects
 		private float maxRadiation = 5000f;
 
 		[SerializeField]
-		private LightSprite light = null;
+		private GameObject WarpEffectFront;
+
+		[SerializeField]
+		private GameObject WarpEffectBack;
+
+		[SerializeField]
+		private new LightSprite light = null;
 
 		[SerializeField]
 		[Tooltip("Allows singularity to be stage 6")]
@@ -82,6 +88,9 @@ namespace Objects
 		private SpriteHandler spriteHandler;
 		private CustomNetTransform customNetTransform;
 		private int objectId;
+
+		private Material WarpEffectFrontMat;
+		private Material WarpEffectBackMat;
 
 		private int lockTimer;
 		private bool pointLock;
@@ -118,6 +127,11 @@ namespace Objects
 			spriteHandler = GetComponentInChildren<SpriteHandler>();
 			lightTransform = light.transform;
 			objectId = GetInstanceID();
+
+
+			WarpEffectFrontMat = WarpEffectFront.GetComponent<MeshRenderer>().materials[0];
+			WarpEffectBackMat = WarpEffectBack.GetComponent<MeshRenderer>().materials[0];
+
 		}
 
 		private void Start()
@@ -125,6 +139,7 @@ namespace Objects
 			if (CustomNetworkManager.IsServer == false) return;
 
 			CurrentStage = startingStage;
+		
 		}
 
 		private void OnEnable()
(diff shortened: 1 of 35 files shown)


corp-0:
does it work nicely over the network? I didn't see any code to network the shader

ThatDan123:
> does it work nicely over the network? I didn't see any code to network the shade

Thats a good point. Singularity stage isnt networked, easy to make it a sync var as its just an enum, and then in the hook make it change the shader stuff.

Bod9001:
my question is, should the shader render over the singularity texture?

jamessimo:
> my question is, should the shader render over the singularity texture?

In can show a version with it over on discord but it make the singularity look deformed and tiny.

jamessimo:
> > does it work nicely over the network? I didn't see any code to network the shade
> 
> Thats a good point. Singularity stage isnt networked, easy to make it a sync var as its just an enum, and then in the hook make it change the shader stuff.

It’s always been like that so not sure if it’s needed since the singularity works currently?

jamessimo:
> does it work nicely over the network? I didn't see any code to network the shader

The shader is just a plane on the singularity, I assumed that as long as it’s inside the prefab it would all still work over network since it’s just inheriting it’s parents pos.

NoooneyDude (commented):
Your `UpdateWarpFX` method will only be called from the server (aside from initial spawn via Awake / OnEnable). So the scale won't be synced. You can probably get away with just moving the method call to the `SyncDynamicScale` method.

Did you test with a client / server? Testing in editor means the client _is_ the server, so it'll appear to work.

UnityProject/Assets/Scripts/Objects/Singularity.cs
```diff
@@ -168,6 +168,7 @@ private void SyncDynamicScale(Vector3 oldScale, Vector3 newScale)
 			}
 
 			gameObject.transform.LeanScale(newScale, updateFrequency);
+			UpdateWarpFX(CurrentStage);
```
ThatDan123: current stage isnt sync'd

ThatDan123: You could link the `scaledRadius` and `scaledEffect` by the `Vector3 newScale` was nooneys idea. 

Or could add sync var to stage and then add new hook for it.

UnityProject/Assets/Scripts/Objects/Singularity.cs
```diff
@@ -209,6 +230,8 @@ private void SingularityUpdate()
 			// will sync to clients
 			dynamicScale = GetDynamicScale();
 
+			UpdateWarpFX(CurrentStage);
```
ThatDan123:
```suggestion
```

UnityProject/Assets/Scripts/Objects/Singularity.cs
```diff
@@ -153,6 +168,12 @@ private void SyncDynamicScale(Vector3 oldScale, Vector3 newScale)
 			}
 
 			gameObject.transform.LeanScale(newScale, updateFrequency);
+			
+		}
+
+		private void SyncCurrentStage(SingularityStages oldStage, SingularityStages newStage)
+		{
+			currentStage = newStage;
```
ThatDan123, line 176:
```suggestion
			currentStage = newStage;
			
			UpdateWarpFX(CurrentStage);
```

ThatDan123 (approved):

corp-0 (approved):

Issue #5819, opened 2021-01-05
Airlock Construction and Department Access Remotes Bounty B:120


PerfectTangent:
- [x] Airlocks should be _constructable_ and _deconstructable._ The process is quite similar to building machines with machine frames, only requiring an Airlock Assembly instead, and should require very little new code. 

However, there are some extra particularities with Airlocks, such as _naming the Airlock with a pen_ and _painting the airlock with a spray painter._ 

- [x] Airlocks also need to have their access be configurable using Access Remotes. Each department head has an Access Remote that they can use to add or take away access requirements to doors in their department, apply emergency access (removing access restrictions), as well as bolt or unbolt the doors. They _cannot_ electrify doors.

construction recipe for airlocks:
![Screenshot 2021-01-04 200957](https://user-images.githubusercontent.com/68730437/103595965-c7f4c680-4eca-11eb-898c-1c9a7b8e226f.png)

access remotes .dm from /tg/
https://github.com/tgstation/tgstation/blob/f5bd6a381d18765624b81b4aca7fa4126b9dc2ca/code/game/objects/items/control_wand.dm

PerfectTangent:
this first requires #5815 to be merged, but it is being promoted to a bounty, amount TBD.

PerfectTangent:
the parts of #5815 relevant to doors has been completed and merged, so there is nothing holding up this bounty from being completed any longer.

Ahion:
Assign me. I will try to do it.

Ahion:
I'm very sorry, but I won't be able to continue working on this task. Please unassign me so that another person can continue the work. Here it remains to do only remote airlock control. Hope I don't let the community down too much

NoooneyDude:
's all good man, thanks for letting us know.

AtnerNT:
Should be closed by #9035?

PerfectTangent:
@Ahion Are you in the discord? What's your username?

Ahion:
> @Ahion Are you in the discord? What's your username?

Yes, I'm in discord. Ahio#4668

PerfectTangent:
Marking this bounty complete. Bounty paid.

PR #7944, opened 2021-12-08
Airlock Painter

Ahion:
### Purpose
part of #5819

Unfortunately, it was not possible to use the right-click menu to select the type of painting. I had to make a separate radial menu that could be used to select something.

### Notes:
There is a bug on the client side. Video with bug:
https://youtu.be/wnKNTjIErVc

### Changelog:
CL: [New] Added airlock painter.


Ahion:
This is what the problem looks like:

https://user-images.githubusercontent.com/58528255/145206965-2686a25f-5e8a-464b-9bc1-30c94c8d52e2.mp4

UnityProject/Assets/Scripts/Objects/Doors/ConstructibleDoor.cs
```diff
@@ -102,6 +105,14 @@ public void ServerPerformInteraction(HandApply interaction)
 					() => WhenDestroyed(null));
 				}
 			}
+			if (Validations.HasUsedComponent<AirlockPainter>(interaction))
+			{
+				AirlockPainter painter = interaction.HandObject.GetComponent<AirlockPainter>();
+				if (painter)
+				{
+					painter.ChoosePainJob(gameObject);
```
ThatDan123: This is being called in ServerPerformInteraction, you'll need to call it in WillInteract instead

UnityProject/Assets/Scripts/Items/Tool/AirlockPainter.cs
```diff
@@ -0,0 +1,31 @@
...
+	public async void ChoosePainJob(GameObject paintableAirlock)
+	{
+		GameObject chosenPaintJob = await UIManager.RadialMenu.ShowRadialMenu(AvailablePaintJobs, paintableAirlock);
+		if(chosenPaintJob != null)
+		{
+			paintTheAirlock(paintableAirlock, chosenPaintJob);
+		}
+	}
+
+	private void paintTheAirlock(GameObject paintableAirlock, GameObject chosenPaintJob)
```
ThatDan123: This will need to be called from a command in PlayerNetworkedActions

ThatDan123: Also do a validation check in the command too, just in case the player has moved away

Ahion: This is a private method of the AirlockPainter class. How can I call it from the PlayerNetworkActions class?

ThatDan123: Guess you could just move the whole method to PlayerNetworkActions since it doesnt rely on this class or make it public

PerfectTangent:
Painter tool should have a progress bar; should take 5 seconds to paint a door. Also would benefit from a sound effect trigger.

Forgive me if I missed it, but I didn't see anything related to the number of uses a painter has. Painters have a limited number of charges available to them before they run out of paint. They use toner cartridges to reload.

UnityProject/Assets/Scripts/Items/Tool/AirlockPainter.cs
Ahion: I probably don't understand something again. Wouldn't it look strange that the airlock painting method is in PlayerNetworkActions? Why transfer this method there? Does this solve the interface issue? ))) sorry for the persistence, I just want to figure it out

Ahion:
> Painter tool should have a progress bar; should take 5 seconds to paint a door. Also would benefit from a sound effect trigger.
> 
> Forgive me if I missed it, but I didn't see anything related to the number of uses a painter has. Painters have a limited number of charges available to them before they run out of paint. They use toner cartridges to reload.


This is a very crude version of airlock painter. I will add all this later when I solve the problem with the radial menu.

ThatDan123:
It might be better to just redesign this so that:

You activate the painter in hand using HandActivate, select the colour using the UI. 

Click on a door to make it that colour.

Means you only need a command to set the painter colour on server then just use the interaction system for the rest.

Ahion:
Ok i'll redo

UnityProject/Assets/Scripts/Objects/Doors/ConstructibleDoor.cs
Ahion: Many thanks. Indeed, the problem was that I was calling the radial menu in ServerPerformInteraction.

corp-0:
Your UIManager is conflicting, is this branch kind of old? Last change to UIManager was like a month ago

corp-0:
Also why wasn't it possible to add the actions to the right click wheel?

Ahion:
> Also why wasn't it possible to add the actions to the right click wheel?

You mean actions with items, like "pick up" and "examine"? If I add a new "paint" action, how does that help me provide the player with a choice of paint types? I don’t understand how it should work.

UnityProject/Assets/Scripts/Items/Tool/AirlockPainter.cs
NoooneyDude: PNA is just our class that holds misc. client -> server requests.

Since it looks like your UI is all client-side, you need a way to inform the server. You could do this through IF2 (interaction system), PNA or a net message. 

PNA is probably fine here.

corp-0:
No need for DNM label since this is a draft.

> You mean actions with items, like "pick up" and "examine"? If I add a new "paint" action, how does that help me provide the player with a choice of paint types? I don’t understand how it should work.

Now that you make me think about it, yeah it doesn't make much sense. Follow the approach Daniel described instead.

Ahion:
I have some bug again. Airlock sprite blink on client-side. I can't understand why this bug is happening and how to fix it. Help me if you can))
Video with bug:
https://youtu.be/wnKNTjIErVc

UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
```diff
@@ -1008,4 +1007,11 @@ public void CmdTriggerStorageTrap(GameObject storage)
 			}
 		}
 	}
+
+	[Command]
+	public void CmdSetPaintJob(int paintJobIndex)
+	{
+		AirlockPainter painter = GetActiveHandItem().GetComponent<AirlockPainter>();
+		painter.CurrentPaintJobIndex = paintJobIndex;
```
ThatDan123:
```suggestion
		var handObject= GetActiveHandItem();
		
		if(handObject == null || handObject.TryGetComponent<AirlockPainter>(out var painter) == false) return;
		
		painter.CurrentPaintJobIndex = paintJobIndex;
```

ThatDan123 (commented):
@Bod9001 Have any idea why the sprite is like that?

Bod9001:
looks like the issue is that, your having to wrestle with the door animator that is ancient as and I barely understand it myself

Ahion:
Except for the bug with the animator, PR is done. I can’t resolve the Codacy review, since these arguments are needed. `eventData `and `pointerEvent ` are used in `AddListener() `and without them an error occurs. I can't solve the bug with the animator yet, I suggest you merge with this bug and solve it in a separate PR. Open for review. Here is a GIF of what the work looks like:
painter

PerfectTangent:
watching the bug footage closely, it looks like the original door sprite isn't being removed properly.

Screenshot 2022-01-01 103017
 
 both sprites are visible simulaneously, overlapping eachother.

Bod9001 (approved):

ThatDan123 (approved):

Issue #2327, opened 2019-11-22
Add folders and file cabinets


StrawberryMoses:
# Description

Folders and file cabinets hold paper to be grouped together. When used, they display a popup where paper can either be removed or viewed.

- [x] add folders
- [x] add file cabinets

https://github.com/tgstation/tgstation/blob/cc1f39661c1cda8de1caa005ea199bb3d8ca591f/code/modules/paperwork/folders.dm
https://github.com/tgstation/tgstation/blob/dd6fa5ededb15c36bb0f0663a4f37c8c699036a5/code/modules/paperwork/filingcabinet.dm

dapocalypse:
Ill take this up

PR #7950, opened 2021-12-10
Paperwork folders

PerfectTangent:
### Purpose

This PR adds paper folders which can be used to store paper sheets, disks, photos, and other thin fileable items. Completes half of #2327. It also maps folders into Square, Fall, Outpost, and Pog. This includes adding the folder that typically contains the Secret Documents HRI to the vault of each map, preparing them for when that item is added.

### Changelog:

CL: [New] Added paperwork folders.


PerfectTangent:
I don't suppose the merging of this PR could be expedited? I realized there's a separate mapping task that needs doing.

corp-0 (approved):

NoooneyDude (approved):

PR #7963, opened 2021-12-11
Checking null is bad

Bod9001:
removes that from staging
assuming that staging is running at 30 fps 
this should bring it up to 40 fps

### DIFF
--- a/UnityProject/Assets/Mirror/Runtime/NetworkServer.cs
+++ b/UnityProject/Assets/Mirror/Runtime/NetworkServer.cs
@@ -1456,9 +1456,8 @@ namespace Mirror
         // broadcasting ////////////////////////////////////////////////////////
         // helper function to get the right serialization for a connection
 	static NetworkWriter GetEntitySerializationForConnection(NetworkIdentity identity,
-			NetworkConnectionToClient connection, out bool IsNull )
+			NetworkConnectionToClient connection)
 		{
-			//CUSTOM UNITYSTATION CODE// IsNull is custom
 			// get serialization for this entity (cached)
 			// IMPORTANT: int tick avoids floating point inaccuracy over days/weeks
 			NetworkIdentitySerialization serialization = identity.GetSerializationAtTick(Time.frameCount);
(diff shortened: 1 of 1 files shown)


Bod9001:
Tested with built client , and built client connection to it, with Big explosion, with re-log and round restart

UnityProject/Assets/Mirror/Runtime/NetworkServer.cs
```diff
@@ -1522,47 +1533,48 @@ static void BroadcastToConnection(NetworkConnectionToClient connection)
...
+	            //       TODO make this more simple / unnecessary later.#
+	            //CUSTOM UNITYSTATION CODE// commented this out \/
+	            //identity.ClearDirtyComponentsDirtyBits();
+                //}
                 // spawned list should have no null entries because we
                 // always call Remove in OnObjectDestroy everywhere.
                 // if it does have null then someone used
                 // GameObject.Destroy instead of NetworkServer.Destroy.
-                else Debug.LogWarning("Found 'null' entry in observing list for connectionId=" + connection.connectionId + ". Please call NetworkServer.Destroy to destroy networked objects. Don't use GameObject.Destroy.");
+                //else Debug.LogWarning("Found 'null' entry in observing list for connectionId=" + connection.connectionId + ". Please call NetworkServer.Destroy to destroy networked objects. Don't use GameObject.Destroy.");
```
corp-0, line 1573: Why comment this?

Bod9001: it's the other side of the null check that I removed

corp-0: Well yeah but checking the IsNull variable conceptually does the same and if it is true then the log is relevant, I think

Bod9001: It's only elseon the null date Needing to be sent

corp-0 (dismissed):

ThatDan123 (approved):

NoooneyDude (approved):

PR #7614, opened 2021-10-19
Bone Setters, Gauze stack fix, Trauma balance improvements

MaxIsJoe:
part of #6272

Adds bone setters, an item required for the upcoming surgery PR for traumas.
Currently you can use it to heal Joint Dislocation without surgery, just let your doctor grab your arm and let him enjoy the sounds of your pain.

Fixes an issue where Gauze was supposed to be stackable but I forgot to assign what objects it can stack with.

Trauma balance improvements, you can now fine tune all body part's damage thresholds for Brute, Bomb and Burn attack types which should fix #7610 
~~Also trauma is now applied after checking your overall health not body part health anymore, this should fix issues where players would sneakily apply traumas before someone goes into crit.~~

### Note
I'll be making smaller PRs like this from now on until the surgery PR instead of cramming all items and features into one large PR that will be a pain to test.

I've also made a lot of hardcoded stuff in HealsTheLiving script tuneable in the inspector

PR #7967, opened 2021-12-12
Stops people being completely gibbed by damage

Bod9001:
tell me what you think,
why it seems a bit BS that you can get gibbed from someone shooting you in chest with a laser
also disembowelling actually works properly now

CL: [balance] players can no longer get completely gibbed by taking too much damage to the chest.
CL: [Fix] fixed disemboweling


MaxIsJoe:
The issue is that the damage thresholds and balance are wrong, I already made a fix for this LONG ago but it's on hold along side with the bone setters.

Trauma is also on hold and we're no longer making any design changes to it because it is essentially frozen. If you want to stop people from getting completely gibbed you'll need to up the damage thresholds for dismemberment on the chest because they were low for some reason.
Does 41 damage sound good as a damage threshold on the chest?

Bod9001:
image
is all good

NoooneyDude (approved):
Related to #7614.

ThatDan123 (approved):

PR #7968, opened 2021-12-12
Added SpillOnThrow item attribute to base Bottle and Can

jamessimo:
### Purpose
This is a very small add but makes a large difference, cans and bottles right now do nothing when you throw them (even bottles which break open on throw). This will make greytiders happy and keep janitors busy. 

image

### Notes:
Tested with quite a few items, I doubt this has any other larger ramifications.

### Changelog:
CL: [Improvement] Cans and Bottles will now spill their contents on the floor when thrown.


Bod9001:
do we have code for opening the cans?

jamessimo:
> do we have code for opening the cans?

No we don't, its assumed as soon as its in your hand its open.

Bod9001:
ok cool

Bod9001 (approved):

NoooneyDude (approved):

PR #7973, opened 2021-12-13
Removes thread sleeps from the sickness manager sickplayers lock

Aranclanos:
### Purpose
Removes thread sleeps from the sickness manager sickplayers lock

### Notes
I don't believe its significant enough to cause a huge issue, but it could freeze the main thread for a tiny bit


NoooneyDude (approved):
I don't suppose you could fix the nesting while you're in there? 😛 

Guess the whole thing needs a refactor at some point anyway...

ThatDan123 (approved):

PR #7974, opened 2021-12-14
squeezing a little bit of performance with `for()` for array in chat relay

MaxIsJoe:
title

for() + arrays = slightly better performance after doing a little bit of research

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Chat/ChatRelay.cs
+++ b/UnityProject/Assets/Scripts/Core/Chat/ChatRelay.cs
@@ -10,6 +10,7 @@ using Systems.Ai;
 using Messages.Server;
 using Objects.Telecomms;
 using UI.Chat_UI;
+using UnityEngine.Profiling;
 
 /// <summary>
 /// ChatRelay is only to be used internally via Chat.cs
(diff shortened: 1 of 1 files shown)


PerfectTangent:
do you think you could demonstrate the performance increase?

MaxIsJoe:
> do you think you could demonstrate the performance increase?

I can't really demonstrate how big of an impact it is locally because there isn't much going on but I can share how much of a difference it does using someone else's tests
https://stackoverflow.com/questions/454916/performance-of-arrays-vs-lists

I also recommend reading the answer below it to understand more on when Array and Lists make sense.

NoooneyDude:
There should be a negligible difference between `for` and `foreach` when iterating over arrays (and the source you linked concludes as much).

`foreach` conveys programming intent better, is slightly more readable and safer to use (admittedly there isn't much to go wrong in this case, but in principle).

MaxIsJoe:
> There should be a negligible difference between `for` and `foreach` when iterating over arrays (and the source you linked concludes as much).
> 
> `foreach` conveys programming intent better, is slightly more readable and safer to use (admittedly there isn't much to go wrong in this case, but in principle).

every bit of performance gain is welcome especially when the game can generate a lot of messages at once
and again, as the PR's title states; this is squeezing for performance.

edit :
I don't like the concept of "make readable code, not performant code". It's the reason why many unoptimized and heavy stuff on the web and chromium powered applications exist. It's bad programming practice and having faster computers isn't a valid excuse to ignore that.
If there's a chance to make code run 1ms faster it's going to be useful in the long run when that code stacks with the other many things that are happening on the same thread.

NoooneyDude:
If you can show an improvement of 1ms or even 0.01ms, then by all means I'll approve this instantly ; )

ThatDan123:
> Array/for: 1864ms (589725196)
> Array/foreach: 1860ms (589725196)

From your source, when iterating over 6000000 entries says that `For` is worse than `Foreach.` This is really negligible difference anyway so the same due to margin of error.

Aranclanos:
unrelated comment, you might find comments saying that `foreach` causes gc in unity but that was fixed many years ago

MaxIsJoe:
I can't actually prove how much of an impact this does due to how random the results are at times but **I do know that the for method seem to constantly bring it down to approximately 0.13ms instead of the usual 0.40~ in `foreach()` when the randomness lands in it's favor spending a lot of time trying to force results and entering and exiting playmode.**

because of it's random nature and maintainers thinking it's not that going to make that much of a difference I've decided I'm not going to put more time into this and focus on something else.

PR #7992, opened 2021-12-16
Adds Pointing to the Right Click Menu

ThatDan123:
CL: [New] Adds Pointing to the Right Click Menu

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Attributes.cs
+++ b/UnityProject/Assets/Scripts/Items/Attributes.cs
@@ -181,12 +181,9 @@ public class Attributes : NetworkBehaviour, IRightClickable, IExaminable
 
 	public RightClickableResult GenerateRightClickOptions()
 	{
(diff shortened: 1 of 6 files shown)


Bod9001 (dismissed):

UnityProject/Assets/Scripts/Items/Attributes.cs
```diff
@@ -176,8 +181,12 @@ public string Examine(Vector3 worldPos)
 
 	public RightClickableResult GenerateRightClickOptions()
 	{
-		return RightClickableResult.Create()
```
corp-0, line 179: it is more elegant to make ``AddElement`` return the object so you can chain as many ``AddElement`` as you need.
```cs
return RightClickableResults.Create()
    .AddElement("Examinable", OnExamine)
    .AddElement("PointTo", OnPointTo)
```

Bod9001 (approved):

corp-0 (approved):

PR #8003, opened 2021-12-18
Persistent Server Logs version 0.00000001

MaxIsJoe:
Adds basic persistent log support, very barebones and uses a string based approach rather than a full pattern model.

A bunch of interactions are currently being logged such as combat, explosions, chat, admin meme-ing, etc.

This is going to be a small side project along side the Signals PR, expect a proper way of handling logs in the future and adding important features to the logs UI such as filtering, priority logs and most importantly; clicking on hyperlinked text that does actions for you such as teleporting to a log's recorded position on the map and bwoinking an offending player who started singlo for no reason as a captain.


corp-0:
Correct me if I'm wrong but I don't think Expensive name is what you want, since that's the IC friendly name and for admins log you need something like AccountName as ICName

MaxIsJoe:
> Correct me if I'm wrong but I don't think Expensive name is what you want, since that's the IC friendly name and for admins log you need something like AccountName as ICName

where else is this an issue?

Bod9001:
tests failed

corp-0:
I'm interested in merging this feature but I need it to be robuster

MaxIsJoe:
> I'm interested in merging this feature but I need it to be robuster

I'll be working on it again soon

moonheart08:
So tidbit of experience from working with SS14's admin logs (both as a coder, and as a game admin):
You do not want to store logs in a text file.
*Ever*.
You need a machine readable log format so that game staff can view logs in-game and *search* those logs.

Look into using a structured logging system for this, ideally you'd be able to use interpolated string handlers to make it dead simple to write, but I'm not sure you guys are on a new enough C# for that. In lieu of that, I'd say use structs with some interface implemented until you move to a newer version of C# that has interpolated string handlers.
From there, you'll probably want to give admins three search options for a round; search by username (OR), search by log type (also OR), and a textual search.
Reading log data outside the game is something you may want to work on later as well, but giving admins immediate access to round log info is of higher priority.

PR #8005, opened 2021-12-18
Makes it so clients load all the scenes before data is sent

Bod9001:
I tested it a bit but I do more testing now

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.Client.cs
+++ b/UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.Client.cs
@@ -102,7 +102,7 @@ public partial class SubSceneManager
 
 		foreach (var Scene in Scenes)
 		{
-			yield return WaitFor.Seconds(0.1f);
+			yield return WaitFor.Seconds(0.1f); //For smooth FPS not necessary technically, but causes freeze For a little bit
 			if (KillClientLoadingCoroutine)
 			{
 				KillClientLoadingCoroutine = false;
@@ -113,7 +113,7 @@ public partial class SubSceneManager
 		}
 
 		clientIsLoadingSubscene = false;
-		yield return WaitFor.Seconds(0.1f);
+		yield return WaitFor.Seconds(0.1f); //For smooth FPS not necessary technically, but causes freeze For a little bit
 		if (KillClientLoadingCoroutine)
 		{
 			KillClientLoadingCoroutine = false;
--- a/UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.Client.cs
+++ b/UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.Client.cs
@@ -96,16 +96,8 @@ public partial class SubSceneManager
 		}
 
 		NetworkClient.PrepareToSpawnSceneObjects();
-		yield return WaitFor.Seconds(0.2f);
-		if (KillClientLoadingCoroutine)
-		{
-			KillClientLoadingCoroutine = false;
-			clientIsLoadingSubscene = false;
-			yield break;
-		}
-
 		RequestObserverRefresh.Send(OriginalScene);
-
+		
 		foreach (var Scene in Scenes)
 		{
 			yield return WaitFor.Seconds(0.1f);
--- a/UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.Client.cs
+++ b/UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.Client.cs
@@ -1,5 +1,8 @@
-﻿using System.Collections;
+﻿using System;
+using System.Collections;
 using System.Collections.Generic;
+using Messages.Client;
+using Mirror;
 using UnityEngine;
 
 // Client
@@ -29,7 +32,7 @@ public partial class SubSceneManager
 		}
 	}
 
-	IEnumerator LoadClientSubScene(SceneInfo sceneInfo)
+	IEnumerator LoadClientSubScene(SceneInfo sceneInfo, bool SynchronisingHandled = true)
 	{
 		if (sceneInfo.SceneType == SceneType.MainStation)
 		{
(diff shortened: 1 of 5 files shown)


UnityProject/Assets/Scripts/Player/JoinedViewer.cs
```diff
@@ -18,26 +20,81 @@
...
+		{
+			if (AlreadyLoaded == Scene.SceneName) continue;
+			SceneS.Add(Scene);
+		}
+
+		RpcLoadScenes(JsonConvert.SerializeObject(SceneS), AlreadyLoaded);
+	}
+
+
+	[ClientRpc]
```
ThatDan123:
```suggestion
	[TargetRpc]
```

corp-0:
You can ignore all codacy suggestions this time

UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.Client.cs
```diff
@@ -29,7 +32,7 @@ void MonitorServerSceneListOnClient()
 		}
 	}
 
-	IEnumerator LoadClientSubScene(SceneInfo sceneInfo)
+	IEnumerator LoadClientSubScene(SceneInfo sceneInfo, bool SynchronisingHandled = true)
```
corp-0: Except this one, what's up with the unused parameter?

UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.Client.cs
```diff
@@ -49,4 +51,36 @@ IEnumerator LoadClientSubScene(SceneInfo sceneInfo)
...
+	}
+
+	IEnumerator LoadClientScenesFromServer(List<SceneInfo> Scenes, string OriginalScene, Action OnFinish)
+	{
+		clientIsLoadingSubscene = true;
+		foreach (var Scene in Scenes)
+		{
+			clientIsLoadingSubscene = true;
+			yield return LoadClientSubScene(Scene, false);
+			clientIsLoadingSubscene = true;
```
corp-0: Why twice?

UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.Client.cs
```diff
@@ -49,4 +51,36 @@ IEnumerator LoadClientSubScene(SceneInfo sceneInfo)
...
+			clientLoadedSubScenes.Add(Scene);
+		}
+
+		NetworkClient.PrepareToSpawnSceneObjects();
+		yield return WaitFor.Seconds(0.2f);
```
corp-0: Oh no, can't we wait until it is reported that all scenes loaded instead of fixed time?

corp-0:
Yeah I don't think going back to IEnumerators and waiting for arbitrary time is the robust way to solve it, I would rather make them Tasks and report back when they are done

UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.Client.cs
Bod9001: I can try removing it I just added it in case it might explode

UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.Client.cs
Bod9001: just to make double sure

UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.Client.cs
Bod9001: it should be used, I think in the original PR it wasn't

Bod9001:
it is using tasks though, it's looping and waiting for each scene to load before proceeding to the next one, the rest is not to overload server with messages

UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.Client.cs
corp-0: Yeah I'd find the way to remove all of them waiting times

UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.Client.cs
corp-0: Well actually thrice, to make triple sure? Why would it change before we want it to?

corp-0:
Then await the tasks instead of hardcoding waiting time

Bod9001:
The Unity coroutines have the Wait stuff built in, yield return SceneManager.UnloadSceneAsync(Scene.SceneName); 
Actually waits for the function to complete before continuing

UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.Client.cs
Bod9001: because the scene manager will try loading in other scenes, while this is loading in scenes

UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.Client.cs
```diff
@@ -29,24 +35,91 @@ void MonitorServerSceneListOnClient()
...
-			yield return StartCoroutine(LoadSubScene(sceneInfo.SceneName));
+			if (SubsceneLoadTimer != null)
+			{
+				SubsceneLoadTimer.IncrementLoadBar($"Loading {sceneInfo.SceneName}");
+			}
+
+			yield return StartCoroutine(LoadSubScene(sceneInfo.SceneName, HandlSynchronising  :HandlSynchronising ));
+		}
+
+		if (OverrideclientIsLoadingSubscene == false)
```
corp-0, line 65: Why here and not just do it before the loop and after the loop? Why does this method need to know about this bool when it is only relevant for the code that uses this method

Bod9001: because it's used for the sub scene manager, To prevent multiple scenes being loaded at once

corp-0 (approved):

ThatDan123:
Has this been properly headless tested?

-Joining game

-Exiting fully then trying to rejoin.

-Ending round and then loading the new maps scenes?

Bod9001:
I'll give it a quick headless test now

Bod9001:
yeah, it's been tested good to go

ThatDan123 (approved):

PR #8006, opened 2021-12-18
Wieldable stuff (2)

### LINKED ISSUE #7254: Add two-handed wielding of certain items.

ilinm1:
### Purpose
Adds ability to carry some items (like spear) with both hands, changing its damage. Fixes #7254

### Notes:
Only glass and bone spears are currently wieldable cause Im lazy and some items like ratvarian spear are missing wielded sprites.

### Changelog:
CL: [New] Some items are wieldable in two hands now for enhanced damage.

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Wieldable.cs
+++ b/UnityProject/Assets/Scripts/Items/Wieldable.cs
@@ -1,0 +1,137 @@
+using Items;
+using Messages.Server;
+using Mirror;
+using UnityEngine;
+
+public class Wieldable : NetworkBehaviour, IServerInventoryMove, ICheckedInteractable<HandActivate>
+{
+	[SerializeField]
+	private int damageUnwielded;
+
+	[SerializeField]
+	private int damageWielded;
+
+	[SerializeField]
+	public ItemsSprites Wielded = new ItemsSprites();
+	public ItemsSprites Unwielded = new ItemsSprites();
+
+	[SyncVar(hook = nameof(SyncState))]
+	private bool isWielded;
+
+	private SpriteHandler spriteHandler;
+	private ItemAttributesV2 itemAttributes;
+
+	private void Awake()
+	{
+		itemAttributes = GetComponent<ItemAttributesV2>();
+		spriteHandler = GetComponentInChildren<SpriteHandler>();
+	}
+
+	private void SyncState(bool oldState, bool newState)
+	{
+		isWielded = newState;
+
+		if (isWielded)
+		{
+			itemAttributes.SetSprites(Wielded);
+		}
+		else
+		{
+			itemAttributes.SetSprites(Unwielded);
+		}
+	}
+
+	public void OnInventoryMoveServer(InventoryMove info)
+	{
+		if (info.InventoryMoveType == InventoryMoveType.Remove)
+		{
+			isWielded = false;
+			ItemSlot hiddenHand = DetermineHiddenHand();
+
+			if (hiddenHand != null)
+            {
+				HideHand(PlayerManager.LocalPlayerScript.connectionToClient, false, hiddenHand);
+			}
+		}
+	}
+
+	[TargetRpc]
+	private void HideHand(NetworkConnection target, bool HideState, ItemSlot HiddenHand)
+	{
+		int hiddenHandSelection = 0;
+
+		if (HiddenHand.NamedSlot.GetValueOrDefault(NamedSlot.none) == NamedSlot.leftHand)
+		{
+			hiddenHandSelection = 1;
+		}
+		else if (HiddenHand.NamedSlot.GetValueOrDefault(NamedSlot.none) == NamedSlot.rightHand)
+		{
+			hiddenHandSelection = 2;
+		}
+
+		HandsController.Instance.HideHands(HideState, hiddenHandSelection);
+	}
+
+	private ItemSlot DetermineHiddenHand()
+	{
+		ItemSlot hiddenHand = null;
+		var playerStorage = PlayerManager.LocalPlayerScript.DynamicItemStorage;
+		var currentSlot = playerStorage.GetActiveHandSlot();
+
+		var leftHands = playerStorage.GetNamedItemSlots(NamedSlot.leftHand);
+		foreach (var leftHand in leftHands)
+		{
+			if (leftHand != currentSlot && Validations.HasComponent<Wieldable>(leftHand.ItemObject) == false)
+			{
+				hiddenHand = leftHand;
+			}
+		}
+
+		var rightHands = playerStorage.GetNamedItemSlots(NamedSlot.rightHand);
+		foreach (var rightHand in rightHands)
+		{
+			if (rightHand != currentSlot && Validations.HasComponent<Wieldable>(rightHand.ItemObject) == false)
+			{
+				hiddenHand = rightHand;
+			}
+		}
+
+		return hiddenHand;
+	}
+
+	public bool WillInteract(HandActivate interaction, NetworkSide side)
+	{
+		if (!DefaultWillInteract.Default(interaction, side)) return false;
+
+		return true;
+	}
+
+	public void ServerPerformInteraction(HandActivate interaction)
+	{
+		ItemSlot hiddenHand = DetermineHiddenHand();
+
+		if (hiddenHand != null)
+        {
+			Inventory.ServerDrop(hiddenHand);
+
+			isWielded = !isWielded;
+
+			if (isWielded)
+			{
+				itemAttributes.ServerHitDamage = damageWielded;
+				itemAttributes.SetSprites(Wielded);
+				Chat.AddExamineMsgFromServer(interaction.Performer, $"You wield {gameObject.ExpensiveName()} grabbing it with both of your hands.");
+				HideHand(PlayerManager.LocalPlayerScript.connectionToClient, true, hiddenHand);
+			}
+			else
+			{
+				itemAttributes.ServerHitDamage = damageUnwielded;
+				itemAttributes.SetSprites(Unwielded);
+				Chat.AddExamineMsgFromServer(interaction.Performer, $"You unwield {gameObject.ExpensiveName()}.");
+				HideHand(PlayerManager.LocalPlayerScript.connectionToClient, false, hiddenHand);
+			}
+
+			PlayerAppearanceMessage.SendToAll(interaction.Performer, (int)interaction.HandSlot.NamedSlot.GetValueOrDefault(NamedSlot.none), gameObject);
+		}
+	}
+}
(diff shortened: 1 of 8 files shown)


UnityProject/Assets/Scripts/UI/Systems/MainHUD/UI Bottom/DoubleHandController.cs
```diff
@@ -24,12 +24,26 @@ public class DoubleHandController : MonoBehaviour
 	public HandsController RelatedHandsController;
 
 
-	public void HideHands(bool HideState)
+	//0 - Hide both hands, 1 - hide left hand, 2 - hide right hand
```
Lizurt: Please use enums in such cases, I don't see any reasons why these values should be ints

ilinm1: good enough?

Lizurt: Better, but I still see many places where you still use ints

UnityProject/Assets/Scripts/Items/Wieldable.cs
```diff
@@ -0,0 +1,180 @@
...
+	{
+		[SerializeField]
+		private int damageUnwielded;
+
+		[SerializeField]
+		private int damageWielded;
+
+		[SerializeField]
+		public ItemsSprites Wielded = new ItemsSprites();
+		public ItemsSprites Unwielded = new ItemsSprites();
```
corp-0, line 16: Public fields don't need the SerializedField attribute, that said prefer to use private fields and add the attribute. If the field is needed elsewhere, make a property so its intention is more explicit

UnityProject/Assets/Scripts/Items/Wieldable.cs
```diff
@@ -0,0 +1,180 @@
...
+					{
+						hiddenHandSelection = (int) hiddenHandValues.rightHand;
+					}
+
+					HideHand(PlayerManager.LocalPlayerScript.connectionToClient, false, hiddenHandSelection);
+				}
+			}
+		}
+
+		[TargetRpc]
```
Bod9001: I'm not too sure, but I think this can be called by any client , so any client could spoof another client's input
@ThatDan123  knows more about it than me

UnityProject/Assets/Scripts/Items/Wieldable.cs
```diff
@@ -0,0 +1,180 @@
...
+				if (isWielded)
+				{
+					itemAttributes.ServerHitDamage = damageWielded;
+					itemAttributes.SetSprites(Wielded);
+					Chat.AddExamineMsgFromServer(interaction.Performer, $"You wield {gameObject.ExpensiveName()} grabbing it with both of your hands.");
+					HideHand(PlayerManager.LocalPlayerScript.connectionToClient, true, hiddenHandSelection);
+				}
+				else
+				{
+					itemAttributes.ServerHitDamage = damageUnwielded;
```
Bod9001, line 152: need to apply this so, when it's removed from inventory it applies Correct damage, for example when the player gets Slipped or something like that

corp-0:
what's up with unused reference to spriteHandler (see codacy issues)

UnityProject/Assets/Scripts/Items/Wieldable.cs
ThatDan123: This is fine since its Server --> Client.

Only issue is the hands stuff isnt a sync var, so if players leave and rejoin they dont have the correct hand values

Bod9001: might be worth making it a  sync var?

UnityProject/Assets/Scripts/Items/Wieldable.cs
```diff
@@ -0,0 +1,182 @@
...
+				}
+			}
+
+
+		}
+
+		[TargetRpc]
+		private void HideHand(NetworkConnection target, bool HideState, int HiddenHandSelection)
+		{
+			HandsController.Instance.HideHands(HideState, HiddenHandSelection);
```
Bod9001: The only issue with this is that if you have multiple hands, it will hide multiple, 
so you say hide the left hand, it would hide all left hands, 
this is not too bad but , is unexpected behaviour

Bod9001:
looks good otherwise just fix up the Useless reference for sprite handler

Bod9001:
and ues synchronise var instead

UnityProject/Assets/Scripts/Items/Wieldable.cs
ilinm1: imma bit confused, is it enough to make SyncVar int to store data about hidden hand and call HideHand every time when player rejoins or do something about HandsController in general so so it remembers which hands are active and which not?

ThatDan123: Make a sync var in here for the hand enum: https://github.com/unitystation/unitystation/blob/develop/UnityProject/Assets/Scripts/Player/PlayerOnlySyncValues.cs

And on the hook call HideHands (add `none` hidden to the enum, so you can get rid of the bool)

Bod9001: only need to change how it's called from server, the hand hiding stuff is handled by dynamic inventory

ilinm1: good enough now?

UnityProject/Assets/Scripts/UI/Systems/MainHUD/UI Bottom/DoubleHandController.cs
```diff
@@ -24,38 +24,31 @@ public class DoubleHandController : MonoBehaviour
...
-
-			if (LeftHandActive)
-			{
-				LeftHand.SetActive(true);
-			}
+			case 0:
+				LeftHand.SetActive(false);
+				RightHand.SetActive(false);
+				break;
+			case 1:
```
Bod9001, line 36: would say you could use the enum throughout but, That's not too much of concern, looks good

Bod9001 (dismissed):

ThatDan123 (approved):
Would be good if someone headless tested this, but the code looks good. Edit: sorry spotted some stuff.

UnityProject/Assets/Scripts/Items/Wieldable.cs
```diff
@@ -0,0 +1,163 @@
...
+		private int damageWielded;
+
+		public ItemsSprites Wielded = new ItemsSprites();
+		public ItemsSprites Unwielded = new ItemsSprites();
+
+		[SyncVar(hook = nameof(SyncState))]
+		private bool isWielded;
+
+		private ItemAttributesV2 itemAttributes;
+		private HandsController HandsController;
```
ThatDan123:
```suggestion
```

This isnt needed

UnityProject/Assets/Scripts/Items/Wieldable.cs
```diff
@@ -0,0 +1,163 @@
...
+			HandsController.OnSwapHand.RemoveListener(OnSwapHands);
+		}
+
+		private void OnSwapHands()
+		{
+			if (isWielded)
+			{
+				Chat.AddExamineMsgFromServer(PlayerManager.LocalPlayerScript.gameObject, $"Your other hand is too busy holding {gameObject.ExpensiveName()}!");
+				HandsController.OnSwapHand.RemoveListener(OnSwapHands);
+				HandsController.SwapHand();
```
ThatDan123, line 52:
```suggestion
				HandsController.Instance.SwapHand();
```

ThatDan123 (changes requested):
Do changes first

UnityProject/Assets/Scripts/Items/Wieldable.cs
```diff
@@ -0,0 +1,162 @@
...
+			HandsController.OnSwapHand.RemoveListener(OnSwapHands);
+		}
+
+		private void OnSwapHands()
+		{
+			if (isWielded)
+			{
+				Chat.AddExamineMsgFromServer(PlayerManager.LocalPlayerScript.gameObject, $"Your other hand is too busy holding {gameObject.ExpensiveName()}!");
+				HandsController.OnSwapHand.RemoveListener(OnSwapHands);
+				HandsController.Instance.SwapHand();
```
ThatDan123:
```suggestion
				HandsController.SwapHand();
```

Ah sorry, didnt see that SwapHand was static

ilinm1: Ill commit everything you propose

ThatDan123 (approved):

UnityProject/Assets/Scripts/UI/Systems/MainHUD/UI Bottom/DoubleHandController.cs
corp-0: actually, use the enum. No need to use int here and it would make the casting also unneeded because this method would receive the enum type anyways.

Bod9001 (approved):

ilinm1:
wait I wanted to fix that enum stuff

ilinm1:
doesnt matter now I suppose

PR #8012, opened 2021-12-19
Adds server rebooting to tackle memory crash issues

MaxIsJoe:
Because of memory issues that happen after a few rounds of playing; this PR aims to add server restarting to the game.
For servers to restart they'll require `systemctl` like how staging currently has one to auto-restart the server.

Incase you as the server host don't have something similar installed; there are two ways to disable this. Either by changing it from the gameConfig on build or using a commandline argument called `-NoReboot` when starting a headless server.

Server restarting will trigger now when one of two conditions are met; server performance has been detected to be low OR the `-AlwaysReboot` argument has been passed to the server to always reboot every round.

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/GameManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/GameManager.cs
@@ -770,7 +770,7 @@ public partial class GameManager : MonoBehaviour, IInitialise
 	{
 		string[] args = Environment.GetCommandLineArgs();
 		if ((ServerShutsDownOnRoundEnd == false || args.Contains("-NoReboot"))
-		    && (ServerAverageFPS >= 45 || GetMemeoryUsagePrecentage() <= 75f) || args.Contains("-AlwaysReboot"))
+		    && (ServerAverageFPS >= 45 || GetMemeoryUsagePrecentage() <= 75f) || args.Contains("-AlwaysReboot") == false)
 		{
 			Logger.Log("Server restarting round now.", Category.Round);
 			Chat.AddGameWideSystemMsgToChat("<b>The round is now restarting...</b>");
--- a/UnityProject/Assets/Scripts/Managers/GameManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/GameManager.cs
@@ -784,7 +784,7 @@ public partial class GameManager : MonoBehaviour, IInitialise
 			StopAllCoroutines();
 			yield break;
 		}
-		Logger.Log("Server is rebooting now. If you don't have a way to automatically restart the " +
+		Logger.LogError("Server is rebooting now. If you don't have a way to automatically restart the " +
 		           "Unitystation process such as systemctl the server won't be able to restart!", Category.Round);
 		Chat.AddGameWideSystemMsgToChat("<size=72><b>The server is now restarting!</b></size>");
 		yield return WaitFor.Seconds(2f);
--- a/UnityProject/Assets/Scripts/Managers/GameManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/GameManager.cs
@@ -763,7 +763,7 @@ public partial class GameManager : MonoBehaviour, IInitialise
 
 	private float GetMemeoryUsagePrecentage()
 	{
-		return (Profiler.GetTotalAllocatedMemory() / 1048576) / SystemInfo.systemMemorySize * 100;
+		return (Profiler.GetTotalAllocatedMemoryLong() / 1048576) / SystemInfo.systemMemorySize * 100;
 	}
 
 	IEnumerator ServerRoundRestart()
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/Managers/GameManager.cs
```diff
@@ -754,18 +761,33 @@ public void RestartRound()
 		StartCoroutine(ServerRoundRestart());
 	}
 
-	IEnumerator ServerRoundRestart()
+	private float GetMemeoryUsagePrecentage()
 	{
-		Logger.Log("Server restarting round now.", Category.Round);
-		Chat.AddGameWideSystemMsgToChat("<b>The round is now restarting...</b>");
+		return (Profiler.GetTotalAllocatedMemoryLong() / 1048576) / SystemInfo.systemMemorySize * 100;
```
Bod9001, line 766: 1048576)?

UnityProject/Assets/Scripts/Managers/GameManager.cs
```diff
@@ -754,18 +761,33 @@ public void RestartRound()
...
+	}
 
-		// Notify all clients that the round has ended
-		EventManager.Broadcast(Event.RoundEnded, true);
+	IEnumerator ServerRoundRestart()
+	{
+		string[] args = Environment.GetCommandLineArgs();
+		if ((ServerShutsDownOnRoundEnd == false || args.Contains("-NoReboot"))
+		    && (ServerAverageFPS >= 45 || GetMemeoryUsagePrecentage() <= 75f) || args.Contains("-AlwaysReboot"))
```
Bod9001: surely this would trigger if there's a lot of people?

UnityProject/Assets/Scripts/Managers/GameManager.cs
MaxIsJoe: GetTotalAllocatedMemoryLong() returns memory in bytes; this converts it into MBs

UnityProject/Assets/Scripts/Managers/GameManager.cs
MaxIsJoe: >Surely this would trigger if there's a lot of people

Sounds like performance issue that needs to be delt with; and this would ensure that everything is cleaned up properly so that in the next round those large number of players don't crash the server for the time being.

UnityProject/Assets/Scripts/Managers/GameManager.cs
Bod9001: ok cool

UnityProject/Assets/Scripts/Managers/GameManager.cs
Bod9001: Wait a minute, its testing if server FPS is higher than 45  surely it should be checking if there is less than 45?

MaxIsJoe: Read the if statement again; it's checking if we don't want to reboot

Bod9001: well, Wouldn't you have to inverse GetMemeoryUsagePrecentage?

MaxIsJoe: if the server FPS is greater 45 then restart the round
if the memory percentage is lower than 75 then restart the round
if -AlwaysReboot is set to true then restart the round (i fixed that)
if any of these are false; reboot the server.

UnityProject/Assets/Scripts/Managers/GameManager.cs
```diff
@@ -754,18 +761,33 @@ public void RestartRound()
...
+			yield return WaitFor.Seconds(0.2f);
 
-		CustomNetworkManager.Instance.ServerChangeScene("OnlineScene");
+			CustomNetworkManager.Instance.ServerChangeScene("OnlineScene");
 
-		StopAllCoroutines();
+			StopAllCoroutines();
+			yield break;
+		}
+		Logger.Log("Server is rebooting now. If you don't have a way to automatically restart the " +
```
Bod9001: make this a log error so we can see it on Discord

MaxIsJoe: Done

Bod9001:
my question is is the error log,  centre medially or is it done in an update loop,
since if it's done in the update loop You might need to Trigger it manually?
since application quit would cancel the update loop

Bod9001:
actually yield return WaitFor.Seconds(2f); Should account for it

Bod9001 (approved):
is there anyway to detect if the restart process is running?

MaxIsJoe:
> is there anyway to detect if the restart process is running?

systemctl

MaxIsJoe:
> is there anyway to detect if the restart process is running?

systemctl

Bod9001:
let's wait for good staging build before merging, E.G interaction fix and right click fix

corp-0:
> is there anyway to detect if the restart process is running?

All this PR does is gracefully quitting the application, it doesn't restart the server. The server operators will have to configure their thing to automatically restart on quit (afaik default behaviour of systemctl)

Bod9001:
humm, You think it's safe to merge?

MaxIsJoe:
> humm, You think it's safe to merge?

It should be safe as long as systemctl is up and running so when the server shutsdown it can bring it backup again.

Bod9001:
systemctl 
idk if stating has that but
 portainer has options for automatically restarting when it closes

corp-0 (approved):

PR #8019, opened 2021-12-23
better handling of IPs since connection is disposed of in mirror

Bod9001:
haven't been able to test it but it should work fine


corp-0:
I didn't understand how it is better, looks like a variable wrapping the ``.address`` field from Mirror.

corp-0:
ah nvm I got it now, it is caching the IP so we don't lose it when Mirror disposes the connection.

corp-0 (approved):

NoooneyDude (approved):

Issue #4658, opened 2020-07-21
Antags need better End Round goal detection


ThatDan123:
- [x] Currently Antags dont green text if the item they have is inside a box  in their inventory
- [x] Currently Antags dont green text if they are in a locker.
- [ ] Additionally a system for detecting being handcuffed on the shuttle so prisoners can fail the escape objective

NoooneyDude:
The box issue was fixed by PR #4797. Is the locker issue still a problem?

PR #8031, opened 2021-12-25
Fix for green text problem for antags hiding in closets

MaxIsJoe:
part of #4658

The idea of this fix is that the game should check if the the shuttle has a tile that the antag is currently on when the matrix check fails

CL: [Fix] fixed antagonists not getting greentext for objectives if they were hiding in closets.

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Antagonists/Objectives/EscapeObjectives/Escape.cs
+++ b/UnityProject/Assets/Scripts/Systems/Antagonists/Objectives/EscapeObjectives/Escape.cs
@@ -24,6 +24,12 @@ namespace Antagonists
 			ValidShuttles.Add(GameManager.Instance.PrimaryEscapeShuttle);
 		}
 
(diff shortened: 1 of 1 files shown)


corp-0:
Is the antag tile returning 0,0 or the closet tile?

MaxIsJoe:
> Is the antag tile returning 0,0 or the closet tile?

Matrix ID is mismatched when in closet so I assume that yeah it thinks that it's in 0,0
`antagTile.WorldPositionServer` should solve this because we're retrieving his position on the closet and checking if the shuttle doesn't have a tile in that position.

corp-0:
And did you test it? Because from what I remember WorldPositionServer is 0,0 when inside a closet. AssumedWorldPosition tho will return the closet pos

MaxIsJoe:
> And did you test it? Because from what I remember WorldPositionServer is 0,0 when inside a closet. AssumedWorldPosition tho will return the closet pos

Tested now and pushed a fix

Bod9001 (approved):

ThatDan123 (approved):

PR #8039, opened 2021-12-27
Fixes lockers leaking out when changing matrices

Bod9001:
~~small fix needed It spawns the contents in when changing matrices~~
Shouldn't spawn initial contents when changing matrices  now

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/ObjectContainer.cs
+++ b/UnityProject/Assets/Scripts/Objects/ObjectContainer.cs
@@ -154,7 +154,7 @@ namespace Objects
 		{
 			if (initialContentsSpawned == false)
 			{
-				TrySpawnInitialContents();
+				TrySpawnInitialContents(true);
 			}
 
 			foreach (var obj in storedObjects.Keys)
--- a/UnityProject/Assets/Scripts/Objects/ObjectContainer.cs
+++ b/UnityProject/Assets/Scripts/Objects/ObjectContainer.cs
@@ -150,9 +150,9 @@ namespace Objects
 			}
 		}
 
-		public IEnumerable<GameObject> GetStoredObjects(bool OnlyInstantiated = false)
+		public IEnumerable<GameObject> GetStoredObjects(bool onlyInstantiated = false)
 		{
-			if (initialContentsSpawned == false && OnlyInstantiated == false)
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Objects/ObjectContainer.cs
```diff
@@ -150,11 +150,11 @@ public void GatherObjects()
 			}
 		}
 
-		public IEnumerable<GameObject> GetStoredObjects()
+		public IEnumerable<GameObject> GetStoredObjects(bool OnlyInstantiated = false)
```
corp-0: All arguments variable names should be camelCase

corp-0:
The intention of the bool argument in all of those 3 method isn't immediately clear

corp-0:
Why did this start happening? I'm pretty sure it wasn't the case since ever

Bod9001:
well, the way it's triggered is pretty self-explanatory,  when changing mattresses, it loops through all the contents and sets the matrix, however due to the default bool ,  if it hasn't been spawned in yet the contentsit instantiates it, and doesn't make it hidden,  I improved it more by make it so you can specify to only include stuff that is spawned inside the locker

UnityProject/Assets/Scripts/Objects/ObjectContainer.cs
NoooneyDude:
```suggestion
		public IEnumerable<GameObject> GetStoredObjects(bool onlyInstantiated = false)
```

UnityProject/Assets/Scripts/Objects/ObjectContainer.cs
```diff
@@ -150,11 +150,11 @@ public void GatherObjects()
...
 		{
-			if (initialContentsSpawned == false)
+			if (initialContentsSpawned == false && OnlyInstantiated == false)
```
NoooneyDude:
```suggestion
			if (initialContentsSpawned == false && onlyInstantiated == false)
```

NoooneyDude (approved):

ThatDan123 (approved):

ThatDan123:
> Why did this start happening? I'm pretty sure it wasn't the case since ever

Guess not many people pushed lockers which hadn't already been opened before onto a different matrix. Rare bug.

PR #8049, opened 2021-12-29
Fix warnings

NoooneyDude:
- Fixes some compiler warnings.
- Namespaces a few classes.


UnityProject/Assets/Scripts/Effects/CustomEffectBehaviour.cs
```diff
@@ -2,7 +2,7 @@
 
 public class CustomEffectBehaviour : MonoBehaviour
 {
-	public ParticleSystem particleSystem;
+	public new ParticleSystem particleSystem;
```
corp-0: why new?

NoooneyDude: MonoBehaviour already has a `particleSystem` member. So we just ignore it and use our (assumed serialized) one instead. I could just rename the variable.

corp-0: Renaming is probably the best solution

ThatDan123 (dismissed):

corp-0:
merge conflict :c

corp-0 (approved):

ThatDan123 (approved):

Issue #5000, opened 2020-08-30
Make dead bodies give off Miasma / give disease


ThatDan123:
Now that miasma gas is in the game, dead bodies should give it off after they have been killed for a while.

And now with the basic sickness system miasma could give players a disease. See #4776 #4777 #4778 

- [x] make bodies give off Miasma.
- [ ] make bodies/miasma give diseases.

AtnerNT:
This was implemented as on #10919

PR #8056, opened 2021-12-29
Bodies Produce Miasma

ThatDan123:
-Bodies produce miasma, 0.02 per second after 2 mins

CL: [Improvement] Mobs which are on fire will now continue to burn after death and eventually go out.

-Made all atmos defines consts, easier to see value now when hovering over in code editior

CL: [New] Bodies older than 2 minutes produce miasma unless they are below -10°C or are in a container

Fixes some of #5000


NoooneyDude (approved):

corp-0:
does Miasma do anything? I didn't see any code for its effect so I think the issue shouldn't be closed yet

corp-0 (approved):

PR #8061, opened 2021-12-30
AI No need to be done on client

Bod9001:


NoooneyDude:
This is a duplicate of #8051 : )

PR #8084, opened 2022-01-03
chemistry reaction optimisation

Bod9001:
~~I still need to do some proper profiling, and doublechecking~~
so, how it works, instead of checking every single reaction, to see if it works

 it now looks for every chemical and checks and adds up all the reactions possible with that chemical, And then loops through those ,
 
### Changelog:

CL: [Improvement] improved chemistry performance with better reaction system
CL: [New] added tests for reagent index and reagent SO


NoooneyDude:
Tests complaining

corp-0:
solve the codacy issues and we gucci

ThatDan123 (approved):

corp-0 (approved):

PR #8068, opened 2021-12-31
Namespace NetUI

NoooneyDude:
- Namespaces Net UI.
- Modifies `Hacking` namespace.

PR #8090, opened 2022-01-03
Revert "Namespace NetUI" sadly namespace changes broke all the buttons on the UI

Bod9001:
Reverts unitystation/unitystation#8068

due to how Unity handles it it has a direct reference to the namespace, in unity actions, so if you change the namespace it breaks The reference


Bod9001:
image
example of the changes needed to fix the reference

ThatDan123 (approved):

corp-0:
Nooney didn't change that in his pr?

corp-0 (approved):

PR #8093, opened 2022-01-03
GetAs fix

ThatDan123:
`where T : RegisterTile` instead

https://github.com/unitystation/unitystation/pull/7621#issuecomment-948771485


corp-0:
I don't get it, didn't Bod say it wasn't compiling like that? I suggested using that very same where restriction

ThatDan123:
> I don't get it, didn't Bod say it wasn't compiling like that? I suggested using that very same where restriction

Bod forgot to change it in both scripts, so it failed to compile

corp-0 (approved):

Bod9001 (approved):

PR #8104, opened 2022-01-05
Caching optimisations on player Sync

Bod9001:
instead of calculating whatmatrix the player is on about a billion times, it should cash it now and reuse it,  
need to double check since target and origin might have got mixed up since from late-night coding, 
need to test with headless but tested with editor

Now tested in headless with client,Works

### DIFF
--- a/UnityProject/Assets/Scripts/Player/PlayerSync.Server.cs
+++ b/UnityProject/Assets/Scripts/Player/PlayerSync.Server.cs
@@ -525,7 +525,7 @@ public partial class PlayerSync
 			if (serverBump == BumpType.Push || serverBump == BumpType.Blocked)
 			{
 				var worldTarget = state.WorldPosition.RoundToInt() + (Vector3Int)action.Direction();
-				var swapee = MatrixManager.GetAs<RegisterPlayer>(worldTarget, true);
+				var swapee = MatrixManager.GetAs<RegisterPlayer>(worldTarget, true) as List<RegisterPlayer>;
 				if (swapee != null && swapee.Count > 0)
 				{
 					swapee[0].PlayerScript.PlayerSync.RollbackPosition();
--- a/UnityProject/Assets/Scripts/Player/PlayerSync.Server.cs
+++ b/UnityProject/Assets/Scripts/Player/PlayerSync.Server.cs
@@ -471,7 +471,7 @@ public partial class PlayerSync
 			TryUpdateServerTarget();
 			return;
 		}
-
+		
 		var newPos = nextState.WorldPosition;
 		var oldPos = serverState.WorldPosition;
 		lastDirectionServer = Vector2Int.RoundToInt(newPos - oldPos);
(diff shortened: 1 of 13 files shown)


UnityProject/Assets/Scripts/Player/PlayerSync.Server.cs
```diff
@@ -787,11 +787,14 @@ private void InteractDoor(Vector3Int currentPos, Vector3Int targetPos)
 	private void InteractEnterable(Vector3Int targetPos)
 	{
 		//Object IPlayerEntersTile
-		List<IPlayerEntersTile> enterables = MatrixManager.GetAt<IPlayerEntersTile>(targetPos, isServer);
-		foreach (IPlayerEntersTile enterable in enterables)
+		var registerTiles = MatrixManager.GetRegisterTiles(targetPos, isServer);
```
ThatDan123, line 783: This change will break the interface as it will be on more than just registertiles?

ThatDan123: Oh nvm, you cached it.

corp-0 (dismissed):

corp-0 (dismissed):

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/Matrix.cs
```diff
@@ -267,17 +267,17 @@ public bool IsNoGravityAt(Vector3Int position, bool isServer)
...
 	}
 
 	//Has to inherit from register tile
-	public IEnumerable<T> GetAs<T>(Vector3Int localPosition, bool isServer) where T : RegisterTile
+	public List<T> GetAs<T>(Vector3Int localPosition, bool isServer) where T : RegisterTile
 	{
 		if (!(isServer ? ServerObjects : ClientObjects).HasObjects(localPosition))
 		{
-			return Enumerable.Empty<T>(); //?
+			return new List<T>(); //?
```
ThatDan123: I think `Enumerable.Empty<T>` doesnt GC, while new `List<T>` does.

https://stackoverflow.com/a/34786905

Bod9001: what solution would be best, since there are a few places that do to list

ThatDan123: > what solution would be best, since there are a few places that do to list

You haven't removed any ToLists?

Bod9001: well, there's places that return lists instead of  Enumerable, so they have to do to list

ThatDan123 (approved):

corp-0 (approved):

NoooneyDude (approved):

PR #8123, opened 2022-01-06
Delete Time Events Bool

ThatDan123:
-Adds bool for deleting time event objects when it is not their time (Bye Christmas tree)

### DIFF
--- a/UnityProject/Assets/Scripts/ScriptableObjects/TimedGameEvents/TimedGameEventSO.cs
+++ b/UnityProject/Assets/Scripts/ScriptableObjects/TimedGameEvents/TimedGameEventSO.cs
@@ -15,6 +15,9 @@ namespace ScriptableObjects.TimedGameEvents
 		public string EventName;
 		[NaughtyAttributes.ResizableTextArea] public string EventDesc;
 		public Sprite EventIcon;
+
+		//Delete the object when the timed event is not happening
+		public bool deleteWhenNotTime = false;
 	}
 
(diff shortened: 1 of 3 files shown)


Bod9001 (dismissed):

UnityProject/Assets/Scripts/ScriptableObjects/TimedGameEvents/TimedGameEventSO.cs
```diff
@@ -12,9 +12,15 @@ public class TimedGameEventSO : ScriptableObject
...
+
 		public string EventName;
-		[NaughtyAttributes.ResizableTextArea] public string EventDesc;
+		[TextArea(10, 20)]
+		public string EventDesc;
+		
 		public Sprite EventIcon;
+
+		//Delete the object when the timed event is not happening
+		public bool deleteWhenNotTime = false;
```
corp-0: I don't understand this, it shouldn't spawn when the even isn't happening

ThatDan123: It's a mapped thing.

corp-0: Uhmm, then I think the default behaviour should be true

Bod9001 (approved):

NoooneyDude (approved):

PR #8136, opened 2022-01-09
Updates Project to Unity 2021.1.28f

MaxIsJoe:
Updating to Unity .28f to avoid crashes when switching branches
Removed un-needed packages that were not used
Updated some packages
Removed the "Deconstruct" component from light fixtures as there _**is**_ deconstruction logic already on the light fixture script that needs to be fixed instead.


corp-0:
it has been running tests for 3 hours, we probably need to update tests as well. Anyways, I'm thinking on putting this on hold until the devlog is released just in case we face any weird bugs from upgrading.

Bod9001:
yeah, tho would be nice to separate the light fixture deconstruction in separate PR

corp-0:
Lol I crashed twice just now doing my latest PR, I think we should give this one some priority

Bod9001:
need to get tests working

PR #8139, opened 2022-01-09
Fixing issue #7874

### LINKED ISSUE #7874: Countdown require ready player
Make the countdown require at least one ready player.

(Need message saying more ready players needed for countdown if not enough)

-Bonus if its a server config option e.g `RequiredPlayersForCountdown : 0` this allows for server to set what player number they want

404-P4rziv4L:
### Purpose

Fixing #7874

CL: [Improvement] the round start countdown won't start until at least 1 player has readied.

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/GameManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/GameManager.cs
@@ -380,9 +380,6 @@ public partial class GameManager : MonoBehaviour, IInitialise
 				{
 					StartRound();
 				}
-				else {
-					Logger.Log("More ready players needed for countdown!")
-				}
 			}
 		}
 		else if (counting)
--- a/UnityProject/Assets/Scripts/Managers/GameManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/GameManager.cs
@@ -30,6 +30,11 @@ public partial class GameManager : MonoBehaviour, IInitialise
 	public int MinPlayersForCountdown { get; set; } = 1;
 
 	/// <summary>
+	/// The minimum number of ready players needed to start the pre-round countdown
+	/// </summary>
+	public int MinReadyPlayersForCountdown { get; set; } = 1;
+
+	/// <summary>
 	/// How long the pre-round stage should last
 	/// </summary>
 	public float PreRoundTime { get; set; } = 120f;
@@ -370,7 +375,13 @@ public partial class GameManager : MonoBehaviour, IInitialise
 		{
 			if (NetworkTime.time >= CountdownEndTime)
 			{
-				StartRound();
+				if (PlayerList.Instance.ReadyPlayers.Count >= MinReadyPlayersForCountdown)
+				{
+					StartRound();
+				}
+				else {
+					Logger.Log("More ready players needed for countdown!")
+				}
 			}
 		}
 		else if (counting)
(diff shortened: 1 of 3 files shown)


Bod9001:
You need to add the references in LoadConfig and 
GameConfig class

Bod9001:
that's easy enough to do

Bod9001:
humm, this might be worth adding the entry to the config so the one that ships with the game has the entry in, but otherwise all good

404-P4rziv4L:
> humm, this might be worth adding the entry to the config so the one that ships with the game has the entry in, but otherwise all good

Which config again? @Bod9001

Bod9001:
gameConfig.json in UnityProject\Assets\StreamingAssets\config

Bod9001 (dismissed):

NoooneyDude:
~~What triggers the round to start once enough players have joined?~~

It's in an UpdateMe : )

NoooneyDude (dismissed):

404-P4rziv4L:
@NoooneyDude I think unless the game starts, it would keep checking for the number of ready players present.

NoooneyDude:
Oh, just noticed tests failing. Could you take a look?

There's also a UX issue where a server has set this new option greater than 1 and there's not enough players, they'll wonder why the round isn't starting once the timer ends, but it's probably not too big of a deal for now.

UnityProject/Assets/Scripts/Managers/GameManager.cs
```diff
@@ -370,7 +376,13 @@ private void UpdateMe()
...
 			if (NetworkTime.time >= CountdownEndTime)
 			{
-				StartRound();
+				if (PlayerList.Instance.ReadyPlayers.Count >= MinReadyPlayersForCountdown)
+				{
+					StartRound();
+				}
+				else {
+					Logger.Log("More ready players needed for countdown!")
+				}
```
NoooneyDude: Found the tests failing issue : )
Missing semicolon.

```suggestion
```

I've suggested to just remove the log instead of fixing it as it is somewhat expected behaviour and could add quite a few entries to the log.

404-P4rziv4L:
> Oh, just noticed tests failing. Could you take a look?
> 
> There's also a UX issue where a server has set this new option greater than 1 and there's not enough players, they'll wonder why the round isn't starting once the timer ends, but it's probably not too big of a deal for now.

What could we do for the UX issue? Or is that a separate issue in itself?

NoooneyDude:
> What could we do for the UX issue? Or is that a separate issue in itself?

I think showing a fraction `Players: 2 / 3` (two online out of three required) somewhere on that lobby window would be sufficient. I would personally only show it if the requirement isn't yet met as seeing, as an example, `Players: 84 / 1` isn't particularly useful.

Edit: if you don't feel like implementing this for this PR, could you make an issue for it once this is merged?

Bod9001 (approved):

NoooneyDude (approved):

NoooneyDude:
Let us know in Discord if you want the Engineer role.

404-P4rziv4L:
Sure, I'd like working on the UX issue as well

mmails:
> Sure, but I don't have access to the Discord server yet...

Here's the [discord server](https://discord.com/invite/tFcTpBp), if you're interested

404-P4rziv4L:
Thanks, I'm in.

mmails:
> Thanks, I'm in.

My pleasure

PR #8141, opened 2022-01-09
Batons and Prods now require powercells, Fixes #5720

### LINKED ISSUE #5720: Stun baton should need power cells in order to stun people.
### Description
In /tg/, stun batons start with high-capacity power cells inside of them. Every time you hit someone with a turned-on stun baton, you use up 1000 watts from the power cell. High-capacity power cells come with 10,000 watts, so it takes ten hits with a turned-on baton to completely deplete the cell by default. Obviously, if a baton runs out of charge for its battery it should automatically turn off.

Since batons have depletable charges, you can put batons with cells inside weapon rechargers to recharge said cells similar to energy guns. 

Additionally, you can use a screwdriver on a stun baton to remove the power cell inside. This changes the stun baton's sprite and lets you put a new power cell inside. You can put cells with higher capacities to give stun batons more charges. Note: if a cell's capacity is less than the amount of charge that would be removed by a baton upon stunning someone, the cell cannot be inserted into a baton.

Finally, examining a stun baton tells you it the charge remaining in the cell in percentage form, similar to energy guns.

AtnerNT:
### Purpose
Like the title says, stun batons, stun prods and teleprods now will require power cells to power them, giving them a limited amount of charges before they must be replenished at a charging station.
Stun prods and teleprods also no longer require a powercell to craft, but one must be installed after the fact.
Power cells can be taken out and exchanged using a screw driver.
Power cells whose capacity is too small will not be able to be installed into the device.
At base level without modifying the power cell, stun batons can go 10 uses before running out, and the stun prod and teleprod can go 5 (Same capacity but twice the usage)

Fixes #5720 

CL: [Balance] stun batons and prods now need power cells with charge to function.


AtnerNT:
Dammit accidentally got the packages file in there

AtnerNT:
There we go that should be good now

AtnerNT:
Okay I forgot to check if there were any assignees, @Pottingmix Are you still planning on working on this? if so I will close this PR and let you do your thing.

Bod9001:
the batteries are compatible with the Charger?

PetMudstone:
Two suggestions that I forgot to bring up when we talked about this prior:
- Remove the power cell from the stun prod and teleprod recipes. They should provide stun prods and teleprods that don't come with cells instead. This will allow people to install cells with higher capacities than high capacity cells (10,000w) without having to remove the hi-cap cell first. Additionally, due to how recipe inheritance works having the high capacity cell be needed means that people won't be able to craft with cells that do have sufficient capacities like bluespace cells.
- Stun batons, stun prods, and teleprods should reject cells whose capacity is less than the charge consumed upon hitting someone. Stun prods and teleprods consume 2,000w per hit, so they will reject normal power cells which only hold 1,000w. Proper stun batons consume 1,000w, so they will accept normal cells.

Text shown on upon trying to install a normal cell in a teleprod.
image

Edit: Suggestions applied, thanks!

UnityProject/Assets/Scripts/Core/Editor/Items/MeleeEffectEditor.cs
```diff
@@ -1,5 +1,6 @@
 using UnityEditor;
 using Weapons;
+using UnityEngine;
```
corp-0: Uhm I wonder why did we do a custom inspector for this

AtnerNT: MeleeEffect has quite a few fields that clog up the inspector, and not all of them are used all the time. So I made a custom inspector to just hide fields when they aren't needed so the inspector isn't as crowded with values.

AtnerNT: (i.e teleport config options are hidden if the weapon cannot teleport)

AtnerNT:
> the batteries are compatible with the Charger?

and yeah, as long as the item has an internal battery component, internal battery trait and an item storage to take from it can charge the item just fine

UnityProject/Assets/Scripts/Core/Editor/Items/MeleeEffectEditor.cs
NoooneyDude: Might be before `NaughtyAttributes` was added.

AtnerNT: It was not- however I only learnt that we had naughty attributes on this project then- Ill make the change now

Bod9001 (dismissed):

UnityProject/Assets/Scripts/Items/Weapons/Melee/MeleeEffect.cs
```diff
@@ -9,30 +10,40 @@ namespace Weapons
...
 
-		[HideInInspector]
+		[ShowIf("canTeleport")]
 		public bool avoidSpace;
-		[HideInInspector]
+		[ShowIf("canTeleport")]
 		public bool avoidImpassable = true;
 
-		[HideInInspector]
+		[ShowIf("canTeleport")]
```
corp-0: change all the ``ShowIf`` to use ``nameof()`` instead of a string so if the field name changes in the future, the code doesn't silently stop working

corp-0: in this case it would be ``ShowIf(nameof(canTeleport))``

corp-0 (approved):

Bod9001 (approved):

PR #8144, opened 2022-01-09
Scene APC Fixes and Some Directional Fixes

ThatDan123:
-There was a major problem with APC connections which has been fixed on all scenes and a test added to catch it. 
(Related APC being set on ApcPoweredDevice but the device wasnt in the APC list meaning it wasnt connected and other tests didnt catch)

-Tweaked the Directional script and reset all scene directionals (This should have now fixed most if not all issues)

-All scenes changed (Scenes Commit only contains scenes no code change)


NoooneyDude:
I'm not sure why, but even after 8 hours I can still only load four of the changed files on the files changed tab.

Bod9001 (dismissed):
I just had to change the filter a little bit to get it to see

ThatDan123:
All good now

Bod9001 (approved):

corp-0 (approved):

PR #8149, opened 2022-01-11
Quick fix for prod cells

AtnerNT:
Something that slipped under the radar when I was testing the PR yesterday that I discovered today on staging. Clients can't put cells into prods or batons- so switching to Inventory.ServerTransfer instead of Inventory.ClientRequestTransfer, this seems to fix the issue.

CL: [Fix] fixed an issue preventing power cells from being placed inside stun prods and batons.


corp-0 (approved):

corp-0:
oh yeah, our bad. We should have found it was a calling a client method inside a server method

ThatDan123 (approved):

PR #8152, opened 2022-01-12
Fixes re-logging on rotated matrices items being displaced

Bod9001:
yeah, 
I've noticed a few wishes but I'm not sure if they're in developed or caused by my changes

Issues + items like banana pie when you eat it and the plate, appears in your inventory if you drop that, it lerps over to you position, this is because it's technically never told to be hidden because it's getting confused between localPOS = hiddenPOS and worldPOS = hiddenPOS ( I think this is in the main repository and it needs to be standardised so it only checks hiddenPOS  for worldPOS )

CL: [Fix] fixed an issue with items appearing in the wrong place when rejoining a round.

### DIFF
--- a/UnityProject/Assets/Mirror/Runtime/NetworkClient.cs
+++ b/UnityProject/Assets/Mirror/Runtime/NetworkClient.cs
@@ -988,7 +988,7 @@ namespace Mirror
             ///CUSTOM UNITYSTATION CODE///
             //Mirror specifically says that they dont support NetworkIdentities nested in normal gameobjects.
             //But, we do it, so we have to override this and make sure we read and send world pos
-            identity.transform.position = message.position;
+            identity.transform.localPosition = message.position;
             ///CUSTOM UNITYSTATION CODE///
 
             if (message.isLocalPlayer)
@@ -1061,6 +1061,10 @@ namespace Mirror
             if (spawnHandlers.TryGetValue(message.assetId, out SpawnHandlerDelegate handler))
             {
                 GameObject obj = handler(message);
+                ///UNITYSTATION CODE///
+                ///so, When the client logs in, The objects that are part of the map get the correct local position, If they've moved
+                obj.transform.localPosition = message.position;
+
                 if (obj == null)
                 {
                     Debug.LogError($"Spawn Handler returned null, Handler assetId '{message.assetId}'");
--- a/UnityProject/Assets/Mirror/Runtime/NetworkClient.cs
+++ b/UnityProject/Assets/Mirror/Runtime/NetworkClient.cs
@@ -987,7 +987,7 @@ namespace Mirror
 
             ///CUSTOM UNITYSTATION CODE///
             //Mirror specifically says that they dont support NetworkIdentities nested in normal gameobjects.
-            //But, we do it, so we have to override this and make sure we read and send world pos
+            //But, we do it, so we have to override this and make sure we read and send local pos
             identity.transform.localPosition = message.position;
             ///CUSTOM UNITYSTATION CODE///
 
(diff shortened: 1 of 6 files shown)


corp-0 (dismissed):

UnityProject/Assets/Mirror/Runtime/NetworkClient.cs
```diff
@@ -988,7 +988,7 @@ internal static void ApplySpawnPayload(NetworkIdentity identity, SpawnMessage me
             ///CUSTOM UNITYSTATION CODE///
             //Mirror specifically says that they dont support NetworkIdentities nested in normal gameobjects.
             //But, we do it, so we have to override this and make sure we read and send world pos
```
ThatDan123:
```suggestion
            //But, we do it, so we have to override this and make sure we read and send local pos
```

Does it need to be that now?

UnityProject/Assets/Mirror/Runtime/NetworkServer.cs
```diff
@@ -938,7 +938,7 @@ internal static void SendSpawnMessage(NetworkIdentity identity, NetworkConnectio
                 ///CUSTOM UNITYSTATION CODE///
                 //Mirror specifically says that they dont support NetworkIdentities nested in normal gameobjects.
                 //But, we do it, so we have to override this and make sure we read and send world pos
```
ThatDan123:
```suggestion
                //But, we do it, so we have to override this and make sure we read and send local pos
``` 
and here

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
```diff
@@ -80,6 +80,8 @@ private static TileLocation GetPooledTile()
 		public BetterBounds? GlobalCachedBounds;
 
 		[NonSerialized] public Matrix4x4 localToWorldMatrix = Matrix4x4.identity;
+		[NonSerialized] public Matrix4x4 worldToLocalMatrix = Matrix4x4.identity;
```
NoooneyDude, line 83: Not used anywhere?

UnityProject/Assets/Scripts/Core/Transform/CustomNetTransform.cs
```diff
@@ -175,21 +175,43 @@ public void SetInitialPositionStates()
...
-		serverLerpState.WorldPosition = pos;
+			predictedState.MatrixId = matrixInfo.Id;
+			predictedState.WorldPosition = pos;
+			serverState.MatrixId = matrixInfo.Id;
+			serverState.WorldPosition = pos;
+			clientState.MatrixId = matrixInfo.Id;
+			clientState.WorldPosition = pos;
+			serverLerpState.MatrixId = matrixInfo.Id;
+			serverLerpState.WorldPosition = pos;
+		}
```
NoooneyDude: Most of these statements are the same in both branches. Could you move the common ones out of the branching? Just makes it a lil easier to read.

Actually, if local and world are the same because the object had no parent, then wouldn't setting `LocalPosition = pos` be the same as `WorldPosition = pos`, so no branching needed? Can just use `WorldPosition = pos` for both cases?

Bod9001: well, it doesn't update it when it changes parent that's the issue

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
Bod9001: useful in future, Since its thread safe and theoretically faster than the current method, tho I did try using it but it hadn't been set by The time stuff started using it

NoooneyDude (approved):

ThatDan123 (approved):

PR #8154, opened 2022-01-13
refactor: separate server public and secret information into 2 different config files

corp-0:
### Purpose
Because of the way we build the docker images for our servers, having both information meant to be public and secret in the same file becomes a security issue as discussed in #conference-room channel in Discord.

Separating both files means we can mount the secrets to the container so they are never set in the docker image.

### Notes
I need to test it a little bit more to make sure I didn't break stuff and I might have to also update the build script.


corp-0:
for easier time reviewing, read the commits instead of all changes at once

NoooneyDude:
Looks good. Perhaps we should be committing example json files e.g. `secrets.json.example` with `secrets.json` gitignored?

I wonder if we could move these files closer to root, too, they're not easily discoverable at the moment.

corp-0:
> Looks good. Perhaps we should be committing example json files e.g. `secrets.json.example` with `secrets.json` gitignored?
> 
> I wonder if we could move these files closer to root, too, they're not easily discoverable at the moment.

Uhmm, Im already committing a file as example but it is called like the real deal

Bod9001:
what's this waiting on?

corp-0:
> what's this waiting on?

Manual tests to make sure I didn't break anything. Also merging this isn't too security now that Redline made me realise I wasn't fixing the issue with this and the first suspect now is the build script alone

Bod9001:
ok cool

corp-0:
found no issue trying locally with two builds

NoooneyDude (approved):

Bod9001 (approved):

Bod9001:
should the build script be updated before merging or after?

corp-0:
> should the build script be updated before merging or after?

Yes it should, thank god you guys didn't merge this yet lol

corp-0:
I have to re do this one, closing for now

PR #8162, opened 2022-01-15
Fixes Round start explosions

Bod9001:
the issue was that the reactor wasn't being removed from the update manager when the Round was ended

CL: [Fix] fixed an issue that was causing the reactor to endlessly explode after a rounded ended, even carrying over into the next round.

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Engineering/Reactor/ReactorGraphiteChamber.cs
+++ b/UnityProject/Assets/Scripts/Objects/Engineering/Reactor/ReactorGraphiteChamber.cs
@@ -77,11 +77,13 @@ namespace Objects.Engineering
 
 		public void OnEnable()
 		{
+			if (CustomNetworkManager.IsServer == false) return;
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Objects/Engineering/Reactor/ReactorGraphiteChamber.cs
```diff
@@ -76,11 +75,17 @@ private void Awake()
 			ReactorPipe = this.GetComponent<ReactorPipe>();
 		}
 
-		public void OnSpawnServer(SpawnInfo info)
+		public void OnEnable()
 		{
 			UpdateManager.Add(CycleUpdate, 1);
```
ThatDan123, line 81: Add check for IsServer

Bod9001: True

ThatDan123 (approved):

NoooneyDude (approved):

PR #8178, opened 2022-01-19
Adds Ministation, a new Mainstation for low-med pop

AtnerNT:
Nothing more to add other than what's in the title, this PR does not add the map into rotation as per request
Capture

CL: [New] Added Ministation, a low-pop map.


PerfectTangent:
very pog. might I recommend we don't immediately add it into the map pool until we get a bit of initial testing done.

AtnerNT:
> very pog. might I recommend we don't immediately add it into the map pool until we get a bit of initial testing done.

Easy fix, just removed it now

ThatDan123:
Theres some mapping tests which are failing. Go here, and download the artifact then open it and open the Index.html

https://github.com/unitystation/unitystation/actions/runs/1716990103

corp-0 (approved):
me like

ThatDan123 (approved):

PR #8189, opened 2022-01-22
Ministation fixes

AtnerNT:
A number of fixes/changes to ministation that were discovered recently

1) Replaces Chapel floor to something better looking
2) Fixes maint door in cargo not having cargo access
3) Adds lights to the emergency O2 room at Evac
4) Stops the Evac shuttle from ramming centcomm at mk 5
5) Adds Library and Xeno biology access to cap spare ID
6) Fixes some doors in science having the wrong access
7) More desk space for bar
8) Gives CE their action figure
9) Replaces implant cat ears with the correct clothing cat ears
10) Should fix gateway
11) Adds a thermomachine to atmos
12) Allows wire placing under decal lines in engi

That's all the issues I have found and/or have been alerted to by players, if you discover any issues specific to Ministation please @ me on discord @AtnerNT  so I can get round to them as quickly as possible.

CL: [Fix] fixed a number of issues with Ministation


PerfectTangent:
> Adds Library and Xeno biology access to cap spare ID

does it not have those accesses? If that's the case, then the  base Spare ID prefab needs to be modified, not the one in the map.

AtnerNT:
> does it not have those accesses? If that's the case, then the base Spare ID prefab needs to be modified, not the one in the map.

It does not- should I do the changes to the base ID on this pr? or another one?

PerfectTangent:
> It does not- should I do the changes to the base ID on this pr? or another one?

splitting up PRs is always a good practice. Question, though. since Gilles hasn't finished the clearance rework, there are two different systems for access in the game. Does this problem actually affect the ID? because it might be the case that those access types are not in the new clearance list, and so the ID should be able to open the doors just fine.

AtnerNT:
> splitting up PRs is always a good practice. Question, though. since Gilles hasn't finished the clearance rework, there are two different systems for access in the game. Does this problem actually affect the ID? because it might be the case that those access types are not in the new clearance list, and so the ID should be able to open the doors just fine.

It does, spare cap Id cannot enter xeno biology or the stations library, however these accesses are present on the scientists, RDs, curator and the captains main ID

PerfectTangent:
alright, then, go ahead.

ThatDan123 (approved):

Bod9001 (approved):

PR #8192, opened 2022-01-23
Disposals Rework [DNM]

### LINKED ISSUE #4654: Disposals Issue and features B:150

### LINKED ISSUE #6568: Disposal bins not initialising correctly
Disposal bins don't seem to be initialising correctly:
off
While the expected sprite state is:
ready
~~In addition, disposal bins spawned in without a disposal pipe terminal underneath are spawning as immediately above, and not rotated, as depicted here:
uninstalled
{spawning over a disposal pipe terminal is as expected)~~

dapocalypse:
### Purpose
Fixes #4654
Fixes #6568

This is a disposals rework.

Disposals bins are on at start.

### Notes:
Edit: Now disposals machines spawn as intended.


NoooneyDude:
There's a problem with SpriteHandler where using PushClear to set no sprites seems a bit buggy, as it thinks the sprite hasn't changed and so doesn't set the new one. I suspect this might be the case here. There's a sprite SO that you can use as a blank, so just switch to this SO in order to clear the sprite.

Might be worth a go?

UnityProject/Assets/Scripts/Objects/Disposals/DisposalBin.cs
```diff
@@ -151,17 +150,10 @@ protected override void UpdateSpriteConstructionState()
 
 		private void UpdateSpriteBinState()
 		{
-			if (MachineUnattached)
-			{
-				baseSpriteHandler.ChangeSprite((int) BinSprite.Sideways);
-				overlaysSpriteHandler.PushClear();
-				return;
-			}
-
```
NoooneyDude, line 160: Why remove this?

UnityProject/Assets/Scripts/Objects/Disposals/DisposalBin.cs
```diff
@@ -151,17 +150,10 @@ protected override void UpdateSpriteConstructionState()
...
 			switch (binState)
 			{
 				case BinState.Disconnected:
-					baseSpriteHandler.ChangeSprite((int) BinSprite.Upright);
```
NoooneyDude, line 164: Disconnected should be upright. Disconnected refers to the state where you first: turn off the power using the UI, then use a screwdriver to disconnect the power cable as per /tg/.

The sideways sprite is for when the disposal bin is unwrenched and moveable.

UnityProject/Assets/Scripts/Objects/Disposals/DisposalBin.cs
dapocalypse: It got stuck in the loop on roundstart. I can just add unconnected to the switch statement

dapocalypse:
> There's a problem with SpriteHandler where using PushClear to set no sprites seems a bit buggy, as it thinks the sprite hasn't changed and so doesn't set the new one. I suspect this might be the case here. There's a sprite SO that you can use as a blank, so just switch to this SO in order to clear the sprite.
> 
> Might be worth a go?

I don't think this is the issue. The ready state isn't using PushClear its changing the sprite.

dapocalypse:
Remove merge conflicts

PR #8194, opened 2022-01-23
Fix Infinite Headset Encryption Keys

### LINKED ISSUE #8186: Encription keys can be removed from a headset indefinitely
using a screw driver on any headset spawns a standrard encription key- but the key can be removed again and again, allowing you to create hundreds of encription keys

Capture

tomasalves8:
### Purpose

Fixed #8186
Encryption key is still using the old sprite system, this PR removes it temporary until we can port it to the new system. It also fixes the issue when putting the encryption keys back in to the headset, by removing unused HeadsetKey, and putting it's behavior inside of the EncryptionKey Class.

### Notes:

For future reference the sprites are in Assets/Textures/items/radio
### Changelog:

CL: [Fix] Infinite Headset Encryption Keys


UnityProject/Assets/Scripts/Items/EncryptionKey.cs
```diff
@@ -175,13 +149,12 @@ public EncryptionKeyType Type
...
 			}
-			UpdateSprite();
 		}
 	}
 
 	private void Start()
 	{
-		UpdateSprite();
+
 	}
```
ThatDan123: If this function is empty please remove it

UnityProject/Assets/Scripts/Items/EncryptionKey.cs
```diff
@@ -309,6 +208,19 @@ public void ServerPerformInteraction(MouseDrop interaction)
...
 		}
 	}
+
+	public bool Interact(InventoryApply interaction)
+	{
+		//insert the headset key if this is used on a headset
+		if (interaction.UsedObject == gameObject
+		    && interaction.TargetObject.GetComponent<Headset>() != null)
+		{
+			UpdateHeadsetKeyMessage.Send(interaction.TargetObject, gameObject);
```
ThatDan123, line 213: Is this the only place where `UpdateHeadsetKeyMessage` is called?

If it is and if you have time could you refactor this to use `ICheckedInteractable` and move the logic of `UpdateHeadsetKeyMessage` here, otherwise I'll make an issue to refactor it later.

tomasalves8: It is also used in the Screwdriver.cs (`Assets\Scripts\Items\Tool\Screwdriver.cs`) file to remove the encryption key from the headset.

Bod9001 (approved):

ThatDan123 (approved):

PR #8196, opened 2022-01-24
Fixes defib sound thread error

### LINKED ISSUE #8155: Defib Bug
```
UnityException: get_gameObject can only be called from the main thread.
Constructors and field initializers will be executed from the loading thread when loading a scene.
Don't use this function in the constructor or field initializers, instead move initialization code to the Awake or Start function.

Items.Medical.DefibrillatorPaddles+<Cooldown>d__14.MoveNext () (at/root/UnityProject/Assets/Scripts/Items/Medical/DefibrillatorPaddles.cs:106)
```

Seems like we're gonna need to remove the async (I havent looked at the code might be an easy fix)

MaxIsJoe:
functions run on other threads can't fetch gameobjects correctly
this fix changes the cooldown function to stop the sound manager from fetching the game object's position and instead the server will provide the position from the main thread to the function.

Fixes #8155

CL: [Fix] fix an error when defibrillators played their sfx.

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Medical/DefibrillatorPaddles.cs
+++ b/UnityProject/Assets/Scripts/Items/Medical/DefibrillatorPaddles.cs
@@ -70,23 +70,24 @@ namespace Items.Medical
 		void Perform()
 		{
 			var livingHealthMaster = interaction.TargetObject.GetComponent<LivingHealthMasterBase>();
+			var objectPos = gameObject.AssumedWorldPosServer();
 			if (CanDefibrillate(livingHealthMaster, interaction.Performer) == false)
 			{
-				_ = SoundManager.PlayNetworkedAtPosAsync(soundFailed, gameObject.AssumedWorldPosServer());
-				Cooldown();
+				_ = SoundManager.PlayNetworkedAtPosAsync(soundFailed, objectPos);
+				Cooldown(objectPos);
 				return;
 			}
 			livingHealthMaster.RestartHeart();
-			_ = SoundManager.PlayNetworkedAtPosAsync(soundZap, gameObject.AssumedWorldPosServer());
+			_ = SoundManager.PlayNetworkedAtPosAsync(soundZap, objectPos);
 			if (livingHealthMaster.IsDead == false)
 			{
 				livingHealthMaster.playerScript.ReturnGhostToBody();
-				_ = SoundManager.PlayNetworkedAtPosAsync(soundSuccsuess, gameObject.AssumedWorldPosServer());
-				Cooldown();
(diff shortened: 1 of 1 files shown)


tomasalves8 (dismissed):

UnityProject/Assets/Scripts/Items/Medical/DefibrillatorPaddles.cs
```diff
@@ -98,12 +99,12 @@ void Perform()
...
 	}
 
-	private async void Cooldown()
+	private async void Cooldown(Vector3 soundPos)
 	{
 		onCooldown = true;
 		await Task.Delay(cooldownMs).ConfigureAwait(false);
 		onCooldown = false;
-		SoundManager.PlayNetworkedAtPos(soundCharged, gameObject.AssumedWorldPosServer());
+		SoundManager.PlayNetworkedAtPos(soundCharged, soundPos);
```
Bod9001: wouldn't this be weird on a moving shuttle?

ThatDan123:
This might have the issue of the sound playing in the wrong place if the person moves before the 5 second cooldown is done?

Not sure whats the best way of fixing that, other than switching to a coroutine. 

I guess if the SoundManager has a method which only takes the object source would also fix it. Could be useful to add and would fix this issue.

UnityProject/Assets/Scripts/Items/Medical/DefibrillatorPaddles.cs
MaxIsJoe: You're right so i fixed this

MaxIsJoe:
> I guess if the SoundManager has a method which only takes the object source would also fix it. Could be useful to add and would fix this issue.

There is actually a method like this already but the issue was that threads didn't play nicely when just using `gameObject` because it would try fetching it off the wrong thread
I fixed it this time after what Bod pointed out by just giving the gameObject to the function as an argument

MaxIsJoe:
review again pls

ThatDan123 (dismissed):
Ah yeah, that also works

Bod9001 (approved):

Bod9001 (changes requested):
it's the users get component in AssumedWorldPos  as ThatDaniel has highlighted

MaxIsJoe:
> it's the users get component in AssumedWorldPos as ThatDaniel has highlighted

I changed that

Bod9001 (dismissed):

ThatDan123:
You don't need to send the gameObject as a parameter now

MaxIsJoe:
> You don't need to send the gameObject as a parameter now

done

Bod9001 (approved):

ThatDan123 (approved):

PR #8217, opened 2022-01-26
Firelock and Windoor Direction only

ThatDan123:
-Firelock and Windoor Direction only

-Cleaned up some of the rotation script

-Sprite will reset to initial variant not 0 on refresh if needed


UnityProject/Assets/Scripts/Objects/Directionals/Rotatable.cs
```diff
@@ -261,7 +261,7 @@ public void ResitOthers()
 				}
 				else
 				{
-					spriteHandler.ChangeSpriteVariant(0, false);
+					spriteHandler.ChangeSpriteVariant(spriteHandler.InitialVariantIndex, false);
```
Bod9001: initial variant index is too confusing

Bod9001: also I have changes here in the PR

ThatDan123: but thats what the default should be??

Bod9001: well, it only runs it if it's not being set Rotational

Bod9001: if (ChangeSprites == false)

corp-0: ChangeSprites sounds like a method, not a bool

ThatDan123: Done, back to 0

corp-0 (approved):

PR #8218, opened 2022-01-26
Adding/Removing Mentor by Admins

ThatDan123:
CL: [New] Admins can add/remove mentorship from players

-Can be temporary (until server is restarted) or permanent (adds to txt file)

-Cleaned up some scripts


corp-0:
why can the sender be null?

Bod9001 (approved):

ThatDan123:
> why can the sender be null?

It needs to be null, you never send it as a parameter. Mirror fills it in automatically

craftxbox (approved):

corp-0 (approved):

PR #8221, opened 2022-01-26
Fixes pipe directions

Bod9001:
inverts pipes so they are aligned properly, note if you've changed it to the new system you have to change it back
 
 Just flip the pipes  so they are inverted,  and added some code to get rid of scene overrides


ThatDan123 (approved):

Bod9001:
moved to #8222

PR #8206, opened 2022-01-25
Refreshes Directionals in Scenes

AtnerNT:
After the rotation pr some scenes needed their directionals refreshing, this just refreshes all the scenes and commits any changes that were made which should help with rotation related bugs

PR #8222, opened 2022-01-26
 #8221 but Test directional changes but includes  #8206

Bod9001:
#8221 + some extra fixes


ThatDan123:
Only scenes left to do are 

- [x] Unreal
- [x] Box
- [x] Test station

Just need to go through and check that the atmos stuff and lights is facing the correct way

-It seems like if they were pointing down before they got switched to pointing up (though it does happen with other direction but rarely)

-I made a tool (under mapping tab) that you use when in playmode to detect the monopipes that arent connected. (Seems the shuttle used to be self sufficient and isnt anymore so that is detected too)

ThatDan123 (dismissed):

ThatDan123 (approved):

ThatDan123:
Ready

ThatDan123 (dismissed):

ThatDan123 (dismissed):

ThatDan123 (approved):

corp-0 (approved):

PR #8228, opened 2022-01-27
Ashwalker Spawn When Lavaland QP First Used

ThatDan123:
CL: [Improvement] Ashwalker ghost role will now only appear when lavaland has been visited

-Allow for no centcom scene

-In the future this might be a good way to trigger the mob spawn as well?

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/EventManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/EventManager.cs
@@ -30,7 +30,8 @@ public enum Event
 	PreRoundStarted,
 	MatrixManagerInit,
 	BlobSpawned,
-	ScenesLoadedServer
+	ScenesLoadedServer,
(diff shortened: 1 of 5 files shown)


UnityProject/Assets/Scripts/Managers/EventManager.cs
```diff
@@ -30,7 +30,8 @@ public enum Event
 	PreRoundStarted,
 	MatrixManagerInit,
 	BlobSpawned,
-	ScenesLoadedServer
+	ScenesLoadedServer,
+	LavalandFirstEntered
```
Bod9001, line 34: would be nice if it wasn't hardcoded but I can't think of any way where it wouldn't be?

Bod9001 (approved):

corp-0 (approved):

PR #8249, opened 2022-01-29
Pump and Flipped Mixer Fix

ThatDan123:
-Flipped mixer had no input set for a connection

-Pumps were working backwards so i flipped the connection

-Idcard jobTitle is still NREing, and so must be being set null somewhere

Just building now


corp-0:
Why?

Bod9001 (dismissed):

Bod9001 (dismissed):

corp-0 (dismissed):

ThatDan123:
Freezers connections don't move correctly as its not the parent which is rotated. 

I tried to get the rotation from the current direction to set the pipes up but that didnt work right. Its a minor issue so its ok for the playtest.

Just for reference the issue is in HeaterFreezer line 286

Bod9001 (approved):

corp-0 (approved):

PR #8250, opened 2022-01-29
last-minute pipe fixes

Bod9001:
last-minute pipe fixes, need to do build to check out


UnityProject/Assets/Scripts/Objects/Directionals/Rotatable.cs
```diff
@@ -51,6 +51,12 @@ public enum RotationMethod
 
 	private void SetDirection(OrientationEnum dir)
 	{
+		if (isServer == false && isLocalPlayer)
+		{
+			CmdChangeDirection(dir);
+			return;
```
ThatDan123, line 57:
```suggestion
```

In the original Directional this actually doesnt have the return. Think it'll be fine to remove as then it's predicted and the client doesnt have to wait for the server to change when moving.

PR #8283, opened 2022-01-30
Rotatable Mapping Fix and Performance

ThatDan123:
-Was calling stuff which was null outside play mode for the mapping issue.

-I think we have a bug in mirror where the SyncHooks are being called on Headless, so it was invoking the event twice

-Stopped headless changing sprites for players as they are not networked.

-Cleaned up the listeners to the rotation event

image

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Disposals/DisposalIntake.cs
+++ b/UnityProject/Assets/Scripts/Objects/Disposals/DisposalIntake.cs
@@ -30,13 +30,16 @@ namespace Objects.Disposals
 		protected override void Awake()
 		{
 			base.Awake();
-
-			if (TryGetComponent<Rotatable>(out var directional))
-			{
-				directional.OnRotationChange.AddListener(OnDirectionChanged);
-			}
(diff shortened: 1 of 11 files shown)


Bod9001 (dismissed):

UnityProject/Assets/Scripts/Objects/Disposals/DisposalIntake.cs
```diff
@@ -162,6 +165,11 @@ protected override void SetMachineInstalled()
 		private void OnDisable()
 		{
 			UpdateManager.Remove(CallbackType.UPDATE, UpdateMe);
+
+			if (TryGetComponent<Rotatable>(out var rotatable))
+			{
+				rotatable.OnRotationChange.AddListener(OnDirectionChanged);
+			}
```
corp-0, line 172: Shouldn't this be RemoveListener?

ThatDan123: yup, fixed

corp-0 (approved):

NoooneyDude (approved):

PR #8290, opened 2022-01-30
Rearrange NetworkServer Broadcast

ThatDan123:
-Old mirror code was loop through all connections and then identities which is expensive

-Because all players currently observe everything we can just loop through all identities and then all connections

-Made some enums bytes
	        
Needs testing

### DIFF
--- a/UnityProject/Assets/Mirror/Runtime/NetworkServer.cs
+++ b/UnityProject/Assets/Mirror/Runtime/NetworkServer.cs
@@ -1529,10 +1529,7 @@ namespace Mirror
             foreach (var identityPair in NetworkIdentity.spawned)
             {
 	            var identity = identityPair.Value;
-	            if (identity.isDirty == false)
-	            {
-		            continue;
-	            }
+	            if (identity.isDirty == false || identity.observers == null) continue;
 
 	            for (int j = 0; j < connectionsCount; j++)
 	            {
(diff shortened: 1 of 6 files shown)


Bod9001:
only issue is when client is loading They are not observing all the scenes?

ThatDan123:
Ah yeah, that will be an issue

ThatDan123:
@Bod9001 i added a check for observers do you think this will be good enough?

UnityProject/Assets/Mirror/Runtime/NetworkServer.cs
```diff
@@ -1541,6 +1538,8 @@ static void BroadcastToConnections()
 		            //Connection disconnected
 		            if (connection.isReady == false) continue;
 
+		            if (identity.observers.ContainsKey(connection.connectionId) == false) continue;
```
Bod9001, line 1541: I think you could just shortcut this with a bool that gets Set once the client has loaded, tho When we do network bubbling We will require this

Bod9001:
Actually noting about the loop performance, imo it's probably better to have one list of Network behaviours so, the list optimisation I did only removes from one list Versus list for everything a player That is immensely slow

UnityProject/Assets/Mirror/Runtime/NetworkServer.cs
```diff
@@ -1520,63 +1520,40 @@ static void ClearSpawnedDirtyBits()
...
+
+		            //Connection disconnected
+		            if (connection.isReady == false) continue;
+
+		            if (identity.observers.ContainsKey(connection.connectionId) == false) continue;
+
+		            ////////////////////////////
+		            // get serialization for this entity viewed by this connection
+		            // (if anything was serialized this time)
+		            NetworkWriter serialization = GetEntitySerializationForConnection(identity, connection);
```
ThatDan123, line 1546: I'm wondering if a can cache this serialization and reuse it for all the connections, as this is what is mainly draining performance too

PR #8292, opened 2022-01-30
Linecast Optimisation

ThatDan123:
-Was doing unnecessary getting of matrix at point

-Added CheckComponent to Matrix for MatrixMove, and removed the nulls checks

### DIFF
--- a/UnityProject/Assets/Scripts/Util/ConverterExtensions.cs
+++ b/UnityProject/Assets/Scripts/Util/ConverterExtensions.cs
@@ -19,7 +19,7 @@ public static class ConverterExtensions
 
 	public static Vector3 ToNonInt3(this Vector3Int other)
 	{
-		return new Vector3(other.x, other.y, 0);
+		return new Vector3(other.x, other.y, other.z);
 	}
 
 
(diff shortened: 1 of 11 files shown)


NoooneyDude (approved):

UnityProject/Assets/Scripts/Util/ConverterExtensions.cs
```diff
@@ -19,7 +19,7 @@ public static Vector3 To3(this Vector2 other)
 
 	public static Vector3 ToNonInt3(this Vector3Int other)
 	{
-		return new Vector3(other.x, other.y, 0);
+		return new Vector3(other.x, other.y, other.z);
```
Bod9001, line 22: This might break some stuff since it was Using it as a safeguard?

ThatDan123: The other thing was something in mobs, not sure why that would need it to be 0

Bod9001: basically just afraid of breaking something, I don't mind either way too much

corp-0 (approved):

PR #8296, opened 2022-01-30
Comment out log in UpdateConnectedPlayersMessage and FootStep Optimisation

ThatDan123:
-String.Join at high player count was causing some frame loss

-Footstep optimisation

### DIFF
--- a/UnityProject/Assets/Scripts/Messages/Server/ServerMessage.cs
+++ b/UnityProject/Assets/Scripts/Messages/Server/ServerMessage.cs
@@ -128,7 +128,7 @@ namespace Messages.Server
 
 			for (int i = players.Count - 1; i > 0; i--)
 			{
-				if (Vector2.Distance(worldPosition, players[i].GameObject.transform.position) > 15f) return;
+				if (Vector2.Distance(worldPosition, players[i].GameObject.transform.position) > 15f) continue;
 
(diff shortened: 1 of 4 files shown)


corp-0:
What's the difference on using matrix and matrixInfo?

ThatDan123:
> What's the difference on using matrix and matrixInfo?

If this is about the footstep var rename, it was named wrong it is supposed to be MatrixInfo

UnityProject/Assets/Scripts/Messages/Server/ServerMessage.cs
```diff
@@ -120,30 +120,17 @@ public static void SendToVisiblePlayers(Vector2 worldPosition, T msg, int channe
...
-				{
-					//Player in the list is too far away for this message, remove them:
-					players.Remove(players[i]);
-				}
-			}
-
-			foreach (ConnectedPlayer player in players)
-			{
-				if (player.Script == null) continue;
+				if (Vector2.Distance(worldPosition, players[i].GameObject.transform.position) > 15f) return;
```
NoooneyDude:
```suggestion
				if (Vector2.Distance(worldPosition, players[i].GameObject.transform.position) > 15f) continue;
```

Shouldn't this be continue?

NoooneyDude (approved):

corp-0 (approved):

PR #8299, opened 2022-01-30
IDCard Fix

ThatDan123:
-Turns out occupation was null

CL: [Fix] fixed captain's spare ID reading as NULL


corp-0:
When is it null?

ThatDan123:
For captains spare Id

corp-0 (approved):

NoooneyDude (approved):

PR #8309, opened 2022-01-31
Adds "Not cuffed" Objective to antag objectives

### LINKED ISSUE #4658: Antags need better End Round goal detection

MaxIsJoe:
For some reason it wasn't in the list from the start.

closes #4658

### DIFF
--- a/UnityProject/Assets/Scripts/Player/Mind.cs
+++ b/UnityProject/Assets/Scripts/Player/Mind.cs
@@ -148,6 +148,14 @@ public class Mind
 		body.OrNull()?.GetComponent<PlayerOnlySyncValues>().OrNull()?.ServerSetAntag(true);
 	}
 
+	public void AddObjectiveToAntag(Objective objectiveToAdd)
+	{
+		var list = new List<Objective>();
+		antag.Objectives.CopyTo<Objective>(list);
+		list.Add(objectiveToAdd);
+		antag.Objectives = list;
+	}
+
 	/// <summary>
 	/// Remove the antag status from this mind
 	/// </summary>
--- a/UnityProject/Assets/Scripts/Player/Mind.cs
+++ b/UnityProject/Assets/Scripts/Player/Mind.cs
@@ -150,6 +150,7 @@ public class Mind
 
 	public void AddObjectiveToAntag(Objective objectiveToAdd)
 	{
+		//TODO : Notify the player that a new objective has been added automatically.
 		var list = new List<Objective>();
 		antag.Objectives.CopyTo<Objective>(list);
 		list.Add(objectiveToAdd);
--- a/UnityProject/Assets/Scripts/Systems/Antagonists/Objectives/NotCuffed.cs
+++ b/UnityProject/Assets/Scripts/Systems/Antagonists/Objectives/NotCuffed.cs
@@ -6,13 +6,9 @@ namespace Antagonists
 	/// Try to be alive and not under arrest at the end of the round.
 	/// </summary>
 	[CreateAssetMenu(menuName="ScriptableObjects/AntagObjectives/NotCuffed")]
-	public class NotCuffed: Objective
+	public class NotCuffed: Escape
 	{
-		[SerializeField] private Objective escapeObject;
-		protected override void Setup()
-		{
-			Owner.AddObjectiveToAntag(escapeObject);
-		}
+		protected override void Setup() { }
 
 		protected override bool CheckCompletion()
 		{
@@ -29,7 +25,7 @@ namespace Antagonists
 				return false;
 			}
 
-			return true;
+			return true || base.CheckCompletion();
 		}
 	}
 }
--- a/UnityProject/Assets/Scripts/Systems/Antagonists/Objectives/NotCuffed.cs
+++ b/UnityProject/Assets/Scripts/Systems/Antagonists/Objectives/NotCuffed.cs
@@ -6,7 +6,7 @@ namespace Antagonists
 	/// Try to be alive and not under arrest at the end of the round.
 	/// </summary>
 	[CreateAssetMenu(menuName="ScriptableObjects/AntagObjectives/NotCuffed")]
-	public class NotCuffed: Escape
+	public class NotCuffed: Objective
 	{
 		protected override void Setup() { }
 
@@ -25,7 +25,6 @@ namespace Antagonists
 				return false;
 			}
 
-			if (base.CheckCompletion() == false) return false;
 			return true;
 		}
 	}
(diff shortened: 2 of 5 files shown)


Bod9001:
so  would this automatic included if you got the objective to escape alive on the shuttle?

MaxIsJoe:
> so would this automatic included if you got the objective to escape alive on the shuttle?

Technically it should as it is in the list now where before it wasn't so no one could ever get this objective

Bod9001:
but could you get the escape station alive objective and not this at same time?

MaxIsJoe:
> but could you get the escape station alive objective and not this at same time?

From my testing you only get one escape objective, so I've made the Not Cuffed objective automatically add the escape objective on setup.
Tho im unsure if what im doing is the correct way to add an item to an IEnuralable list, review pls

Bod9001:
imo, having it part of the escape on the shuttle objective is good enough for now, saves you having to write code, and then if you need it later on you can add it

MaxIsJoe:
> imo, having it part of the escape on the shuttle objective is good enough for now, saves you having to write code, and then if you need it later on you can add it

I already wrote the code, it didn't take much or long.

UnityProject/Assets/Scripts/Systems/Antagonists/Objectives/NotCuffed.cs
```diff
@@ -8,7 +8,11 @@ namespace Antagonists
 	[CreateAssetMenu(menuName="ScriptableObjects/AntagObjectives/NotCuffed")]
 	public class NotCuffed: Objective
 	{
-		protected override void Setup() {}
+		[SerializeField] private Objective escapeObject;
+		protected override void Setup()
+		{
+			Owner.AddObjectiveToAntag(escapeObject);
```
Bod9001: surely this would only add If the person rolled  not cuffed?

MaxIsJoe: wdym?

MaxIsJoe: Objective setups are independent, it will only call the setup() function on the objective that exists for an antag.

ThatDan123: I don't get why you are adding to the antag objectives. It will already be in there is it is rolled.

ThatDan123:
> imo, having it part of the escape on the shuttle objective is good enough for now, saves you having to write code, and then if you need it later on you can add it

Yeah, i think this is a better solution.

UnityProject/Assets/Scripts/Systems/Antagonists/Objectives/NotCuffed.cs
MaxIsJoe: There can only be one escape objective, if the not cuffed objective is rolled im also adding the escape the shuttle objective with it

VernaSaurus: what i don't understand is why you don't just make NotCuffed extend the escape objective?

corp-0: in my opinion they conceptually have nothing to do with each other

NoooneyDude: Yeah don't need inheritance, but it makes sense that in order to "escape" you can't be in cuffs as you're not free.

Does /tg/ also check if you're in the jail cell? Or is it just if you're cuffed?

MaxIsJoe: Completely forgot that that was a thing, done.

corp-0:
two devs agreed on it to be better part of the escape objetive, so add it there. I'm still not completely convinced and I think making the checks as small as possible and pile them is better and flexible but then we would need for an objective system that allows a list of "all of these must pass and pick any of this other list" or something like that. Go with Bod's and Daniel's solution for now.

MaxIsJoe:
> two devs agreed on it to be better part of the escape objetive, so add it there. I'm still not completely convinced and I think making the checks as small as possible and pile them is better and flexible but then we would need for an objective system that allows a list of "all of these must pass and pick any of this other list" or something like that. Go with Bod's and Daniel's solution for now.

.. a bit too late for this comment since I merged the not cuffed objective with the escape objective already

corp-0:
can you explain the logic behind this extension? Is that ``||`` supposed to be an ``&&`` ?

MaxIsJoe:
> can you explain the logic behind this extension? Is that `||` supposed to be an `&&` ?

you're right, it should be && actually.
fixed it

corp-0:
so now that it is an AND you can see it is making logic with a literal ``true`` so that isn't needed actually and you could simply return the base completion check

MaxIsJoe:
> so now that it is an AND you can see it is making logic with a literal `true` so that isn't needed actually and you could simply return the base completion check

you mean like this?
```
if (base.CheckCompletion() == false) return false;
return true;
```

corp-0:
no, I meant ``return Base.Completion()`` hehe. You already checked the person wasn't cuffed before, now you would just have to check if they escaped. That said, yeah no... I will have to ask you to add a cuffed check on Escape objetive and leave NotCuffed unchanged

MaxIsJoe:
> no, I meant `return Base.Completion()` hehe. You already checked the person wasn't cuffed before, now you would just have to check if they escaped. That said, yeah no... I will have to ask you to add a cuffed check on Escape objetive and leave NotCuffed unchanged

Done bossman 👍

UnityProject/Assets/Scripts/Player/Mind.cs
```diff
@@ -148,6 +148,14 @@ public void SetAntag(SpawnedAntag newAntag)
 		body.OrNull()?.GetComponent<PlayerOnlySyncValues>().OrNull()?.ServerSetAntag(true);
 	}
 
+	public void AddObjectiveToAntag(Objective objectiveToAdd)
```
Bod9001, line 151: this still needed?

MaxIsJoe: Might have some use in the future because we currently don't have a way to add specific objectives to antags from admins or coders.

Bod9001: only thing I would say would maybe add some code for notifying the player of added objectives, you can just add a "TODO needs to notify players/client"

MaxIsJoe: done

corp-0 (approved):

Bod9001 (approved):

PR #8310, opened 2022-01-31
Fixed Volume Slider

### LINKED ISSUE #8302: Main Menu: Music volume is set to max when volume slider is set to lowest possible setting.
### Environment
- Game Version: 3921 (Personal development build)
- OS Version: Windows 10 Pro Build 19043.1466

While on the main menu, a volume control slider is present in the bottom right for the ambient menu music. Sliding left reduces volume and sliding right increases it. Sliding all the way to the left, however, at the very end of the slider, returns to full volume.

### Location
 - Main Menu

### Steps to reproduce

1. Open Game
2. Observe volume slider in bottom right.
3. Adjust slider slightly to left.
4. Observe that volume is reduced.
5. Adjust slider to leftmost point.
6. Observe that volume returns to maximum.

### Screenshots attached:
Reduced Volume:
image
Full Volume:
image
Nearly Silent:
image
Full Volume:
image

### Additional Notes
 - Worth noting that it looks like the slider itself corresponds to another, slightly differently sized bar with which it overlaps, visible in the Reduced Volume and Nearly Silent screenshots. I'm assuming we're allowing the User to go past zero on this slider because of this disconnect.

Nianmi:
Mutes Music in the Main Menu when volume slider is 0

### Purpose

Fixes Bug where volume goes to 100% when the slider is a 0. Fixes it by muting the music. Mute Toggle button also reacts to change.
resolves #8302 

CL: [Fix] Fixed a bug where setting in game volume to 0 would cause it to jump to 100%

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Systems/Lobby/SongTracker.cs
+++ b/UnityProject/Assets/Scripts/UI/Systems/Lobby/SongTracker.cs
@@ -163,7 +163,18 @@ namespace Audio.Containers
 
 		public void OnVolumeSliderChange()
 		{
-			MusicManager.Instance.ChangeVolume(volumeSlider.value);
+			if (volumeSlider.value == 0)
+            {
+				ToggleMusicMute();
+			}
+			else
+			{
+				if (PlayerPrefs.GetInt(PlayerPrefKeys.MuteMusic) == 0)
+                {
+					ToggleMusicMute();
+				}
+				MusicManager.Instance.ChangeVolume(volumeSlider.value);
+			}
 		}
 
 		public void OnTrackButtonClick()


UnityProject/Assets/Scripts/UI/Systems/Lobby/SongTracker.cs
```diff
@@ -163,7 +163,18 @@ private async Task PlayRandomTrack()
 
 		public void OnVolumeSliderChange()
 		{
-			MusicManager.Instance.ChangeVolume(volumeSlider.value);
+			if (volumeSlider.value == 0)
```
corp-0, line 166: no entirely sure but I imagine ``volumeSlider.value`` is a float, maybe it would be worth to compare it with some tolerance

Nianmi: I think it makes only sense to mute the music when the volume slider is at 0 witch is all the way down. The number should also never go below 0 cause the slider is limited at 0.

corp-0: well yes, if the value is ``0.000000000096f`` it might be very low but not muted, but t hat isn't like it will break anything. I will accept it as is.

corp-0 (approved):

ThatDan123 (approved):

UnityProject/Assets/Scripts/UI/Systems/Lobby/SongTracker.cs
Nianmi: I will test this tomorrow. But should not be de case 0.00096f should still be very silent. The Volume was only set to full blast when the variable was exactly 0. Of curse 0.00096f would not mute the music entirely but in this case there is still the Mute button. Will test tomorrow if 0.00096f breaks anything i will fix it. thx. 

Only that we understand each other: 0.1 is silent 1 is the loudest. Scale breaks at 0. 0 should be muted in my opinion.

PR #8313, opened 2022-02-01
SBR lag fix

### LINKED ISSUE #8312: Big Lag With Station Bounced Radios.
### Environment
- Game Version: UnityStationDevelop22013106
- OS Version: gen2

When there are many station bounced radios next to eachother, a lag spike is caused whenever someone speaks. 

### Steps to reproduce

1. Be admin.
2. Spawn a bunch of station bounced radios (15 should be good).
3. Speak. 
4. Lag.
5. Cry.

MaxIsJoe:
Should fix #8312

CL: [Fix] fixed an issue with station bounced radios causing recursive messages, leading to a lot of lag.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Chat/ChatRelay.cs
+++ b/UnityProject/Assets/Scripts/Core/Chat/ChatRelay.cs
@@ -224,6 +224,7 @@ public class ChatRelay : NetworkBehaviour
 	private void CheckForRadios(ChatEvent chatEvent)
 	{
 		HandleRadioCheckCooldown();
+		var SBRSpamCheck = false;
 		// Only spoken messages should be forwarded
 		if (chatEvent.channels.HasFlag(ChatChannel.Local) == false)
 		{
@@ -234,6 +235,7 @@ public class ChatRelay : NetworkBehaviour
 		foreach (Collider2D coll in Physics2D.OverlapCircleAll(chatEvent.position,
 			radioCheckRadius, itemsMask))
 		{
+			if (SBRSpamCheck == true) break;
 			if (chatEvent.originator == coll.gameObject) continue;
 			if (coll.gameObject.TryGetComponent<LocalRadioListener>(out var listener) == false) continue;
 
@@ -242,6 +244,7 @@ public class ChatRelay : NetworkBehaviour
 				layerMask, radioPos).ItHit == false)
 			{
 				listener.SendData(chatEvent);
+				SBRSpamCheck = true;
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Core/Chat/ChatRelay.cs
```diff
@@ -252,13 +255,15 @@ private void CheckForRadios(ChatEvent chatEvent)
...
 				{
+					if (SBRSpamCheck == true) break;
 					foreach (var slot in slots)
 					{
 						if (slot.IsEmpty) continue;
 						if (slot.Item.TryGetComponent<LocalRadioListener>(out var listener)
 						    && listener != chatEvent.originator)
 						{
 							listener.SendData(chatEvent);
+							SBRSpamCheck = true;
```
Bod9001, line 250: where is this Reset?

UnityProject/Assets/Scripts/Core/Chat/ChatRelay.cs
```diff
@@ -234,6 +235,7 @@ private void CheckForRadios(ChatEvent chatEvent)
 		foreach (Collider2D coll in Physics2D.OverlapCircleAll(chatEvent.position,
 			radioCheckRadius, itemsMask))
 		{
+			if (SBRSpamCheck == true) break;
```
NoooneyDude: ~~`SBRSpamCheck` is always false at this point.~~ Nevermind, read it as `== false` : )

UnityProject/Assets/Scripts/Core/Chat/ChatRelay.cs
```diff
@@ -252,13 +255,15 @@ private void CheckForRadios(ChatEvent chatEvent)
 			{
 				foreach (var slots in playerScript.DynamicItemStorage.ServerContents.Values)
 				{
+					if (SBRSpamCheck == true) break;
```
ThatDan123, line 242: You can add this check to line 251 as well, in case it got blocked before

UnityProject/Assets/Scripts/Core/Chat/ChatRelay.cs
ThatDan123: It doesnt need to be.

The whole spam prevention is that it only finds one nearby radio and thats it, don't bother searching for more

ThatDan123:
The other way to fix the issue would be to have some static HashSet of radios and add a radios as they are used and clear at the end of the frame

MaxIsJoe:
Alright, it has been tested and after spawning exactly 30 SBRs i no longer find any framerate drops.
approvals pls

MaxIsJoe:
> The other way to fix the issue would be to have some static HashSet of radios and add a radios as they are used and clear at the end of the frame

the issue seems to be fixed right now so i don't see a point of trying to do that currently.

Bod9001 (approved):

corp-0 (approved):

PR #8317, opened 2022-02-02
Renamed Materials

### LINKED ISSUE #8281: Materials stacks in the cargo listings should be renamed
Currently, when you look in the "Materials" list of the cargo console, the materials have the size of stack first. Because the list is alphabetised, this means the top of the list looks like this:

- 10 sheets of x
- 10 sheets of y
- 50 sheets of z

I would propose we rename the listings to be like this, for more sensible layout and overview:

- x (10 sheets)
- y (10 sheets)
- z (50 sheets)

Nianmi:
### Purpose
 <br>

Fixes Cargo Martials Names so that they can be sorted Alphabetically
Resolves #8281


corp-0 (approved):

corp-0:
nice, thank you

ThatDan123 (approved):

PR #8337, opened 2022-02-08
Finally fixes Explosives being uninterractable

MaxIsJoe:
The issue this whole time was happening due to variables being out of sync.
Now that they are; C4s are interactable once again on headless servers.

CL: [Fix] fixed C4 being uninteractable

Edit : yes, this has been tested on a headless server.

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Weapons/Explosive.cs
+++ b/UnityProject/Assets/Scripts/Items/Weapons/Explosive.cs
@@ -16,8 +16,8 @@ namespace Items.Weapons
 	{
 		[Header("Explosive settings")]
 		[SerializeField] private ExplosiveType explosiveType;
-		[SerializeField] private bool detonateImmediatelyOnSignal;
-		[SerializeField] private int timeToDetonate = 10;
+		[SerializeField, SyncVar] private bool detonateImmediatelyOnSignal;
+		[SerializeField, SyncVar] private int timeToDetonate = 10;
 		[SerializeField] private int minimumTimeToDetonate = 10;
 		[SerializeField] private float explosiveStrength = 150f;
 		[SerializeField] private SpriteDataSO activeSpriteSO;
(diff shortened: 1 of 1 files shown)


corp-0:
is that enough? Don't you need to do a sync hook method?

MaxIsJoe:
> is that enough? Don't you need to do a sync hook method?

from my testing it just works as it is
also from what im reading in the docs I think syncvar hooks are only meant for telling clients extra stuff to do when a value changes which I don't think applies here because we only want to keep track if the variable changes and thats it.

MaxIsJoe:
https://mirror-networking.gitbook.io/docs/guides/synchronization/syncvars

Bod9001:
[https://unitystation.github.io/unitystation/development/SyncVar-Best-Practices-for-Easy-Networking/ ](https://unitystation.github.io/unitystation/development/SyncVar-Best-Practices-for-Easy-Networking/ )

corp-0:
yeah, I remember that guide and that's what I believed but maybe that was the case back then when we wrote it and it isn't anymore.

ThatDan123:
Was the bug with net tabs not showing the right values? If so this isnt the way to fix it.

See line 360 in:

https://github.com/unitystation/unitystation/blob/develop/UnityProject/Assets/Scripts/Objects/Engineering/ParticleAcceleratorControl.cs

Thats how you can update values for the nettabs.

(I'm happy with the sync vars on `isArmed` and `isOnObject` just the others should be looked at again)

NoooneyDude (dismissed):
It would also be acceptable to just skip those particular checks on the client (by checking `side == NetworkSide.Server`), no syncing needed, as the client otherwise doesn't need to know about them.

MaxIsJoe:
> Was the bug with net tabs not showing the right values? If so this isnt the way to fix it.
> 
> See line 360 in:
> 
> https://github.com/unitystation/unitystation/blob/develop/UnityProject/Assets/Scripts/Objects/Engineering/ParticleAcceleratorControl.cs
> 
> Thats how you can update values for the nettabs.
> 
> (I'm happy with the sync vars on `isArmed` and `isOnObject` just the others should be looked at again)

No, the issue is that C4s weren't interactable once you attached them on a source because the variables WillInteract needed were out of sync

NoooneyDude:
I think the extra SyncVars not used in WillInteract need explaining what they're for? Not immediately obvious. I think that's where ThatDan's going.

MaxIsJoe:
Oh you mean the extra syncvar i put on the timetodetonate variable? i gotcha i gotcha
will fix it

MaxIsJoe:
Done, the signal bool should only be set by the netui as well so if it doesn't for whatever reason we can re-add it again later.

Bod9001 (approved):

corp-0 (approved):

PR #8340, opened 2022-02-09
Fixes accumulating ability icons

### LINKED ISSUE #7441: Picking up some items makes you accumulate ability icons
### Environment
- Game Version: UnityStationDevelop-21091918

image
Items that cause this:
- any bag
- bag of dice
- trash bag

Picking up same item does not duplicate ability. Each item only adds one.

### Steps to reproduce

1. Pick up any listed item
2. Drop it

MaxIsJoe:
CL: [Fix] - Fixed accumulating ability icons (Fixes #7441)

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Core/Action/UIActionManager.cs
+++ b/UnityProject/Assets/Scripts/UI/Core/Action/UIActionManager.cs
@@ -186,6 +186,12 @@ namespace UI.Action
 
 		public static void Show(IActionGUI iActionGUI)
 		{
+			foreach (var actionButton in Instance.DicIActionGUI)
+			{
+				//If there is a duplicate of this button, don't spawn it!
+				if (actionButton.Value.ActionData != iActionGUI.ActionData) continue;
+				return;
+			}
 			UIAction _UIAction;
 			if (Instance.PooledUIAction.Count > 0)
 			{
--- a/UnityProject/Assets/Scripts/UI/Core/Action/UIActionManager.cs
+++ b/UnityProject/Assets/Scripts/UI/Core/Action/UIActionManager.cs
@@ -189,8 +189,7 @@ namespace UI.Action
 			foreach (var actionButton in Instance.DicIActionGUI)
 			{
 				//If there is a duplicate of this button, don't spawn it!
-				if (actionButton.Value.ActionData != iActionGUI.ActionData) continue;
-				return;
+				if (actionButton.Value.ActionData == iActionGUI.ActionData) return;
 			}
 			UIAction _UIAction;
 			if (Instance.PooledUIAction.Count > 0)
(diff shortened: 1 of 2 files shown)


riquelme-A:
so this means it will still pile up but it is hidden?

MaxIsJoe:
> so this means it will still pile up but it is hidden?

Yeah

Bod9001 (dismissed):
need to make  This server side some point

corp-0:
I'm not sure I'm happy with this approach to fix the bug

CdrCross:
Once we've moved intents off the number keys, I'm eyeing a chance to redesign the ability bar entirely, to act more like it does in an MMO or ARPG, and crucially to eat less space. @corp-0

MaxIsJoe:
> I'm not sure I'm happy with this approach to fix the bug

I'll look for the function that spawns them then and do the check there instead and delete them if they are not available.

Though an Idea that came to my mind would be action usage counters, so imagine an item that has 5 usages and it generates 5 action buttons, we can hide 4 and make one display a counter of how many usages it has and delete the other buttons once they are no longer needed.

CdrCross:
> > I'm not sure I'm happy with this approach to fix the bug
> 
> I'll look for the function that spawns them then and do the check there instead and delete them if they are not available.
> 
> Though an Idea that came to my mind would be action usage counters, so imagine an item that has 5 usages and it generates 5 action buttons, we can hide 4 and make one display a counter of how many usages it has and delete the other buttons once they are no longer needed.

A charge counter is a good idea, but make sure to give it context for the user. We don't need yet another unknowable number floating around. I would recommend either "x uses" or "x/x".

MaxIsJoe:
> I'm not sure I'm happy with this approach to fix the bug

Done, it won't spawn anymore now if there's another button with the same action data.
Review pls

UnityProject/Assets/Scripts/UI/Core/Action/UIActionManager.cs
```diff
@@ -186,6 +186,12 @@ public static void SetCooldown(IActionGUI iActionGUI, float cooldown, GameObject
 
 		public static void Show(IActionGUI iActionGUI)
 		{
+			foreach (var actionButton in Instance.DicIActionGUI)
+			{
+				//If there is a duplicate of this button, don't spawn it!
+				if (actionButton.Value.ActionData != iActionGUI.ActionData) continue;
```
corp-0: Isn't it simpler to invert this logic so no continue is needed?

MaxIsJoe: Done

corp-0 (approved):

Bod9001 (approved):

PR #8341, opened 2022-02-09
Allows AI to interact with consoles that require IDs

### LINKED ISSUE #7172: Ai can't properly access console functions that require IDs
Currently Cargo, Comms, Hop and security consoles require ID to use and therefore cant be used by Ai.

- [x] Cargo Console
- [x] ID Console
- [x] Comms Console
- [x] Security Console
- [x] ACU's

-Modify them to allow for Ai use by doing a check for Ai when login button pressed

MaxIsJoe:
Fixes #7172

CL: [Improvement] Allows AI to interact with consoles that require IDs

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Atmospherics/AirController.cs
+++ b/UnityProject/Assets/Scripts/Objects/Atmospherics/AirController.cs
@@ -263,8 +263,7 @@ namespace Objects.Atmospherics
 
 		public void ServerPerformInteraction(HandApply interaction)
 		{
-			if (accessRestrictions.CheckAccessCard(interaction.HandObject) ||
-				interaction.PerformerPlayerScript.mind.occupation.JobType == JobType.AI)
+			if (accessRestrictions.CheckAccessCard(interaction.HandObject))
 			{
 				IsLocked = !IsLocked;
 
--- a/UnityProject/Assets/Scripts/Objects/Atmospherics/AirController.cs
+++ b/UnityProject/Assets/Scripts/Objects/Atmospherics/AirController.cs
@@ -263,7 +263,8 @@ namespace Objects.Atmospherics
 
 		public void ServerPerformInteraction(HandApply interaction)
 		{
-			if (accessRestrictions.CheckAccessCard(interaction.HandObject))
+			if (accessRestrictions.CheckAccessCard(interaction.HandObject) ||
+				interaction.PerformerPlayerScript.mind.occupation.JobType == JobType.AI)
 			{
 				IsLocked = !IsLocked;
 
--- a/UnityProject/Assets/Scripts/UI/Objects/Cargo/GUI_CargoPageCart.cs
+++ b/UnityProject/Assets/Scripts/UI/Objects/Cargo/GUI_CargoPageCart.cs
@@ -47,7 +47,7 @@ namespace UI.Objects.Cargo
 
 		public void ConfirmCart()
 		{
-			if (CanAffordCart() == false || (cargoGUI.cargoConsole.CorrectID == false && cargoGUI.IsAIInteracting() == false) == true) return;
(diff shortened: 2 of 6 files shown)


corp-0:
It should be an almost insignificant difference in most cases, but checking for AI first should be faster

MaxIsJoe:
> It should be an almost insignificant difference in most cases, but checking for AI first should be faster

where?

UnityProject/Assets/Scripts/UI/Objects/Cargo/GUI_CargoPageCart.cs
```diff
@@ -47,7 +47,7 @@ private void CheckTotalPrice()
 
 		public void ConfirmCart()
 		{
-			if (CanAffordCart() == false || cargoGUI.cargoConsole.CorrectID == false) return;
+			if (CanAffordCart() == false || (cargoGUI.cargoConsole.CorrectID == false && cargoGUI.IsAIInteracting() == false) == true) return;
```
corp-0: wait what

corp-0:
```suggestion
			if (CanAffordCart() == false || (cargoGUI.cargoConsole.CorrectID == false && cargoGUI.IsAIInteracting() == false)) return;
```

UnityProject/Assets/Scripts/Objects/Atmospherics/AirController.cs
```diff
@@ -263,7 +263,8 @@ public bool WillInteract(HandApply interaction, NetworkSide side)
 
 		public void ServerPerformInteraction(HandApply interaction)
 		{
-			if (accessRestrictions.CheckAccessCard(interaction.HandObject))
+			if (accessRestrictions.CheckAccessCard(interaction.HandObject) ||
```
corp-0: I meant here, to check for AI before the HandObject but honestly it doesn't really matter. Ignore the comment.

NoooneyDude: ~~Looks like all this will do is toggle the lock on the ACU, but they still can't open the UI. I think the fix will unfortunately be a little more involved.~~

Actually, does `WillInteract` even return true?

ThatDan123: Ai doesnt trigger HandApply, so this will do nothing

ThatDan123:
@tomasalves8 are you still working on your changes for Ai console stuff?

tomasalves8:
@ThatDan123 Been a bit busy, so I wasn't able to finish it.

UnityProject/Assets/Scripts/Objects/Atmospherics/AirController.cs
MaxIsJoe: What is this this? 🤔 

image

MaxIsJoe: actually wait nvm, different component

MaxIsJoe:
another review pls

NoooneyDude (commented):
This will cause some unexpected bugs, like players being able to use restricted functions when an AI is observing. ~~I think particularly the ID console is the worst offender: it isn't out of the question an AI wants to observe what accesses are being given to someone, but then the HoP can't remove the card~~. Sorry, wrong console.

NoooneyDude (approved):

ThatDan123 (approved):

ThatDan123:
merge after the playtest build?

MaxIsJoe:
not sure whats the point of a playtest if we don't test changes we make

PerfectTangent:
Did this PR fix AI access for ACUs? I don't see any modifications to them in the changes.

MaxIsJoe:
> Did this PR fix AI access for ACUs? I don't see any modifications to them in the changes.

I must have accidently reverted something related to it, will post another PR later.

PR #8353, opened 2022-02-13
Observer optimisation

Bod9001:
fixes expensive Destroying drawing of network objects ( brings back Slower loop)
makes it so  external Nodes only get update if the matrix is moving


corp-0:
this looks like a revert, what happened?

Bod9001:
observingList.Remove(netIdentity); is slow as balls

NoooneyDude (approved):

corp-0 (approved):

PR #8356, opened 2022-02-13
Fix hostile mobs attack message always using default body part

### LINKED ISSUE #6695: Hostile mobs attack text shows incorrectly.
This happens on staging

1. Allow a hostile mob to attack you
2. Notice the chat message says "xxx has attacked yyy in the head"
3. Notice on the health doll that the attacked bodypart is most likely not the head
    3a. if it is the head, allow the mob to attack multiple times and note the health doll each time

phxmask:
Fixes #6695

The chat message after a mob attack was always being sent using the default body type target, despite the actual attack using a randomized value. Now the actual value chosen is used in the chat too.

In commit 57f043c9ac91bf644c448f70947cfc858f1fd96d I've added some human readable names for the BodyPartType enums (so that we don't print "RightHand" in chat for example) using Descriptions, if that's not the desired approach for this I can revert it.

CL: [Fix] fixed hostile mob attack messages always using default body part name.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Chat/Chat.Process.cs
+++ b/UnityProject/Assets/Scripts/Core/Chat/Chat.Process.cs
@@ -565,7 +565,7 @@ public partial class Chat
 
 	private static string InTheZone(BodyPartType hitZone)
 	{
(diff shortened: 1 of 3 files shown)


corp-0:
oh that GetDescription is very nice

Bod9001 (approved):

corp-0 (approved):

PR #8358, opened 2022-02-14
Adds server history to connection screen

MaxIsJoe:
I've always found it annoying that when I disconnect from a server I'd have to close the game and open it again from the hub or have to manually enter the IP address, so I've added a panel that keeps track of the last servers you have connected to and a button next to their IP to quickly connect to them.

CL: [Improvement] added a button to quick join the last server played in the lobby.


MaxIsJoe:
oh yeah, I forgot to mention but i've also removed two scripts that didn't do anything

corp-0:
what does it look like?

MaxIsJoe:
> what does it look like?

The green button is what toggles the panel

image

PerfectTangent:
is it possible to retrieve the server's name?

MaxIsJoe:
> is it possible to retrieve the server's name?

at the moment it's not possible because the log only tracks the inputs you gave to the text fields and not anything related to the server itself.
I tried looking into a way to log that data _after_ the player connects but it turned quickly into a headache and just focused on what I wanted to do.

corp-0:
might be an overkill, but you could compare it with the serverlist lol

MaxIsJoe:
> might be an overkill, but you could compare it with the serverlist lol

hm, maybe that's a good solution.
Though is it reeeaallly needed if we have server names in the history list? The top first server on the list is also going to be the last one you joined and we only have three servers at all times on the hub and only two of them people play on. So I doubt the extra work is needed for that.

PerfectTangent:
> The top first server on the list is also going to be the last one you joined and we only have three servers at all times on the hub and only two of them people play on.

the first thing is not obvious to someone who isn't explicitly told that that is the case. The second one is true now, but won't always be true.

...

hopefully.

MaxIsJoe:
yeah screw it, I've been trying to get server name changes going for the past 5 hours and I've been completed demotivated and frustrated with unity and how i can't get the data i need to translate correctly from the API to a json format.

pls merge, i do not plan on doing any further changes anymore.

Bod9001 (dismissed):

corp-0:
to be honest I wouldn't merge this as it is just because it looks way too unpolished

MaxIsJoe:
> to be honest I would merge this as it is just because it looks way too unpolished

We can polish the UI later, functionality is what matters atm.

MaxIsJoe:
if I hide the history menu until we polish it and just make the button connect to the last server the player has connected to, would you merge this @corp-0 ?

corp-0:
with the promise of making the ui better eventually? Why do you want to merge it? I haven't seen people complaining about something that would be solved by this feature

MaxIsJoe:
> with the promise of making the ui better eventually? Why do you want to merge it? I haven't seen people complaining about something that would be solved by this feature

..because I'm the one who wants this feature because I find it annoying that I have to close the game then re-open it again from the hub if i get disconnected for whatever reason?
I think I've clearly stated in the PR's desc why I made this change and who it's intended for.

PerfectTangent:
I would get use out of this feature, I can say that. Like when the server times out for some reason and boots you out. Having a quick solution to rejoin rather than close and reopen the game would save some time.

PerfectTangent:
> if I hide the history menu until we polish it and just make the button connect to the last server the player has connected to, would you merge this @corp-0 ?

if these changes are made, I think this should be merged.

NoooneyDude:
> I've always found it annoying that when I disconnect from a server I'd have to close the game and open it again from the hub or have to manually enter the IP address

Don't really understand this, it always has the server IP populated for me when I'm disconnected.

Edit: oh, but only when disconnected by the server and not logging out. Interesting.

corp-0:
> if I hide the history menu until we polish it and just make the button connect to the last server the player has connected to, would you merge this @corp-0 ?

Ok, I will approve then

MaxIsJoe:
> Ok, I will approve then

Done

Bod9001 (approved):

corp-0 (approved):

PR #8359, opened 2022-02-14
makes wrapping paper wrap an object with alt-click

### LINKED ISSUE #6840: Cannot put wrapping paper inside a backpack or any other wrappable container
### Environment
- Game Version: latest

if you try to place a roll of wrapping paper inside a backpack on help intent, you begin to wrap the backpack, even if the backpack is equipped. If you use any other intent, nothing happens.

### Steps to reproduce

1. get a roll of wrapping paper
2. try to put it in the backpack
3. observe how none of the intents have the desired effect of stowing the wrapping paper in the backpack.

MaxIsJoe:
Adds in the fix that was suggested in #6840

CL: [Fix] made wrapping paper wrap an object with alt-click to fix an interaction conflict with containers.

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Cargo/Wrapping/WrappingPaper.cs
+++ b/UnityProject/Assets/Scripts/Items/Cargo/Wrapping/WrappingPaper.cs
@@ -30,6 +30,7 @@ namespace Items.Cargo.Wrapping
 
 		public bool WillInteract(HandApply interaction, NetworkSide side)
 		{
+			if (interaction.IsAltClick == false) return false;
 			return DefaultWillInteract.Default(interaction, side) &&
 			       CommonWillInteract(interaction, side);
 		}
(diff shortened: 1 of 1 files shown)


corp-0:
I would append the alt click check to the rest of the return statement

MaxIsJoe:
> I would append the alt click check to the rest of the return statement

I personally think it's best to not call any extra checks if the main requirement is not met, just a tinsy smol performance squeeze.

corp-0:
yeah, that's what you do with ``&&``. If the first condition isn't true it won't try the second one

MaxIsJoe:
done

Bod9001 (dismissed):

vlandemart (dismissed):

corp-0:
Please don't do == true, == false is ugly enough

corp-0:
Also tests be failing

corp-0 (approved):

Bod9001 (approved):

PR #8368, opened 2022-02-14
Make it so wrenched down chairs can be rotated

Bod9001:
as name says

CL: [Fix] wrenched down chairs can be rotated.

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/PlayerRotatable.cs
+++ b/UnityProject/Assets/Scripts/Objects/PlayerRotatable.cs
@@ -45,7 +45,7 @@ namespace Objects
 		public bool WillInteract(ContextMenuApply interaction, NetworkSide side)
 		{
 			if (!DefaultWillInteract.Default(interaction, side)) return false;
-			if (TryGetComponent(out ObjectBehaviour behaviour) && (!behaviour.IsPushable && CanRotateIfNotMovable == false)) return false;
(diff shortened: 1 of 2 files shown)


corp-0:
why does it need to check if it is movable? Why not also rotate movable stuff?

Bod9001:
it does the try get component, if it has a moving component then there's a bool to override it to allow it to be rotated
otherwise if it doesn't have a Moving component then it allows it

NoooneyDude (dismissed):

UnityProject/Assets/Scripts/Objects/PlayerRotatable.cs
```diff
@@ -43,7 +45,7 @@ public RightClickableResult GenerateRightClickOptions()
 		public bool WillInteract(ContextMenuApply interaction, NetworkSide side)
 		{
 			if (!DefaultWillInteract.Default(interaction, side)) return false;
-			if (TryGetComponent(out ObjectBehaviour behaviour) && !behaviour.IsPushable) return false;
+			if (TryGetComponent(out ObjectBehaviour behaviour) && (!behaviour.IsPushable && CanRotateIfNotMovable == false)) return false;
```
vlandemart:
```suggestion
			if (TryGetComponent(out ObjectBehaviour behaviour) && (behaviour.IsPushable == false && CanRotateIfNotMovable == false)) return false;
```

Just to be coherent

NoooneyDude (approved):

vlandemart (approved):

PR #8377, opened 2022-02-16
Regen meshs' stacks are now visually visible + require opening before use.

MaxIsJoe:
Title


NoooneyDude (changes requested):
The code from the other PR (#8380) to require opening an edible (like soda can) is very similar to the code here. Therefore, that and this code here should be in a new component, say: `HandPreparable` or something.

In this PR, you would not create a new component that inherits from `HealsTheLiving` but rather add a `TryGetComponent()` within `HealsTheLiving` that checks for the `HandPreparable` component says it is ready for use.

In the edible PR: similar story, checking if that component says it is ready to use.

Also, move the check to `WillInteract` if you want the interaction to propagate properly to the `HandPreparable` component. That is, the check to see if something is ready to eat / apply should be done in WillInteract (can't remember if required client-side for propagation - would need to test) and return false if not, so that the HandActivate event will propagate correctly to the `HandPreparable` component regardless of component order.

~~In light of this I'm unapproving #8380 and marking as changes required.~~ Too slow 😛

MaxIsJoe:
Done, review pls

UnityProject/Assets/Scripts/Items/HandPreparable.cs
```diff
@@ -0,0 +1,47 @@
...
+		{
+			var rightClickResult = new RightClickableResult();
+			if (isPrepared) return rightClickResult;
+			rightClickResult.AddElement("Open This", Open);
+			return rightClickResult;
+		}
+
+		public bool WillInteract(HandActivate interaction, NetworkSide side)
+		{
+			return isPrepared ? false : true;
```
Bod9001, line 46: is this networked?

MaxIsJoe: no it's not, I probably should.

MaxIsJoe: Done

Bod9001 (approved):

NoooneyDude (approved):

PR #8378, opened 2022-02-16
Adds the option for disabling autostacking for some items

### LINKED ISSUE #8361: Not all stackable items should merge together when occupying the same tile.
### Description

The system added in #7285 for making stackable items merge together when on the same tile currently affects all items, but there are some exceptions that need to be added. For example, regen meshes should not merge together when occupying the same tile. 

This quality of stackable items needs to be a toggleable option for stackable items. 

The items that don't automatically merge can't be manually split either using the stack splitting added in #7870. They should stay discrete.

MaxIsJoe:
Some items do not automatically stack when dropped on the floor or when spawning, this adds the option to disable this behavior.

closes #8361

### DIFF
--- a/UnityProject/Assets/Prefabs/Items/Medical/Trauma Treatment/Regen Mesh.prefab
+++ b/UnityProject/Assets/Prefabs/Items/Medical/Trauma Treatment/Regen Mesh.prefab
@@ -1,39 +1,5 @@
 %YAML 1.1
 %TAG !u! tag:unity3d.com,2011:
---- !u!114 &-6495101291114284292
-MonoBehaviour:
-  m_ObjectHideFlags: 0
-  m_CorrespondingSourceObject: {fileID: 0}
-  m_PrefabInstance: {fileID: 0}
-  m_PrefabAsset: {fileID: 0}
-  m_GameObject: {fileID: 1791647728409730387}
-  m_Enabled: 1
-  m_EditorHideFlags: 0
-  m_Script: {fileID: 11500000, guid: e7c30ca55ffa4467a421abf7a832a6a1, type: 3}
-  m_Name: 
-  m_EditorClassIdentifier: 
-  syncMode: 0
-  syncInterval: 0.1
-  initialAmount: 15
-  maxAmount: 15
-  stacksWith: []
---- !u!114 &2710468740513023014
-MonoBehaviour:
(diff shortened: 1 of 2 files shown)


corp-0:
when do we want stackable items to don't stack on drop?

MaxIsJoe:
> when do we want stackable items to don't stack on drop?

I honestly have no idea
Another example that comes to my mind would probably be papers because not all of them should; Either way the option is there to disable it from the inspector so someone else can come and change it.

CdrCross:
> when do we want stackable items to don't stack on drop?

Any time the individual object contains individual data. Papers is a good example, but forensics is also an interesting case.

MaxIsJoe:
When we start working on forensics and detective stuff we can probably just merge that data into the stack item itself.

NoooneyDude (approved):

Fr-Dae:
> when do we want stackable items to don't stack on drop?

Pizza part.
Pastec 
Paper

PerfectTangent:
> Any time the individual object contains individual data. Papers is a good example, but forensics is also an interesting case.
> Pizza part. Pastec Paper

none of those items have the stackable component. This PR is about _stackable_ items merging when placed on the same tile, not just any items. Currently, there aren't many things except the regen mesh that need this.

Fr-Dae:
> none of those items have the stackable component. This PR is about _stackable_ items merging when placed on the same tile, not just any items. Currently, there aren't many things except the regen mesh that need this.

Pizza slices, if placed on the same tile, merge into a stack.
which is a pain in the ass to arrange the 6 slices of pizza for 6 people.
and not one person eating a pizza by himself.

Bod9001:
added to the pizza!

MaxIsJoe:
> > Any time the individual object contains individual data. Papers is a good example, but forensics is also an interesting case.
> > Pizza part. Pastec Paper
> 
> none of those items have the stackable component. This PR is about _stackable_ items merging when placed on the same tile, not just any items. Currently, there aren't many things except the regen mesh that need this.

I haven't changed anything except regen meshes to use this change.
I'm saying that if anyone wants to change any other items to prevent autostacking on spawn/drop, the change is so easy because it's a simple toggle from the inspector. No code changes required.

Bod9001 (approved):

PR #8385, opened 2022-02-17
Server wont add or notify disconnected players anymore #6665

MaxIsJoe:
fixes #6665

~~Need @Bod9001 to confirm that it solves the issue 100% incase I missed another queue function or a more important area that needed this fix.~~

Edit : we talked, this seems like it does the trick by making sure all disconnected players are cleared from the list before doing anything + the server wont waste time sending stuff to offline players.

PR #8390, opened 2022-02-18
Quickfix for things Dan pointed in #8385

MaxIsJoe:
title


ThatDan123 (approved):

NoooneyDude (approved):
This is ok for now, though I would like to see a disconnect event, or something similar, raised when a ConnectedPlayer disconnects so that any interested scripts can perform cleanup.

That will avoid needing to check against the lists at arbitrary positions.

PR #8394, opened 2022-02-18
Asking for more materials than there is just dumps all materials now

### LINKED ISSUE #6956: Attempting to withdraw more sheets from the ore redemption machine than are available should just give you all the sheets.
### Description
Right now, the ore redemption machine doesn't let you withdraw sheets of material if the amount you are trying to withdraw. For example, if you are trying to withdraw 10 sheets worth of plasma but there is only enough for 8 sheets then clicking on the button to withdraw 10 sheets of plasma won't do anything. This is quite annoying when trying to withdraw large amounts of material and I believe that when you attempt to overdraw sheets, you should just completely empty the ORM of whatever material you're attempting to withdraw.

MaxIsJoe:
closes #6956

CL: [Improvement] attempting to withdraw more sheets from the mineral silo and autolathes will dump all sheets they have now

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Machines/MaterialStorage.cs
+++ b/UnityProject/Assets/Scripts/Objects/Machines/MaterialStorage.cs
@@ -40,9 +40,8 @@ namespace Objects.Machines
 
 		private void ConsumeMaterial(ItemTrait material, int quantity)
 		{
-			var quantityToSubtract = (int) Mathf.Clamp(quantity, 0, Single.PositiveInfinity);
-			MaterialList[material] -= quantityToSubtract;
-			currentResources -= quantityToSubtract;
+			MaterialList[material] -= quantity;
+			currentResources -= quantity;
 		}
 
 		public bool TryAddSheet(ItemTrait material, int quantity)
--- a/UnityProject/Assets/Scripts/Objects/Machines/MaterialStorage.cs
+++ b/UnityProject/Assets/Scripts/Objects/Machines/MaterialStorage.cs
@@ -60,23 +60,8 @@ namespace Objects.Machines
 
 		public int TryRemoveSheet(ItemTrait material, int quantity)
 		{
+			quantity = Mathf.Min(quantity, MaterialList[material] / Cm3PerSheet);
 			var Cm3Used = Cm3PerSheet * quantity;
-			foreach (var materialToTake in MaterialList)
-			{
-				if(materialToTake.Key != material) continue;
-				if(materialToTake.Value > Cm3Used || materialToTake.Value == 0) continue;
-				ConsumeMaterial(material, materialToTake.Value);
-				return quantity;
-			}
-			if (MaterialList[material] < Cm3PerSheet)
-			{
-				return 0;
-			}
-			if (MaterialList[material] < Cm3Used)
-			{
-				quantity = MaterialList[material] / Cm3PerSheet;
-				Cm3Used = Cm3PerSheet * quantity;
-			}
 			ConsumeMaterial(material, Cm3Used);
 			UpdateGUIs.Invoke();
 			return quantity;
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Objects/Machines/MaterialStorage.cs
```diff
@@ -36,8 +40,9 @@ private void AddMaterial(ItemTrait material, int quantity)
 
 		private void ConsumeMaterial(ItemTrait material, int quantity)
 		{
-			MaterialList[material] -= quantity;
-			currentResources -= quantity;
+			var quantityToSubtract = (int) Mathf.Clamp(quantity, 0, Single.PositiveInfinity);
```
Bod9001: surely clamping to how much material you have would be better way of doing it?
so if you got 50 clamps maximum to 50?

MaxIsJoe: It's a safety precaution to make sure quantity never falls below 0. It's basically doing the same exact thing as before but making sure that quantity is never negative which the machine will try to do sometimes.

Bod9001: how does it become negative?

MaxIsJoe: 20 materials - 50 materials to consume (quantity) = -30 materials/resources

Bod9001: meant as in quantity , and surely you could just do 99999
 and then that would still equal
 MaterialList[material] = -99994

MaxIsJoe: I reverted it because I can't replicate the issue anymore.

Bod9001 (dismissed):

UnityProject/Assets/Scripts/Objects/Machines/MaterialStorage.cs
```diff
@@ -56,11 +60,18 @@ public bool TryAddSheet(ItemTrait material, int quantity)
...
+				return quantity;
+			}
 			if (MaterialList[material] < Cm3PerSheet)
 			{
 				return 0;
 			}
-			var Cm3Used = Cm3PerSheet * quantity;
 			if (MaterialList[material] < Cm3Used)
 			{
 				quantity = MaterialList[material] / Cm3PerSheet;
```
NoooneyDude, line 77: Seems a little complex, and you're not poking the UI.

Wouldn't it be simpler instead to just remove the `if` condition on line `63` (before changes)? The method could be as little as five lines.

NoooneyDude (changes requested):

UnityProject/Assets/Scripts/Objects/Machines/MaterialStorage.cs
NoooneyDude: Seems a little complex: you're not poking the UI and you'll get more sheets out than is possible (e.g. 500 cm3 stored, player requests 50 sheets and gets 50 sheets even though there's only a quarter of a sheet stored).

Wouldn't it be simpler instead to just remove the `if` condition on line `63` (before changes) and use either the requested amount of sheets or the amount of available sheets, whichever is lower?
```suggestion
			quantity = Mathf.Min(quantity, MaterialList[material] / Cm3PerSheet);
```
Can't suggest changes on lines too far away so there's an exta `}`... @Github...

MaxIsJoe: I don't understand what you want here?

NoooneyDude: Just do 

```
		public int TryRemoveSheet(ItemTrait material, int quantity)
		{
			quantity = Mathf.Min(quantity, MaterialList[material] / Cm3PerSheet);
			var Cm3Used = Cm3PerSheet * quantity;
			ConsumeMaterial(material, Cm3Used);
			UpdateGUIs.Invoke();
			return quantity;
		}
```

MaxIsJoe: Done and tested

MaxIsJoe:
review pls

NoooneyDude (approved):

Bod9001 (approved):

PR #8402, opened 2022-02-20
Cleanup workflows

NoooneyDude:
- Autoformats workflows to indentation of two spaces.
- Moves license file generation to its own manually-run workflow.
- Uncomments playmode tests workflow (was still trying to run when commented, so workflow is being logged as failed). Instead, will disable workflow via Actions tab on GitHub.
- Renames integration workflow to be consistent with the naming of others.


corp-0:
lovely, thank you! Would you mind removing the artifact janitor? We don't use it anymore

NoooneyDude:
Sure thing. Done.

corp-0 (approved):

Bod9001 (approved):

PR #8404, opened 2022-02-20
Update YamlDotNet plugin

NoooneyDude:
- Required for Unity 2021.2 (which updates C# language version to 9).


corp-0:
Where do you we use this plugin?

Bod9001:
in the ThemeManager, imo json would be better  **is known to be biased**

corp-0 (approved):

NoooneyDude:
Yep, used by the ThemeManager and chemistry importer.

Bod9001 (approved):

PR #8413, opened 2022-02-23
makes it so null entries are filled up in Sprite Catalogue 

Bod9001:
there's probably a good reason it was commented out but let's see if everything explodes, 
I had a look through the code such is the test code and it looks like it would be okay with this change


Bod9001:
rebased to get tests working, yeah me get it rebased!

NoooneyDude (approved):

corp-0 (approved):

PR #8426, opened 2022-02-24
Surplus rifle resprite and rework

PerfectTangent:
### Purpose

This PR somewhat reworks the Surplus rifle and gives it a new sprite courtesy of Meme machine and myself.

The point of the changes is to make the surplus rifle somewhat more effective to use, partially inspired by it's new asthetic. The surplus rifle now uses 7.12x82mm stripper clips instead of a detachable magazine. The weapon's damage remains the same, but the total ammo capacity of the weapon has been halved from 10 rounds to 5. To counterbalance, the rate of fire is no longer absolutely horseshit, being reduced from 3 seconds to 1. 

### Notes:

The reduced capacity means the weapon has a lower total killing potential, but the rate of fire means it actually has a chance in a gunfight, as the slow rof before made it basically unusable in any circumstances except shoot'n'scooting. I think this will end up being a very positive change for the weapon. For example, at the old rate of fire it would take a minimum of 6 seconds from the first shot to even have a chance to get the two additional shots on target to down an _unarmored_ opponent assuming you _don't miss._ That's pretty terrible, as even non-combat weapons like the toolbox has higher dps.

### Changelog:

CL: [Balance] The surplus rifle has been reworked; it now reloads with 7.12x82mm stripper clips and has had it's ammo capacity reduced from 10 to 5. It's rate of fire has also been tripled. Nukies can now buy ammo for the surplus rifle at 1TC per clip.


PerfectTangent:
new surplus rifle sprite compared to old:
Screenshot 2022-02-23 230050

Bod9001:
rebased to get tests working

corp-0 (approved):

corp-0:
ask @Bod9001  what's the process to unfuck these PRs now that the catalogue has been changed to do id's on runtime

PerfectTangent:
I'll redo them when I get some time.

Bod9001:
Theoretically just make sure the elements get added properly

PR #8429, opened 2022-02-24
Stops integration tests causing errors on Builds

Bod9001:
Tess Singleton was in the, resources folder so made a custom class for it since its meant to be editor only


UnityProject/Assets/Tests/PlayModeTests/GameRunTests.cs
```diff
@@ -17,6 +17,9 @@ public IEnumerator NewTestScriptWithEnumeratorPasses()
 			}
 			GameManager.Instance.QuickLoad = true;
 
+			yield return null;
```
riquelme-A: why two nulls?

corp-0 (approved):

NoooneyDude (approved):

PR #8430, opened 2022-02-25
Updates addressables Maybe fixes Mac error

Bod9001:
need someone with a Mac to try this


corp-0:
what mac error?

Bod9001:
on mac os it doesn't seem to find the addressable bundles in the editor, meaning that there is no sounds or music and It just spams a bunch of errors

JustinTrotter:
Tested. Fixes the problem when running the game on the editor on my Mac. Thanks a bunch Bod.

corp-0 (approved):

NoooneyDude:
Tests are having a lot of trouble with this, not completing. Also noticed that version isn't available in the package manager on 2021.1.14f, but you're on .28, right? Probably available on that version, might need to update to .28?

NoooneyDude:
Addressables updated, no longer necessary.

PR #8434, opened 2022-02-26
Update Mirror to our packaged version

NoooneyDude:
- Removes unpackaged Mirror 42.2.11 and replaces it with our packaged version found [here](https://github.com/unitystation/Mirror) (64.0.4 currently).
Updates the project to comply with this version.
- Updates Ignorance to 1.4.0 LTS Release 1.
- Updates packages.

> Tested with two clients on headless.

>  `CustomNetSceneChecker` is commented out as the Mirror update means interest management is done with a global script and not with a component on every prefab. Will create a followup PR to remove this component from all prefabs.


Bod9001:
should we have this built after the playtest?

NoooneyDude:
The Mirror / Telepathy updates might improve performance. Up to you. This is a somewhat large change. I could successfully play with two clients on headless locally. I would personally make a build now before merging this, then make a second build. That means we can revert the staging servers to the previous image if things go awry.

Bod9001:
ok cool

Bod9001:
tests fail rip

ThatDan123:
You cant remove CustomNetSceneChecker as theres stuff needed for joining which uses it, the observer stuff is just a mess atm. And the global interest management can't be used yet for the same reason and tbh the global one isnt that useful as we need to check stuff like inventory.

NoooneyDude:
Hmm I'm not sure what you mean. The `CustomNetSceneChecker` logic was moved to `CustomInterestManagement` with not much difficulty. We are passed the `NetworkIdentity` component as a parameter and so can do anything the old system could - right?

I opted to create a new script instead of updating `CustomNetSceneChecker` because Mirror was complaining about using the script more than once and was very laggy. Didn't want to make a lot of prefab changes in this PR by removing that component from everything else.

NoooneyDude:
Don't know how to fix this. Tests pass locally with no `NetworkManager` prefab changes to commit.
Removing a few entries and running the test again correctly updates the prefab to what it was.

```
Assets AmbientTile, Power Sink, solar panel transformer, SolarPanel, SolarPanelController, MimeBox, MimeChair, MimeSpeech, MimeWall, SimpleSpell weren't in the NetworkManager spawnPrefabs list.
If the lists were updated you NEED to commit the changed NetworkManager prefab.at Tests.Asset.NetworkManagerTests.SpawnableListTest
```

NoooneyDude:
Turns out Mirror was adding a NetworkIdentity component to those prefabs as they didn't already have it and had networked components, but wasn't correctly saving the prefab so I couldn't commit the changes.

corp-0 (approved):

ThatDan123 (approved):

PR #8435, opened 2022-02-26
Fixes antags stealing objectives failing

### LINKED ISSUE #7814: Some Objectives are broken
Reported on discord 

image

Player said they had the items and were on the shuttle when docked.

-Might be an issue with the recursive container checking, for the items.
-Unsure of the escape one as that code seems like it should work, needs debugging.

Needs investigating, unsure if you can just spawn as traitor using admin tools and then seeing if objectives complete.

MaxIsJoe:
fixes #7814

Also adds in a new console command for quick debugging without having to end the round every single time to test objectives.

CL: [Fix] fixes a bug causing steal objectives to fail.

### DIFF
--- a/UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
+++ b/UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
@@ -4,10 +4,8 @@ using System.Collections.Generic;
 using UnityEngine;
 using UnityEditor;
 using Systems.Atmospherics;
-using Systems.Cargo;
 using Random = UnityEngine.Random;
 using DatabaseAPI;
-using Items;
 using Messages.Client;
 using Messages.Server;
 using Messages.Server.HealthMessages;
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
```diff
@@ -20,6 +18,28 @@ namespace IngameDebugConsole
...
+			if (playerSpawned == false)
+			{
+				Logger.LogError("Player has not spawned yet to be able to check for their objectives!");
+				return;
+			}
+			if (PlayerManager.LocalPlayerScript.mind.IsAntag == false)
+			{
+				Logger.LogError("Player is not an antagonist!");
+				return;
+			}
```
NoooneyDude, line 34: Does this still work if you're a ghost?

NoooneyDude (approved):

UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
Bod9001: works if you're a ghost

Bod9001 (approved):

PR #8442, opened 2022-03-01
Off line editor

Bod9001:
should hopefully make builds and editor work when there is no Internet or HUB API is down


Bod9001:
note : didn't let me start from lobby but then I tried it again and it works fine so, wasn't able to test what was stopping me logging on the first time

NoooneyDude (approved):

corp-0 (approved):

PR #8453, opened 2022-03-09
Fixes infinite firelocks / doors

### LINKED ISSUE #8432: Seems to be infinite firelocks
### Environment
- Game Version: 22022318
- OS Version: Catalina

When North to South facing airlock is broken it leaves behind what seems to be an infinite amount of airlocks. Oh and it doesn't really break either.

### Steps to reproduce

1. Robust a North to South facing airlock until it "breaks"
2. Observe the leftover airlocks. 
Observe the non-broken airlock. 
image

MaxIsJoe:
fixes #8432 

CL: [Fix] fixed duplicating firelocks and airlocks

### DIFF
--- a/UnityProject/Assets/Scripts/Health/Objects/Integrity.cs
+++ b/UnityProject/Assets/Scripts/Health/Objects/Integrity.cs
@@ -126,7 +126,7 @@ public class Integrity : NetworkBehaviour, IHealth, IFireExposable, IRightClicka
 	private RegisterTile registerTile;
 	public RegisterTile RegisterTile => registerTile;
 	private IPushable pushable;
-	public Meleeable Meleeable => GetComponent<Meleeable>();
+	public Meleeable Meleeable;
 
 	//The current integrity divided by the initial integrity
 	public float PercentageDamaged => integrity.Approx(0) ? 0 : integrity / initialIntegrity;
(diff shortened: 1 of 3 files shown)


Bod9001 (dismissed):

UnityProject/Assets/Scripts/Health/Objects/Integrity.cs
```diff
@@ -125,6 +126,7 @@ public class Integrity : NetworkBehaviour, IHealth, IFireExposable, IRightClicka
 	private RegisterTile registerTile;
 	public RegisterTile RegisterTile => registerTile;
 	private IPushable pushable;
+	public Meleeable Meleeable => GetComponent<Meleeable>();
```
corp-0: it is better to cache this reference. Like it is right now it will call GetComponent every time the ``Meleeable`` property is used anywhere in the consumer code. Yeah, currently it is only used in ``WhenDestroyed`` from ``ConstructibleDoor`` component, so we won't notice it until it becomes a problem later in the future.

corp-0 (changes requested):

MaxIsJoe:
review pls

Bod9001 (approved):

corp-0 (approved):

PR #8464, opened 2022-03-14
Fixes emote sounds being global

MaxIsJoe:
title

CL: [Fix] Fixed emote sounds being global. No more screams of the damned.


NoooneyDude (dismissed):

Fr-Dae:
if you can add a craft and an object
to be able to gag/muzzle someone.

to silence the prisoners without having to knock them out.

MaxIsJoe:
> if you can add a craft and an object to be able to gag/muzzle someone.
> 
> to silence the prisoners without having to knock them out.

The muzzle is already something in-game that should be mapped into the security department, i don't have any plans to add a crafting recipe for them and if i wanted it would be in a separate PR.

corp-0:
Did you test it actually works? The global parameter isn't what you would imaginr initially

MaxIsJoe:
> Did you test it actually works? The global parameter isn't what you would imaginr initially

It was set to true by default, this sets it to false.

corp-0:
Yes but again, the global parameter just has a confusing name. Making global true just means it will send the message to everyone and don't use the hardcoded 15 tiles distance. It doesn't mean everyone will hear it, it doesn't mean it will make the sound not spacial, it only means the message will be sent to everyone and each one of them will evaluate if they hear it and how.

Bod9001:
any changes wanted?

MaxIsJoe:
Alright fixed this for real this time

sounds were marked as 2D instead of 3D which made them global

corp-0 (approved):

Bod9001 (approved):

PR #8472, opened 2022-03-20
Restore radial menu look

Copern:
It's been a long time since I've checked on UnityStation, hence why this was not out much, much sooner. It looks like the initial issues with the menu happened after the prefabs were moved around. Somehow the references in the prefabs were lost during the move, thanks unity? At some point in the patching process to fix the issue, the colours, sizes of the images, and references were set incorrectly.

Fixes the colours and the image sizes to reflect how the radial used to look.

Replaces a still missing reference.

Now verifies more references and logs what reference is missing and in which prefab/object. Attempts to find the missing component by searching the prefab's children if the reference isn't set.

Removed a couple obsolete/unused things.

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Core/Radial/Radial.cs
+++ b/UnityProject/Assets/Scripts/UI/Core/Radial/Radial.cs
@@ -91,7 +91,7 @@ namespace UI.Core.Radial
 
 			var prefabName = gameObject.name.Replace("(Clone)", string.Empty);
 
-			Logger.LogError($"{prefabName} prefab is missing a reference for {missingReference}. Functionality may be hindered or outright broken.", Category.UI);
+			Logger.LogError($"{prefabName} prefab is missing a reference for {missingReference}. Functionality may be hindered or broken.", Category.UI);
 
 			return component;
 		}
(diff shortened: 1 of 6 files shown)


NoooneyDude (approved):

UnityProject/Assets/Scripts/UI/Core/Radial/Radial.cs
```diff
@@ -84,6 +85,35 @@ public virtual int MaxShownItems
 
 		public int ShownItemsCount => Math.Min(TotalItemCount, MaxShownItems);
 
+		public V VerifyNonChildReference<V>(V component, string missingReference) where V : UnityEngine.Object
```
corp-0, line 88: These two look very useful. Couldn't we move them to somewhere more general?

Bod9001 (approved):

UnityProject/Assets/Scripts/UI/Core/Radial/Radial.cs
Copern: I was thinking of having them as extension methods but wasn't sure if they would be used outside of the radials. I could look into that tomorrow.

corp-0: Ok, let's merge this for now then

PR #8487, opened 2022-03-22
Ensure verification only happens once.

Copern:
If a reference is missing, a simple component is added to the offending object. This is to ensure that, in the event that a child component isn't found, accessing the component reference doesn't keep going through the hierarchy to find a component that doesn't exist. The helper component is only ever added or checked if a reference has failed verification. Verification will only log the event and try to find a valid child component once.


NoooneyDude:
Nice touch!

corp-0 (approved):

PR #8490, opened 2022-03-22
Game window title will now change to reflect what server you are on

### LINKED ISSUE #3292: Display Server Name In-Game
Players currently have no indication of what their current server is. We might want to display this information somerwhere, maybe on the pause screen.

The server name should be retrieved from ```ServerData.ServerStatus.cs``` and displayed somewhere. ~~Alternatively we could make the in-game name separate from ```ServerData.ServerStatus.cs``` and store it in a ```.txt``` file instead. I'm not sure which option is better.~~ nnaaaah.

MaxIsJoe:
while browsing through older issues i found an issue saying that players can't know what server they are on, this is of course very old as you can see what server you are on when you join a server now with the MOTD window. 
However I've decided to add an extra thing to this so it can be more in parity with SS13 with how detailed some servers' make the game's window title be some times.

### Notes :
At some point we should make this method public and let admins change the title themselves like how admins can do on UnityStation
Works sadly only with windows because Unity is not as advanced as other cooler game engines like Godot.

closes #3292

### DIFF
--- a/UnityProject/Assets/Scripts/Util/WindowNameChanger.cs
+++ b/UnityProject/Assets/Scripts/Util/WindowNameChanger.cs
@@ -33,7 +33,8 @@ namespace Util
 			//Get the window handle.
 			var windowPtr = FindWindow(null, Application.productName);
 			//Set the title text using the window handle.
-			SetWindowText(windowPtr, $"{Application.productName} Build v{Application.version} - {ServerInfoUI.Instance.ServerName.text} ||" +
+			var serverName = ServerInfoUI.Instance.ServerName.text != "" ? ServerInfoUI.Instance.ServerName.text : "Nameless Server";
+			SetWindowText(windowPtr, $"{Application.productName} Build v{Application.version} - {serverName} ||" +
 			                         $" {GameManager.Instance.GetGameModeName()} on {SubSceneManager.ServerChosenMainStation}");
 		}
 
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Util/WindowNameChanger.cs
```diff
@@ -0,0 +1,44 @@
...
+			EventManager.AddHandler(Event.PlayerRejoined, ChangeTile);
+			EventManager.AddHandler(Event.RoundStarted, ChangeTile);
+		}
+
+		private void ChangeTile()
+		{
+			//Get the window handle.
+			var windowPtr = FindWindow(null, Application.productName);
+			//Set the title text using the window handle.
+			var serverName = ServerInfoUI.Instance.ServerName.text != "" ? ServerInfoUI.Instance.ServerName.text : "Nameless Server";
```
Bod9001, line 36: is there anything that could happen from malicious strings from server with this?

MaxIsJoe: I looked into it and I couldn't find anything. There is probably zero way you can do anything malicious by changing the window's title and even if there is we'd probably solve it when we get to the sandboxing/security development phase and open up stationhub to community servers.

Bod9001: ok cool

Bod9001: the only issue by looks like is if you spam change it causing Windows to blue screen but doesn't seem likely with this code so as all good

Bod9001 (approved):

PR #8491, opened 2022-03-22
Adds working file cabinets

### LINKED ISSUE #2327: Add folders and file cabinets

MaxIsJoe:
finishes and closes #2327

CL: [New] Working file cabinets are a thing now, mappers can map them in later.


UnityProject/Assets/Scripts/Util/SweetExtensions.cs
```diff
@@ -16,6 +16,11 @@ public static IPushable Pushable(this GameObject go)
 		return go.GetComponent<IPushable>();
 	}
 
+	public static Pickupable Pickupable(this GameObject go)
```
corp-0: uuuuuhm why?

Copern: I think it's used often enough to be useful as a convenience method. Maybe a better name would be "AsPickupable". Since "as" is a conversion operator that returns null if conversion fails. I think As[SomeComponent] would be enough to convey the idea that it could be null. Also, keep in mind that "go" could technically be null too. Might want to add an "OrNull()?." to it. Has the added benefit that if someone is only trying to get the pickupable component, they don't have to check the gameobject for null too.

MaxIsJoe: Same thing i said on discord

NoooneyDude: Copern's `OrNull()?` suggestion is sound and is what most of the other similar extension methods use. Can you change that too? Bonus points if you update `Pushable()` as well!

MaxIsJoe: done

NoooneyDude (approved):

Bod9001:
merged into develops if this pokes tthe tests

Bod9001:
looks like it doesn't want to work, can you run the tests locally?

MaxIsJoe:
It was an issue from github not the PR, tests passed here as well.
merge

PR #8480, opened 2022-03-21
Turns all explosives into bounced signal devices

MaxIsJoe:
part of #6149 

No more pairing to get results, fire in the hole!

Issue #6149, opened 2021-03-01
C4, Syndicate bombs, and Power Sinks Bounty B:120


PerfectTangent:
### Description

This bounty calls for the addition of C4 explosives (and their more powerful variant, X4), syndicate bombs, and power sinks to the syndicate TC uplink. 

C4 is a stickable brick of explosives that comes integrated with a configurable timer and remote signalling device. It can be applied to pretty much any object or item to stick it to them after 3 uninterrupted seconds. Once it has been stuck to an item, the timer starts (minimum of 10 seconds) before it detonates, unless it has been switched to the remote signalling mode in which it can be paired with a remote signaller to be detonated remotely.

Syndicate bombs are large, bulky explosives that must be dragged around. When they are order from the TC uplink, players get a small beacon that, when activated, teleports the Syndicate bomb to the player's position. Syndicate bombs can be wrenched down to make them immobile like a locker, and have a configurable timer (minimum 90 seconds) and remote signalling device. The bomb makes a loud beeping as it counts down before it explodes.

Power sinks are devices that, when screwed down into an electrical grid, can be turned on either manually or remotely with a remote signaller to start draining power from the grid at a rapidly increasing speed. Once it absorbs enough energy, it explodes.

Pizza bombs are pizza boxes that have a bomb attached to the lid. Once the pizza bomb is armed, it explodes when the lid is opened. As a departure from the implementation of the bomb in /tg/, the pizza bomb should be able to be configured to blow up either after a set time, when the lid is opened, or with a remote signaller.

### what is needed to fulfill the bounty:

- [x] Add C4 and X4 explosives
- [x] Add syndicate bombs
- [x] Add power sinks
- [x] Add pizza bombs
- [x] Fix remote signallers and ensure they are compatible with everything above.

### Notes:

It should also be noted that as a general change in gameplay, all four devices here should have the ability to accept and receive remote signals from signallers without having to first attach a remote signaller to the device.

AutumnNova:
ill take this bounty if thats alright

PerfectTangent:
Since Autumn's War Ops bounty was finished a bit ago and only got merged after the freeze because of delays with the reviews, I think it's fair that she can work on this one. She had previously expressed interest with it several times before hand, and mtk's initial work is close to being finished, so It's being let through.

MaxIsJoe:
I'll work on this

MaxIsJoe:
syndicate bombs are already in so idk why it hasn't been checked off

[...]

PR #8493, opened 2022-03-24
Bomb signal pairing

MaxIsJoe:
work on #6149

Bombs by default can receive any bounced signal from remote signaling devices as of #8480 at the request of Jack. This PR aims to add the ability for antags to make bombs only accept a signal from a specific remote signaler(s).

CL: [Balance] Bombs have a random frequency on spawn now.
Bombs will only work with paired remote signalers, the bomb will automatically copy the remote's frequency on pair.

On the backend side a bit, the `ReceiveSignal()` function is now aware of the emitter regardless if it was a PING or BOUNCED signal.

~~Tip : If a traitor doesn't pair their bombs with their a remote, you can sabotage their bombs by pairing any kind of emitter to the bomb. However, you can't override any devices it was paired to.~~

### DIFF
--- a/UnityProject/Assets/Scripts/Communications/SignalReceiver.cs
+++ b/UnityProject/Assets/Scripts/Communications/SignalReceiver.cs
@@ -29,7 +29,7 @@ namespace Communications
 		/// <summary>
 		/// Logic to do when
 		/// </summary>
-		public abstract void ReceiveSignal(SignalStrength strength, ISignalMessage message = null);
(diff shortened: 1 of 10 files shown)


MaxIsJoe:
Tests passed, Jack's requested changes have been made. It's review ttiimmmeeee

NoooneyDude (approved):

Bod9001 (approved):

PR #8497, opened 2022-03-26
Nerfs hunger slightly because it's annoying as fuck

MaxIsJoe:
title

CL: [Balance] Nerfed the rate at which hunger builds. It now takes an hour and 15 minutes to become starving.

### DIFF
--- a/UnityProject/Assets/Prefabs/Player/Resources/BodyParts/Felinid/Felinid.asset
+++ b/UnityProject/Assets/Prefabs/Player/Resources/BodyParts/Felinid/Felinid.asset
@@ -66,3 +66,4 @@ MonoBehaviour:
     RootImplantProcedure: {fileID: 11400000, guid: 3de247ed9ca572941975e45430051197,
       type: 2}
     BodyPartsThatShareTheSkinTone: []
+    NumberOfMinutesBeforeStarving: 75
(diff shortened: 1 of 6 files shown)


corp-0:
I just wonder if 75 is the number we're looking for

MaxIsJoe:
> I just wonder if 75 is the number we're looking for

As discussed on discord : 

- anything above 60 is the acceptable range due to how annoying hunger is as a mechanic and how it can lead people to their death sometimes
- the time isn't actually 75 as a lot of food is usually not fuffling enough to stop people's hunger fully
- chefs currently cannot feed over 20 players at once and if they die or are slow this problem gets worse for everyone
- chef's lack of materials and easy access to ingredients means that once what's in his kitchen gets consumed he literately has to rely on cargo or botany to bring the ingredients and food for him and sometimes both these jobs are easier busy with other stuff OR not interested in helping the kitchen at all.
- Vending machines have a limited amount of resources, if someone eats everything in a vendor then no one has a way to quickly get back on their feet when they're not playing a role that involves them being on the station 24/7 (i.e nukies or shaft miners)

corp-0:
Alright, fair enough

corp-0 (approved):

PR #8498, opened 2022-03-26
Animates crit indicator

MaxIsJoe:
CL: [New] instead of instantly appearing or change values, the crit indicator now blinks between it's last value and newest one.


MaxIsJoe:
Just a heads up about this, i don't know how to animate the alpha so only the radius is animated.

if anyone knows how to do this before this is merged, please tell. If not, it can be done in another day.

corp-0:
What does it look like?

MaxIsJoe:
> What does it look like?

It's basically the crit indicator but instead of the radius popping in when you get hit, it is now lerped (aka animated)

NoooneyDude (approved):
Would be nice to not constantly be lerping if we don't need to.

A momentary damage-received overlay would be a nice touch at some point in the future.

PR #8514, opened 2022-03-27
Fixes Mob attack cooldowns

### LINKED ISSUE #7830: Mobs attack cooldown is broken, making mobs way too deadly. B:40
### Environment
- Game Version: Staging 21111506
- OS Version: Arch Linux

Mobs when they attack you can hit you twice in a single second and some are incredibly overpowered.

here you can see a live CCTV footage of your average spaceman getting decimated by a goat.

https://user-images.githubusercontent.com/80334192/142127345-f63ca864-f024-4583-b7b7-19ed4fe1cc39.mp4

### Steps to reproduce

1. go to the kitchen/spawn a goat
2. punch goat
3. die by the goat

MaxIsJoe:
Fixes #7830

this is the silliest coding mistake i've seen lol, Mobs had a cooldown variable but it was used on their animations.. not their actual actions.

This actually introduces a proper cooldown to their melee actions and will no longer hit you 10000000000 times a second.

CL: [Fix] fixed mob attack cooldowns.

### DIFF
--- a/UnityProject/Assets/Prefabs/Mobs/Living/SimpleAnimals/Hostile/Xenomorphs/Facehugger.prefab
+++ b/UnityProject/Assets/Prefabs/Mobs/Living/SimpleAnimals/Hostile/Xenomorphs/Facehugger.prefab
@@ -7,6 +7,21 @@ PrefabInstance:
   m_Modification:
     m_TransformParent: {fileID: 0}
     m_Modifications:
+    - target: {fileID: 941816724250882487, guid: cf769d410d94f0e47bab5eee03a3b07d,
+        type: 3}
+      propertyPath: m_LocalRotation.x
+      value: -0
+      objectReference: {fileID: 0}
+    - target: {fileID: 941816724250882487, guid: cf769d410d94f0e47bab5eee03a3b07d,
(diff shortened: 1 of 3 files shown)


Bod9001 (dismissed):

waffielz:
https://user-images.githubusercontent.com/15893891/160304684-930455e3-e9c8-4eb8-8ebf-3d0df44aae9f.mp4

corp-0:
From what I remember, MobMeleeAction does an action when reaching the target and was only used for facehuggers. Does MobMeleeAttack extend this class? If not, you will sadly have to recreate this cooldown there too

MaxIsJoe:
> From what I remember, MobMeleeAction does an action when reaching the target and was only used for facehuggers. Does MobMeleeAttack extend this class? If not, you will sadly have to recreate this cooldown there too

Let me test this and if that's the case i'll fix facehuggers as well

MaxIsJoe:
Yup, fixed huggers. Extended their attack cooldown to 3 seconds as well instead of 1 since their attack is technically not a "melee" attack.

Bod9001 (approved):

Issue #5289, opened 2020-10-12
Blood miner, Leaper, and Drake do not have properly assigned sprites and are invisible in the game.


Akronymus-git:
### Detailed Bug Report
build: latest staging

### Steps to Reproduce

1. Find a blood miner, Leaper, or Drake.

NoooneyDude:
Related to #5325.

PetMudstone:
I have given the NPCs in question new prefabs that from testing, have working sprites. I think this issue can be closed now.

PerfectTangent:
K.

PetMudstone:
Scratch that, the mobs mentioned above do not have working sprites out of editor, even as local host. I do not know why their sprites only work in editor but I suspect it may be due to the large dimensions of said sprites. That doesn't explain why the xenomorph queen sprites work out of editor while having similar dimensions, however.

PR #8516, opened 2022-03-27
Fixes ash drake and blood miner broken prefabs

MaxIsJoe:
Fixes ash drake and blood miner sprites being invisible and not having a prefab tracker


PetMudstone:
How did you fix the sprites? I tried looking at it before but couldn't figure out what was wrong.

MaxIsJoe:
> How did you fix the sprites? I tried looking at it before but couldn't figure out what was wrong.

Simple, there were no sorting groups so they just looked invisible. Adding one and setting them to NPC fixes them

MaxIsJoe:
image

look at them!

corp-0 (approved):

PerfectTangent:
is the leaper sprite still invis, as described in #5289? I presume so. Perhaps it needs the same treatment.

MaxIsJoe:
> is the leaper sprite still invis, as described in #5289? I presume so. Perhaps it needs the same treatment.

ah, i didn't look at the leaber, will open a PR for it soon.

PR #8521, opened 2022-03-28
Expanded acccess

Bod9001:
why? 
since we don't have 60+ pops most of the time you end up having to cover two jobs in one, 
so I've expanded the access to

Engineering, + atmospherics
chemist, + surgery
Medical doctor, + chemistry
bartender, + kitchen
cook, + bar + botany
botanist, + kitchen

### DIFF
--- a/UnityProject/Assets/ScriptableObjects/Occupations/Engineering/Engineer.asset
+++ b/UnityProject/Assets/ScriptableObjects/Occupations/Engineering/Engineer.asset
@@ -16,11 +16,12 @@ MonoBehaviour:
   isCrewmember: 1
   lateSpawnIsArrivals: 1
   inventoryPopulator: {fileID: 11400000, guid: aeb66e4be690d4b7e9ca83169dc2871a, type: 2}
+  useStandardPopulator: 1
   limit: 5
   priority: 88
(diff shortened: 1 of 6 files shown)


ilinm1:
Won't it be better idea to check if it's actually lowpop on round start and give access only then?

NoooneyDude:
Could we have a comment on why this is on hold? Thanks.

Bod9001:
I think Jack wanted it to be a game mode not sure though?

CdrCross:
> Won't it be better idea to check if it's actually lowpop on round start and give access only then?

This is certainly what TG does. If the number of players  is below the "Skeleton crew" threshold at roundstart, expanded access is granted to everyone. Ideally, we should make it so that everyone who joins until the crew exceeds skeletal levels get expanded access, but everyone after does not.

ilinm1:
Ik, just don't like this temporary solution, we might get medpop at playtest, and with "hardcoded" expanded access it will be uncomfortable

craftxbox:
This feels like it should be configurable server-to-server, rather than hard-coded

PerfectTangent:
closing because of the merging of #8599.

PR #8524, opened 2022-03-28
Update ChatRelay.cs

MaxIsJoe:
CL: [Fix] Fixed radio sounds being played when hitting stuff

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Chat/ChatRelay.cs
+++ b/UnityProject/Assets/Scripts/Core/Chat/ChatRelay.cs
@@ -310,28 +310,10 @@ public class ChatRelay : NetworkBehaviour
 			case ChatChannel.Security:
 				_ = SoundManager.Play(Chat.Instance.commonSecurityChannelSound);
 				break;
-			case ChatChannel.Common:
-				_ = SoundManager.Play(Chat.Instance.commonRadioChannelSound);
-				break;
-			case ChatChannel.Engineering:
-				_ = SoundManager.Play(Chat.Instance.commonRadioChannelSound);
-				break;
-			case ChatChannel.Science:
-				_ = SoundManager.Play(Chat.Instance.commonRadioChannelSound);
-				break;
-			case ChatChannel.CentComm:
-				_ = SoundManager.Play(Chat.Instance.commonRadioChannelSound);
-				break;
-			case ChatChannel.Supply:
-				_ = SoundManager.Play(Chat.Instance.commonRadioChannelSound);
-				break;
-			case ChatChannel.Command:
-				_ = SoundManager.Play(Chat.Instance.commonRadioChannelSound);
-				break;
-			case ChatChannel.Medical:
-				_ = SoundManager.Play(Chat.Instance.commonRadioChannelSound);
-				break;
-			case ChatChannel.Binary:
+			case var c when c.HasFlag(ChatChannel.Binary) ||
+			                c.HasFlag(ChatChannel.Medical) || c.HasFlag(ChatChannel.Command)
+			                || c.HasFlag(ChatChannel.Supply) || c.HasFlag(ChatChannel.CentComm) ||
+			                c.HasFlag(ChatChannel.Science) || c.HasFlag(ChatChannel.Engineering) || c.HasFlag(ChatChannel.Common):
 				_ = SoundManager.Play(Chat.Instance.commonRadioChannelSound);
 				break;
 		}
--- a/UnityProject/Assets/Scripts/Core/Chat/ChatRelay.cs
+++ b/UnityProject/Assets/Scripts/Core/Chat/ChatRelay.cs
@@ -310,10 +310,28 @@ public class ChatRelay : NetworkBehaviour
 			case ChatChannel.Security:
 				_ = SoundManager.Play(Chat.Instance.commonSecurityChannelSound);
 				break;
-			case var c when c.HasFlag(ChatChannel.Binary) ||
-			                c.HasFlag(ChatChannel.Medical) || c.HasFlag(ChatChannel.Command)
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Core/Chat/ChatRelay.cs
```diff
@@ -304,29 +304,34 @@ public void AddMentorPrivMessageToClient(string message)
...
+				_ = SoundManager.Play(Chat.Instance.commonRadioChannelSound);
 				break;
-			case ChatChannel.Security:
-				_ = SoundManager.Play(Chat.Instance.commonSecurityChannelSound);
+			case ChatChannel.Medical:
+				_ = SoundManager.Play(Chat.Instance.commonRadioChannelSound);
 				break;
-			default:
+			case ChatChannel.Binary:
 				_ = SoundManager.Play(Chat.Instance.commonRadioChannelSound);
```
Alainx277: It should be possible to collapse all these cases with something like:
```csharp
case ChatChannel.Common:
case ChatChannel.Engineering:
case ChatChannel.Science:
    SoundManager.Play(Chat.Instance.commonRadioChannelSound);
    break;
```

UnityProject/Assets/Scripts/Core/Chat/ChatRelay.cs
```diff
@@ -304,29 +304,34 @@ public void AddMentorPrivMessageToClient(string message)
 
 		switch (channels)
```
Alainx277, line 305: It should be noted that the switch doesn't understand combinations of channels. I'm not sure if that can happen with the listed ones, can anyone else confirm that it doesn't?

Alainx277 (changes requested):

UnityProject/Assets/Scripts/Core/Chat/ChatRelay.cs
corp-0:
```cs
switch (channels)
{
    case var c when c.HasFlag(ChatChannel.Common):
        DoCommonThing();
        break;
    case var c when c.HasFlag(ChatChannel.Action):
        DoActionThing();
        break;
    ...
}
```

Something like that should work. I imagine you could use the fancier switch expressions as well

MaxIsJoe: there we go

UnityProject/Assets/Scripts/Core/Chat/ChatRelay.cs
```diff
@@ -304,29 +304,16 @@ public void AddMentorPrivMessageToClient(string message)
...
-			case ChatChannel.Warning:
-				break;
-			case ChatChannel.Blob:
-				break;
-			case ChatChannel.OOC:
-				break;
 			case ChatChannel.Syndicate:
 				_ = SoundManager.Play(Chat.Instance.commonSyndicteChannelSound);
 				break;
 			case ChatChannel.Security:
```
NoooneyDude, line 310: These two cases haven't been updated to use `HasFlag()`. Speaking of which, from what I recall using `HasFlag()` has performance issues - might be allocating or something. Probs not too much of an issue given we're using it throughout chat but something to keep in mind.

Copern: HasFlag boxes the values twice. That was an issue a while back with the IsInLayer method used in a hot loop that caused dozens to hundreds of megabytes of garbage. Probably not going to be an issue with chat.

corp-0: We can always go safe and use bitwise operations instead

MaxIsJoe: i didn't want to create any more problems by using bitwise operators since i know ill screw it up somehow so i chose the safest path

corp-0:
why don't you collapse all cases that play the common sound?

MaxIsJoe:
> why don't you collapse all cases that play the common sound?

HasFlag() was said to create a lot of GC and i dont know how to do bitwise operations correctly without screwing it up

corp-0:
yeah but that doesn't change the fact that a lot of those cases have the same result so you can collapse them together. Also if you want to change it to use bitwise operation if would be like this 
```cs
case var c when (c & Channel.Common) == Channel.Common;
```
...I think, would hav to test it xd

corp-0:
according to this thing: https://docs.microsoft.com/en-us/dotnet/api/system.enum.hasflag?redirectedfrom=MSDN&view=net-6.0#remarks that's exactly what HasFlag does so I don't know if we should bother to use the ugly bitwise operation syntax directly

Copern:
> according to this thing: https://docs.microsoft.com/en-us/dotnet/api/system.enum.hasflag?redirectedfrom=MSDN&view=net-6.0#remarks that's exactly what HasFlag does so I don't know if we should bother to use the ugly bitwise operation syntax directly

It doesn't quite do just that, at least in the current version of unity we use. There's still two allocations it makes per call. It sure would be nice if unity moved over to CoreCLR.

I don't think we need the bit checks for chat stuff though. If by some miracle garbage becomes a problem with chat, it can be changed later.

corp-0:
Hey Max, collapse the cases that use the common sound and we will be ready to merge this

MaxIsJoe:
@corp-0 like this?

corp-0:
that works but doing like this is cleaner imo:
```cs
switch (stuff) 
{
    case firstCase:
    case secondCase:
    case thirdCase:
        commonResult();
        break;
    case differentCase:
        diffferentResult();
        break;
}

MaxIsJoe:
now?

Bod9001 (approved):

PR #8526, opened 2022-03-29
Radial consolidation

Copern:
Adds in the ability to use the same right click radial for a quick single ring radial with some tweaks.

Airlock painter selection looks similar and opens the same way, but makes use of the right click radial and removes the duplicated radial.

Fixes Airlock Science Assembly (it used the command door prefab).

CL: [Improvement] If the radial is opened and there is only one object on the tile, the actions for that object are placed directly on the main ring (no action ring). This includes action bar objects. There is no delay when hovering over items when it is a single ring.

Moves some right click related scriptable objects around and renames a couple things.


UnityProject/Assets/Scripts/Items/Tool/AirlockPainter.cs
```diff
@@ -45,16 +58,29 @@ public void OnSpawnServer(SpawnInfo info)
...
-			GameObject chosenPaintJob = await UIManager.RadialMenu.ShowRadialMenu(AvailablePaintJobs, performer);
-			if (chosenPaintJob)
+			var result = new List<RightClickMenuItem>();
+
+			for (var i = 0; i < objects.Count; i++)
 			{
-				int chosenPaintJobIndex = AvailablePaintJobs.IndexOf(chosenPaintJob);
-				PlayerManager.LocalPlayerScript.playerNetworkActions.CmdSetPaintJob(chosenPaintJobIndex);
+				var index = i; // Copy for action function
+				Action setPaint = () => PlayerManager.LocalPlayerScript.playerNetworkActions.CmdSetPaintJob(index);
```
corp-0, line 78: I don't get it, why couldn't you just use 'i'?

Copern: Had a feeling that would come up. TLDR: If 'i' was used, all the actions would choose the last airlock in the list.

The reason has to do with the way variables are captured in anonymous functions/lambdas. The lambdas would capture a reference to 'i' and all the actions would use that reference. Once the loop finishes, 'i' will be the final value in the list. This means all the setPaint actions would be say, 10, after the loop runs. Instead, we create a new variable 'index' on each iteration and copy the value of 'i' into it. Then the setPaint lambdas capture the individual 'index' references which have the values we want (0, 1, 2... etc).

A foreach doesn't have the same issue with capturing the loop variable as a for loop does. The loop variable is created on each iteration in foreach. I could use a foreach using objects.Select((obj, i) => i) but I figured I would get a similar response ("why not just use a for loop"). I might be able to use a foreach with Enumerable.Range if it's available in the current version of unity. I'll check if unity has it when I get a chance.

corp-0: nah, the for loop is fine. I see why you did it now.

corp-0 (approved):

PR #8528, opened 2022-03-29
Add "neutrally" colored floor tiles.

PetMudstone:
### Purpose
- Adds "neutrally" colored floor tiles. They have lighter tiles. Done at the request of Kavik on Discord.

image

They're located at the bottom of the "Floors" palette.
image


MaxIsJoe:
I wonder if there's a way to let admins or players color these during play..

PetMudstone:
> I wonder if there's a way to let admins or players color these during play..

I think we should wait for a decal rework for that, since otherwise we'd have a staggering amount of colored floor variants available. This is already the case in the tile palette.

SluggyWuggies:
***LETS FUCKING GO.***

corp-0 (approved):

PerfectTangent:
yeah, the future of colored tiling should be overlays. Same with damage to the tiles and dirt, frost, and everything else, even misplaced tiles.

An art project for a later date.

PR #8531, opened 2022-03-31
Added option to not consume healing item on use + Fixes Door crowbar ( thought this branch was already merged oops )

Bod9001:
pretty self-explanatory title
Fixes Door crowbar 

CL: [Balance] organ treatment no longer consumes gauze or ointment when used.

### DIFF
--- a/UnityProject/Assets/ScriptableObjects/Surgery/HealBruteProcedure.asset
+++ b/UnityProject/Assets/ScriptableObjects/Surgery/HealBruteProcedure.asset
@@ -28,3 +28,4 @@ MonoBehaviour:
   Affects: 0
   HeelStrength: 15
   FailAttackType: 0
+  UseUpItem: 1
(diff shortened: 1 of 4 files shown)


corp-0:
So you still need to interact with the item but it isn't consumed, I'm understanding that correctly?

Bod9001:
yeah

corp-0:
I might have not read a discussion about it but why do  we need it to work like this now? It is odd using a consumable item but not actually consuming it.

CdrCross:
> I might have not read a discussion about it but why do we need it to work like this now? It is odd using a consumable item but not actually consuming it.

If the item is consumed, surgery is 100% inferior to just using the item for heals.

corp-0:
I thought surgery was needed to heal damage of internal parts and you couldn't heal those with the regular method of simply clicking body part with item In hand

Bod9001:
you can't heal internal organs by, clicking them with the bandage

corp-0:
ok so I'm back to the first question. Why do we need this?

Bod9001:
Because, it doesn't give any advantage to surgery for doing healing stuff

corp-0 (approved):

PR #8534, opened 2022-04-02
Removes the character screen on joining a server from stationhub

MaxIsJoe:
Makes the game join the server immediately after connecting to one from stationhub.

The idea behind this is that people should edit their characters *after* they have joined instead of *before* so players can get their correct character settings from the server's firebase/django database.
If you edit your characters offline or on another server the character menu in the pre-lobby screen actually loads your **offline** character sheets not the server's character sheets.

Also it makes joining a server now require less clicks on the user's part and they don't have to be forced to make a character to just join a server. (Though they cant join the round still until they have made a character)


corp-0:
very nice

Bod9001 (approved):

Bod9001 (approved):

PR #8536, opened 2022-04-03
Expanded Voting

MaxIsJoe:
Updated voting code to allow for more types of voting so in the future admins can host their own votes and we can replace our current system with a more modular system.

Added two new things to vote for; Game Mode and Map voting.
The voting UI has been updated to be dynamic and is less static now.

CL: [New] admins can now call votes for maps and game modes.

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Voting/VotingButton.cs
+++ b/UnityProject/Assets/Scripts/Systems/Voting/VotingButton.cs
@@ -1,0 +1,22 @@
+﻿using TMPro;
+using UnityEngine;
+
+namespace UI
+{
+	public class VotingButton
+	{
+		public TMP_Text btnText;
+		private VotePopUp popUp;
+
+		public void Initlize(string optionTxt, VotePopUp pop)
+		{
+			btnText.text = optionTxt;
+			popUp = pop;
+		}
+
+		public void OnClick()
+		{
+			popUp.Vote(btnText.text);
+		}
+	}
+}
--- a/UnityProject/Assets/Scripts/Systems/Voting/VotingButton.cs
+++ b/UnityProject/Assets/Scripts/Systems/Voting/VotingButton.cs
@@ -3,19 +3,25 @@ using UnityEngine;
 
 namespace UI
 {
-	public class VotingButton
+	public class VotingButton : MonoBehaviour
 	{
 		public TMP_Text btnText;
 		private VotePopUp popUp;
 
 		public void Initlize(string optionTxt, VotePopUp pop)
 		{
-			btnText.text = optionTxt;
 			popUp = pop;
+			btnText.text = optionTxt;
+			gameObject.SetActive(true);
 		}
 
 		public void OnClick()
 		{
+			if (popUp == null)
+			{
+				Logger.LogError("ITS FAKIN NOT SET YA DAFT CUNT");
+				return;
+			}
 			popUp.Vote(btnText.text);
 		}
 	}
--- a/UnityProject/Assets/Scripts/Systems/Voting/VotingButton.cs
+++ b/UnityProject/Assets/Scripts/Systems/Voting/VotingButton.cs
@@ -17,11 +17,6 @@ namespace UI
 
 		public void OnClick()
 		{
-			if (popUp == null)
-			{
-				Logger.LogError("ITS FAKIN NOT SET YA DAFT CUNT");
-				return;
-			}
 			popUp.Vote(btnText.text);
 		}
 	}
--- a/UnityProject/Assets/Scripts/Systems/Voting/VotingButton.cs
+++ b/UnityProject/Assets/Scripts/Systems/Voting/VotingButton.cs
@@ -8,7 +8,7 @@ namespace UI
 		public TMP_Text btnText;
 		private VotePopUp popUp;
 
-		public void Initlize(string optionTxt, VotePopUp pop)
+		public void Initialize(string optionTxt, VotePopUp pop)
 		{
 			popUp = pop;
 			btnText.text = optionTxt;
--- a/UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
+++ b/UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
@@ -251,7 +251,7 @@ public class VotingManager : NetworkBehaviour
 			var winner = GetHighestVote();
 			if (winner == "")
 			{
-				Chat.AddGameWideSystemMsgToChat($"<color=blue>Voting failed! Not enough people voted!</color>");
+				Chat.AddGameWideSystemMsgToChat($"<color=blue>Voting failed! vote has somehow passed but no winner was written!</color>");
 				return;
 			}
 			switch (voteType)
--- a/UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
+++ b/UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
@@ -1,6 +1,9 @@
-﻿using System.Collections;
+﻿using System;
+using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 using Castle.Core.Internal;
+using Messages.Client.Admin;
 using Messages.Server;
 using Mirror;
 using UnityEngine;
@@ -46,6 +49,10 @@ public class VotingManager : NetworkBehaviour
 	/// </summary>
 	private Coroutine cooldown;
 
+	private List<string> MapList = new List<string>();
+	private List<string> GameModeList = new List<string>();
+	private List<string> yesNoList = new List<string>();
+
 	public enum VoteType
 	{
 		RestartRound,
@@ -63,6 +70,14 @@ public class VotingManager : NetworkBehaviour
 		{
 			Destroy(gameObject);
 		}
+	}
+
+	private void Start()
+	{
+		MapList = SubSceneManager.Instance.MainStationList.MainStations;
+		GameModeList = GameManager.Instance.GetAvailableGameModeNames();
+		yesNoList.Add("Yes");
+		yesNoList.Add("No");
 	}
 
 	void OnEnable()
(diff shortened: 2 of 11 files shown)


MaxIsJoe:
Note about map votes :

I'm 1000% sure that if a server hard reboot due to low performance or it crashing; the game will forget about the vote.

UnityProject/Assets/Scripts/Systems/Voting/VotingButton.cs
```diff
@@ -0,0 +1,23 @@
...
+using UnityEngine;
+
+namespace UI
+{
+	public class VotingButton : MonoBehaviour
+	{
+		public TMP_Text btnText;
+		private VotePopUp popUp;
+
+		public void Initlize(string optionTxt, VotePopUp pop)
```
Alainx277: I think this might be a typo of `Initialize`?

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
```diff
@@ -102,18 +126,62 @@ public void TryInitiateRestartVote(GameObject instigator, NetworkConnection send
...
+			return;
+		}
+
+		votes.Clear();
+		countTime = 0f;
+		prevSecond = 0;
+		voteType = VoteType.NextGameMode;
+		votePolicy = VotePolicy.MajorityRules;
+		voteInProgress = true;
+		RpcOpenVoteWindow("Vote restart initiated by", instigator.name, CountAmountString(), (30 - prevSecond).ToString(), GameModeList);
```
Alainx277: Copy paste mistake? This is a different type of vote.

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
```diff
@@ -224,17 +304,35 @@ private string CountAmountString()
 	}
 
 	/// <summary>
-	/// Tallies the number of players who have voted 'yes'.
+	/// Gets number of coutns
 	/// </summary>
-	/// <returns>The number of 'yes' votes.</returns>
 	private int ForVoteCount()
```
Alainx277: This name does not make sense anymore. You could change it to something like `TotalVotes`.

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
```diff
@@ -224,17 +304,35 @@ private string CountAmountString()
...
+
+	/// <summary>
+	/// Gets the highest vote count on the list
+	/// </summary>
+	/// <returns></returns>
+	private string GetHighestVote()
+	{
+		Dictionary<string, int> count = new Dictionary<string, int>();
+		var highestVote = 0;
+		var winner = "";
```
Alainx277, line 319: This should be `null` by default. Otherwise there will be odd results when no one votes.
Example: With the current code, the game mode could be set to `""`.

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
```diff
@@ -256,18 +354,19 @@ private void RpcFinishVote()
...
 	}
 
 	[TargetRpc]
-	private void RpcVoteCallerDefault(NetworkConnection target)
+	private async void RpcVoteCallerDefault(NetworkConnection target)
 	{
-		if (GUI_IngameMenu.Instance == null) return;
+		if (GUI_IngameMenu.Instance == null || PlayerList.Instance.AllPlayers.Count == 1) return;
 
+		await Task.Delay(500); //Away for the list to be generated before automatically voting
```
Alainx277, line 366: This seems very error-prone. Could you instead work with an event?
If not, at least a to do comment would be nice.

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
```diff
@@ -102,18 +126,62 @@ public void TryInitiateRestartVote(GameObject instigator, NetworkConnection send
...
+			return;
+		}
+
+		votes.Clear();
+		countTime = 0f;
+		prevSecond = 0;
+		voteType = VoteType.NextMap;
+		votePolicy = VotePolicy.MajorityRules;
+		voteInProgress = true;
+		RpcOpenVoteWindow("Vote restart initiated by", instigator.name, CountAmountString(), (30 - prevSecond).ToString(), MapList);
```
Alainx277: Copy paste mistake? This is a different type of vote.

Alainx277 (changes requested):

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
MaxIsJoe: ?

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
MaxIsJoe: it cant because it has to loop through all votes.

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
MaxIsJoe: this is un-needed and i will remove it later

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
```diff
@@ -102,18 +126,62 @@ public void TryInitiateRestartVote(GameObject instigator, NetworkConnection send
...
 		RpcVoteCallerDefault(sender);
 		Logger.Log($"Vote restart initiated by {instigator.name}", Category.Admin);
 	}
 
 	[Server]
-	public void RegisterVote(string userId, bool isFor)
+	public void RegisterVote(string userId, string isFor)
```
Alainx277, line 177: Is `isFor` validated against the possible choices anywhere?

corp-0:
> Note about map votes :
> 
> I'm 1000% sure that if a server hard reboot due to low performance or it crashing; the game will forget about the vote.

only way to solve this would be using the persistent file configuration, maybe a "next_map" field in the json

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
corp-0: the method is for NextGameMode but the text says "restart"

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
```diff
@@ -102,18 +126,62 @@ public void TryInitiateRestartVote(GameObject instigator, NetworkConnection send
 		voteType = VoteType.RestartRound;
 		votePolicy = VotePolicy.MajorityRules;
```
corp-0: all these enums are perfect for a strategy pattern https://refactoring.guru/design-patterns/strategy

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
corp-0: what if no one votes?

MaxIsJoe: > what if no one votes?

it wont pass because the "IsSuccess" function needs a majority to vote. if there is no votes then nothing will happen

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
MaxIsJoe: wdym?

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
MaxIsJoe: This seems like the modular system i was talking about, but right now i think it's good as it is until someone picks this up again and implements admin voting and other things.

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
MaxIsJoe: the `isFor` is recorded and then at the end the game checks who voted for what.
the one string that has been detected the most is the winner.

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
```diff
@@ -305,13 +305,13 @@ private void FinishVote()
...
 	{
-		return $"{ForVoteCount()} / {PlayerList.Instance.AllPlayers.Count}";
+		return $"{TotalVotes()} / {PlayerList.Instance.AllPlayers.Count}";
 	}
 
 	/// <summary>
 	/// Gets number of coutns
 	/// </summary>
-	private int ForVoteCount()
+	private int TotalVotes()
```
corp-0: why make this a method?

MaxIsJoe: I wanted to check if it was null but forgot to add that bit, fixed now.

MaxIsJoe: actually does votes.Clear() even turn it null?

corp-0: no, clear only cleans the elements that the list contains.

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
corp-0: you sure you don't want to test implementing it? It is fun!

MaxIsJoe: Maybe in an another PR, right now im kinda stuck on this because Unity really hates when i make any prefab changes like this then switch PRs and I want to fix other things like the give item tab for admins.

corp-0: *sad design patterns fanboy noises*

UnityProject/Assets/Scripts/Systems/Voting/VotingManager.cs
Alainx277: What is stopping a hacker from sending "trolled" as a vote?

MaxIsJoe: nothing... Until now! I've added a way for the server to check for 'sus' votes

corp-0:
I have one last request before I merge this. Can you move the vote window away from the chat?

MaxIsJoe:
> I have one last request before I merge this. Can you move the vote window away from the chat?

I actually did

corp-0 (approved):

PR #8537, opened 2022-04-04
should hopefully log thread crashes now

Bod9001:
I'm still unsure if it doesn't log it properly
 is there a thread safe log?

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Threading/ThreadBehaviour.cs
+++ b/UnityProject/Assets/Scripts/Core/Threading/ThreadBehaviour.cs
@@ -65,7 +65,7 @@ namespace Core.Threading
 				}
 				catch (Exception e)
 				{
-					Logger.LogError(e.ToString());
+					Logger.LogError(e.ToString(), Category.Threading);
 					throw;
 				}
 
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Core/Threading/ThreadBehaviour.cs
```diff
@@ -58,7 +59,16 @@ private void UpdateMe()
...
 				mainThreadTimer.Restart();
-				RunTick();
+				try
+				{
+					RunTick();
+				}
+				catch (Exception e)
+				{
+					Logger.LogError(e.ToString(), Category.Threading);
+					throw;
```
corp-0, line 69: can you do this? Don't you have to say what exception to throw?

Bod9001: if I put e on the end of it, it just recommends says rethrowing
image

image

corp-0: ah cool, it must infere it from the context then

corp-0 (approved):

PR #8551, opened 2022-04-09
Fixes editor device linking tool

Bod9001:
it was forgetting to set the interface master dirty

### DIFF
--- a/UnityProject/Assets/Scripts/EditorButInAssetsAssembly/DeviceLinker.cs
+++ b/UnityProject/Assets/Scripts/EditorButInAssetsAssembly/DeviceLinker.cs
@@ -179,7 +179,7 @@ namespace Core.Editor.Tools.Mapping
 				}
 
 
-				if (slave.Master != null && relinkConnected == false) return false;
+				//if (slave.Master != null && relinkConnected == false) return false;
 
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/EditorButInAssetsAssembly/DeviceLinker.cs
```diff
@@ -179,7 +179,7 @@ private int LinkSlaves(bool relinkConnected)
 				}
 
 
-				if (slave.Master != null && relinkConnected == false) return false;
+				//if (slave.Master != null && relinkConnected == false) return false;
```
corp-0: why is this check not needed anymore?

Bod9001: it seems to cause issues if stuff is partially linked, so might as well do re-complete link every time

corp-0: I see, can you remove it? Don't comment stuff out, someone in the future might be wondering why is it commented out and think uncommenting might solve something

Bod9001: ok

corp-0 (approved):

PR #8565, opened 2022-04-14
Created SecBarriers (#8265)

### LINKED ISSUE #8265: Sec barriers
image

-Can be wrenched down and then locked with Id

mkuppers:
Took object_110 which was a blank object with the SecBarrier sprite, and gave it the wrench securable and access restrictions components

### Purpose
Closes #8265 

### Changelog:
Created Sec Barrier Prefab in Prefabs/Objects/Security
Added the Wrench Securable With Access Restriction script to the codebase


corp-0:
I need to open the editor to check some stuff that isn't easily readable in the code and I haven't been able to do that recently. Here are a list of stuff that worries me without actually checking:

1. Is the prefab properly parented? That is, is it a prefab variant of the base object prefab?
2. What functionality is access restricted? Just adding the component isn't enough, the AccessRestriction component is supposed to be used by your other component that handles the logic of whatever you want to restrict. For example, the door's behaviour asks the AccessRestricted component if they should open or play the rejected animation.

There probably is something the barrier can do other than being an obstacle you just wrench down.

mkuppers:
> I need to open the editor to check some stuff that isn't easily readable in the code and I haven't been able to do that recently. Here are a list of stuff that worries me without actually checking:
> 
> 1. Is the prefab properly parented? That is, is it a prefab variant of the base object prefab?
> 2. What functionality is access restricted? Just adding the component isn't enough, the AccessRestriction component is supposed to be used by your other component that handles the logic of whatever you want to restrict. For example, the door's behaviour asks the AccessRestricted component if they should open or play the rejected animation.
> 
> There probably is something the barrier can do other than being an obstacle you just wrench down.

My bad I probably misunderstood a little bit in regards to functionality/requirements.
In regards to point 1, I just checked and it is not a variant, I can redo this and make it such. 
In regards to point 2, I just added the access restricted component and did not think about what specifically it would restrict. I will look at the doors to see how that works and make it only affect the Wrench Securable functionality
I imagine there is more the barrier can do but am not sure what other functionalities to give it

Thanks again for all the help with getting me started to contributing, I greatly appreciate it!

corp-0:
I wouldn't suggest doors as example, they are rather complicated. But inspect the AccessRestriction component directly instead, it shouldn't be hard to understand the API.

mkuppers:
> I wouldn't suggest doors as example, they are rather complicated. But inspect the AccessRestriction component directly instead, it shouldn't be hard to understand the API.

Gotchya, I understand now! I think I misread the original issue and assumed that id meant job (ex: the access restriction itself) but I added the id locked pin component and made the secbarrier a variant of the object prefab so I think it should all be good now!

corp-0:
@PetMudstone I invoke you to tell me about barrier, what's the deal with them, what they do

PerfectTangent:
alright, I took a look at the PR. First, thanks for working on this @mkuppers.

it's properly parented, all of the values are correct, thanks for doing that. However, I don't think the ID Locked Pin component is the correct thing you want to put on here. That's for gun firing pins. You will have to make your own solution to get the Wrench Securable component to check for access.

Second note, one of the main utilities of the deployable barricade in /tg/ is that it provides cover; unless you are standing _directly adjacent_ to the barrier, any projectile you fire that passes through the barricade has a chance to hit the barricade (20% chance in this case for the sec barrier). Otherwise, it passes through the barrier without a problem. You will need to make a solution for that as well, perhaps in the object behavior or the object attributes components.

corp-0:
first feature you mention is ezpz, the other one might be a challenge

PerfectTangent:
relevant stats from /tg/ regarding the sec barrier. 
```
/obj/structure/barricade/security
	name = "security barrier"
	desc = "A deployable barrier. Provides good cover in fire fights."
	icon = 'icons/obj/objects.dmi'
	icon_state = "barrier0"
	density = FALSE
	anchored = FALSE
	max_integrity = 180
	proj_pass_rate = 20
	armor = list(MELEE = 10, BULLET = 50, LASER = 50, ENERGY = 50, BOMB = 10, BIO = 100, FIRE = 10, ACID = 0)
```

corp-0:
do other objects also have the proyectile pass rate or is it unique to this object?

mkuppers:
Thanks for taking a look @PerfectTangent !

Just wanted to make sure but when you say "make your own solution" should I create an external script or edit the WrenchSecurable script to have an option to check for access restriction?

I'll come back to the cover utility since I do not have enough familiarity with the code base to know where to get started with that

corp-0:
I would make a custom script that intercepts the interaction and makes the barrier play a denied sound or message if you don't have access, otherwise it would let the interaction chain go and it will eventually get to the wrench securable. 

So we have this interaction framework we use to handle clicking stuff, the WrenchSecurable component itself is an implementation of the interaction interface. All checked interactions have a ``WillInteract`` method that makes a check to see if the click the player did meets the requirements to handle the interaction, if they don't they simply do nothing and the interaction falls to the next interaction component.

I can guide you better in Discord, just not right now, if you need any help that is.

PerfectTangent:
> do other objects also have the proyectile pass rate or is it unique to this object?

lots do.

corp-0:
> > do other objects also have the proyectile pass rate or is it unique to this object?
> 
> lots do.

in that case it is something we need to add to ``ObjectBehaviour `` and maybe not part of this particular PR.

mkuppers:
> I would make a custom script that intercepts the interaction and makes the barrier play a denied sound or message if you don't have access, otherwise it would let the interaction chain go and it will eventually get to the wrench securable.
> 
> So we have this interaction framework we use to handle clicking stuff, the WrenchSecurable component itself is an implementation of the interaction interface. All checked interactions have a `WillInteract` method that makes a check to see if the click the player did meets the requirements to handle the interaction, if they don't they simply do nothing and the interaction falls to the next interaction component.
> 
> I can guide you better in Discord, just not right now, if you need any help that is.

Gotchya, I think that makes sense. I'll message more on discord probably tomorrow to ask a couple clarifying questions with that to make sure I fully understand. Thanks again!

mkuppers:
> in that case it is something we need to add to ObjectBehaviour and maybe not part of this particular PR.

Ok, so I did the first part of the functionality in regards to checking access. Should I try to tackle projectile pass rate or should that be left for an entirely different task/PR?

UnityProject/Assets/Scripts/Objects/Other/SecBarrier.cs
```diff
@@ -0,0 +1,68 @@
...
+		public async void ServerPerformInteraction(HandApply interaction)
+		{
+			if (Validations.HasItemTrait(interaction.HandObject, CommonTraits.Instance.Wrench))
+			{
+				//Notify player that they are unable to wrench down barrier
+				Chat.AddActionMsgToChat(interaction, "You try to wrench down the barrier...",
+					"The barrier denies you access.");
+			}
+			else {
+				Chat.AddActionMsgToChat(interaction, "ACCESS DENIED","");
```
PerfectTangent: perhaps this message should be editable, as the script is not necessarily always going to be used for sec barriers.

mkuppers: Gotchya, in that case should I generalize the script for wrenchable objects that have access restrictions? I assumed I was just making a sec barrier only script

corp-0: > perhaps this message should be editable, as the script is not necessarily always going to be used for sec barriers.

the conditions of ``WillInteract`` might be way too specific to be used for anything else, but still parametrising the string with a serialized field is always good.

corp-0:
> > in that case it is something we need to add to ObjectBehaviour and maybe not part of this particular PR.
> 
> Ok, so I did the first part of the functionality in regards to checking access. Should I try to tackle projectile pass rate or should that be left for an entirely different task/PR?

I think it is should be tackled in another task

UnityProject/Assets/Scripts/Objects/Other/SecBarrier.cs
PerfectTangent: > Gotchya, in that case should I generalize the script for wrenchable objects that have access restrictions? I assumed I was just making a sec barrier only script

generalized scripts are * always * preferred I would say, though I'm not an arbiter for this stuff.

UnityProject/Assets/Scripts/Objects/Other/SecBarrier.cs
```diff
@@ -0,0 +1,68 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Objects.Security
+{
+	[RequireComponent(typeof(AccessRestrictions))]
+	[RequireComponent(typeof(Construction.WrenchSecurable))]
+	public class SecBarrier : MonoBehaviour, IPredictedCheckedInteractable<HandApply>
```
corp-0: make it just a ``CheckedInteractable`` if you're not going to implement logic for the prediction.

mkuppers: Done!

UnityProject/Assets/Scripts/Objects/Other/SecBarrier.cs
```diff
@@ -0,0 +1,68 @@
...
+		public bool WillInteract(HandApply interaction, NetworkSide side)
+		{
+			//start with the default HandApply WillInteract logic.
+			if (!DefaultWillInteract.Default(interaction, side)) return false;
+
+			//Only go through with interaction if correct access restriction
+			if (AccessRestrictions.CheckAccess(interaction.Performer))
+			{
+				//Player needs wrench to interact
+				if (Validations.HasItemTrait(interaction.HandObject, CommonTraits.Instance.Wrench))
```
corp-0: Uhm wouldn't this mean that if you have the acces but not the tool you suddenly get a true? Maybe it is better to join both conditions with an ``&&`` and return the opposite of it.

``return !(AccessRestrictions.CheckAccess(interaction.Performer) && (Validations.HasItemTrait(interaction.HandObject, CommonTraits.Instance.Wrench))``

mkuppers: Ah yes, thanks and resolved!

UnityProject/Assets/Scripts/Objects/Other/SecBarrier.cs
mkuppers: Just renamed and generalized it further!

UnityProject/Assets/Scripts/Objects/WrenchSecurableWithAccessRestriction.cs
```diff
@@ -0,0 +1,73 @@
...
+
+		//invoked when the server recieves the interaction request and WIllinteract returns true
+		public async void ServerPerformInteraction(HandApply interaction)
+		{
+			if (Validations.HasItemTrait(interaction.HandObject, CommonTraits.Instance.Wrench))
+			{
+				string objectName = ObjectAttributes.name;
+				//Notify player that they are unable to wrench down barrier
+				Chat.AddActionMsgToChat(interaction, "You try to wrench down the " + objectName + "...",
+					"The " + objectName + " denies you access.");
```
NoooneyDude: The third parameter for this method is for observers of the action, not the performer. So others would see: "The secbarrier denies you access.", which is rather confusing 😛. The performer also never gets messages intended for observers, so they'll be left wondering what happened.

mkuppers: Oops my bad!! Just fixed

NoooneyDude:
Small tip: if you mention the issue in a particular way, GitHub will close it automatically when this is merged.

Closes #8265, close #8265, fixes #8265, fix #8265 are the magic words I know about.

I've updated the description : )

mkuppers:
> Small tip: if you mention the issue in a particular way, GitHub will close it automatically when this is merged.
> 
> Closes #8265, close #8265, fixes #8265, fix #8265 are the magic words I know about.
> 
> I've updated the description : )

Thanks for that!! Was not aware I could do that

corp-0 (dismissed):

corp-0 (dismissed):

corp-0:
you tests failed, if you go to details then summary you can download the report in html and see why

corp-0:
what did you change in boxstation?

PerfectTangent:
perhaps removal of instances of objects_110? But boxstation isn't the only map that has them.

mkuppers:
> perhaps removal of instances of objects_110? But boxstation isn't the only map that has them.

Yeah that's what I was going for since in the test I failed it was a missing prefab in boxstation

mkuppers:
> > perhaps removal of instances of objects_110? But boxstation isn't the only map that has them.
> 
> Yeah that's what I was going for since in the test I failed it was a missing prefab in boxstation

Are there other scenes I should remove the prefabs from or should I revert to a version with the prefabs and just add back the sec barrier?

corp-0:
That is probably th cleanest and easiest. Just add the sec barrier and delete nothing, let mappers add your object

PR #8567, opened 2022-04-14
Networking update attempt 3

Bod9001:
ignorance update, mirror update, changes to allow multithreading to work okay, 

notes :a message fail in the editor, with his current version but that happened only once and then I connected 7 clients and it didn't appear again
I've done a code change to the ignorance code so we have some custom code there

> `Ignorance.cs` modification: add lock on `Server.Outgoing.Enqueue(dispatchPacket);`


UnityProject/Packages/packages-lock.json
```diff
@@ -236,11 +236,11 @@
       "url": "https://packages.unity.com"
     },
     "org.unitystation.spacemirror": {
-      "version": "https://github.com/unitystation/Mirror.git?path=/Assets/Mirror#dev",
+      "version": "https://github.com/unitystation/Mirror.git?path=/Assets/Mirror#master",
       "depth": 0,
       "source": "git",
       "dependencies": {},
-      "hash": "2afb70b72c7d784faee0ba893b93147ac91f3dad"
+      "hash": ""
```
corp-0, line 243: I think I know why this is needed. The actual way to change the branch in ``manifest.json`` would be using the graphical interface of the package manager and not manually changing the json file.

Bod9001: I'm leaving it blank so that It will pull the latest version

NoooneyDude (approved):

PR #8574, opened 2022-04-16
Better multithreading logging

Bod9001:
this should now properly is Display in the discord web hook,

note I removed the throw from the thread because, the other try catches didn't have it

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Threading/ThreadBehaviour.cs
+++ b/UnityProject/Assets/Scripts/Core/Threading/ThreadBehaviour.cs
@@ -82,7 +82,15 @@ namespace Core.Threading
 			Profiler.BeginThreadProfiling("Unitystation", threadName);
 			while (running && threadMode == ThreadMode.Threaded && midTick == false)
 			{
-				RunTick();
+				try
+				{
+					RunTick();
+				}
+				catch (Exception e)
+				{
+					ThreadLogger.AddLog(e.ToString(), Category.Threading);
+				}
+
 				Thread.Sleep(tickDelay);
 				ticker++;
 			}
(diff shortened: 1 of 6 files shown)


UnityProject/Assets/Scripts/Core/Threading/ThreadBehaviour.cs
```diff
@@ -82,7 +82,15 @@ private void ThreadLoop()
...
 			while (running && threadMode == ThreadMode.Threaded && midTick == false)
 			{
-				RunTick();
+				try
+				{
+					RunTick();
+				}
+				catch (Exception e)
+				{
+					ThreadLogger.AddLog(e.ToString(), Category.Threading);
```
corp-0, line 91: throw as well so we notice stuff breaking

Bod9001: atmospherics and electrical, both didn't have throw and those were the most likely to generate a lot of errors so

corp-0: I'm just saying it is fine to catch to log, but you should throw the exception at the end of the handling

Bod9001: But the system that would have varied a lot already just had a try catch, with no throw, and the server has been fine, do you think it requires the throw being added?

corp-0: I would rather have it so we can see stuff breaking if it breaks. You first see the efect of stuff breaking then you read the logs.

corp-0 (approved):

PR #8581, opened 2022-04-19
Fixes missing item in R&D Production Machines

AtnerNT:
As it turns out:
Capture

As a result of this Protolathes and Circuit Imprinters couldn't load their items. This takes The JsonImport out of a task back onto the main thread. 
Also Includes changes to the metafiles of the new textures from the Protolathe PR that weren't properly included.

CL: [Fix] fixed missing items in the R&D machines


AtnerNT:
*(Alot of lines were changes to correct the indention levels)*

corp-0:
that doesn't look like correct indentation level. If you want to have classes at the outer most indentation level, you can remove the brackets from the ``namespace`` block, I think it works in our current c# version.

Copern:
> that doesn't look like correct indentation level. If you want to have classes at the outer most indentation level, you can remove the brackets from the `namespace` block, I think it works in our current c# version.

Unfortunately, file scoped namespaces aren't available in our version of C#. It's a C# 10 feature. I believe we have C# 8 and most of C# 9 available to us.

corp-0:
Aww damn

UnityProject/Assets/Scripts/Systems/Research/Designs.cs
```diff
@@ -9,152 +9,170 @@
...
-		};
-		public static Dictionary<string,Design> InternalIDSearch;
 
-	}
-	private static void JsonImportInitialization ()
+	public class Designs : MonoBehaviour
 	{
-		Globals.InternalIDSearch =  new Dictionary<string,Design>();
-		foreach (string FilePath in Globals.PathsList) {
-			string json = (Resources.Load (FilePath) as TextAsset).ToString ();
```
Bod9001, line 99: you could just move this outside of the thread task?

UnityProject/Assets/Scripts/Systems/Research/Designs.cs
```diff
@@ -9,152 +9,170 @@
...
+		{
+			if (!(Globals.IsInitialised))
+			{
+				JsonImportInitialization();
+			}
+		}
+		public static class Globals
+		{
+			public static bool IsInitialised = false;
+			public static List<string> PathsList = new List<string> {
```
corp-0: I'm not sure I like this hardcoded list of strings

UnityProject/Assets/Scripts/Systems/Research/Designs.cs
```diff
@@ -9,152 +9,170 @@
...
-					DesignPass.Category = new List<string> ();
-				}
+		void Awake()
+		{
+			if (!(Globals.IsInitialised))
+			{
+				JsonImportInitialization();
+			}
+		}
+		public static class Globals
```
corp-0, line 90: why do we need a static ''Globals''?

UnityProject/Assets/Scripts/Systems/Research/Designs.cs
AtnerNT: I've changed it to get the files inside the specified folder. The folder itself is hardcoded to Metadata/Designs. But we shouldn't be putting designs outside that folder as it is so I think its fine.

UnityProject/Assets/Scripts/Systems/Research/Designs.cs
AtnerNT: Ummm- not entirely sure. However this script wasn't written by me and I don't want to interfere with any future plans by deleting it incase it has a planned purpose.

corp-0: ok fair

UnityProject/Assets/Scripts/Systems/Research/Designs.cs
```diff
@@ -4,157 +4,177 @@
...
+			{
+				var JsonTechnologies = JsonConvert.DeserializeObject<List<Dictionary<String, System.Object>>>(json);
+				for (var i = 0; i < JsonTechnologies.Count(); i++)
+				{
+					Design DesignPass = new Design();
+					DesignPass.InternalID = JsonTechnologies[i]["Internal_ID"].ToString();
+					DesignPass.Name = JsonTechnologies[i]["name"].ToString();
+					DesignPass.ItemID = JsonTechnologies[i]["Item_ID"].ToString();
+
+					if (JsonTechnologies[i].ContainsKey("Description"))
```
corp-0, line 111: is it needed to check for every key individually?

AtnerNT: I think it's best. If a specific item doesn't have a description set for instance- you don't want that throwing an error stopping every other design from importing as well

corp-0: There must be a way to make missing keys a null on the deserialized object

AtnerNT: The mere act of trying to get the key if it doesnt exist causes an NRE and stops it form imporintg- always have to check with .Contains() first

corp-0: https://docs.microsoft.com/en-us/answers/questions/291110/json-deserialization-behaviour-regarding-missing-p.html

Bod9001: json newtonsoft could also use that idk , I think it's okay to have the checking thing

corp-0: Any repetitive thing like that is prone to mistakes and it is also annoying to maintain

corp-0 (changes requested):
do you think you could do what I asked in the conversation so we merge this?

UnityProject/Assets/Scripts/Systems/Research/Designs.cs
AtnerNT: I tested it in editor- the checks have to be there- the problem isnt the Json getting a null field- the mere act or trying to get an invalid key from the dictionary causes the erorr and breaks it. It doesn't work without the check. (Getting the wrong key doesn't make the dictionary return null- it just makes it cause an error). Not sure what we are doing different to that link you sent- all I know- it has been tested to not work without the checks

corp-0: I would have to see how you did it, you are probably missing giving the model a default value, but I will merge for now

corp-0 (approved):

PR #8583, opened 2022-04-19
Fixes cargo admin manager

MaxIsJoe:
Title

For the future, if you want to add commands for admins make it inside the AdminCommandsManager which is similar to the PlayerCommandsManager

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/AdminCommandsManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/AdminCommandsManager.cs
@@ -497,8 +497,9 @@ namespace AdminCommands
 		#region CargoControlCommands
 
 		[Command(requiresAuthority = false)]
-		public void CmdRemoveBounty(int index, bool completeBounty)
+		public void CmdRemoveBounty(int index, bool completeBounty, NetworkConnectionToClient sender = null)
 		{
+			if (IsAdmin(sender, out var admin) == false) return;
 			if (completeBounty)
 			{
 				CargoManager.Instance.CompleteBounty(CargoManager.Instance.ActiveBounties[index]);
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Managers/AdminCommandsManager.cs
```diff
@@ -490,5 +493,48 @@ public static void LogAdminAction(string msg, string userName = "")
 		}
 
 		#endregion
+
+		#region CargoControlCommands
+
+		[Command(requiresAuthority = false)]
+		public void CmdRemoveBounty(int index, bool completeBounty)
```
Bod9001: security concerns?

MaxIsJoe: fixed now?

UnityProject/Assets/Scripts/UI/Systems/AdminTools/AdminBountyManager.cs
```diff
@@ -43,7 +52,7 @@ private void OnDisable()
 		public override void OnEnable()
 		{
 			base.OnEnable();
-			RefreshBountiesList();
+			AdminCommandsManager.Instance.CmdRequestCargoServerData(PlayerManager.PlayerScript.connectedPlayer.UserId);
```
Bod9001: I don't think you need to specify it I think it Fills it out automatically, not to sure the need to ask noonly

MaxIsJoe: It works so sure, but if the entries dont get updated after we merge still then we know where the issue is.

Bod9001 (approved):

PR #8599, opened 2022-04-23
Expanded access (Max edition)

MaxIsJoe:
basically the same as #8521 but access is giving to everyone when they join while a server is or was in a lowpop state ~~and after 25+ players join any new player that joins/respawns will not receive that extra access.~~

CL: [New] low pop settings can be configured in the gameConfig file and there are announcements that tell the station when low pop in happening


PerfectTangent:
could you list out what jobs gained what accesses? The list looks different than #8521.

MaxIsJoe:
> could you list out what jobs gained what accesses? The list looks different than #8521.

It's the same as Bod's but with the only difference is that the shaft miner has access to the cargo console.
It may look like i have edited a bunch of extra roles but thats because I was manually editing every single occupation asset before adding a check that would fill the low pop access for me for every role

Bod9001:
is using the new access system?

corp-0:
It is not. New one uses Clearance, this uses Access

MaxIsJoe:
From what ive checked doors dont use the new clearance system

UnityProject/Assets/Scripts/Managers/CentComm.cs
```diff
@@ -158,6 +165,21 @@ IEnumerator WaitToGenericReport()
...
 
+		private IEnumerator LowpopCheck()
+		{
+			yield return WaitFor.Seconds(Application.isEditor ? 60 : 300);
+			if(PlayerList.Instance.GetAlivePlayers().Count > gameManager.LowPopLimit) yield break;
+			IsLowPop = true;
+			MakeAnnouncement(ChatTemplates.CentcomAnnounce,
+				"Due to the shortage of staff on the station; We have granted additional access to all crew members until further notice."
+				, UpdateSound.Announce);
+			IDCards.Remove(null);
```
corp-0: uhm?

corp-0:
I think it would be a lot simpler if instead of adding and removing accesses to the ID, you pass the script that checks for access both lists and that script asks GameManager if it is lowpop so it uses the lowpop list.

UnityProject/Assets/Scripts/Managers/CentComm.cs
MaxIsJoe: what does "uhm" mean? whats the issue?

MaxIsJoe:
Done.

UnityProject/Assets/Scripts/Managers/CentComm.cs
```diff
@@ -158,6 +162,16 @@ IEnumerator WaitToGenericReport()
 			MakeAnnouncement(ChatTemplates.CentcomAnnounce, PlayerUtils.GetGenericReport(), UpdateSound.Announce);
 		}
 
+		private IEnumerator LowpopCheck()
+		{
+			yield return WaitFor.Seconds(Application.isEditor ? 60 : 300);
```
Bod9001: doesn't this mean most of the crew wouldn't spawn without low pop access?

MaxIsJoe: Jack said he doesn't want everyone to spawn immediately with it so i added this timer before checking because people tend to late join 5 minutes usually so people dont sit on the lobby intentionally to cause everyone to get access.

craftxbox:
Enhanced access enabling/disabling (specifically disabling) *during* the round on existing players sounds like a recipe for trapped players.

PerfectTangent:
> Enhanced access enabling/disabling (specifically disabling) _during_ the round on existing players sounds like a recipe for trapped players.

well, this doesn't disable access, so all good.

UnityProject/Assets/Scripts/Items/IDCard.cs
```diff
@@ -213,6 +222,10 @@ public void SyncIDCardType(IDCardType oldCardType, IDCardType cardType)
 	/// <returns></returns>
 	public bool HasAccess(Access access)
```
corp-0: wait a second, what the fuck is this method. What uses it? I think most of stuff uses https://github.com/unitystation/unitystation/blob/develop/UnityProject/Assets/Scripts/Systems/Occupations/AccessRestrictions.cs this one to check for access

UnityProject/Assets/Scripts/Managers/CentComm.cs
Bod9001: but doesn't that mean that, it doesn't update the people then after, since it's not tracking the ID anymore due to Gilles asking to change it?

corp-0: it doesn't ever update people. It asks GameManager if it is lowpop or not and uses either list of accesses accordingly. Now, reading the rest of the conversation, it sounds like you guys would rather have people keep their extra access even if the round isn't low pop anymore, is that correct?

PerfectTangent: > it sounds like you guys would rather have people keep their extra access even if the round isn't low pop anymore, is that correct?

yes, taking it away mid round is a bad idea.

NoooneyDude:
When the round is no longer lowpop a CentComm announcement encouraging crew to get their access reset via HoP (perhaps don't announce if the HoP does not exist or is not alive) might add a bit of flavour.

UnityProject/Assets/Scripts/Managers/CentComm.cs
```diff
@@ -158,6 +162,25 @@ IEnumerator WaitToGenericReport()
...
+			if(PlayerList.Instance.GetAlivePlayers().Count > gameManager.LowPopLimit) yield break;
+			IsLowPop = true;
+			MakeAnnouncement(ChatTemplates.CentcomAnnounce,
+				"Due to the shortage of staff on the station; We have granted additional access to all crew members until further notice."
+				, UpdateSound.Announce);
+
+			var idsSpawned = FindObjectsOfType<IDCard>();
+			foreach (var card in idsSpawned)
+			{
+				foreach (var access in card.Occupation.AllowedLowPopAccess)
```
corp-0, line 178: occupation might be null, add a null check and skip those

UnityProject/Assets/Scripts/Managers/CentComm.cs
```diff
@@ -158,6 +162,25 @@ IEnumerator WaitToGenericReport()
 			MakeAnnouncement(ChatTemplates.CentcomAnnounce, PlayerUtils.GetGenericReport(), UpdateSound.Announce);
 		}
 
+		private IEnumerator LowpopCheck()
+		{
+			yield return WaitFor.Seconds(Application.isEditor ? 60 : 300);
```
corp-0: add the amount of time to the config file

MaxIsJoe:
Done.

MaxIsJoe:
> When the round is no longer lowpop a CentComm announcement encouraging crew to get their access reset via HoP (perhaps don't announce if the HoP does not exist or is not alive) might add a bit of flavour.

Doesnt make sense for NT to be able to remotely give access to all crew members but cant take those back

MaxIsJoe:
also we already agreed that taking away access away from players mid round is bad even for people who are late joining

corp-0 (approved):

PR #8603, opened 2022-04-23
improvements, 

Bod9001:
remove the default sound from radio communications, it was a bit loud and obnoxious and not particularly useful
hopefully makes it so null username makes it so the client is actually informed, the player knows what to do to fix it

CL: [Fix] added audio slider for radio message notification


MaxIsJoe:
> Remove

image

corp-0:
I also dislike the plain out removal, I'd rather do the option to disable/lower its volume instead

Bod9001:
Though it is kept for the security and other channels? Should I have it as a volume slider?

corp-0:
> Though it is kept for the security and other channels? Should I have it as a volume slider?

Yes, a slider would be great

Bod9001:
done

NoooneyDude (approved):

PR #8613, opened 2022-04-25
maybe changelogs will work again?

MaxIsJoe:
mayhabs?

Just need a test repo called "changelog"

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -47,21 +47,20 @@ jobs:
         with:
           categories: Fix;New;Improvement
 
-      - name: Commit files
-        if: ${{ steps.pr2changelog.outputs.generated_changelog == 1}}
+      - name: prepare files to push
+        if: ${{ steps.pr2changelog.outputs.generated_changelog == 1 }}
         run: |
-          git config --local user.email "action@github.com"
-          git config --local user.name "GitHub Action"
-          git add *
-          git commit -m "misc: update Changelog" -a
+          mkdir output
+          cp CHANGELOG.md output/CHANGELOG.md
+          # cp CHANGELOG.json output/CHANGELOG.json
 
-      - name: Pushes to another repository
+      - name: Push to changelog repo
         uses: cpina/github-action-push-to-another-repository@main
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/changelog.yml
```diff
@@ -55,9 +55,13 @@ jobs:
...
-        uses: ad-m/github-push-action@master
+      - name: Pushes to another repository
+        uses: cpina/github-action-push-to-another-repository@main
+        env:
+          API_TOKEN_GITHUB: ${{ secrets.API_TOKEN_GITHUB }}
         with:
-          force: true
-          github_token: ${{ secrets.PR2CHANGELOG_TOKEN }}
+          source-directory: 'output'
+          destination-github-username: 'US13-BOT'
```
corp-0: use the same user we were using, also is this github username or git username? Because those are different and we need an account if the former.

.github/workflows/changelog.yml
```diff
@@ -55,9 +55,13 @@ jobs:
...
+          destination-repository-name: 'changelog'
```
corp-0: you sure you don't need to add the organization

MaxIsJoe: now?

corp-0:
do you mind if I take over?

MaxIsJoe:
go ahead

corp-0:
just realized there is an issue with this approach. In order to get previous changes, the action needs to read the changelog file and the issue is we can't write the updated version to this repo. I need to give this more time to think.

Bod9001:
surely the change log file is on a different repository?

corp-0:
this is how it works now:
1. Read PR description and find the tags for changelog.
2. Build changelog python object.
3. Build a string representation of the changelog based in markdown format.
4. Read current changelog and append the string to the beginning of the file (so it goes from newest to oldest)
5. commit and push to repo

Issue is step number 4, since the action is run here it can only read the changelog file we currently have so the result will always be current changelog + the last change at the beginning.

Bod9001:
What prevents you from reading from different repositories?

corp-0:
Maybe I can set the checkout action to do that, I need a moment to test it

corp-0:
this one should also be ready for review now

NoooneyDude (approved):

PR #8619, opened 2022-04-28
Antags will no longer receive the steal objective with an item they have immediate access to in their job

### LINKED ISSUE #4857: Certain roles shouldn't get the traitor objective to steal certain items.
## Description
Right now, it is possible for traitorous chief engineers and chief medical officers to be tasked with stealing their own advanced magboots or hypospray respectively. This results in traitorous CEs and CMOs being able to get pathetically easy objectives to steal items that already belong to them without opposition. If the steal objective SO disallowed a list* of roles from stealing a particular item, then oddities like this wouldn't happen anymore.

*A list of roles instead of selecting a singular one would help with the steal plasma objective for example. It's easy to openly obtain solid plasma for _anyone_ in engineering and supply without anyone getting suspicious.

MaxIsJoe:
Fixes #4857

Adds the ability to blacklist occupations from receiving the steal objective that have immediate access to their objective

image

### Changelog

CL: [Fix] - Antags will no longer receive the steal objective with an item they have immediate access to in their job


corp-0 (approved):
lgtm

PR #8620, opened 2022-04-28
Makes reboot value configurable

MaxIsJoe:
title

### DIFF
--- a/UnityProject/Assets/StreamingAssets/config/gameConfig.json
+++ b/UnityProject/Assets/StreamingAssets/config/gameConfig.json
@@ -37,5 +37,5 @@
 
 	"LowPopCheckTimeAfterRoundStart" : 300,
 
-	"RebootOnAverageFPSOrLower" : 35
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/StreamingAssets/config/gameConfig.json
```diff
@@ -35,5 +35,7 @@
 	
 	"LowPopLimit" : 20,
 
-	"LowPopCheckTimeAfterRoundStart" : 300
+	"LowPopCheckTimeAfterRoundStart" : 300,
+
+	"RebootOnAverageFPSOrLower" : 35
```
corp-0: make it MUCH lower. Server fps aren't the same as the fps you see while playing, I think even 10 is good server fps when  a lot of stuff is happening.

MaxIsJoe: 20 seems like a better value since thats where the server dips a lot most of the time especially after a profile.

corp-0 (approved):

PR #8624, opened 2022-04-29
RDProductionMachine fix iteration3

AtnerNT:
Last one had some weird things around the directory, don't know why - using Application.streamingAssetsPath now- hopefully this should *actually* fix it.


Bod9001 (approved):
Let's hope this works

PR #8634, opened 2022-05-02
51 new cargo bounties

PerfectTangent:
### Purpose

With MaxIsJoe working on #6150, it's about time to add more bounties than just the placeholder ones. This PR adds 51 new cargo bounties, organized by 6 job types (engineering, medical, science, service, mining, and misc). None of the bounties have been put in CargoData yet, as modifications need to be made to that. This means they can't be done yet, but they are ready to go for when the random bounty system is done.

### Notes:

- added a lot of new traits and added a lot of traits to items.
- fixed a few typos in descriptions, as well as in the name of the harebell flower
- most of the bounties are ported from /tg/, though with some tweaks to make them a bit more flexible (like, collect x number of any flowers instead of only one specific kind of flower).
- adds an export price to produce (3 per product. Not very nuanced, but it was an easy baseline to start with).
- nerfs the sell price of hot ice by 100 (to 300 per).
- some of the flavor text is incomplete for some bounties as it doesn't handle long descriptions very well.

These new additions to the cargo bounties has highlighted 4 problems with how objects, bounties, and export permissions are handled;
- first, the export properties of objects are mercurial to me as the export stats are a part of _item_ attributes, which objects don't have.
- second, it is not possible to assign traits to objects, as traits are a part of item attributes only. This makes making bounties affecting objects not possible.
- third, if an object is part of a bounty but has no export price (e.g., the default situation for most items), when it is sent in the cargo shuttle, it will count as part of the bounty, but it won't actually be sold. The default cargo setting added in #7664 should be tweaked to check if the item in question is part of a bounty, and if it is, then it should be sold even if it is not normally sellable.
- fourth, the bounty system cannot handle reagents or gases, which heavily limits the kinds of possible bounties, preventing a lot of the ones typically available in /tg/. Bartenders cannot participate in bounties at all because of this.

### Changelog:

CL: [New] 51 new random cargo bounties have been added for the station to fulfill.
CL: [Balance] The export price of all hydroponics produce has been increased from 0 to 3.
CL: [Balance] The export price of Hot ice has been reduced from 400 to 300.
CL: [Fix] fixed a typo with Harebells.


UnityProject/Assets/ScriptableObjects/Cargo/Bounties/MiningBounties/Metal.asset
```diff
@@ -0,0 +1,20 @@
+%YAML 1.1
```
corp-0, line 1: pretty sure we have already a ``Metal`` trait that can be used to identify the material

corp-0: nevermind just read the path of the file

corp-0 (approved):

corp-0:
looks good. I woul rather have what you describe as stuff that is needed to add these to the system as a separated issue tho, even maybe more than one describing each task. Biggest one being making a common parent component for attributes between objects and items so they both have access to the common properties they should have.

PR #8635, opened 2022-05-03
Mining Diversification Part 1

PerfectTangent:
### Purpose

This PR lays the foundation for new mining gameplay in the form of _mining hardness._ When attempting to mine a mineable tile or object, the hardness of the pickaxe will be compared to the target's hardness, and if the object is too hard, the pickaxe will fail to mine it.

The goal of this additional stat for mining tools is to make the choice of which tool you use to mine more nuanced than just the fastest tool science has unlocked. Nothing currently has changed with the inclusion of hardness for miners in the game right now; all mining tools can still mine what they could before. I plan to follow this PR up with more content that will provide new things to mine that will use the hardness system.

### Changelog:

[New] Mining tools and mineable tiles now have a hardness stat that decided what objects can and cannot be mined with a given tool.


Bod9001:
what would PKA do? and mining laser?

PerfectTangent:
> what would PKA do? and mining laser?

eventually they'd be restricted from mining certain tiles.

Bod9001:
Maybe you could change the power mode on them, but it takes different times to charge up and  you have to have the exact level?

UnityProject/Assets/Scripts/Items/Tool/Pickaxe.cs
```diff
@@ -18,6 +18,11 @@ public class Pickaxe : MonoBehaviour, ICheckedInteractable<PositionalHandApply>,
 		[SerializeField]
 		private float timeMultiplier = 1f;
 
+		[Range(1, 10)]
+		[Tooltip("what degree of hardness this mining tool can overcome. Higher means the pick can mine more types of things.")]
+		[SerializeField]
+		private int pickHardness = 5;
```
corp-0, line 24: would you like this to be an arbitrary integer or would it be better to use a finite amount of hardness levels to define it?

PerfectTangent: isn't a limited range integer a finite amount of hardness levels?

corp-0: yeah, I guess 10 levels might be too many levels. I don't know, wouldn't a drop down where you pick between say 4 levels be easier to manage?

PerfectTangent: I don't think four values is enough for the effect I'm trying to achieve. I mean, we're dealing with 5 tiers of pickaxe, 2 drills, the jackhammer, PKA, PKC, 2 versions of plasma cutter, and the mech drill. The goal of this hardness scale is to try to (along with later additions) create a dynamic where you get utility out of all kinds of mining tools because they have different strengths and weaknesses. 

I originally drafted around 5 points of hardness, but I felt like it was too limiting for how many tools we have. As an example, with a 5 point scale, you'd naturally end up with a pickaxe in each tier (improvised, compact, standard, silver-tipped, diamond-tipped), and then the drills and jackhammer being 3, 4, and 5, because 1 and 2 are too low for high tier tools created by R&D. Thing is, this defeats the point of the PR as the drills would be the same hardness as the picks in their tier, and the point is to try to make drills _not automatically better_ than picks in all situations, like they currently are.

tl;dr, I thought about it, and I think the scale needs to be higher than 4-5 values. I went with 10 because it provides enough granularity to do the job and as a bonus, the real mohs scale of hardness is based on a 1 to 10 scale.

corp-0 (approved):

PR #8636, opened 2022-05-03
 unity been complaining about some stuff so  time to reserialise

Bod9001:


corp-0:
if we merge this before what other peopel are working on, will their local break?

Bod9001:
shouldn't do , the only worry is merge conflicts?

corp-0:
from what I understand reserialising gives all tracked files a new id and all, doesn't it? I'm just worried about people losing their work because something stupid like missing references

Bod9001:
it just causes unity to look at every file and do the changes that haven't been tracked over time, such as new fields being added to scripts and stuff

Bod9001:
image
all stuff like this

corp-0 (dismissed):

corp-0:
merging, if shit gets utterly broken I backpedal

corp-0 (approved):

PR #8637, opened 2022-05-04
Serialisation changes from unity being lazy (WIP)

Bod9001:
Mostly consists of starting Lobby and then hosting and then Waiting for it to do some changes, and then do it again until it's done all the changes idk I'm still doing it


UnityProject/Assets/Scripts/Core/Sprite Handler/SpriteHandler.cs
```diff
@@ -790,6 +790,7 @@ public void UpdateMe()
 		if (PresentFrame == null)
 		{
 			TryToggleAnimationState(false);
+			return;
```
corp-0, line 793: what is this fixing?

PR #8659, opened 2022-05-06
ReagentContainer inspector options for right click menu.

Copern:
Adds option to ReagentContainer inspector to choose what to show on the right click menu. Default is to show "Show Contents" and "Pour Out" options.

Food base and welding tool variants will show neither option on the menu.

I don't know if you are supposed to be able to "pour out" the fuel in a welder via right click options, so I just set it to none.

### DIFF
--- a/UnityProject/Assets/Scripts/ChemistryComponents/ReagentContainer.cs
+++ b/UnityProject/Assets/Scripts/ChemistryComponents/ReagentContainer.cs
@@ -76,9 +76,6 @@ namespace Chemistry.Components
 			}
 		}
 
-		[Tooltip("Can you empty out its contents?")]
-		public bool canPourOut = true;
-
 		[Tooltip("Initial mix of reagent inside container")]
 		[FormerlySerializedAs("reagentMix")]
 		[SerializeField] private ReagentMix initialReagentMix = new ReagentMix();
@@ -496,7 +493,7 @@ namespace Chemistry.Components
 
 			var pourOutContext = ContextMenuApply.ByLocalPlayer(gameObject, "PourOut");
 			//Pour / add can only be done if in reach
-			if (WillInteract(pourOutContext, NetworkSide.Client) && canPourOut && menuOptions.HasFlag(ShowMenuOptions.PourOut))
+			if (WillInteract(pourOutContext, NetworkSide.Client) && menuOptions.HasFlag(ShowMenuOptions.PourOut))
 			{
 				result.AddElement("PourOut", OnPourOutClicked);
 			}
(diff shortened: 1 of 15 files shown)


corp-0 (dismissed):

UnityProject/Assets/Scripts/ChemistryComponents/ReagentContainer.cs
```diff
@@ -18,6 +18,15 @@ namespace Chemistry.Components
 	public partial class ReagentContainer : MonoBehaviour, IServerSpawn, IRightClickable, ICheckedInteractable<ContextMenuApply>,
 		IEnumerable<KeyValuePair<Reagent, float>>
 	{
+		[Flags]
+		private enum ShowMenuOptions
+		{
+			None = 0,
+			ShowContents = 1 << 0,
+			PourOut = 1 << 1,
```
Bod9001, line 26: I suppose it could be handled by a bool for example like is solid, like you can't   pour the contents out of a burger because it's solid?

Copern: A flagged enum was used as I figured there might be rare cases where you should be able to see the contents of something but not able to pour it out via the menu. It doesn't seem like a good idea to base it off specific features of a reagent container as then we might end up with a bunch of special cases. IMO, it's much easier for a maintainer to set the flag for the prefab/variant in the editor.

Bod9001: true, though having it set up automatically is a bonus, 

do you think it would be to be complicated if it used the bool and the enum would override it?

Bod9001 (dismissed):

UnityProject/Assets/Scripts/ChemistryComponents/ReagentContainer.cs
Copern: We have a "canPourOut" bool on ReagentContainer. Oddly things like "_FoodBase" and the base "WeldingTool" have it set to true. I don't know if most foods override that and set it to false though. Think we should use that bool and have the enum as an override for the options?

Edit: Actually the ReagentContainer right click option generation already used that bool which puts us back to what this PR was meant for.

Copern: Well then, looking at that bool again, turns out the only purpose it serves is for the right click menu. I figured it had something to do with spill and other non-menu interactions for reagent containers too, but no it doesn't. That bool's tooltip probably could have been labeled better. So, two options. We could scrap that bool and use the enum instead. Or we could scrap the enum, relabel the tooltip, and try to find out which foods need that bool fixed. Scrapping the enum would mean we couldn't disable the contents/pouring options separately though.

Bod9001: It properly should do, I think having the Enumerator would be best option if we can't have both

Bod9001 (dismissed):

UnityProject/Assets/Scripts/ChemistryComponents/ReagentContainer.cs
Copern: I'll see about removing the old bool tonight and adjusting the prefabs that had changed it. If there ends up being a significant number of prefab changes caused by the removal, then I'll leave it alone for this PR.

Bod9001 (approved):

PR #8663, opened 2022-05-07
fix addressable url

corp-0:
ez mistake, urls were missing the branch so it 404'ed on latest version of the game (which also makes it impossible to spawn btw)


corp-0:
I will commit admin abooze and speed merge this, given the urgency of having a play test tomorrow

PR #8666, opened 2022-05-07
Security fix for cmd in admin commands manager

MaxIsJoe:
Prevents none admins from changing a very important bool in Cargo manager


corp-0 (dismissed):

MaxIsJoe:
can you re-run tests? its complaining about something that i didn't even do

corp-0:
it is an issue on develop,  I  think. Be the hero we need, add the prefab to the list

MaxIsJoe:
> it is an issue on develop, I think. Be the hero we need, add the prefab to the list

Doesn't sound like it's an issue on develop if my other PR had passing tests

corp-0:
those were probably old enough to not get affected. Two of recent Jack's PRs had the same test failing with the same prefab

corp-0 (approved):

PR #8673, opened 2022-05-09
fix pr2changelog categories

corp-0:
It was missing ``Balance`` on the step that wrote the chagenlog


Bod9001:
does this mean change log works now?

corp-0:
> does this mean change log works now?

yeah, they do. Go to https://github.com/unitystation/changelog

Bod9001:
nice

PR #8684, opened 2022-05-14
Fixes issue where detention times were out of sync for bombs

MaxIsJoe:
When you change the timer on a bomb you only changed the value client side so when you arm the bomb it would actually use the value on the server's not the one you adjusted. Part of #6149

## Changelog:

CL: [Fix] - Fixed bomb timers being out of sync when adjusting them.

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Weapons/ExplosiveBase.cs
+++ b/UnityProject/Assets/Scripts/Items/Weapons/ExplosiveBase.cs
@@ -26,7 +26,7 @@ namespace Items.Weapons
 		[Header("Explosive settings")]
 		[SerializeField] protected ExplosiveType explosiveType;
 		[SerializeField] protected bool detonateImmediatelyOnSignal;
-		[SerializeField] protected int timeToDetonate = 10;
+		[SerializeField, SyncVar] protected int timeToDetonate = 10;
 		[SerializeField] protected int minimumTimeToDetonate = 10;
 		[SerializeField] protected float explosiveStrength = 150f;
 		[SerializeField] protected SpriteDataSO activeSpriteSO;


UnityProject/Assets/Scripts/Items/Weapons/ExplosiveBase.cs
```diff
@@ -26,7 +26,7 @@ public class ExplosiveBase : SignalReceiver
 		[Header("Explosive settings")]
 		[SerializeField] protected ExplosiveType explosiveType;
 		[SerializeField] protected bool detonateImmediatelyOnSignal;
-		[SerializeField] protected int timeToDetonate = 10;
+		[SerializeField, SyncVar] protected int timeToDetonate = 10;
```
Bod9001, line 29: I thought you could only change SyncVar from server side?

MaxIsJoe: NetUI does the change server sided

Bod9001: I'm confused since in the description you say it was only being done client side ? 
so that means it surely isn't communicating with the server?

MaxIsJoe: oH, you're right. Yeah something doesn't add up here and I noticed that the cooldown function is only called on the server and has the `[Server]` attribute above it. something is odd here.

MaxIsJoe: Yeah I don't know what exactly the issue is then because everything seems to be fine prefab and code wise, going to wait until next playtest until we get this happening again.

PR #8688, opened 2022-05-14
fixes 100 reycast error

Bod9001:
<don't mind me, I'm just trying to make pr2changelog not shit itself>


corp-0:
in this case I think it is fine to don't register the fix in changelogs. The changelog file is supposed to be for players and player had no idea of the error anyways

corp-0 (approved):

PR #8697, opened 2022-05-18
Fixes signal manager getting bricked by the despawn call

MaxIsJoe:
Ever wondered why all devices that use signal emitters/receivers would suddenly stop working on staging until the server fully reboots?
The reason behind this is the use of the `Despawn` class which doesn't trigger `OnDisable` at all; causing the signal manager to get stuck on an object that has been destroyed but was never properly removed from the list on the server. This fixes it

This should fix issues regarding C4/X4s/Synidcate bombs/etc from being only used once in a lifetime until the server completely reboots and accidentally destroying a handheld radio will no longer cause this.

## Changelog:

CL: [Fix] - Fixed an issue that caused the signal manager to completely break when despawning objects.

### DIFF
--- a/UnityProject/Assets/Prefabs/Items/Weapons/Explosive/C4.prefab
+++ b/UnityProject/Assets/Prefabs/Items/Weapons/Explosive/C4.prefab
@@ -300,8 +300,7 @@ MonoBehaviour:
   Frequency: 122
   Emitter: {fileID: 0}
   DelayTime: 3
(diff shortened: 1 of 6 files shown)


Bod9001 (approved):

MaxIsJoe:
tests froze. re-run pls

PerfectTangent:
rerunning.

PR #8699, opened 2022-05-18
Adds stun immunity to armor

### LINKED ISSUE #4364: Armours immune to tazer
## Description

Make thicker armour vests, riot suits and hardsuits immune to tazer shots.

Syndicate is pretty underwhelming when an assistant can take them out in one shot from a weapon with extremely high firing speed.

I'm honestly surprised we didn't have an issue made for this already. At least, not one I could find.

MaxIsJoe:
closes #4364

When stunning a player, the game will now check how much armor (Bomb and/or energy) does the player have and if it's greater than a set amount (default is 50) then the player wont get stunned.

### Changelog:
CL: [New] - Armors now have a parameter to prevent stuns.
CL: [Improvement] - Highlanders have full stun immunity on their kilts now, you can no longer cheese fights with stuns anymore.
CL: [Improvement] - Highlanders have full bullet immunity now and strong resistance to bombs.
CL: [Fix] - Fixes unity un-referencing a ProTip on tasers.

### DIFF
--- a/UnityProject/Assets/Prefabs/Items/Clothing/Outerwear/Armor/ArmorVests/WardensArmoredJacket.prefab
+++ b/UnityProject/Assets/Prefabs/Items/Clothing/Outerwear/Armor/ArmorVests/WardensArmoredJacket.prefab
@@ -54,6 +54,11 @@ PrefabInstance:
       propertyPath: armoredBodyParts.Array.data[0].armor.Bullet
       value: 40
       objectReference: {fileID: 0}
+    - target: {fileID: 5480096206054138487, guid: 259354123db311145a0ccc3922005d0f,
+        type: 3}
+      propertyPath: armoredBodyParts.Array.data[0].armor.StunImmunity
+      value: 0
+      objectReference: {fileID: 0}
     - target: {fileID: 8076471909809007804, guid: 259354123db311145a0ccc3922005d0f,
         type: 3}
       propertyPath: m_AssetId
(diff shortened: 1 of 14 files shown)


wraith-54321:
this will have to get reworked at some point but I think this will be good for now

PerfectTangent:
I'd much rather if it were just, you know, a bool that we can flip on a piece of armor rather than a quality that is co-opted in with other elements of balance (how much damage a person takes).

MaxIsJoe:
> I'd much rather if it were just, you know, a bool that we can flip on a piece of armor rather than a quality that is co-opted in with other elements of balance (how much damage a person takes).

There is a bool you can flip programmatically; it's enabled by default.

PerfectTangent:
> There is a bool you can flip programmatically; it's enabled by default.

then why have the armor check? It's still is a loss of control in our ability to balance the game, just from the top rather than from the bottom.

MaxIsJoe:
I dont see how its a loss of control tho?
depending on how thicc your armor is you should be able to logically be able to stop stuns from getting to you

PerfectTangent:
yes, and that should be in control of the balance team.

> I dont see how its a loss of control tho?

let's print out a hypothetical. Armor X and Y together pass the threshold to gain stun immunity. They are widely available and easily pair together, granting immunity trivially and disrupting game balance in an undesirable way. What can be done to fix this? The armor value of one or both sets of armor can be cut to get it lower than the stun immunity threshold. But now, those armors are significantly _underperforming_ because they are not blocking as much damage as they used to. We could increase the default threshold, but then that starts affecting armor A and B, which together don't pass the stun immunity threshold anymore, so they need to be tweaked and etc.

Linking the two different aspects of damage protection and stun protection together reduces control in a substantial way. If we are to have this threshold, it is going to require an immense amount of thought and balancing around how the current existing sets of armor in game could interact in ways that may create unforeseen powergaming strategies.

PerfectTangent:
The concept of a quality or stat of armor used to calculate stun immunity is not inherently a bad idea. Tying it to the same mechanic that also controls damage received is the problem.

MaxIsJoe:
Updated this based on our discussions on discord. Review time.

corp-0 (approved):

Issue #6906, opened 2021-07-02
You can't pick up mined ores on Asteroid09.unity by hand.


PetMudstone:
### Environment
- Game Version: Production 4119
- OS Version: Windows 10

### Detailed description of bug
When you mine ores on Asteroid09.unity, you cannot pick them up with your hands. Ore boxes and mining satchels can still pick up mined ores on Asteroid09.unity. Strangely this issue seems to be exclusive to that particular scene, as testing with another asteroid and lavaland let me pick up ores by hand no problem.

Images of Asteroid09.unity in 4119
![image](https://user-images.githubusercontent.com/38266309/124235734-7b312780-daca-11eb-8b53-2a90b53f324f.png)
![image](https://user-images.githubusercontent.com/38266309/124235761-82583580-daca-11eb-89e8-9c969d1394ed.png)

Images of Asteroid09.unity as of #6903.
![image](https://user-images.githubusercontent.com/38266309/124236132-f561ac00-daca-11eb-9ae5-64f10d6ade84.png)
![image](https://user-images.githubusercontent.com/38266309/124236162-fc88ba00-daca-11eb-8188-f65c4f32620c.png)

Note: Asteroid may appear rotated.

### Steps to reproduce
1. Find Asteroid09.unity (use above images for reference).
2. Mine ores on said asteroid.
3. Try to pick up the ores by hand, fail.

PR #8709, opened 2022-05-23
Takes Caverna's base layer out of objects

AtnerNT:
Somehow Caverna's base tilemap layer found its way into its objects. Hasn't been reported to cause any problems- but probably best its taken out to be safe.
Capture


Copern:
Think that is something we should add a test for? We have a test for correct parents on matrix syncs but not a test for layers.

PerfectTangent:
> Think that is something we should add a test for? We have a test for correct parents on matrix syncs but not a test for layers.

An excellent idea. Maybe it will help catch some of the random bugs a few scenes have like #6906.

Bod9001 (approved):

PR #8715, opened 2022-05-24
Fix viewing certain SpriteDataSO assets and use less memory

Copern:
Previously any SpriteDataSO's viewed in the editor would create and store texture copies of the sprite's frames. Depending on how many textures you viewed, this could add anywhere from just a couple MBs up to over 100 MBs of extra memory used up by the editor. This was changed so that textures will no longer be copied in order to preview SpriteDataSOs.

This PR also fixes the various issues with drawing large/small sprites or drawing sprites not matching 1:1 size ratios.

Some random samples before:
spritedatabefore

And after:
spritedataafter

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Editor/EditorDrawPreview.cs
+++ b/UnityProject/Assets/Scripts/Core/Editor/EditorDrawPreview.cs
@@ -1,207 +1,193 @@
-﻿using System;
+﻿#if UNITY_EDITOR
 using System.Collections.Generic;
-using System.Linq;
-using UnityEngine;
-#if UNITY_EDITOR
 using UnityEditor;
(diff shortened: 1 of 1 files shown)


MaxIsJoe:
oh my god yes, thank you Copern your last few PRs have been phenomenal

Bod9001:
Looks good I think this will work well with addressables to, if we have addressable sprites

Bod9001 (approved):

Bod9001:
you could Probably take a crack at the Highlight stuff as well if you want, since that's a bit similar though it needs to copy the complete texture of the item, to add the highlight effect idk, you can Probably just come up with some Hat-trick that makes it a million times better

PR #8720, opened 2022-05-27
Adds more diseases and basic random cures

MaxIsJoe:
This PR adds a bunch of new diseases that are actually far more threatening than the ones in-game so far as they can actually cause damage to you over time. This also adds a basic "random cure" system where at the start of each life cycle of a sickness a random reagent will be assigned as a cure. Players will have to hunt for reagents and test them before everyone gets infected and slowly start dying if they ever get one of those new threatening illnesses.

### Space Cancer 

Space cancer is a sensitive disease that does damage every 24 seconds unless the user is wearing protective bio gear.
It does 2 damage to a random organ and is contagious when not wearing something like a hazmat suit.

### Bone Consumer Disease

Remains of an old bio weapon traveling through space, It targets the infected's bones every 120 seconds and does 30 internal damage.
Can target prosthetics.

### Solar Dust Disease

A disease that infects the bloodstream of the infected and slowly release toxins in the user's body.
Does 2 internal toxin damage every 60 seconds.

## Notes :

I wanted to expand a little more on the cure part but I'd like to discuss howmest should we approach finding the cure first on discord because currently the idea is that the player(s) are just going to be injecting random reagents they find around until they find the cure... which they don't have any indication for.
More discussion is required so I'll leave it simple like this until we expand upon it later.

# Changelog:

CL: [New] - Added three new threatening diseases that actually harm the player over time.
CL: [New] - Some diseases can have random cures assigned when they first appear.


corp-0:
Space cancer doesn't sound much like cancer, tho. I think it should first not be contagious, and second target an organ, damage it a lot over time and then target another organ (so it spreads).

For a contagious illness you could do Covid-13, spread over air, makes you cough and sneeze and destroys lungs over time.

MaxIsJoe:
> Space cancer doesn't sound much like cancer, tho. I think it should first not be contagious, and second target an organ, damage it a lot over time and then target another organ (so it spreads).
> 
> For a contagious illness you could do Covid-13, spread over air, makes you cough and sneeze and destroys lungs over time.

Space cancers mainly targets organs already
They're contagious because space, video game logic.

speaking on contagious, symptoms are completely borked in editor so i cant change anything about them.

corp-0:
> speaking on contagious, symptoms are completely borked in editor so i cant change anything about them.

That's an issue. Players will start dying without knowing and that isn't fun

MaxIsJoe:
> That's an issue. Players will start dying without knowing and that isn't fun

No worries, I have a better thing.
I made it so that it manually uses the emote system, which means that if the cough emote has sounds it will play it.

Bod9001:
NetworkManagerHasSpawnListManager

MaxIsJoe:
Should be fixed now

Bod9001 (approved):

PR #8726, opened 2022-05-28
[Critical Fix] - Fixes game no longer generating ores

### LINKED ISSUE #8710: Ores are no longer generating
Pretty self explanatory but ores are no longer generating on lavaland or asteroids.

MaxIsJoe:
Fixes #8710

## Notes :

- **NEVER USE HARDCODED STRINGS FOR SEARCHING FOR ANYTHING BECAUSE THEY WILL CHANGE SOMETIME IN THE FUTURE WITHER YOU LIKE THAT OR NOT**
- I've moved calls to the `IServerSpawn` interface because from my last few PRs in the past few months i've learned the hard way that functions that are meant only for the server usually just cause issues when they're on Awake or Start even if a check from the custom net manager thingy is placed.

# Changelog:

CL: [Fix] - Fixed ore generation on lavaland and asteroids.

### DIFF
--- a/UnityProject/Assets/Scripts/Map/LavaLand/LavaLandRandomGenScript.cs
+++ b/UnityProject/Assets/Scripts/Map/LavaLand/LavaLandRandomGenScript.cs
@@ -38,7 +38,7 @@ namespace Systems.Scenes
 		[SerializeField]
 		private RandomItemSpot mobPools = null;
 
-		private void Start()
+		private void Awake()
 		{
 			LavaLandManager.Instance.randomGenScripts.Add(this);
 
--- a/UnityProject/Assets/Scripts/Map/LavaLand/LavaLandRandomGenScript.cs
+++ b/UnityProject/Assets/Scripts/Map/LavaLand/LavaLandRandomGenScript.cs
@@ -38,7 +38,7 @@ namespace Systems.Scenes
 		[SerializeField]
 		private RandomItemSpot mobPools = null;
 
-		private void Awake()
+		private void Start()
 		{
 			LavaLandManager.Instance.randomGenScripts.Add(this);
 
--- a/UnityProject/Assets/Scripts/Systems/Ore Generation/OreGenerator.cs
+++ b/UnityProject/Assets/Scripts/Systems/Ore Generation/OreGenerator.cs
@@ -14,7 +14,7 @@ using Tiles;
 /// <summary>
 /// Component which should go on a Matrix and which generates ore tiles in any mineable tiles of that matrix.
 /// </summary>
-public class OreGenerator : MonoBehaviour
+public class OreGenerator : MonoBehaviour, IServerSpawn
 {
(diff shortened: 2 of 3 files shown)


UnityProject/Assets/Scripts/Map/LavaLand/LavaLandRandomGenScript.cs
```diff
@@ -38,7 +38,7 @@ public class LavaLandRandomGenScript : MonoBehaviour
 		[SerializeField]
 		private RandomItemSpot mobPools = null;
 
-		private void Start()
+		private void Awake()
```
NoooneyDude: Is it guaranteed that LavaLandManager's Awake is called prior to this one? Seems fragile to me.

UnityProject/Assets/Scripts/Systems/Ore Generation/OreGenerator.cs
```diff
@@ -122,7 +113,7 @@ private void NodeScatter(Vector3Int location, OreProbability materialSpecified)
 			var chosenLocation = locations[RANDOM.Next(locations.Count)];
 			var ranLocation = chosenLocation + DIRECTIONS[RANDOM.Next(DIRECTIONS.Count)];
 			var tile = metaTileMap.GetTile(ranLocation);
-			if (tile != null && tile.name.Contains("rock_wall"))
+			if (tile != null)
```
NoooneyDude, line 116: A tile adjacent to a mineable tile might not be mineable.

NoooneyDude:
This will mean ore can spawn in any tile of type wall, which isn't great.

UnityProject/Assets/Scripts/Map/LavaLand/LavaLandRandomGenScript.cs
MaxIsJoe: You should always use `Awake()` for initializing data in unity and `Start()` for running things that affect the prefab or other game objects.
Also we use `Awake()` everywhere in the project so if it starts causing trouble now that means there's a million other scripts that are also fragile.

MaxIsJoe:
> This will mean ore can spawn in any tile of type wall, which isn't great.

Take a look a few lines ~~below~~ above that

UnityProject/Assets/Scripts/Map/LavaLand/LavaLandRandomGenScript.cs
NoooneyDude: Look carefully at LavaLandManager. It uses Awake() too (via parent), in order to instantiate the singleton. Therefore, there might be possibility that this script awakes before the manager awakes (probably hierarchy or scene-dependent) and so we'd get an NRE. That's why this particular case looks fragile - because it depends on load order.

MaxIsJoe: Changed

Bod9001 (approved):

PR #8727, opened 2022-05-28
Clears receivers in signal manager at round end

MaxIsJoe:
This should be the last time tackling the signal manager getting bricked.

CL: [Fix] - The signal manager will now clear it's receivers list at the end of the round.

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/EventManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/EventManager.cs
@@ -17,6 +17,7 @@ public enum Event
 	LoggedOut,
 	RoundStarted,
 	PostRoundStarted,
+	SceneUnloading,
 	RoundEnded,
 	DisableInternals,
 	EnableInternals,
--- a/UnityProject/Assets/Scripts/Managers/GameManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/GameManager.cs
@@ -821,6 +821,7 @@ public partial class GameManager : MonoBehaviour, IInitialise
 			Chat.AddGameWideSystemMsgToChat("<b>The round is now restarting...</b>");
 			// Notify all clients that the round has ended
 			EventManager.Broadcast(Event.RoundEnded, true);
+			EventManager.Broadcast(Event.SceneUnloading, true);
 
 			yield return WaitFor.Seconds(0.2f);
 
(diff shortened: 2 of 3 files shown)


corp-0 (dismissed):

NoooneyDude:
This doesn't cover the case where someone adds a receiver after the round has ended but before restart.

MaxIsJoe:
> This doesn't cover the case where someone adds a receiver after the round has ended but before restart.

what do you mean

NoooneyDude:
> what do you mean

e.g. while the crew are at CentCom

MaxIsJoe:
> > what do you mean
> 
> e.g. while the crew are at CentCom

what if you need to use the signal manager while the crew at are centcom?
what if the admin created a thing that requires signals and he was planning to trigger it before the round completely restarts?
what if a traitor had a bunch of explosions on him and he wanted to blow it up once the shuttle arrives?

I dont get why you want to signal manager to stop working when there's still gameplay going on; it should only clear it when the old receivers are no longer present in-game anymore because a new fresh round has started.

MaxIsJoe:
It should also be noted that on some MRP and HRP servers people still roleplay even after the round has ended in centcom and the round end could go up to 3+ minutes for admins to wrap things up and for antags to pull out their last stunt. Stopping an entire system from working while people are playing is a bad idea.

NoooneyDude:
You misunderstand me. You are clearing while players are still able to use the manager. They can still use it after the round has ended, (thus issue still exists). You want to clear the list when the scene is unloaded instead.

MaxIsJoe:
> You want to clear the list when the scene is unloaded instead.

How do i do that?

NoooneyDude:
> > You want to clear the list when the scene is unloaded instead.
> 
> How do i do that?

Hey Max. Sorry, this is partially my mistake. It turns out that the event is broadcast twice; once when the round ends and once just before the round restarts, which isn't good. Would you mind changing this by adding a `RoundRestarting` event and changing one of the broadcasts to broadcast this instead? 

`GameManager.cs` lines ~`567` ~`823` and `GUI_InGameMenu` line ~`201` (broadcast both events for the latter). You might need to check all the current listeners for the `RoundEnded` event and make sure they listen to the appropriate event.

MaxIsJoe:
> GameManager.cs

Done.

MaxIsJoe:
hewwooo? wewview pwease

corp-0:
it looks fine, but Nooney was reviewing it

corp-0 (approved):

PR #8735, opened 2022-05-29
ConnectedPlayer changes

NoooneyDude:
- `ConnectedPlayer` now provides role / online information to reduce lookups.
- Rename `ConnectedPlayer` to `PlayerInfo` now that these objects have been made available for offline players.
- Rename `PlayerManager.LocalPlayer` to `PlayerManager.LocalPlayerObject` (I initially had `LocalPlayer` referring to the local player `PlayerInfo` object instead, though I had some difficulties - may reintroduce it later).
- Remove duplicate `PlayerManager.PlayerScript` property, in favour of `PlayerManager.LocalPlayerScript`.
- Use `PlayerInfo` for parameters instead of `userID` strings.
- `PlayerList` method tweaks.

Prefer to use  `PlayerInfo` for parameters and method return types (particularly in public methods) instead of user ID strings or what-not, in order to reduce somewhat expensive ID lookups (there are more than enough cases where each method in a call stack would do their own `ConnectedPlayer` lookup from the ID string to get some information - we just need to do this once, early on).

> Testing in progress...


Bod9001:
how would we approach merging this with the movement since since I changed quite a few functions, and there might be some clashes? I merge this first and then I fixed the changes or merge movement first and then fixed the clashes?

NoooneyDude:
> how would we approach merging this with the movement since since I changed quite a few functions, and there might be some clashes? I merge this first and then I fixed the changes or merge movement first and then fixed the clashes?

Most of my changes are simple, so probably best to merge movement first.

Bod9001:
ok cool

corp-0:
no longer on hold as movement  has been merged. Oh boy quite a list of conflicts

Bod9001:
you need any help sorting out conflicts?

corp-0 (approved):

Bod9001 (approved):

NoooneyDude:
Oof I don't think this was ready to be merged 😛 was running into issues with testing but haven't figured if my changes were broken or something else. Sorry, should've added DNM.

corp-0:
oh oof, we can revert it tho. We didn't even make a build out of this. Would you rather it keep merged and fix issues or just revert?

PR #8742, opened 2022-05-29
Fixes stun immunity spawning particle feedback when not needed

MaxIsJoe:
CL: [Fix] - Fixes stun immunity spawning particle feedback when not needed


Alainx277 (dismissed):
Should be refactored, but this is a quick fix.

Alainx277 (approved):
Even better, thanks Max!

Bod9001 (approved):

PR #8749, opened 2022-06-01
Fixes issue regarding Jukebox playing more than one track at once

### LINKED ISSUE #8738: Spamming the jukebox music change buttons causes multiple tracks to play
### Environment
- Game Version: 22052105
- OS Version: Windows 10

### Detailed description of the bug
I have only observed this with two players simultaneously, but it may also happen with one player. If you spam the heck out of the "switch song" buttons, you'll break the jukebox so that  it'll play multiple songs at the same time, delayed by milliseconds next to each other, so it sounds awful

### Steps to reproduce

1. Start the jukebox playing
2. Play cookie clicker on the "Switch song" buttons (maybe with two people)
3. **Pain**

MaxIsJoe:
### Purpose
Fix #8738

### Notes

I've added a new audio bus for juke boxes so songs dont sound muffled, now they sound like they are coming from a jukebox with a highpass filter and a tiny bit of distortion 
I've also added a new way for us poor programmers to stop audio
we no longer have to use strings for stopping audio, just give the play function a integer sound ID and you can now stop multiple different audio clips that have the same integer ID 
much more flexiable and important for cases like in the jukebox where there could be multiple audio sources playing on one object but each audio clip has it's own string ID that's assoicated with the addressable audio source itself
by manually giving them an integer ID we have more control over them even if they have different addressable audio sources

# Changelog:

CL: [Fix] - Fixed issue where Jukeboxes would play more than two tracks at once.
CL: [Improvement] - The Jukebox now has it's own audio bus and audio effect. No longer will you listen to muffled music.


corp-0:
after reading, I'm still not quite convinced of the usufulness of the integer ID when the spawn token does exactly the same thing. I will wait for what others have to say.

Alainx277:
It seems like `SoundSpawnToken` is exactly what you want. It references a specific sound being played, and allows you to stop it.

Using an integer is not great either, as there's no guarantee it won't be accidentally reused in two places. The existing system uses GUIDs, which is much safer.

MaxIsJoe:
IDexplanationdigram

I've mentioned this on Discord but the problem with the SoundSpawnToken is that it's associated with the the sound clip that's being played, not the actual object or whats it being called under exactly.
If you don't know the sound token of two or more sounds you're going to run into issues like the jukebox currently where it plays 2+ sounds at the same time and because you only know the sound token of one sound you can only stop one sound while the other continues playing.

corp-0:
like I said in discord, adding the integer ID is not the fix to stop jukebox from playing two songs when it is first interacted with (not to mention it didn't happen before, so it is a new bug and has nothing to do with the spawnToken and its implementation).

MaxIsJoe:
review pls

corp-0 (approved):
Lgtm

PR #8757, opened 2022-06-02
movement fixes

Bod9001:
### Changelog:

fixes #8751
fixes #8755

CL: [Fix] Fixes ghost movement only working in two directions. #8751
CL: [Fix] You are not able to Pushing objects you are dragging  #8755
CL: [Fix]  error on start-ups fixed
CL: [Fix]  fixes being able to swap through window

PR #8759, opened 2022-06-02
Readjusts the speeds of several mobs

MaxIsJoe:
CL: [Balance] - A lot of mobs no longer move like snails.

### DIFF
--- a/UnityProject/Assets/Prefabs/Mobs/Living/SimpleAnimals/Friendly/Butterfly.prefab
+++ b/UnityProject/Assets/Prefabs/Mobs/Living/SimpleAnimals/Friendly/Butterfly.prefab
@@ -24,6 +24,11 @@ PrefabInstance:
       value: 
       objectReference: {fileID: 11400000, guid: a3ec407f0fc55924b94ec88a58681c7d,
         type: 2}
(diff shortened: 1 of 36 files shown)


corp-0:
#8757 also has some speed adjustements, but honestly I'd rather have them all here in a single pr and the other one only code fixes.

MaxIsJoe:
> #8757 also has some speed adjustements, but honestly I'd rather have them all here in a single pr and the other one only code fixes.

Bod only adjusted the speed of animals

corp-0:
only sloth, it seems. Fair enough

corp-0 (approved):

corp-0:
will you add any other mob here? Otherwise I will merge it now

PerfectTangent:
I think the base mobs should be modified first before modifying individual mobs. That way it's more consistent and less mobs get omitted.

MaxIsJoe:
> I think the base mobs should be modified first before modifying individual mobs. That way it's more consistent and less mobs get omitted.

I've modifed some base prefabs but i gave some unique speeds (i.e hunters being faster than other xenos and queens being slower)

MaxIsJoe:
> will you add any other mob here? Otherwise I will merge it now

nah not planning, merge

PR #8771, opened 2022-06-04
mob limit

Bod9001:
why because the thing that always is responsible for tanking performances bloody mobs, 
so adds a limit to the number of mobs that can be active , can change limit is needed
Looking at the profiles there was 81 hostile mobs looking for players at the same time

### DIFF
--- a/UnityProject/Assets/Scripts/NPC/AI/MobController.cs
+++ b/UnityProject/Assets/Scripts/NPC/AI/MobController.cs
@@ -32,7 +32,7 @@ namespace Systems.MobAIs
 		{
 			if (CustomNetworkManager.IsServer == false) return;
 			SceneManager.activeSceneChanged += OnRoundRestart;
-			if (CurrentMobs > 10)
+			if (CurrentMobs > 20)
 			{
 				return;
 			}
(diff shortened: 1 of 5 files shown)


corp-0:
the codacy warnings make me uneasy with the static solution, not going to lie.

Bod9001:
it is a static cap Though, applying across all classes so it makes sense to be static

UnityProject/Assets/Scripts/NPC/AI/MobController.cs
```diff
@@ -3,30 +3,56 @@
...
 		}
 
 		private void OnEnable()
 		{
 			if (CustomNetworkManager.IsServer == false) return;
-		
+			SceneManager.activeSceneChanged += OnRoundRestart;
+			if (CurrentMobs > 20)
+			{
+				return;
```
NoooneyDude, line 40: Could we send this to admin chat? Just to keep them informed

Bod9001: ok cool will do

corp-0 (approved):

PR #8773, opened 2022-06-04
Improvements again

Bod9001:
basically looked at profiles and Looked at what was slow


NoooneyDude (approved):
Thanks for fixing parameter style 👍

PR #8798, opened 2022-06-07
Fallstation medical rework + NanoDrug

PerfectTangent:
### Purpose
 
This PR expands Fallstation's medical department, rearranging the space to fix a number of problematic parts.

### Notes:

- fixed the several null elements in the HoP's button door line. there should probably be a test to spot these sorts of things.
- the dorms now have direct access to the maintenance tunnel to the south. shrunk the morgue and the back kitchen room slightly to allow for this.
- fixed several cases of duplicate pipes stacked on the same tile. For some reason tests weren't catching this problem until I started editing pipes, then duplicates showed up all across the map. Thanks to whoever added coordinates to the map tests. I couldn't have fixed it without them.

### Changelog:

CL: [New] Fallstation's chemistry room has gained a NanoDrug+ vendor and a chemistry closet stocked with basic medicines.
CL: [Balance] Fallstation's medical department has been reworked. The department battery is now inside an engineering maintenance room. The CMO's office has been moved to the center and been given window shutters for privacy. Virology has been expanded and rearranged. An extra medical storage room has been added to the back. The Morgue was shrunken slightly and has gained a medical supplies closet.


PerfectTangent:
How it looks. The new room between medbay and virology is intentionally left empty so that future medical tools such as cryo can be put there.
Fall medical makeover

MaxIsJoe:
This is a very welcome change because I've abused the map layout before as a traitor by putting X4s outside the cloning room while outside in space which frustrated people because that meant they couldn't be cloned again for a while and they couldn't catch who did it because the culprit was always in the darkness of space.
The new window placements also make it harder for antags to try and sabotage medical from outside so thats good.

PerfectTangent:
I'm not sure that genetics is going to be substantially more secure with these changes, as it's bordered by maintenance at the bottom.

MaxIsJoe:
Oh right, we dont even have much content for geneticists to be there in the first place.. welp, thats a problem that will be solved in the future.

corp-0 (approved):

PR #8806, opened 2022-06-08
Fixes admin getting kicked due to out of sync data in the cargo manager tool (again) + Other fixes

MaxIsJoe:
For admins with slow connections and servers that are running at 0FPS, i've finally found the perfect solution to prevent all of you from getting kicked.
You cant use wrong outdated data if there is none, the game will now completely empty the entire cargo bounty list when a change is made and you'll have to wait until you receive the new changes.
Also the `CompleteBounty()` function now has a check for stopping admins from using index values that are higher than the ones that exist on the active bounties list.

# Changelog:

CL: [Fix] - Admins can no longer get kicked when experiencing lag while editing the cargo bounties.
CL: [Fix] - When the cargo console goes back online it will now switch pages to a the cart instead of staying on the offline page.
CL: [Improvement] - Admins have more logs now when using the cargo manager tool.

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/AdminCommandsManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/AdminCommandsManager.cs
@@ -516,7 +516,7 @@ namespace AdminCommands
 		public void CmdRemoveBounty(int index, bool completeBounty, NetworkConnectionToClient sender = null)
 		{
 			if (IsAdmin(sender, out var admin) == false) return;
-			if (CargoManager.Instance.ActiveBounties.Count - 1 < index) return; 
+			if (CargoManager.Instance.ActiveBounties.Count <= index) return; 
 			if (completeBounty)
 			{
 				CargoManager.Instance.CompleteBounty(CargoManager.Instance.ActiveBounties[index]);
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Managers/AdminCommandsManager.cs
```diff
@@ -516,6 +516,7 @@ public static void LogAdminAction(string msg, string userName = "")
 		public void CmdRemoveBounty(int index, bool completeBounty, NetworkConnectionToClient sender = null)
 		{
 			if (IsAdmin(sender, out var admin) == false) return;
+			if (CargoManager.Instance.ActiveBounties.Count - 1 < index) return;
```
Alainx277:
```suggestion
			if (CargoManager.Instance.ActiveBounties.Count <= index) return; 
```

MaxIsJoe: that is wrong because count doesn't return the value of the last index, it instead returns the number of entries in a list.

Alainx277: Yes, that's what the equals sign is for. `x - 1 < y` is the same as `x <= y`.

MaxIsJoe: then this seems like an unnecessary nitpick if it does the same exact thing

Bod9001: it's easier to read with the <=

MaxIsJoe:
e

Bod9001 (approved):

PR #8826, opened 2022-06-12
fixes MagBoots not working

### LINKED ISSUE #8812: magboot are sticky
Magboot (all) sticks you to the ground

Bod9001:
fixes ghost moovement of long ranges
fixes maximum speed
fixes : #8812

### DIFF
--- a/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
+++ b/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
@@ -1078,7 +1078,7 @@ public class UniversalObjectPhysics : NetworkBehaviour, IRightClickable
 		{
 			worldDirection.Normalize();
 
-			SetNewtonianMovement = newtonianMovement + worldDirection * speed;
+			SetNewtonianMovement += worldDirection * speed;
 
 			if (float.IsNaN(nairTime) == false)
 			{
@@ -1103,7 +1103,7 @@ public class UniversalObjectPhysics : NetworkBehaviour, IRightClickable
 			}
 
 			worldDirection.Normalize();
-			SetNewtonianMovement = newtonianMovement + worldDirection * speed;
+			SetNewtonianMovement += worldDirection * speed;
 
 		}
 
--- a/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
+++ b/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
@@ -61,7 +61,7 @@ public class UniversalObjectPhysics : NetworkBehaviour, IRightClickable
 
 	[PrefabModeOnly] public bool Intangible = false;
 
-	[PlayModeOnly] public bool CanNotBeWindPushed = false;
+	[PlayModeOnly] public bool CanBeWindPushed = true;
 
 	public Vector3WithData SetLocalTarget
 	{
--- a/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
+++ b/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
@@ -61,6 +61,8 @@ public class UniversalObjectPhysics : NetworkBehaviour, IRightClickable
 
 	[PrefabModeOnly] public bool Intangible = false;
 
+	[PlayModeOnly] public bool CanNotBeWindPushed = false;
+
 	public Vector3WithData SetLocalTarget
 	{
 		set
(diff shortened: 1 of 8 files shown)


UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
```diff
@@ -176,6 +178,21 @@ public ObjectContainer ContainedInContainer
 
 	[PlayModeOnly] public Vector2 newtonianMovement; //* attributes.Size -> weight
 
+
+	public Vector2 SetNewtonianMovement
```
Alainx277: It seems misleading to prefix a property with `Set`. Why not just call it `NewtonianMovement`?

If the name is too similar to `newtonianMovement`, the convention is to prefix the backing field with an underscore.

UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
```diff
@@ -61,6 +61,8 @@ public class UniversalObjectPhysics : NetworkBehaviour, IRightClickable
 
 	[PrefabModeOnly] public bool Intangible = false;
 
+	[PlayModeOnly] public bool CanNotBeWindPushed = false;
```
Alainx277: Booleans with `Not` in their names tend to cause confusion. Could it be something like `WindImmunity` instead?

If it were inverted, it could just be called `CanBeWindPushed`.

UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
```diff
@@ -1060,16 +1077,8 @@ public void PullSet(UniversalObjectPhysics toPull, bool byClient, bool synced =
...
-			{
-				//Logger.LogError("  momentum boost");
-			}
 
-			newtonianMovement += worldDirection * speed;
-			if (newtonianMovement.magnitude > MAX_SPEED)
-			{
-				newtonianMovement *= (MAX_SPEED / newtonianMovement.magnitude);
-			}
+			SetNewtonianMovement = newtonianMovement + worldDirection * speed;
```
Alainx277: Why not use the new property here too? It could then be simplified to just:
```suggestion
			SetNewtonianMovement += worldDirection * speed;
```

UnityProject/Assets/Scripts/Player/MovementV2/MovementSynchronisation.cs
```diff
@@ -330,11 +330,17 @@ public void Update()
...
 			CheckWindOtherPush();
 		}
-
+		else
+		{
+			if (PushingData.magnitude != 0)
+			{
+				PushingData = Vector2.zero;
+			}
+		}
```
Alainx277:
```suggestion
		else if (PushingData.magnitude != 0)
		{
			PushingData = Vector2.zero;
		}
```

UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
Bod9001: find inverted would also be more confusing maybe?

UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
Bod9001: it's the one you want to use when you want to set newtonianMovement,   Since it has the cap on it

UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
Alainx277: I think `CanBeWindPushed` is clearer than `CanNotBeWindPushed == false`, in the second case you have two negations.

UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
Alainx277: It's a property, so you can read from it too, in that case `Set` makes no sense. I assume `newtonianMovement` should never exceed the `MAX_SPEED`, so making it private and only using the property seems better?

corp-0: yeah, that's exactly what properties are for. Make the field private and wrap it with the property which has the cap when you set a value to it, return the private field when you get it.

UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
corp-0: agree

Alainx277 (approved):

PR #8827, opened 2022-06-12
Removes highlander from the list of gamemodes

MaxIsJoe:
For some reason the game-mode kept triggering more frequently than it should so now it's just an event that admins trigger for the time being.

### Changelog:

CL: [Balance] - Highlander is now an event only and not a gamemode.


MaxIsJoe:
Yes this also means you can no longer vote for it as well. You'll have to bug admins to trigger it for you or wait for the random event system start it for you when enough players are on.

Bod9001 (approved):

Issue #8824, opened 2022-06-11
a multitool cannot be placed on a table, or thrown on top, 


Fr-Dae:
a multitool cannot be placed on a table, or thrown on top,

corp-0:
is it only the multitool? This issue needs more info

Fr-Dae:
the multitool tries to interact with the table, and does not want to be placed on it.

Fr-Dae:
atmo tool too

PR #8829, opened 2022-06-13
throwing over tables

### LINKED ISSUE #8754: can't throw items to tables
### Environment
- Game Version: 22060122
- OS Version: win10

When throwing an item to a table, the expected behaviour is to land on top of it. Now it will instead bounce off the table.

### Steps to reproduce

1. Pick item and throw it aiming to a table
2. cry when it bounces off instead of landing on top of it.

Bod9001:
Pretty much as it says


Fr-Dae:
https://github.com/unitystation/unitystation/issues/8824

PR #8869, opened 2022-06-18
CharacterSettings changes

### LINKED ISSUE #7801: Null Username
-No idea how to reproduce but the issue is here:

image

Clients are sending null there, so something is broken in their current character settings

Not sure why we store the username in their as it should be the same per character, in the future we should just change it so when we do the account validation we get the username sent back from the auth

As of #7784 users cant join with null username and will get kicked (Says they have emtpy/null username in the kick message)

NoooneyDude:
- Renames `CharacterSettings` to `CharacterSheet` to make it more clear this is for details on the player's character.
- Removes `Username` field from `CharacterSettings`. It doesn't belong there. Previous usages of this member now source the username from somewhere more appropriate. Fixes #7801 (an assumption, not tested).
- Random character generation & related code refactors.
- Misc. changes  I had lying around from many months ago. 🤷 
- More instances of user ID strings converted to use `PlayerInfo`.
- Fixes an exploit where players (and admins/mentors) with modified clients could masquerade as another player in admin/mentor chats.
- Adds convenient `PickRandom()` extension to enums (generates garbage; use sparingly).

> Unable to test on headless due to unrelated issues.

### DIFF
--- a/UnityProject/Assets/Scripts/Clothing/RaceSOSingleton.cs
+++ b/UnityProject/Assets/Scripts/Clothing/RaceSOSingleton.cs
@@ -1,11 +1,25 @@
-﻿using System.Collections;
-using System.Collections.Generic;
-using ScriptableObjects;
+﻿using System.Collections.Generic;
 using UnityEngine;
+using ScriptableObjects;
 
 [CreateAssetMenu(fileName = "RaceSOSingleton", menuName = "Singleton/RaceSOSingleton")]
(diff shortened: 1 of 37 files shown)


Bod9001 (approved):

Bod9001:
"Adds convenient PickRandom() extension to enums (generates garbage; use sparingly)." 
I remember being somewhere in the project already idk tho

corp-0 (approved):

PR #8870, opened 2022-06-18
Guns eject magazine into hand and loading + unloading sfx for guns.

### LINKED ISSUE #8763: unloading a gun's magazine should place it in your hand and should have sfx.
### Description

Allow guns with magazines attached to them drop the magazine into your hand if you alt-click the gun with an empty hand, instead of having it drop to the floor regardless of what you do. On top of that, give the gun a mag-loading sound, to indicate it is loaded.
As a bonus step, press the gun again to cock it if it has no bullets in the chamber.

PerfectTangent:
### Purpose

This PR makes it so guns try to eject into a player's empty hand rather than always drop on the floor. In addition, sounds have been added for guns loading and unloading.

closes #8763.

changes tested in editor.

### Changelog:

CL: [Improvement] when removing a magazine from a gun, the magazine will be put in the player's empty hand if they have one.
CL: [Improvement] guns have sound effects for when they load and unload magazines.
CL: [Balance] Stalker SMG is now a smart weapon; it ejects empty magazines automatically onto the floor.
CL: [Balance] tactical energy carbine has double the energy capacity and fires 20% faster than the standard energy carbine.


Bod9001 (dismissed):

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
```diff
@@ -281,6 +287,7 @@ public virtual bool WillInteract(HandActivate interaction, NetworkSide side)
 			//try ejecting the mag if external
 			if (CurrentMagazine != null && allowMagazineRemoval && !MagInternal)
 			{
+				UnloadMagSound();
```
corp-0: this method, ``WillInteract`` runs once in the client requesting the interaction and then on server to ensure it is a valid interaction. Did you test this change in client & server environment? Does it play the sound twice?

It is also not the best place to call the sound, imo. It would be better to call it inside the method that handles unloading and loading mags respectively.

corp-0 (changes requested):
move the call to playing the sound to the methods that handle unloading and loading, I saw there is a ``ServerHandleUnloadRequest`` which is better for this than the ``WillInteract`` method. I assume there is also a ``ServerHandleLoadRequest``

PerfectTangent:
done.

NoooneyDude (approved):

Bod9001 (approved):

PR #8871, opened 2022-06-18
Staging fixes2

Bod9001:
more in-depth stuff tomorrow

### Changelog:
CL: [Fix] Fixes wind movement
CL: [Fix] fixes conveyor belts being funny
CL: [Fix] Fixes space walking not updating momentum on client
CL: [Fix] fixes surgical table  you can put people on it now
CL: [Fix] fixes  heal organ procedure, was buggy out with the usual interaction
CL: [Fix] fixes extra lag from lots of failing move messages, causing the server to lag more


corp-0:
can you elaborate a little bit on the first two fixes in the changelog?

NoooneyDude:
Maybe leave the logs uncommented but set as trace level for `Category.Movement`? Assuming the server log level is normal or higher, but shouldn't be too hard to change.

Bod9001:
my only, concern is stuff like a lot of strings being made for logging that gets discarded?

corp-0 (approved):

PR #8873, opened 2022-06-18
Increases conveyor belt speed

MaxIsJoe:
CL: [Improvement] - Increases conveyor belt speed from 0.8 to 4

### DIFF
--- a/UnityProject/Assets/Prefabs/Objects/Construction/ConveyorSwitch.prefab
+++ b/UnityProject/Assets/Prefabs/Objects/Construction/ConveyorSwitch.prefab
@@ -329,7 +329,7 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   conveyorBelts: []
-  ConveyorBeltSpeed: 0.8
+  ConveyorBeltSpeed: 4
   conType: 2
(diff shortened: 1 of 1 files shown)


NoooneyDude (approved):
Which Factorio belt colour is this equivalent to? : P

PerfectTangent:
Let's see. Factorio game tile is 1m. SS13 game tile is 2.5m. Assuming this belt speed is tile per second, then the old speed of 0.8 is 6% faster than yellow factorio belts and the new speed is 533% faster than yellow factorio belts, or 77% faster than blue belts, if my math is correct.

PerfectTangent:
actually, how fast is this in game? Is there enough time for people to get up from being ejected from disposals before they are swept away to the mass driver? With the old speed (pre movement rework, at least) it was already pretty close.

corp-0:
give a gif, I also want to see how fast it looks now

MaxIsJoe:
> give a gif, I also want to see how fast it looks now

What you'll see in the gif wont carry over to staging terms because the server shows the movement much more smoothly than on an actual live server.
However looks doesn't matter, functionality does; the 4 speed is the equveilent of someone switching their movement speed to walking instead of running.

PR #8875, opened 2022-06-19
Authentication move

NoooneyDude:
- Moves our client -> server authentication code to `Authenticator.cs`, a Mirror-derived component attached to the `NetworkManager` prefab.
This now happens _before_ we register the player on the player lists.
- Reduces `Ignorance` logging level now that the Mac issue is resolved.


corp-0 (dismissed):

Bod9001:
does this affect the request open scenes thing? 
Since currently the client Loads all the scenes before requesting data

NoooneyDude:
I think that's in `JoinedViewer`, right? Authentication is now done before the JoinedViewer object is spawned and authority given to the client so shouldn't be affected.

Bod9001:
well, the spawn Scenes was a bit intermingled with the authentication

Bod9001:
had a look at the code it looks good

Bod9001 (approved):

PR #8879, opened 2022-06-20
synchronising slippery tiles

Bod9001:
Done

### Changelog:

 CL: [Fix] fixes slippery floor tiles not being synchronised to the client properly

### DIFF
--- a/UnityProject/Assets/Scripts/Messages/Server/MetaDataLayerMessage.cs
+++ b/UnityProject/Assets/Scripts/Messages/Server/MetaDataLayerMessage.cs
@@ -1,0 +1,131 @@
+using System.Collections;
+using System.Collections.Generic;
+using Messages.Server;
+using Mirror;
+using Tilemaps.Behaviours.Layers;
+using UnityEngine;
+
+public class MetaDataLayerMessage : ServerMessage<MetaDataLayerMessage.NetMessage>
+{
+	public struct NetMessage : NetworkMessage
+	{
+		public List<DelayedData> Changes;
+		public uint MatrixSyncNetID;
+	}
+
+	//just a best guess, try increasing it until the message exceeds mirror's limit
+	private static readonly int MAX_CHANGES_PER_MESSAGE = 350;
+
+
+	public override void Process(NetMessage msg)
+	{
+		if (CustomNetworkManager.IsServer) return;
+		LoadNetworkObject(msg.MatrixSyncNetID);
+
+		var MetaDataLayer = NetworkObject.transform.parent.GetComponentInChildren<MetaDataLayer>();
+		foreach (var Change in msg.Changes)
+		{
+			MetaDataLayer.Get(Change.Position).IsSlippery = Change.IsSlippy;
+		}
+	}
+
+	public static void SendTo(GameObject managerSubject, NetworkConnection recipient,
+		Dictionary<Vector3Int, MetaDataNode> changeList)
+	{
+		if (changeList == null || changeList.Count == 0) return;
+		var netID = managerSubject.transform.parent.GetComponent<NetworkedMatrix>().MatrixSync.netId;
+		foreach (var changeChunk in changeList.Chunk(MAX_CHANGES_PER_MESSAGE))
+		{
+			List<DelayedData> Changes = new List<DelayedData>();
+
+			foreach (var metaData in changeChunk)
+			{
+				Changes.Add(new DelayedData()
+				{
+					Position = metaData.Key,
+					IsSlippy = metaData.Value.IsSlippery
+				});
+			}
+
+			NetMessage msg = new NetMessage
+			{
+				MatrixSyncNetID = netID,
+				Changes = Changes
+			};
+
+			SendTo(recipient, msg);
+		}
+	}
+
+	public static void Send(GameObject managerSubject, List<MetaDataNode> changeList)
+	{
+		var netID = managerSubject.transform.parent.GetComponent<NetworkedMatrix>().MatrixSync.netId;
+		// foreach (var changeChunk in changeList.Chunk(MAX_CHANGES_PER_MESSAGE)) //TODO Check it's not too big maybe
+		// {
+		List<DelayedData> Changes = new List<DelayedData>();
+
+		foreach (var metaData in changeList)
+		{
+			Changes.Add(new DelayedData()
+			{
+				Position = metaData.Position,
+				IsSlippy = metaData.IsSlippery
+			});
+		}
+
+		NetMessage msg = new NetMessage
+		{
+			MatrixSyncNetID = netID,
+			Changes = Changes
+		};
+		SendToAll(msg);
+		// }
+	}
+
+	public struct DelayedData
+	{
+		public Vector3Int Position;
+		public bool IsSlippy;
+	}
+}
+
+public static class UpdateTileMessageReaderWriters
+{
+	public static MetaDataLayerMessage.NetMessage Deserialize(this NetworkReader reader)
+	{
+		var message = new MetaDataLayerMessage.NetMessage();
+		message.Changes = new List<MetaDataLayerMessage.DelayedData>();
+		message.MatrixSyncNetID = reader.ReadUInt();
+		while (true)
+		{
+			var Continue = reader.ReadBool();
+			if (Continue == false)
+			{
+				break;
+			}
+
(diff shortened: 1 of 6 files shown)


UnityProject/Assets/Scripts/Messages/Server/MetaDataLayerMessage.cs
```diff
@@ -0,0 +1,131 @@
...
+				Position = metaData.Position,
+				IsSlippy = metaData.IsSlippery
+			});
+		}
+
+		NetMessage msg = new NetMessage
+		{
+			MatrixSyncNetID = netID,
+			Changes = Changes
+		};
```
Alainx277, line 81: This is duplicated code from `SendTo` above. Can you extract it into a method or similar?

UnityProject/Assets/Scripts/Messages/Server/MetaDataLayerMessage.cs
```diff
@@ -0,0 +1,131 @@
...
+		var message = new MetaDataLayerMessage.NetMessage();
+		message.Changes = new List<MetaDataLayerMessage.DelayedData>();
+		message.MatrixSyncNetID = reader.ReadUInt();
+		while (true)
+		{
+			var Continue = reader.ReadBool();
+			if (Continue == false)
+			{
+				break;
+			}
```
Alainx277, line 106: Instead of writing many bools, you could instead just write the length of the list at the beginning.

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaDataLayer.cs
```diff
@@ -1,22 +1,39 @@
...
 public class MetaDataLayer : MonoBehaviour
 {
-	private SerializableDictionary<Vector3Int, MetaDataNode> nodes = new SerializableDictionary<Vector3Int, MetaDataNode>();
 
+	//Easiest to synchronise?
+	//Attribute
+	//What needs to be done
+
+	//set current state on clients
+	//Only send the different node
```
Alainx277: I assume these are leftovers, if yes they should be removed.

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaDataLayer.cs
```diff
@@ -1,22 +1,39 @@
...
+
+
+	private Dictionary<Vector3Int, MetaDataNode> nodes = new Dictionary<Vector3Int, MetaDataNode>();
+
+
+	public Dictionary<Vector3Int, MetaDataNode> ChangedNodes = new Dictionary<Vector3Int, MetaDataNode>(); //Used for networking
```
Alainx277: Please use doc comments.
I think it would also help to specify how they are used (ex. keeping track of nodes that need to be synchronized)

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaDataLayer.cs
```diff
@@ -1,22 +1,39 @@
...
+
+	public List<MetaDataNode> nodesToUpdate = new List<MetaDataNode>();
```
Alainx277, line 31: What's the difference between `ChangedNodes` and this member?

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/MetaDataNode.cs
```diff
@@ -326,6 +326,12 @@ private void SyncNeighbors()
 			}
 		}
 	}
+
+	public void ForceUpdateClient()
+	{
+		PositionMatrix.MetaDataLayer.nodesToUpdate.Add(this);
+		PositionMatrix.MetaDataLayer.ChangedNodes[Position] = this;
```
Alainx277: If this were a method on `MetaDataLayer`, you could make these members private and ensure the lists are always consistent.

Alainx277 (changes requested):

UnityProject/Assets/Scripts/Messages/Server/MetaDataLayerMessage.cs
Bod9001: I can't really because one uses dictionary one uses list

UnityProject/Assets/Scripts/Messages/Server/MetaDataLayerMessage.cs
Bod9001: it's going to grow in complexity and have, conditional components so a fixed list wouldn't make sense

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaDataLayer.cs
Bod9001: more efficient have a list

Bod9001: you also need to be separate so, changes per frame can be worked out

UnityProject/Assets/Scripts/Messages/Server/MetaDataLayerMessage.cs
Alainx277: You could do it by taking a `IEnumerable<MetaDataNode>` as a parameter.
I noticed the commented out code that's actually present in the above method, which makes me want this to be factored out even more.

UnityProject/Assets/Scripts/Messages/Server/MetaDataLayerMessage.cs
Alainx277: You will need to encode each optional field with a boolean anyway, that's not incompatible with a length. I don't know any protocol that doesn't prefix lists with a length (except fixed length lists).

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaDataLayer.cs
Alainx277: I'm skeptical about it being faster (unless it's queried more than inserted into) but I'm going to assume you've profiled it.

UnityProject/Assets/Scripts/Messages/Server/MetaDataLayerMessage.cs
Bod9001: idk how slow is chunk also it's a dictionary versus a list you can't make it a IEnumerable<MetaDataNode>

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaDataLayer.cs
Bod9001: list is just there because you're inserting changes, you never query it

UnityProject/Assets/Scripts/Messages/Server/MetaDataLayerMessage.cs
Bod9001: nah, the plan is just use  bytes

Bod9001:
I'll Merge this tomorrow before build for playtest

UnityProject/Assets/Scripts/Messages/Server/MetaDataLayerMessage.cs
Alainx277: You can get an `IEnumerable` from `dictionary.Values`.
Then both methods will be exactly the same except the `SendTo` and `SendToAll` and it can be factored out.

UnityProject/Assets/Scripts/Messages/Server/MetaDataLayerMessage.cs
Alainx277: Okay, that's irrelevant to the original comment. I guess it's fine to leave as is for now.

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaDataLayer.cs
Alainx277: Looking at the code again I understand it now. `ChangedNodes` contains the changes for the entire lifetime of the map, `nodesToUpdate` only the changes per frame. I think a comment explaining this would be great.

PR #8884, opened 2022-06-22
Updates packages and Unity to 2021.3.3f1 + Adds UniTask package

MaxIsJoe:
Updates packages so newcomers don't get confused by a package error on newer LTS releases of the version we're currently on.
Adds the UniTask package to allow us to use `async` and Tasks finally since the native approach is not fully supported by Unity and cannot grab gameobjects from the main thread (game loop) 

https://github.com/Cysharp/UniTask#getting-started


MaxIsJoe:
Gonna leave this note here :

The reason why on this version specfically of unity we had errors with conflicting packages with the Newsoft Json package was because the visual scripting lang package included the Json package by default inside of it. Apprently a lot of people complained about this because they usually manually install that inside their own projects and when they updated they suddenly were unable to compile their projects because another duplicate of that package had appeared out of nowhere and were unable to remove it so Unity removed it in a later update for the VSL package.

corp-0:
update the test workflow too

MaxIsJoe:
> update the test workflow too

Updated 👍 
tho im not sure why you guys didn't edit that one line of text yourselves while i was away for my exams

corp-0:
I'm also in my exams. This seems to work fine, I'd merge. @Bod9001 do you have any concerns about this one?

corp-0 (approved):

PR #8887, opened 2022-06-26
Fix headless auth

NoooneyDude:
- Fixes #8875 breaking authentication on headless servers (presumably with no `PlayerPrefs` login token set).
- Fixes harmless NRE on startup from network manager not being available yet (dead code, removed).
- Adds an extra UI step about joining a game after logging into your account, instead of hiding the UI.
- Adds a tool from https://forum.unity.com/threads/editor-utility-player-prefs-editor-edit-player-prefs-inside-the-unity-editor.370292/ to manage your `PlayerPrefs` within the editor (I used the delete-all button but my local headless server still managed to authenticate fine 🤔).

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Networking/Authenticator.cs
+++ b/UnityProject/Assets/Scripts/Core/Networking/Authenticator.cs
@@ -43,6 +43,9 @@ namespace Core.Networking
 			public string Message;
 		}
 
+		public struct ServerClientAuthRequestMessage : NetworkMessage { }
+		public struct ServerClientAuthResponseMessage : NetworkMessage { }
+
 		public enum ResponseCode
 		{
 			Success,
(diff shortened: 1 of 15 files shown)


Bod9001 (approved):

UnityProject/Assets/Scripts/Core/Networking/Authenticator.cs
```diff
@@ -61,15 +64,36 @@ public enum ResponseCode
...
+				return;
+			}
+
+			if (conn != NetworkServer.localConnection)
+			{
+				ServerReject(conn); // In the unlikely case of a modified client beating the server's client's request.
+				return;
+			}
+
+			NetworkServer.UnregisterHandler<ServerClientAuthRequestMessage>(); // Should only be one of these.
```
ThatDan123, line 91: This correct? Don't we need the handler for other connections?

ThatDan123: Ah, nvm. Got confused between Request and Response

PR #8897, opened 2022-07-03
A bunch of fixes

Bod9001:
staging fixes

### Changelog:

CL: [Fix] fixed on station nuke Being pullble
CL: [Fix] fixes cases where medicine would apply NaN
CL: [Fix] you will no longer die from one packet of crisps
CL: [Fix] medicine wasn't working correctly due to kidneys filtering out the reagent that is now fixed
CL: [Fix] random item spawners in lockers should not appear any more
CL: [Improvement] grinder has better interactions and has appropriate sprites for state
CL: [Fix] fixes disconnected cable in command in fallstation
CL: [Fix] fixes low oxygen warning not appearing
CL: [Fix] Fixes issues with diagonal movement
CL: [Fix] Fixes hydroponics running on clients
CL: [Fix] fixes fat not filling up

### DIFF
--- a/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
+++ b/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
@@ -63,6 +63,8 @@ public class UniversalObjectPhysics : NetworkBehaviour, IRightClickable
 
 	[PlayModeOnly] public bool CanBeWindPushed = true;
 
+	[PrefabModeOnly] public bool IsPlayer = false;
+
 	public Vector3WithData SetLocalTarget
 	{
 		set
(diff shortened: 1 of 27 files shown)


PerfectTangent:
what was modified in fallstation?

Bod9001:
I added missing cable that was missing in command

UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
```diff
@@ -1269,7 +1271,29 @@ public void SetRegisterTileLocation(Vector3Int localPosition)
...
 	}
 
-	[PlayModeOnly] public bool IsFlyingSliding = false; //Is animating with space flying
+	[PlayModeOnly] private float SecondsFlying;
+
+	public bool IsFlyingSliding
+	{
+		get
+		{
+			return false;
```
NoooneyDude: Just a little confused on why we always return false - what's the point of this then?

Bod9001: oh yeah, forgot to pass it through

PR #8903, opened 2022-07-08
Showing more love for the sickness system

MaxIsJoe:
### Purpose
Removes sickness thread as it's unused 99.99% of the time and now sicknesses get updated once every 10+ seconds.
Removed a lot of unused code and simplifed the system to continue the move towards containing all the sickness logic locally inside sickness classes.
Add sickness immunity for races.
And much more! This PR is mainly groundwork for other stuff coming in the future but there are still some important stuff addressed here which can be seen in the changelog.

### Notes:
- Syringes cannot be properly cleaned right now due to the lack of sinks.
- - Using a syringe on the BioChem Scanner cleans all sicknesses from it.
- Debating if i should nuke `SicknessAffliction` because I honestly see no point in it and would rather keep everything on the sickness class.
- BioChems Scanners are currently not mapped in and are only admin spawned.
- BioChems Scanners are not final and only exist for the chance a sickness event triggers and medbay wants to find a cure quickly before the entire department is full of dead bodies.
- Sicknesses physically spawn and are parented to the player's gameobject
- - No, I do not care about how ugly my solution is for setting consistent cures currently because of this and I will not fix it.

# Changelog:
CL: [Fix] - Fixed sicknesses getting random cures for the same contagious zone spawned. Now they all share the same cure.
CL: [Improvement] - Some species can now be immune to specfic sicknesses.
CL: [Improvement] - Sickness stages have been re-implemented and now are tick dependent.
CL: [Improvement] - Sicknesses that physically harm the player now are much more threatening with each additonal stage.
CL: [Improvement] - Syringes can now spread sicknesses.
CL: [Improvement] - Made sickness event annoucement clearer so people stop getting confused about it being blob.
CL: [Improvement] - All sicknesses have cooldowns by default now. They can be disabled by setting the amount of tickrate needed to progress stages to something below 2 seconds.
CL: [New] - The Paranoia sickness now actually has some functionality.

- Paranoid Players will get random colorful thoughts about everything around them.
- Paranoid Players will constantly lose and regain the ability to recognize faces upong entering the 4th stage of paranoia
- Paranoid Players will start twitching on the 3rd stage.
- Paranoid Players will randomly get defensive and their intent will be set to harm on the 2nd stage.
- GET OUT OF MY HEAD. GET OUT OF MY HEAD. GET OUT OF MY HEAD. GET OUT OF MY HEAD.

CL: [New] - Added the ability for players to lose the ability to recognize faces. Marking all players that talk infront of them as "Unknown".
CL: [New] - Added a new machine for virology, the BioChem Scanner.


PerfectTangent:
wow, paranoia's legit now. Really creative, Max. I like it.

MaxIsJoe:
This should be the last thing
Review pls (but ignore the SpawnResult parts)

Bod9001 (approved):

UnityProject/Assets/Scripts/Health/Sickness/Contagion.cs
```diff
@@ -67,7 +67,18 @@ public virtual bool WillAffectPlayer(PlayerScript playerScript)
 
 		public void OnPlayerStep(PlayerScript playerScript)
 		{
-			playerScript.playerHealth.AddSickness(Sickness);
+			SpawnResult sickNessResult = Spawn.ServerPrefab(transform.GetChild(0).gameObject, Vector3.zero, playerScript.gameObject.transform);
```
Bod9001, line 70: probably best use instantiate since it's not a network identity?

PR #8912, opened 2022-07-12
Adds Emergency shuttle launch vote for command

MaxIsJoe:
CL: [New] - Adds the ability for station command to vote for an emergancy launch for the shuttle.
CL: [Improvement] - Emagged shuttles depart in 5 seconds now instead of 10.

- For alert levels lower than Red, the shuttle requires 4 votes to launch early.
- For alert levels greater or equal to Red, the shuttle requires 2 votes to launch early.

### DIFF
--- a/UnityProject/Assets/Scripts/Items/IDCard.cs
+++ b/UnityProject/Assets/Scripts/Items/IDCard.cs
@@ -225,6 +225,15 @@ public class IDCard : NetworkBehaviour, IServerInventoryMove, IServerSpawn, IInt
 		return accessSyncList.Contains((int) access);
 	}
 
+	public bool HasAccess(List<Access> access)
+	{
+		foreach (var accessToCheck in access)
+		{
+			if (accessSyncList.Contains((int)accessToCheck)) return true;
+		}
+		return false;
+	}
+
 	public string GetJobTitle()
 	{
 		if (jobTitle.IsNullOrEmpty())
--- a/UnityProject/Assets/Scripts/Items/IDCard.cs
+++ b/UnityProject/Assets/Scripts/Items/IDCard.cs
@@ -225,6 +225,11 @@ public class IDCard : NetworkBehaviour, IServerInventoryMove, IServerSpawn, IInt
 		return accessSyncList.Contains((int) access);
 	}
 
+	/// <summary>
+	/// Checks if this id card has the indicated access from a list of accesses.
+	/// </summary>
+	/// <param name="access"></param>
+	/// <returns></returns>
 	public bool HasAccess(List<Access> access)
 	{
 		foreach (var accessToCheck in access)
--- a/UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
+++ b/UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
@@ -119,8 +119,8 @@ namespace Objects
 		{
 			var swpiesRequired = GameManager.Instance.CentComm.CurrentAlertLevel is CentComm.AlertLevel.Red or CentComm.AlertLevel.Delta ? 2 : 4;
 			var remainingSwipes = swpiesRequired - registeredIDs.Count;
-			Chat.AddSystemMsgToChat($"\n\n<color=#FF151F><size={ChatTemplates.LargeText}><b>Escape Shuttle Emergency Launch has been request! need {remainingSwipes} more votes.</b></size></color>\n\n",
-				MatrixManager.MainStationMatrix);
+			string announcemnt = $"\n\n<color=#FF151F><size={ChatTemplates.LargeText}><b>Escape Shuttle Emergency Launch has been request! need {remainingSwipes} more votes.</b></size></color>\n\n"
+			Chat.AddSystemMsgToChat(announcemnt, MatrixManager.MainStationMatrix);
 
 			Chat.AddSystemMsgToChat($"\n\n<color=#FF151F><size={ChatTemplates.LargeText}><b>Escape Shuttle Emergency Launch has been request! need {remainingSwipes} more votes.</b></size></color>\n\n",
 				GameManager.Instance.PrimaryEscapeShuttle.MatrixInfo);
@@ -131,11 +131,11 @@ namespace Objects
 		{
 			if (GameManager.Instance.ShuttleSent) return;
 			var departTime = beenEmagged ? 5 : 10;
-			Chat.AddSystemMsgToChat($"\n\n<color=#FF151F><size={ChatTemplates.LargeText}><b>Escape Shuttle Emergency Launch Triggered! Launching in {departTime} seconds..</b></size></color>\n\n",
-				MatrixManager.MainStationMatrix);
+			string announcement =
+				$"\n\n<color=#FF151F><size={ChatTemplates.LargeText}><b>Escape Shuttle Emergency Launch Triggered! Launching in {departTime} seconds..</b></size></color>\n\n";
+			Chat.AddSystemMsgToChat(announcement, MatrixManager.MainStationMatrix);
 
-			Chat.AddSystemMsgToChat($"\n\n<color=#FF151F><size={ChatTemplates.LargeText}><b>Escape Shuttle Emergency Launch Triggered! Launching in {departTime} seconds..</b></size></color>\n\n",
-				GameManager.Instance.PrimaryEscapeShuttle.MatrixInfo);
+			Chat.AddSystemMsgToChat(announcement, GameManager.Instance.PrimaryEscapeShuttle.MatrixInfo);
 
 			_ = SoundManager.PlayNetworked(CommonSounds.Instance.Notice1);
 			GameManager.Instance.ForceSendEscapeShuttleFromStation(departTime);
--- a/UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
+++ b/UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
@@ -145,8 +145,7 @@ namespace Objects
 		{
 			if (AdminCommandsManager.IsAdmin(PlayerManager.LocalPlayerScript.netIdentity.connectionToServer,
 				    out var _) == false) return null;
-			var result = RightClickableResult.Create().AddElement("Launch Early", DepartShuttle, Color.red);
-			return result;
+			return RightClickableResult.Create().AddElement("Launch Early", DepartShuttle, Color.red);
 		}
 	}
 }
--- a/UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
+++ b/UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
@@ -143,8 +143,7 @@ namespace Objects
 
 		public RightClickableResult GenerateRightClickOptions()
 		{
-			if (AdminCommandsManager.IsAdmin(PlayerManager.LocalPlayerScript.netIdentity.connectionToServer,
-				    out var _) == false) return null;
+			if (CustomNetworkManager.IsServer == false) return null;
 			return RightClickableResult.Create().AddElement("Launch Early", DepartShuttle, Color.red);
 		}
 	}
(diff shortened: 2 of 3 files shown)


AtnerNT:
I feel like required votes should be done as a % of alive players, maybe 40% and 20%, as opposed to a fixed value, otherwise in a match of 30 people, 2 people can leave the rest of the station behind.

MaxIsJoe:
Are there more than 4 command jobs out there?
There are 6 command jobs, only 2-4 are playable currently and some are not even filled most of the time due to low pop. If an emergancy happens we shouldn't wait for multiple people to show up (who might be dead or non-existent) to vote for something as important as this.

There is no reason to make this dynamic or add weird prerequsits for this. 4 for normal, 2 for high alert. Anything else adds more complications for no reason.

AtnerNT:
I didn't see command my bad, I thought this was a genral vote

MaxIsJoe:
Nah its okay, good to explain things still

UnityProject/Assets/Scripts/Items/IDCard.cs
```diff
@@ -225,6 +225,15 @@ public bool HasAccess(Access access)
 		return accessSyncList.Contains((int) access);
 	}
 
+	public bool HasAccess(List<Access> access)
```
Alainx277, line 233: Nitpick, but could the method name more clearly state what this is actually checking? Something like `HasAnyAccess`?

UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
```diff
@@ -21,6 +23,10 @@ public class EscapeShuttleConsole : MonoBehaviour, ICheckedInteractable<HandAppl
 
 		private RegisterTile registerTile;
 
+		[SerializeField] private List<Access> validAccess = new List<Access>();
+
+		private List<IDCard> registeredIDs = new List<IDCard>();
```
Alainx277: If you want unique ID cards this should be a `HashSet<IDCard>`.

UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
```diff
@@ -72,17 +83,70 @@ private void FinishHack(HandApply interaction)
...
+				DepartShuttle();
+				return;
+			}
+
+			AnnounceRemainingSwipesRequired();
+		}
+
+		private void AnnounceRemainingSwipesRequired()
+		{
+			var swpiesRequired = GameManager.Instance.CentComm.CurrentAlertLevel is CentComm.AlertLevel.Red or CentComm.AlertLevel.Delta ? 2 : 4;
```
Alainx277: The same logic is used above, refactor it into a method.

UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
```diff
@@ -72,17 +83,70 @@ private void FinishHack(HandApply interaction)
...
+			var remainingSwipes = swpiesRequired - registeredIDs.Count;
+			Chat.AddSystemMsgToChat($"\n\n<color=#FF151F><size={ChatTemplates.LargeText}><b>Escape Shuttle Emergency Launch has been request! need {remainingSwipes} more votes.</b></size></color>\n\n",
+				MatrixManager.MainStationMatrix);
+
+			Chat.AddSystemMsgToChat($"\n\n<color=#FF151F><size={ChatTemplates.LargeText}><b>Escape Shuttle Emergency Launch has been request! need {remainingSwipes} more votes.</b></size></color>\n\n",
```
Alainx277: The text can be templated once and assigned to a variable.

UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
```diff
@@ -72,17 +83,70 @@ private void FinishHack(HandApply interaction)
...
+			GameManager.Instance.ForceSendEscapeShuttleFromStation(departTime);
+		}
 
-			GameManager.Instance.ForceSendEscapeShuttleFromStation(10);
+		public RightClickableResult GenerateRightClickOptions()
+		{
+			if (AdminCommandsManager.IsAdmin(PlayerManager.LocalPlayerScript.netIdentity.connectionToServer,
+				    out var _) == false) return null;
+			var result = RightClickableResult.Create().AddElement("Launch Early", DepartShuttle, Color.red);
+			return result;
```
Alainx277: `result` can be removed and directly returned.

UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
```diff
@@ -72,17 +83,70 @@ private void FinishHack(HandApply interaction)
...
-			Chat.AddSystemMsgToChat($"\n\n<color=#FF151F><size={ChatTemplates.LargeText}><b>Escape Shuttle Emergency Launch Triggered!</b></size></color>\n\n",
+		private void DepartShuttle()
+		{
+			if (GameManager.Instance.ShuttleSent) return;
+			var departTime = beenEmagged ? 5 : 10;
+			Chat.AddSystemMsgToChat($"\n\n<color=#FF151F><size={ChatTemplates.LargeText}><b>Escape Shuttle Emergency Launch Triggered! Launching in {departTime} seconds..</b></size></color>\n\n",
 				MatrixManager.MainStationMatrix);
 
-			Chat.AddSystemMsgToChat($"\n\n<color=#FF151F><size={ChatTemplates.LargeText}><b>Escape Shuttle Emergency Launch Triggered!</b></size></color>\n\n",
+			Chat.AddSystemMsgToChat($"\n\n<color=#FF151F><size={ChatTemplates.LargeText}><b>Escape Shuttle Emergency Launch Triggered! Launching in {departTime} seconds..</b></size></color>\n\n",
```
Alainx277: The text can be templated once and assigned to a variable.

UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
```diff
@@ -72,17 +83,70 @@ private void FinishHack(HandApply interaction)
...
 			_ = SoundManager.PlayNetworked(CommonSounds.Instance.Notice1);
+			GameManager.Instance.ForceSendEscapeShuttleFromStation(departTime);
+		}
 
-			GameManager.Instance.ForceSendEscapeShuttleFromStation(10);
+		public RightClickableResult GenerateRightClickOptions()
+		{
+			if (AdminCommandsManager.IsAdmin(PlayerManager.LocalPlayerScript.netIdentity.connectionToServer,
+				    out var _) == false) return null;
+			var result = RightClickableResult.Create().AddElement("Launch Early", DepartShuttle, Color.red);
```
Alainx277: Does this work in multiplayer? I think right click options are executed on the client.

Alainx277 (changes requested):

UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
MaxIsJoe: Even if it doesn't it doesn't really matter

UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
MaxIsJoe: ?

UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
MaxIsJoe: No, because each one id different.

UnityProject/Assets/Scripts/Items/IDCard.cs
MaxIsJoe: Why? It's more confusing having differently named functions. Having one function that can have different arguments but is expected to return the same result makes more sense

UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
Alainx277: `$"\n\n<color=#FF151F><size={ChatTemplates.LargeText}><b>Escape Shuttle Emergency Launch has been request! need {remainingSwipes} more votes.</b></size></color>\n\n` is duplicated. It can be assigned to a variable and then used in both calls.

UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
Alainx277: How does it not matter if it doesn't work in the actual game?

UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
Alainx277: The logic that calculates whether 2 or 4 ID cards are required is in two places. It should be defined in one spot.

UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
MaxIsJoe: > How does it not matter if it doesn't work in the actual game?

Because its mainly meant for testing, i don't think admins should be able to push the shuttle early from a console that they probably cant even access from their ghost view. It should be from their tools or a console command.

Alainx277: Huh, I guess in that case it's fine.

UnityProject/Assets/Scripts/Items/IDCard.cs
Alainx277: It's not clear from the signature if it needs to have only one or all accesses in the list. Alternatively a doc comment could explain it.

MaxIsJoe:
review pls

UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
```diff
@@ -72,17 +83,75 @@ private void FinishHack(HandApply interaction)
...
+
+		private void AnnounceRemainingSwipesRequired()
+		{
+			var swpiesRequired = IsHighAlert() ? 2 : 4;
+			var remainingSwipes = swpiesRequired - registeredIDs.Count;
+			string announcemnt =
+				$"\n\n<color=#FF151F><size={ChatTemplates.LargeText}><b>Escape Shuttle Emergency Launch has been request! need {remainingSwipes} more votes.</b></size></color>\n\n";
+			Chat.AddSystemMsgToChat(announcemnt, MatrixManager.MainStationMatrix);
+
+			Chat.AddSystemMsgToChat($"\n\n<color=#FF151F><size={ChatTemplates.LargeText}><b>Escape Shuttle Emergency Launch has been request! need {remainingSwipes} more votes.</b></size></color>\n\n",
```
Alainx277:
```suggestion
			Chat.AddSystemMsgToChat(announcemnt,
```

UnityProject/Assets/Scripts/Shuttles/EscapeShuttleConsole.cs
```diff
@@ -72,17 +83,75 @@ private void FinishHack(HandApply interaction)
...
+			}
+
+			AnnounceRemainingSwipesRequired();
+		}
+
+		private bool IsHighAlert()
+		{
+			return GameManager.Instance.CentComm.CurrentAlertLevel is CentComm.AlertLevel.Red
+				or CentComm.AlertLevel.Delta;
+		}
```
Alainx277: I meant something like this:
```suggestion
		private int RequiredSwipesEarlyLaunch => GameManager.Instance.CentComm.CurrentAlertLevel is CentComm.AlertLevel.Red or CentComm.AlertLevel.Delta ? 2 : 4;
```
Then you can just compare against that in both cases, and you don't configure the `2` and `4` in two places.

Alainx277 (changes requested):
Almost done 😅

MaxIsJoe:
e

Alainx277 (approved):
Nice!

PR #8916, opened 2022-07-14
More fixes

AtnerNT:
### Changelog:

CL: [Fix] Makes Turrets and Gateway unable to be pushed and pulled.
CL: [Fix] Fixes disposalbin construction/deconstruction

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Disposals/DisposalMachine.cs
+++ b/UnityProject/Assets/Scripts/Objects/Disposals/DisposalMachine.cs
@@ -207,7 +207,7 @@ namespace Objects.Disposals
 
 		private void UseWrench()
 		{
-			objectPhysics.SetIsNotPushable(objectPhysics.IsNotPushable);
+			objectPhysics.SetIsNotPushable(!objectPhysics.IsNotPushable);
 
 			if (objectPhysics.IsNotPushable == false)
 			{
--- a/UnityProject/Assets/Prefabs/Objects/Gateway/StationGateway.prefab
+++ b/UnityProject/Assets/Prefabs/Objects/Gateway/StationGateway.prefab
@@ -844,6 +844,7 @@ MonoBehaviour:
     Magic: 0
     Bio: 0
     DismembermentProtectionChance: 0
+    StunImmunity: 0
   Resistances:
     LavaProof: 0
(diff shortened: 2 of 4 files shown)


UnityProject/Assets/Scripts/Objects/Disposals/DisposalMachine.cs
```diff
@@ -207,7 +207,7 @@ protected void TryUseWrench()
 
 		private void UseWrench()
 		{
-			objectPhysics.SetIsNotPushable(objectPhysics.IsNotPushable);
+			objectPhysics.SetIsNotPushable(!objectPhysics.IsNotPushable);
```
corp-0, line 210: this double negation will bite us in the ass eventually

AtnerNT: What do you mean double? its only toggling once

corp-0: read it in English, ``not is not pushable``. Issue isn't your code directly but the bool property

AtnerNT: should I rename the property? will most likely have to change a few scripts but should be fine

corp-0: would have to see all usages and find if it is fine to invert

AtnerNT: Visual studio says 31 references across 21 scripts which would have to be switched around

AtnerNT: also all inspector vlaues on prefabs would have to be inverted

Bod9001 (approved):

UnityProject/Assets/Scripts/Objects/Disposals/DisposalMachine.cs
NoooneyDude: It's really about whether an object can move.

I suggest `IsAnchored` which wouldn't require updating prefabs or if that sounds too machiney, `IsMovable`.

PR #8925, opened 2022-07-15
Sickness improvements and fixes

MaxIsJoe:
More backend stuff, nothing new this time except some fixes

You now have to use reactions for the possible cures list to make it more accurate what players can use to cure sicknesses.

CL: [Fix] - Dead bodies will no longer trigger sickness logic.
CL: [Fix] - Sickness will only check for players when trying to spread it now.
CL: [Fix] - Replaced possible cures for some sicknesses using reagents that can be actually made in-game rather than placeholder ones that can't be obtained.
CL: [Balance] - Sickness cure hints uses reactions now to more correctly get the cure.
CL: [Balance] - The body will now check for the cure when consuming nutriant reagents.
CL: [Balance] - The paranoia sickness now has a cooldown of 25 seconds instead of 10, requires 40 ticks to progress now rather than 50.

### DIFF
--- a/UnityProject/Assets/Scripts/Health/Sickness/Sickness.cs
+++ b/UnityProject/Assets/Scripts/Health/Sickness/Sickness.cs
@@ -4,6 +4,8 @@ using Core.Chat;
 using HealthV2;
 using ScriptableObjects.RP;
 using System.Collections.Generic;
+using System.Linq;
+using Chemistry;
 using UnityEngine;
 using UnityEngine.Serialization;
 using Random = UnityEngine.Random;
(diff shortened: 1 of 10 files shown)


UnityProject/Assets/Scripts/Health/Sickness/Sickness.cs
```diff
@@ -33,16 +35,21 @@ public class Sickness : MonoBehaviour
...
 
 		[SerializeField, Tooltip("basic Symptomp feedback")] protected EmoteSO emoteFeedback;
 
 		[SerializeField, Range(10f,60f)] private float cooldownTime = 10f;
 		public bool IsOnCooldown = false;
 
+		public LayerMask PlayerMask;
+
+		private int cureIndex = 0;
+		public int CureIndex => cureIndex;
```
Alainx277, line 51: Is there a reason to provide the index here instead of the `Reaction` directly?

UnityProject/Assets/Scripts/Objects/Machines/BioChemScanner.cs
```diff
@@ -27,17 +28,29 @@ void Scan()
...
+					if (CustomNetworkManager.IsServer && Application.isEditor)
+					{
+						stringBuilder.AppendLine("--DEBUG--");
+						stringBuilder.AppendLine(sickness.Sickness.CureForSickness.Name);
+						foreach (var key in sickness.Sickness.PossibleCures[sickness.Sickness.CureIndex].ingredients.Keys)
+						{
+							stringBuilder.AppendLine(key.Name);
+						}
+						stringBuilder.AppendLine("--DEBUG OVER--");
+					}
```
Alainx277, line 40: Does this need to be left in?

Alainx277 (changes requested):

UnityProject/Assets/Scripts/Health/Sickness/Sickness.cs
MaxIsJoe: we want to remember which index we chosen from the possible cure list

UnityProject/Assets/Scripts/Objects/Machines/BioChemScanner.cs
MaxIsJoe: yes, its only triggered in the editor

Bod9001 (approved):

PR #8926, opened 2022-07-15
Fixes lava acting as a wall

MaxIsJoe:
CL: [Fix] - Fixes lava tiles acting as walls.


MaxIsJoe:
ready for review

Alainx277 (dismissed):
Looks good on the code side.

Bod9001:
if (registerTile.Matrix.MetaTileMap.ObjectLayer.FloorHazardList == null) return; Will return before  the lava actually gets triggered, need to think of optimal way of doing this, since it's triggered every step

MaxIsJoe:
I don't think there is an "optimal" way at all with how tilemaps are handled currently but at the same time the foreach almost never gets triggered as long as theres nothing to check for.

Bod9001 (approved):

PR #8939, opened 2022-07-17
APC null removal serializing and test description change

Copern:
I meant to PR this awhile ago but got distracted by other things...

Fix the APC test descriptions and add some additional information.

APC's that have nulls removed when a scene loads will serialize the changes. The scene will then save after it has completed loading. However, the scene will only save if the list was changed, if it's in the editor, and the editor is not in play mode. This should prevent any unintentional scene modifications.

#### Notes:
In order to allow it to save before tests run, the connected devices null removal had to be moved from Start to Awake.


corp-0 (approved):
very nice

PR #8953, opened 2022-07-18
Fixes objects not being intractable After tile is blown up underrneath them

Bod9001:
and objects no longer count towards matrices when doing atpoint

### Changelog:

 CL: [Fix] Fixes objects not being intractable After tile is blown up underneath them

### DIFF
--- a/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
+++ b/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
@@ -1036,7 +1036,9 @@ public class UniversalObjectPhysics : NetworkBehaviour, IRightClickable
 			MatrixManager.AtPoint(transform.position, true, registerTile.Matrix.MatrixInfo);
 		if (registerTile.Matrix.Id != newMatrix.Id)
 		{
+			var WorldPOS = transform.position;
 			SetMatrix(newMatrix.Matrix);
+			SetRegisterTileLocation(WorldPOS.ToLocal(newMatrix.Matrix).RoundToInt());
 			ResetLocationOnClients();
 		}
 	}
--- a/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
+++ b/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
@@ -1036,9 +1036,9 @@ public class UniversalObjectPhysics : NetworkBehaviour, IRightClickable
 			MatrixManager.AtPoint(transform.position, true, registerTile.Matrix.MatrixInfo);
 		if (registerTile.Matrix.Id != newMatrix.Id)
 		{
-			var WorldPOS = transform.position;
+			var worldPOS = transform.position;
 			SetMatrix(newMatrix.Matrix);
-			SetRegisterTileLocation(WorldPOS.ToLocal(newMatrix.Matrix).RoundToInt());
+			SetRegisterTileLocation(worldPOS.ToLocal(newMatrix.Matrix).RoundToInt());
 			ResetLocationOnClients();
 		}
 	}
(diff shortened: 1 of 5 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
```diff
@@ -1127,19 +1127,6 @@ public bool IsEmptyAt(Vector3Int position, bool isServer)
...
-					foreach (RegisterTile o in isServer
-						         ? ((ObjectLayer) LayersValues[index]).ServerObjects.Get(position)
-						         : ((ObjectLayer) LayersValues[index]).ClientObjects.Get(position))
-					{
-						if (o.IsPassable(isServer) == false)
-						{
-							return false;
-						}
-					}
-				}
```
ThatDan123, line 1142: Removing this changes the function a lot as objects no longer count?

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterTile.cs
```diff
@@ -297,9 +297,9 @@ public virtual void OnDespawnServer(DespawnInfo info)
 	#endregion
 
 
-	public void ServerSetLocalPosition(Vector3Int value)
+	public void ServerSetLocalPosition(Vector3Int value, bool OverRideCheck = false)
```
corp-0: remember that all variables that are a parameter of a function will be in ``camelCase``

UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
```diff
@@ -1036,7 +1036,9 @@ public void CheckMatrixSwitch()
 			MatrixManager.AtPoint(transform.position, true, registerTile.Matrix.MatrixInfo);
 		if (registerTile.Matrix.Id != newMatrix.Id)
 		{
+			var WorldPOS = transform.position;
```
corp-0: everytime you type a ``var`` the variable is a private local to the context it is being defined in, so you should use ``camelCase``

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaTileMap.cs
Bod9001: yeah, but the only way to test it thoroughly is to merge it so

ThatDan123: Im just concerned about how much code relies on the fact that IsEmptyAt assumes there should be no objects not just no tiles.

ThatDan123: Whats the purpose of removing it?

Bod9001: optimisation plus, when the player gets blown up, it checks for matrix and the player finds itself, and therefore thinks it's on the station matrix still, and then when they Fall over that they become IsPassable,  So therefore the at point matrix calculations think it space

ThatDan123: > optimisation plus, when the player gets blown up, it checks for matrix and the player finds itself, and therefore thinks it's on the station matrix still, and then when they Fall over that they become IsPassable, So therefore the at point matrix calculations think it space

Can you just make another function then which only checks for tiles and then change the matrix check to use that function, as looking through the code some other stuff is using the function to check for objects too.

Bod9001: what's checking for objects?

Bod9001: because I think a matrix Shouldn't be determined by objects

ThatDan123: Was the supermatter and energy ball but looking through what they're doing it seems like it would be fine if objects weren't checked

Could you change the comment above the function then i think it'll be fine to merge

ThatDan123 (approved):

PR #8956, opened 2022-07-19
Adds LazyFindObject helper

Copern:
Pulling some random stuff out into small PRs. Nothing fancy here, just converting the constant Instance null check + FindObjectOfType into a helper method.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/AutoMod/AutoMod.cs
+++ b/UnityProject/Assets/Scripts/Core/AutoMod/AutoMod.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Text.RegularExpressions;
 using Initialisation;
 using UnityEngine;
+using Util;
 //using Telepathy;
 using Debug = UnityEngine.Debug;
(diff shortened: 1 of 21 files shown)


Copern:
No codacy, I will not change it.

Copern:
Out of curiosity, some of these are MonoBehavior managers. Any reason they don't use the SingletonManager?

ThatDan123 (commented):
These should eventually be changed to SingletonManagers<> but there's something weird initialisation issues (being null even though being set in awake in the base class) so they've been kept on the old find object method. Thx anyway for cleaning those up.

However i'm sorta agreeing with the issue codacy, i wouldnt have thought you'd need to use ref for reference types?

Copern:
If you simply pass the object (pass by value), then it would set the local value (if it's null) instead. The object's instance field would never change. We want to set the reference of what we pass in instead.

ThatDan123 (approved):
Yeah, just one of those situations that doesn't happen often then.

PR #8971, opened 2022-07-20
Crawl under players

### LINKED ISSUE #8499: You are unable to crawl past other players.
### Environment
- Game Version: Staging 22032418
- OS Version: Windows 10

Title. Other players still block you even if you are crawling. They won't budge if they're off help intent and if they are on help intent you'll swap places with them as normal. If you are crawling you should be able to crawl onto the same tile as other players and move past them.

### Steps to reproduce

1. Start crawling.
2. Try to move onto a tile another player is occupying 
3. Either get blocked, or swap places with that player.

MaxIsJoe:
### Purpose
Fixes #8499

### Notes:
Cleaned up the `OnBump()` function a little bit to avoid nested if statements.

### Changelog:
CL: [Improvement] - You can now crawl under players without swapping places or getting blocked.

### DIFF
--- a/UnityProject/Assets/Scripts/Player/MovementV2/MovementSynchronisation.cs
+++ b/UnityProject/Assets/Scripts/Player/MovementV2/MovementSynchronisation.cs
@@ -432,10 +432,11 @@ public class MovementSynchronisation : UniversalObjectPhysics, IPlayerControllab
 	{
 		Pushing.Clear();
 		Bumps.Clear();
-		if (intent == Intent.Help && CurrentMovementType == MovementType.Crawling) return;
+		if(intent != Intent.Help) return;
 		if (bumpedBy.TryGetComponent<MovementSynchronisation>(out var move))
 		{
-			if (move.intent != Intent.Help || MatrixManager.IsPassableAtAllMatricesV2(bumpedBy.AssumedWorldPosServer(),
+			if (move.CurrentMovementType == MovementType.Crawling) return;
+			if (MatrixManager.IsPassableAtAllMatricesV2(bumpedBy.AssumedWorldPosServer(),
 				    this.gameObject.AssumedWorldPosServer(), SetMatrixCache, this, Pushing, Bumps) == false) return;
 			var pushVector = (bumpedBy.transform.position - this.transform.position).RoundToInt().To2Int();
 			ForceTilePush(pushVector, Pushing, client, move.TileMoveSpeed);
--- a/UnityProject/Assets/Scripts/Player/MovementV2/MovementSynchronisation.cs
+++ b/UnityProject/Assets/Scripts/Player/MovementV2/MovementSynchronisation.cs
@@ -432,7 +432,7 @@ public class MovementSynchronisation : UniversalObjectPhysics, IPlayerControllab
 	{
 		Pushing.Clear();
 		Bumps.Clear();
-		if (intent != Intent.Help || CurrentMovementType == MovementType.Crawling) return;
+		if (intent == Intent.Help && CurrentMovementType == MovementType.Crawling) return;
 		if (bumpedBy.TryGetComponent<MovementSynchronisation>(out var move))
 		{
 			if (MatrixManager.IsPassableAtAllMatricesV2(bumpedBy.AssumedWorldPosServer(),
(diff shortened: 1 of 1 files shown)


MaxIsJoe:
this isn't a draft, accidently made it a draft

corp-0:
oh, is it ready for review?

MaxIsJoe:
yup

MaxIsJoe:
Wait actually I found a problem while testing this
you can crawl under people if you're on any intent but help for some reason. Any clue why?

AtnerNT:
> Wait actually I found a problem while testing this you can crawl under people if you're on any intent but help for some reason. Any clue why?

if (intent != Intent.Help || CurrentMovementType == MovementType.Crawling) return;

UnityProject/Assets/Scripts/Player/MovementV2/MovementSynchronisation.cs
```diff
@@ -432,26 +432,17 @@ public void OnBump(GameObject bumpedBy, GameObject client)
 	{
 		Pushing.Clear();
 		Bumps.Clear();
-		if (intent == Intent.Help)
+		if (intent != Intent.Help || CurrentMovementType == MovementType.Crawling) return;
+		if (bumpedBy.TryGetComponent<MovementSynchronisation>(out var move))
 		{
-			if (bumpedBy.TryGetComponent<MovementSynchronisation>(out var move))
-			{
-				if (move.intent == Intent.Help)
```
ThatDan123, line 439: is ```intent == Intent.Help``` and ```move.intent == Intent.Help``` the same thing? 

You've removed ```move.intent == Intent.Help```

MaxIsJoe:
Either I can't read or I probably shouldn't rely on auto-refactor in an IDE from now on while working on changes at the same time.

I'd wager on the former.

UnityProject/Assets/Scripts/Player/MovementV2/MovementSynchronisation.cs
MaxIsJoe: Yeah I noticed, thanks for pointing that out ❤️

MaxIsJoe:
ok now i solved everything and works as intended. review pwease.

ThatDan123:
Just concerned that this could let you crawl through objects. But i'm not sure how @Bod9001 has used Ibumpable here.

It might be better to add an exclusion to IsPassableAtAllMatricesV2 and have it ignore players when you are crawling. The old ```MatrixManager.IsPassableAtAllMatrices``` has a parameter for whether to check for players.

Bod9001:
Theoretically you should just say that the player is passable, And it should work fine,  since I removed the thing that pushes players if From something that Is passable

corp-0:
@Bod9001 I'm not entirely sure if you are asking for changes or what. I left this pr to be reviewed by you since I know nothing about the movement code.

MaxIsJoe:
> Just concerned that this could let you crawl through objects. 

Have been testing it, no issues with objects.

> Theoretically you should just say that the player is passable, 

Technically already do because the component we're trying to grab is used only on players (maybe mobs as well?)
It's not grabbing UniversalPhysics but rather grabbing the component the player uses already and is looking at the other player's intent when checking for this.

Objects do not have intent.

Bod9001:
shouldn't register as a bump in the first place?

Bod9001:
if (inPushing.Count > 0 && registerTile.IsPassable(isServer) == false) //Has to push stuff
	
	so, this handles if the object is not passpossible, I don't think it should be extended to bumps though?

MaxIsJoe:
> so, this handles if the object is not passpossible

where even is this line?

also I don't see a point of using the matrix passable stuff because crawling is something that only players can do, putting that sort of logic elsewhere looks like spegettii code and I don't think checking for a player one more time is a good idea because that means we're creating more references for the game to cache to use for only once.
Here it already makes sense because we already got it once and it's in a script that houses all movement interactions for players

Bod9001:
you don't need to check if they are crawling, because they are passable, so you shouldn't need to worry about that, 
but would stuff that is passable trigger bumps on stuff that isn't?

MaxIsJoe:
idk, all i know is that currently passable players trigger `OnBump()` on other players.

Bod9001 (approved):

PR #8973, opened 2022-07-20
Lavaland Room Optimise

ThatDan123:
-Rooms detection now done in a coroutine so can be done over multiple frames, makes the game more responsive

-Fix NRE on spawn for FieldGenerators

-Fixes Lavaland walls not containing gasmix, so mining would create wind (Added an override on the AtmosSystem of that matrix as the issue was the asteroid wall tile is set to have no gas which is good for asteroids but bad for lavaland)


MaxIsJoe:
aww, now miners can mine in peace without worrying about pressure and lack of oxygen when mining too deep.

Bod9001 (approved):

PR #8988, opened 2022-07-22
Should fix item giver tool

AtnerNT:
Works fine locally, and in editor - hopefully it works on headless too.

### Changelog:

CL: [Fix] Fixes Item Giver tool in playermanager


UnityProject/Assets/Scripts/UI/Systems/AdminTools/PlayerManagePage.cs
```diff
@@ -213,18 +213,15 @@ private void SendTeleportAllPlayersToPlayer()
...
 		{
 			adminTools.giveItemPage.selectedPlayer = null;
-			var players = FindObjectsOfType<PlayerScript>(); //since this is client sided it's fiinnnneee
-			foreach (var possiblePlayer in players)
-			{
-				if(possiblePlayer.PlayerInfo.Username != PlayerEntry.PlayerData.accountName) continue;
-				adminTools.giveItemPage.selectedPlayer = possiblePlayer.gameObject;
-			}
+
+			adminTools.giveItemPage.selectedPlayer = PlayerList.Instance.GetPlayerByID(PlayerEntry.PlayerData.uid).GameObject;
```
ThatDan123: The removed comment says this is clientside, so i don't think this will work as Playerlist stuff is serverside

UnityProject/Assets/Scripts/UI/Systems/AdminTools/PlayerManagePage.cs
```diff
@@ -213,18 +213,15 @@ private void SendTeleportAllPlayersToPlayer()
 		public void GiveItemToPlayerButton()
 		{
 			adminTools.giveItemPage.selectedPlayer = null;
```
ThatDan123: Confused as to why you are setting this to null? Isnt this storing the playerdata you want right now?

potatoalienof13:
Are you going to finish this sometime?

AtnerNT:
> Are you going to finish this sometime?

Still am, just be very busy IRL as of late

ThatDan123:
This approach also wont work but it should also not be needed, this is my understanding of what you are needing to do:

1) Select the player, save the players id somewhere so you know who you selected (its the ```PlayerEntry.PlayerData.uid```, this might already be happening?)

2) Open the prayer menu

3) Press the give item button

4) Admin sends command to server to give item to that player (all you need to send is the player id you saved earlier and the item prefab name)

- i'm very confused by this whole finding the player gameobject on the admin client not sure why you'd need to do that?

AtnerNT:
> This approach also wont work but it should also not be needed, this is my understanding of what you are needing to do:
> 
>     1. Select the player, save the players id somewhere so you know who you selected (its the `PlayerEntry.PlayerData.uid`, this might already be happening?)
> 
>     2. Open the prayer menu
> 
>     3. Press the give item button
> 
>     4. Admin sends command to server to give item to that player (all you need to send is the player id you saved earlier and the item prefab name)
> 
> 
>     * i'm very confused by this whole finding the player gameobject on the admin client not sure why you'd need to do that?

The reason we have to find the player and pass it through is because this function isn't actually what gives the player the item, it merely opens the item giver window, but the item giver window needs the player gameObject. The item giver menu requires a player gameobject and is run on client, and the player list only has access to the player uids, so I have to get the player from the uid, and then give that to the item menu.

AtnerNT:
did that make sense?

ThatDan123:
It seems theres a fundamental flaw with the current item giver code anyway:

image

It is also using serverside code clientside. The whole thing needs redoing from when you click the button in AdminGiveItemEntry.cs

ThatDan123:
The GiveItem function needs to be movede into a Command in AdminCommandsManager sending the player ID and the prefab name which can then do the current logic. Then all you need to save is the playerId clientside not the object.

ThatDan123:
Superseded by #9030

PR #8989, opened 2022-07-22
The mining scanner is now affordable

MaxIsJoe:
2000 labor points is too damn high + most miners currently would probably not need to mine much further if they managed to grab enough ores to redeem 2000 points

Edit : It's also not a luxary item, it's a tool that miners are expected to obtain at the start to make their jobs easier.

### Changelog:

CL: [Balance] - Mining scanner price decreased to 500 from 2000


PerfectTangent:
The mining scanner should appear in miner's lockers. It's standard issue equipment for them in /tg/.

ThatDan123:
Could make a basic (in lockers, shorter range) and advanced version (longer range, have to buy)?

NoooneyDude (approved):
Approving as we can still map them - just need to pay for the replacements if you lose your freebies, I suppose.

An upgrade path would be nice to have as ThatDan suggests, too.

PR #8992, opened 2022-07-22
Remove Quickload Auto Join

ThatDan123:
- I always want to just pick where i spawn sorry bod


Bod9001:
why not have separate option for it?

ThatDan123:
You just need to press the job you want, doesnt really seem like we need another toggle for that

Bod9001:
ok

Bod9001 (approved):

PR #8996, opened 2022-07-23
Adds a touch-based medicine

Bod9001:
Adds Styptic Powder  and Silver Sulfadiazine and Synthflesh
interactions into the game

### Notes:
more interactions need to be added such as acid and that type of thing

### Changelog:

CL: [New] Adds touch-based medicine


UnityProject/Assets/Scripts/HealthV2/Living/CirculatorySystem/IAreaReactionBase.cs
```diff
@@ -0,0 +1,8 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public interface IAreaReactionBase
```
Alainx277: Could you add a doc comment here? I can't really tell what this interface represents.

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -315,9 +322,75 @@ private void BodyPartTransfer(Pickupable prevImplant, Pickupable newImplant)
...
+
+				storage.Value.Take(0.2f); //Evaporation
+			}
+		}
+
+		public List<ExternalBodyHealthEffect>
+			ALLExternalMetabolismReactions = new List<ExternalBodyHealthEffect>(); //TOOD Move somewhere static maybe
+
+		public List<MetabolismReaction> metabolismReactions = new List<MetabolismReaction>();
+		public List<MetabolismReaction> MetabolismReactions => metabolismReactions;
```
Alainx277:
```suggestion
		public List<MetabolismReaction> MetabolismReactions { get; } = new();
```

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -315,9 +322,75 @@ private void BodyPartTransfer(Pickupable prevImplant, Pickupable newImplant)
...
 		{
 			CirculatorySystem.OrNull()?.BodyPartListChange();
-			//
+			SurfaceBodyPartChanges();
 		}
 
+		public void SurfaceBodyPartChanges()
+		{
+			PrecalculatedMetabolismReactions.Clear();
+			foreach (var MR in ALLExternalMetabolismReactions)
```
Alainx277: Can you name this variable something more descriptive? Maybe something like `reaction`?

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -315,9 +322,75 @@ private void BodyPartTransfer(Pickupable prevImplant, Pickupable newImplant)
...
+							PrecalculatedMetabolismReactions[MR] = new List<BodyPart>();
+						}
+
+						PrecalculatedMetabolismReactions[MR].Add(bodyPart);
+					}
+				}
+			}
+		}
+
+		public List<BodyPart> UseList = new List<BodyPart>();
```
Alainx277: I don't think this needs to be public?
Could it be named more clearly (ex. `temporaryPartList`)

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -315,9 +322,75 @@ private void BodyPartTransfer(Pickupable prevImplant, Pickupable newImplant)
...
+
+					Reaction.React(UseList, storage.Value, ProcessingAmount);
+				}
+
+				storage.Value.Take(0.2f); //Evaporation
+			}
+		}
+
+		public List<ExternalBodyHealthEffect>
+			ALLExternalMetabolismReactions = new List<ExternalBodyHealthEffect>(); //TOOD Move somewhere static maybe
```
Alainx277:
```suggestion
			AllExternalMetabolismReactions = new List<ExternalBodyHealthEffect>(); //TODO Move somewhere static maybe
```

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -315,9 +322,75 @@ private void BodyPartTransfer(Pickupable prevImplant, Pickupable newImplant)
...
+
+		public void ExternalMetaboliseReactions()
+		{
+			foreach (var  storage in SurfaceReagents)
+			{
+				if (storage.Value.Total == 0) continue;
+
+				metabolismReactions.Clear();
+
+				foreach (var Reaction in PrecalculatedMetabolismReactions)
```
Alainx277, line 364: Shouldn't this check if the reaction matches the body part?

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -315,9 +322,75 @@ private void BodyPartTransfer(Pickupable prevImplant, Pickupable newImplant)
...
+			}
+		}
+
+		public List<ExternalBodyHealthEffect>
+			ALLExternalMetabolismReactions = new List<ExternalBodyHealthEffect>(); //TOOD Move somewhere static maybe
+
+		public List<MetabolismReaction> metabolismReactions = new List<MetabolismReaction>();
+		public List<MetabolismReaction> MetabolismReactions => metabolismReactions;
+
+		public  Dictionary<MetabolismReaction, List<BodyPart>> PrecalculatedMetabolismReactions =
```
Alainx277: Can this be private?

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaDataLayer.cs
```diff
@@ -193,6 +193,16 @@ public void MakeSlipperyAt(Vector3Int position, bool canDryUp = true)
...
 	{
+		var mobs = MatrixManager.GetAt<LivingHealthMasterBase>(worldPosInt, true);
+		if (mobs is List<LivingHealthMasterBase>)
+		{
+			reagents.Divide((mobs as List<LivingHealthMasterBase>).Count + 1);
+			foreach (var mob in mobs)
+			{
+				mob.ApplyReagentsToSurface(reagents, BodyPartType.None);
+			}
+		}
```
Alainx277:
```suggestion
		reagents.Divide(mobs.Count() + 1);
		foreach (var mob in mobs)
		{
			mob.ApplyReagentsToSurface(reagents, BodyPartType.None);
		}
```

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -854,6 +930,38 @@ public void RestartHeart()
 			}
 		}
 
+		[Server]
+		public void ApplyReagentsToSurface(ReagentMix Chemicals, BodyPartType bodyPartAim)
+		{
+
+			foreach (var reaction in ALLExternalMetabolismReactions)
```
Alainx277: This is an `O(n)` operation. I think it's only called rarely, so it's fine. We should just keep it in mind.

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -854,6 +930,38 @@ public void RestartHeart()
...
+								TouchCharacteristics.AttackType,
+								TouchCharacteristics.DamageEffect, bodyPartAim);
+						}
+					}
+				}
+			}
+
+			SurfaceReagents[bodyPartAim].Add(Chemicals);
+
+			if (SurfaceReagents[bodyPartAim].Total  > 5)
```
Alainx277: I think the `5` here should be a class variable or constant, as you're using it twice (and maybe it should be configurable).

Alainx277 (changes requested):
A few architectural aspects seem suboptimal to me, but I'm not familiar enough to work something out. Maybe someone who knows it better can comment on it?

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
Bod9001: damn you with your c#8 fancys

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaDataLayer.cs
Bod9001: the  if (mobs is List<LivingHealthMasterBase>) checks if it's not empty

Alainx277: I get the idea, but you could get an empty list too.
You can just get the count and check that against 0, no need for casting.

Bod9001: don't think you can get count on enumerator though?

Copern: You can. It will call the Enumerable Count extension. It has some built in optimizations to check if the underlying collection is an ICollection and calls the ICollection's count. If it's not an ICollection, then it will enumerate through each item in the list to get the count. You might get a "possible multiple enumeration" warning in your IDE in either case though.

MatrixManager.GetAt returns an IEnumerable but it's actually a List (ICollection) so it should just return the list's count.

Alainx277: > don't think you can get count on enumerator though?

Did you look at the suggested code?

Bod9001: "Did you look at the suggested code?" Just because you written it doesn't mean it compile : P

Bod9001: GIF164
 Ryder complains

Copern: > Ryder complains

See my post above. "You might get a "possible multiple enumeration" warning in your IDE in either case though." 

Use Count, it should avoid actually enumerating twice and won't create garbage.

Alainx277: > "Did you look at the suggested code?" Just because you written it doesn't mean it compile : P

Yeah, but I obviously checked that `Count()` exists.

The warning is not of concern in this context, as there's no expensive operation. You can either suppress the warning here, or return an `ICollection` from `MatrixManager.GetAt`. I'd just suppress it.

Bod9001: possible multiple enumeration? what could that cause?

Alainx277: You can technically have an `IEnumerable` that does a db query or similar. So it would be executed twice.
Shouldn't be relevant here though.

Alainx277 (dismissed):

PR #9000, opened 2022-07-25
[Draft] State machine for NPC AI

MaxIsJoe:
The very first step into modernizing mobs and improving them.

This rework allows us to write more complex and sequenced mob behaviors and supposedly should give us an increase in performance when dealing with large number of mobs.
It also allows us to do parallel states, a feature that has always existed from waaay back in 2004 though our current setup is meant to force the game to mimic SS13's slow action times which makes mobs much more predicatble and less likely to [cause too much damage quickly](https://cdn.discordapp.com/attachments/676153930492805123/957762125139435551/Screenshot_130.png?size=4096)

# Todo :

- [x] Add basic state machine API
- [ ] Reimplament Hostile mobs
- [x] Reimplament mob wandering
- [ ] Add pathfinding/movement base script for mobs (To Reduce code duplicates and have re-usability)

### Overscope / Things to deal with in other PRs:
- Reimplament friendly mob interactions
- Reimplament "Come, Ian~" command and behavior
- Reimplament mob food eating
- Add Goliath Tentacle grab behavior to demonstrate new rework
- Add a "SequenceState" script to make the creation of complex connected behaviors easier.
- Attempt to Update mobs to HealthV2

### DIFF
--- a/UnityProject/Assets/Prefabs/Mobs/_MobBase.prefab
+++ b/UnityProject/Assets/Prefabs/Mobs/_MobBase.prefab
@@ -41,6 +41,7 @@ Transform:
   m_Children:
   - {fileID: 4152194380817414287}
   - {fileID: 2593793382972812258}
+  - {fileID: 1727311274271427194}
   m_Father: {fileID: 0}
   m_RootOrder: 0
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
@@ -222,7 +223,9 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 5a02af87cf44eaf4b81945b406d2757f, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
-  MobObjectives: []
+  RegisterTile: {fileID: 114713267190622314}
+  Active: 0
+  StateMachine: {fileID: 7297751433223728779}
 --- !u!114 &8869584225780784631
 MonoBehaviour:
   m_ObjectHideFlags: 0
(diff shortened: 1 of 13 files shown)


corp-0:
before I read your approach, I have to ask. Why not just use one of the hundreds of free assets we could import to implement behaviour trees instead?

MaxIsJoe:
> before I read your approach, I have to ask. Why not just use one of the hundreds of free assets we could import to implement behaviour trees instead?

From a glance at some of the free ones on the asset store I have a few issues with how they're setup, The top free 2 use a scriptable object approach or force the designer to always have a state in motion which is not something I really like.
This approach is much more cleaner imo and opens better oppritunities to do stuff since we're dealing with full on gameobjects and classes and not scriptable objects stuck in one component.

The way it's designed here is like this :
Mob controller (Base Mob Transform) -> Blackboard and State Machine -> State modules

When the game starts the mob controller tells the state machine to sort all of it's modules from the highest priortity to the lowest and then cycle through them, once a state has been picked the state tells the game that it's active until `OnComplete()` is called which afters that game will repeat the cycle of checking all states and if they're eligiable to be triggered and the cycle goes on.

There is also another type of states and they're independent states, AIs that run alone on the UpdateManager and great for things like shooting while the NPC chases the player. It's not checked through the state machien cycle and instead has it's own update loop that can be programmed to check for other states to work hand in hand with each other.
I don't think other state machines on the asset store can do that because fininate state machines are linear usually.

corp-0:
I think you are using the term behaviour tree and finite state machine interchangeable when they aren't https://www.youtube.com/watch?v=gXrKGTPwfO8

PR #9001, opened 2022-07-25
Fix conflicting types warnings

Copern:
Combined and moved some DeviceLinker and related classes into a shared asmdef.

Move some unitystation independent conversion extensions into shared. Strongly hint to inline some short extensions. Renamed a couple extensions for consistency. Add a very brief summary to each (public) extension.

Some miscellaneous clean up.

### Notes:
There might be some codacy warnings. Some things I simply moved into the shared asmdef so there might be some old stuff in there.

TrySetMaster takes a GameObject (the performer) instead of the interaction. Nothing has made use of the anything else in interactions yet, just the performer. So to make it easy to move out into an asmdef, it now just takes the performer GameObject.


ThatDan123:
Thank you so much, those warnings are really annoying.

UnityProject/Assets/Scripts/Util/ConverterExtensions.cs
```diff
@@ -337,58 +153,30 @@ public static Vector2 GetRandomRotatedVector2(float minimum, float maximum)
...
-		{
-			return OrientationEnum.Up_By0;
-		}
-
-		return OrientationEnum.Down_By180;
-	}
-
-
+			// Based on what was checked and returned before. These are very specific and don't account for anything
+			// greater than one or one direction greater than another for example. Is it supposed to be that way?
```
NoooneyDude: CC @Bod9001 thoughts?

NoooneyDude:
Looks good so far. What's remaining?

Copern:
Just wanted to do some further testing when I get a chance.

UnityProject/Assets/Scripts/Util/ConverterExtensions.cs
Bod9001: yeah,   Because you're only supposed to be putting in Vector2Int direction ,
 1,0 
 0,1
 1,1
and then the minus equivalents

Fr-Dae: for coordinate values it would be better to use ; (semi-comma) 
(comma) it's more for sub-division

like x;y
x;y;z
x,xxx;y,xxx

Fr-Dae:
.

UnityProject/Assets/Scripts/Util/ConverterExtensions.cs
Copern: Okay, I'll add that note in a summary for that method tonight. Is it supposed to return the Down_Left, Down_Right, and so on orientations too? The original only returned the Up, Right, Down, or Left orientations.

Copern:
Should be good to go.

UnityProject/Assets/Scripts/Core/Editor/Tools/Mapping/ObjectNameValidator.cs
```diff
@@ -163,7 +163,7 @@ private void SectionReviewNames()
 				}
 
 				EditorGUILayout.Space(10);
-				if (modifiedObjects.Count > 0 && EditorUIUtils.BigAssButton("Save All"))
+				if (modifiedObjects.Count > 0 && EditorUIUtils.BigButton("Save All"))
```
corp-0, line 166: how could you

Copern: I'm evil and no fun.

ThatDan123 (approved):

PR #9013, opened 2022-07-27
General purpose fixes2

Bod9001:
### Changelog:

CL: [Improvement] walk in disposals now grabs a piece of air when closing

 CL: [Fix] fixes glass duplication glitch with glass table deconstructing while climbing onto it


corp-0:
> walk in disposals now grabs a piece of their when closing

I did not understand

Bod9001:
walk in disposals now grabs a piece of air when closing
*

corp-0:
Shouldn't the IsSealed be set the other way around then? First false, equalize, then true

Bod9001:
the problem is if it's not Sealed that grabs the tile mix as internal mix

PR #9015, opened 2022-07-27
Refactors limb script

### LINKED ISSUE #8987: [Refactor Needed] Seperate arms and legs into seperate limb scripts
Related to #8986

Currently all limb logic is housed in one script that houses both arm and leg variables, we need to split them into a base limb script and a leg and arm scripts that inherit from limb.
This will help us avoid stuff like enum and bool checks and we can just directly ask for the type in code like this `if(limb is arm c)`

This isn't just for cleaning up code, it can allow us to more easily build things like a Lamia tail or an energy powered limb.
This is a relatively easily refactor and can be done quickly but incase I get distracted by other PRs or a squrriel in the coming days I'll leave this issue here for others to acknowladge it's existence.

MaxIsJoe:
fixes #8987 

a pinky promise is a pinky promise :) @corp-0


UnityProject/Assets/Scripts/Items/Implants/Limb.cs
```diff
@@ -7,73 +7,24 @@ namespace HealthV2
 {
 	public class Limb : BodyPartFunctionality, IMovementEffect
```
corp-0: you should be able to remove ``IMovementEffect`` from here and the arm script

MaxIsJoe: actually you need the IMovementEffect on the arm script for crawling

corp-0: oh, crawling uses arms speed? Aight kinda cool.

corp-0: you shuold still remove the interface from the base limb class though

UnityProject/Assets/Scripts/Items/Implants/Limbs/HumanoidLeg.cs
```diff
@@ -0,0 +1,51 @@
+using Player.Movement;
+using UnityEngine;
+
+namespace HealthV2.Limbs
+{
+	public class HumanoidLeg : Limb, IMovementEffect
```
corp-0, line 6: maybe just call this ``Leg`` since the other limb is simply ``Arm``

UnityProject/Assets/Scripts/Items/Implants/Limb.cs
MaxIsJoe: can't remove it because the base function has this functionality
`playerHealth.OrNull()?.PlayerMove.AddModifier(this);`

UnityProject/Assets/Scripts/Items/Implants/Limbs/HumanoidLeg.cs
MaxIsJoe: It makes sense to call it "Humanoid[limb]" because there can be many types of legs in the future.

MaxIsJoe: renamed arm to HumanoidArm

UnityProject/Assets/Scripts/Items/Implants/Limb.cs
corp-0: don't resolve conversations by yourself, please. You have 3 courses of action then:
A - completely remove speed from arms and move this functionality to legs which is the only place it would matter.
B - Remove the interface from the inheritos of limb, they are are IMovementEffect since they inherit from Limb.
C - Remove the interface from Limb and move this method call to both inheritors.

MaxIsJoe: done

corp-0 (approved):

Issue #8166, opened 2022-01-15
[Research Foundation] Ordnance lab objects


CdrCross:
As part of the research foundation, building a basic functionality implementation of the ordnance lab is a first-line task. This issue concerns the introduciton the machines and mechanics a full implementation of Ordnance requires. Proper documentation of the core implementation is expected, as others will build upon it.

- [ ] Tank Transfer Valve, which allows the combination of two gas canisters into a single linked and triggerable object, which releases the content of one tank into the other upon triggering. Consider borrowing the GUI from the C4 for the detonation controls.
- [x] Blast Yield Detector, which detects explosions in a massive cone shaped area in front of it, and calculates the blast yield in front of it. If #8163 is already completed, add additional functionality for calculating research points based off the blast yield. Rather than the old method of sending the recorded data to the radio chat, instead have machine have a screen where it logs blast yields and details.

pyrosshade:
I will begin working on this

PR #9027, opened 2022-08-01
Boom goes research

pyrosshade:
### Purpose
Adds the Blast Yield Detector outlined in #8166. 
Minor reorganization of MachineParts SOs into subfolders.
Added a machine parts SO and Circuit board to research server construction.
Removed a stray pixel in the Doppler sprite being used for the BYD.
Fixed HyperwaveFilter not having It's ItemTrait.

### Notes:
For Players: 
- The Blast Yield Detector, or BYD, detects certain explosions within a certain range or direction. The BYD can have It's detection Direction changed via a wrench. It's default range is 150 tiles. 
- There are two randomized targets that blasts must have a certain strength tested against to award points. One that has a more forgiving range of yields that can award _some_ points, and another that has a narrow range of yields but is your only chance at gaining the max capacity of points from the Ordnance lab. 
- Currently only a few explosives are actually detectable, Including C4, X4 and the Syndicate Macrobomb. 
- The BYD could be built, but getting the parts and circuitboard might be difficult, or not possible without admin intervention until later updates. Actual points awarded end up being any points over the current maximum, if any. 
- Currently a maximum of 140 points can be awarded from the BYD, if nothing is altered from map makers or VV. 
- The Detector must have a connected server that can receive points, which if not already connected, can be connected to an existing server via a multitool. The connection can be checked for in examine text or noting the graph bounds labels. 
- When the BYD is powered (currently self-powered until mappers add this in), and has a connection to a research server, the BYD will be on standby to check for explosions. When explosions are detected, they will be recorded on the graph that can be viewed by interacting with the server. 
- One can scroll through the graph data via the bottom right buttons, or by clicking on visible nodes. This will update the display on the right with the power of the explosion recorded and the expected total points that would be added to a fresh tech web living on the connected research server. 
- Data recorded can be lost if disconnected from a server, wiping clean the graph being built. Points are not lost unless that server that was previously connected gets destroyed or tampered with. This means you should probably play nice with your engineers. At least for a little bit. 
- The random yield targets differ from server to server however. So even if you nail the target for one server, you'll need a different size bomb for another, depending on the play style you're aiming for.

For Map Makers: Blast Yield Detector has a 90 degree cone of detection, this is configurable for direction and range. 
It might be considered in the future that range would be standardized and changeable by parts, but that is outside of the scope of this pr and is not super important for now. Currently the base prefab for the BYD is self-powered. For whomever adds this to the mainstations, please make sure to uncheck this. 

For Contributors: The BYD's cone detects explosions that Use ExplosiveBase as a class, and as such several explosives are not detected and should stand to be updated and adapted to use ExplosiveBase.
The ExplosiveStrength is then compared with target yields to determine research points which theoretically would be awarded based on proximity to the target yield.
The ResearchServer needs to be fleshed out some more with it's UI, and some technologies to shoot for in the tech web to really make this machine truly matter.

For Future Science Lab Contributors: For machines that will award research points, the BYD inherits from a base class of ResearchPointMachine. This class contains logic needed for machines to be powered, and to rely on a research server connection. Any machine inheriting and not overriding relevant methods will have the machine lose connection to the research server on power loss. As research points, it comes with three methods for adding research points to the tech web of a connected research server. These are:
- `AwardResearchPoints(ResearchPointMachine source,int points)` For adding points based on difference beyond the tracked total of a given source. This is to differentiate and limit how many points a lab produces.
- `AddResearchPoints(int points)` For freeform, wild, potentially limitless untracked adding of points, depending on context.
- `AddResearchPoints(ResearchPointMachine source, int points)` For adding a flat amount of points on top of a tracked total, no stipulations. What's done with this total outside of the Award method is beyond me, but it's there for future use.

For admemes: As stated previously; the ResearchServer needs to be fleshed out some more with it's UI, and some technologies to shoot for in the tech web to really make this machine truly matter. It is fairly limited in gameplay and would need some fairly creative and perhaps even sadistic consideration to make this a worthwhile element of round shenanigans in it's current state. It works, it just needs other elements to be interesting. That being said when there becomes some compelling reason to use the machine, there are some VV elements on the research server and the BYD itself that can be toyed with. Proceed with consideration.

For everyone else: Cheers.

### Changelog:

CL: [New] Adds Blast Yield Detector
CL: [New] Adds Blast Yield Detector circuitboard
CL: [New] Adds ResearchServer circuitboard


MaxIsJoe:
fantastic work

wraith-54321:
nice

PerfectTangent:
bro, _legendary._ Can you share some photos, please?

pyrosshade:
`An Unconnected Blast Yield Detector facing station east`
image

`That same Blast Yield Detector facing station south, now connected`
image

`Some trial and error required to reach the peak of science`
image

`This is truly some nice science`
image

Courtesy of ~~Cuban Pete~~ Roman Langston

pyrosshade:
Oh, and before i forget, @CdrCross

UnityProject/Assets/Scripts/Objects/Research/BlastYieldDetector.cs
```diff
@@ -0,0 +1,325 @@
...
+			if (distance > range) return;
+
+			//Math to check for if our explosion falls within a certain angle away from the center of the cone
+			Vector2 coneToQuery = pos.To2Int() - thisMachine;
+			coneToQuery.Normalize();
+
+			Vector2 coneCenterVector = coneDirection.CurrentDirection.ToLocalVector2Int();
+			coneCenterVector.Normalize();
+
+			float angle = Math.Abs((Mathf.Acos(Vector2.Dot(coneToQuery, coneCenterVector)) * 180) / Mathf.PI);
```
Alainx277: Could `Vector2.Angle` be used here?

UnityProject/Assets/Scripts/Objects/Research/BlastYieldDetector.cs
```diff
@@ -0,0 +1,325 @@
...
+				return interaction.Intent != Intent.Harm;
+			}
+			return false;
+		}
+
+		public void ServerPerformInteraction(HandApply interaction)
+		{
+			if (stateSync == BlastYieldDetectorState.Connected)
+			{
+				if (Validations.HasUsedItemTrait(interaction, CommonTraits.Instance.Wrench))
```
Alainx277: Can you invert these two conditions to reduce nesting?

UnityProject/Assets/Scripts/Objects/Research/ResearchPointMachine.cs
```diff
@@ -0,0 +1,120 @@
+using Mirror;
+using Systems.Electricity;
+using Shared.Systems.ObjectConnection;
+using UnityEngine;
+
+namespace Systems.Research.Objects
+{
+	public class ResearchPointMachine : NetworkBehaviour, IAPCPowerable, IMultitoolSlaveable, IExaminable
+	{
+		public ResearchServer researchServer;
```
Alainx277, line 10: Can this only have `protected` visibility?

UnityProject/Assets/Scripts/Systems/Research/Objects/ResearchServer.cs
```diff
@@ -1,20 +1,18 @@
...
 using Systems.Research.Data;
 using Shared.Systems.ObjectConnection;
+using Random = UnityEngine.Random;
 
 namespace Systems.Research.Objects
 {
 	public class ResearchServer : NetworkBehaviour, IMultitoolMasterable
 	{
-		[SerializeField] private int researchPointsTrickl = 25;
+		[SerializeField] private int researchPointsTrickle = 25;
```
Alainx277, line 15: Does this not need a `FormerlySerializedAs` attribute?

Alainx277 (commented):
Great code quality, just a few nits.

UnityProject/Assets/Scripts/Systems/Research/Objects/ResearchServer.cs
pyrosshade: The value was never changed in inspector, so there's no need to keep the old reference.

UnityProject/Assets/Scripts/Objects/Research/ResearchPointMachine.cs
pyrosshade: I end up getting a few things complaining at me when I change it to protected. The research point methods used on `BlastYieldDetector.cs`, `ResearchServer.cs`, and `ResearchPointMachine.cs` could be rewritten in regards to this, however, `BlastYieldDetector.cs` needs to stay public for `GUI_BlastYieldDetector.cs` and I get a complaint about inconsistent accessibility from the base class of `BlastYieldDetector.cs` being less accessible.

UnityProject/Assets/Scripts/Objects/Research/BlastYieldDetector.cs
pyrosshade: done

UnityProject/Assets/Scripts/Objects/Research/BlastYieldDetector.cs
pyrosshade: done

UnityProject/Assets/Scripts/Objects/Research/ResearchPointMachine.cs
pyrosshade: Unless you are refering to `public ResearchServer researchServer;`
That can be done simple enough

Alainx277 (dismissed):

UnityProject/Assets/Scripts/Objects/Research/ResearchPointMachine.cs
pyrosshade: I stand corrected, `GUI_BlastYieldDetector.cs` needs reference to the research server, and values from it, from `BlastYieldDetector.cs`.

Bod9001 (approved):

PR #8376, opened 2022-02-16
[DNM] Access remote(#5819)

vlandemart:
### Purpose
Adds Access Remote, that allows to remotely control airlocks - open, bolt or set them to emergency access mode. 
Closes last task in #5819

https://user-images.githubusercontent.com/19672958/154288954-0a0ca49a-c2bc-4ffc-867b-e7bd587c928d.mp4

### Notes:
1. It doesn't check access as for now, this will be fixed in future commit
2. I haven't added them to any scenes. This either needs to be done in separate PR or I can add access remotes to default loadout of department heads

PR #9035, opened 2022-08-04
Finishes Access Remote PR #8376

MaxIsJoe:
Since #8376 is abandoned, I've decided to take it and finish it myself.
Part of #5819

### Changelog:
CL: [New] - Adds in Access Remotes for head of departments. Item can be found in their lockers.


MaxIsJoe:
whoops, forgot to update the custom network manager
will do that when i open unity again

UnityProject/Assets/Scripts/Items/Devices/AccessRemote.cs
```diff
@@ -0,0 +1,135 @@
...
+					break;
+				default:
+					TryOpenDoor(doorController, interaction.Performer);
+					break;
+			}
+		}
+
+		private void SyncCurrentRemoteState(AccessRemoteState oldState, AccessRemoteState newState)
+		{
+			//(Max) - Why do we need to keep track of the old state?
```
ThatDan123: The oldState parameter is necessary for Mirror as the function is used as a hook

UnityProject/Assets/Scripts/Items/Devices/AccessRemote.cs
```diff
@@ -0,0 +1,135 @@
...
+using Mirror;
+using UnityEngine;
+
+
+namespace Items.Devices
+{
+	/// <summary>
+	/// An item that controls department doors remotely.
+	/// </summary>
+	public class AccessRemote : NetworkBehaviour, IInteractable<HandActivate>, ICheckedInteractable<HandApply>
```
ThatDan123: ICheckedInteractable with the handActivate too so it is validated

UnityProject/Assets/Scripts/Items/Devices/AccessRemote.cs
```diff
@@ -0,0 +1,135 @@
...
+			Chat.AddExamineMsg(interaction.Performer, $"Remote mode is set to: {currentState.ToString()}.");
+		}
+
+		public bool WillInteract(HandApply interaction, NetworkSide side)
+		{
+			if (interaction.IsHighlight || interaction.IsAltClick)
+				return false;
+			if (Validations.HasComponent<DoorMasterController>(interaction.TargetObject) == false)
+				return false;
+
```
ThatDan123, line 77: Need this as well, i think theres a range parameter or else it might be on Validations.CanApply():

UnityProject/Assets/Scripts/Items/Devices/AccessRemote.cs
MaxIsJoe: yes but why do we need it? `[SyncVar]` is good enough alone

UnityProject/Assets/Scripts/Items/Devices/AccessRemote.cs
MaxIsJoe: why do you need to validate handactive? We don't need to check for where it's position in the world or anything like that

ThatDan123: we check to make sure its in their hand

UnityProject/Assets/Scripts/Items/Devices/AccessRemote.cs
ThatDan123: true, so can be removed

UnityProject/Assets/Scripts/Items/Devices/AccessRemote.cs
MaxIsJoe: Remotes are meant to be used from a distance so I'm not adding a distance check.

ThatDan123: use Validations.CanApply() then as that has a range parameter

MaxIsJoe: Done

UnityProject/Assets/Scripts/Objects/Doors/Modules/AccessModule.cs
```diff
@@ -89,10 +90,37 @@ private bool ProcessCheckAccess(GameObject player)
...
+
+			//If the door is in low voltage, there's a very low chance the access check fails and opens anyway.
+			//Meant to represent the kind of weird flux state bits are when in low voltage systems.
+			if (master.Apc.State == PowerState.LowVoltage)
+			{
+				if (Random.value < lowVoltageOpenChance)
+				{
+					return true;
+				}
+			}
```
NoooneyDude: Cool idea, but as it is I think it would easily be interpreted as a bug by players. An accommodating chat message might work.

MaxIsJoe: not really my idea, it's been here forever.
adding the chat message tho, do you know /tg/ has one that we can copy?

NoooneyDude: Not that I'm aware of. Hmmmm, "The airlock's control panel flickers for a moment..." 
Just something so that players can correlate the unexpected behaviour to something.

MaxIsJoe: The sad thing about this is that the station rarely ever goes under powered because our generators are too strong, and department batteries keep most of these things well powered for a long while. So the likelyhood of someone encountering this is very.. veeeeeeeeeery low.

Also done 👍

NoooneyDude: Yeah. Electricity isn't quite right at the moment. Microwaves are overvolting-capable but we just can't do that yet, unfortunately : P
I want to hear the microwave pitch scream!

UnityProject/Assets/Scripts/Items/Devices/AccessRemote.cs
```diff
@@ -0,0 +1,134 @@
...
+		}
+
+		public bool WillInteract(HandApply interaction, NetworkSide side)
+		{
+			if (side == NetworkSide.Client)
+			{
+				if (interaction.IsHighlight || interaction.IsAltClick) return false;
+			}
+
+			return Validations.HasComponent<DoorMasterController>(interaction.TargetObject) && Validations.CanApply(interaction.PerformerPlayerScript, interaction.TargetObject, side, false, ReachRange.Unlimited);;
```
ThatDan123:
```suggestion
			return Validations.HasComponent<DoorMasterController>(interaction.TargetObject) && Validations.CanApply(interaction.PerformerPlayerScript, interaction.TargetObject, side, false, ReachRange.Unlimited);
```

ThatDan123:
Could you just do the codacy default case issue, don't worry about the spritehandler one. Then this is good to go i think.

MaxIsJoe:
i forgot to say but done.
merge pls

UnityProject/Assets/Scripts/Items/Devices/AccessRemote.cs
```diff
@@ -0,0 +1,137 @@
...
+
+
+namespace Items.Devices
+{
+	/// <summary>
+	/// An item that controls department doors remotely.
+	/// </summary>
+	public class AccessRemote : NetworkBehaviour, ICheckedInteractable<HandActivate>, ICheckedInteractable<HandApply>
+	{
+		[SyncVar] private AccessRemoteState currentState;
```
ThatDan123:
```suggestion
		private AccessRemoteState currentState;
```

This doesnt need to be a sync var as the client isnt using it anywhere.

MaxIsJoe:
done

ThatDan123 (approved):

PR #9043, opened 2022-08-04
Storage CoolDown

ThatDan123:
-Adds cooldown to storage to prevent chat spam and server lag


UnityProject/Assets/Scripts/Clothing/BackPack/InteractableStorage.cs
```diff
@@ -175,6 +182,8 @@ public bool WillInteract(InventoryApply interaction, NetworkSide side)
 		if (!Validations.CanPutItemToStorage(interaction.Performer.GetComponent<PlayerScript>(),
 			itemStorage, interaction.UsedObject, side, examineRecipient: interaction.Performer)) return false;
 
+		if (Cooldowns.TryStart(interaction, cooldown, side) == false) return false;
```
Bod9001: Will cause issues with highlight system, since highlight system calls will interact

ThatDan123: fixed, InventoryApply is fine as highlight doesnt use that. I've checked for highlight in the normal HandApply.

Bod9001 (approved):

PR #9047, opened 2022-08-05
Players crawl now when slipped

### LINKED ISSUE #4665: You should be able to crawl around after slipping over.
## Description
Right now if you slip for whatever reason be it a banana, lube, atmospherics, coming out of a disposals outlet, or after extinguishing yourself by rolling, you will be laying on the ground unable to move. This is quite unfun due to the length of time you have to wait to get back up (especially after stop, dropping, and rolling) and potentially dangerous (you can end up on a conveyor belt that leads directly into outer space).

In /tg/, if you fall over due to the above reasons (atmos not included) you'll still be able to crawl around till you automatically (or manually) get back up. This lets you get off conveyor belts, and lets you reposition should you slip for whatever reason. Another thing to note is that you can actually crawl in Unitystation at the moment, but only while in (soft) crit for some reason. It's possible to slip over and rendered unable to move until you get put into crit and suddenly become able to crawl.

MaxIsJoe:
closes #4665

## Changelog:

CL: [New] - Players will now crawl when slipped and are no longer stunned.

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterPlayer.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterPlayer.cs
@@ -291,11 +291,32 @@ public class RegisterPlayer : RegisterTile, IServerSpawn, RegisterPlayer.IContro
 			return;
 		}
 
-		ServerStun(StopMovement : false);
+		ServerSlip();
 		AudioSourceParameters audioSourceParameters = new AudioSourceParameters(pitch: Random.Range(0.9f, 1.1f));
 		SoundManager.PlayNetworkedAtPos(CommonSounds.Instance.Slip, WorldPositionServer, audioSourceParameters, sourceObj: gameObject);
 		// Let go of pulled items.
 		playerScript.objectPhysics.StopPulling(false);
+	}
+
+	private void ServerSlip()
+	{
+		ServerCheckStandingChange(true);
+		OnSlipChangeServer.Invoke(IsSlippingServer, true);
+		ServerDropItemsInHand();
+	}
+
+	[Server]
+	public void ServerDropItemsInHand()
+	{
+		foreach (var itemSlot in playerScript.DynamicItemStorage.GetNamedItemSlots(NamedSlot.leftHand))
+		{
+			Inventory.ServerDrop(itemSlot);
+		}
+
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterPlayer.cs
```diff
@@ -291,13 +291,34 @@ public void ServerSlip(bool slipWhileWalking = false)
...
+		foreach (var itemSlot in playerScript.DynamicItemStorage.GetNamedItemSlots(NamedSlot.leftHand))
+		{
+			Inventory.ServerDrop(itemSlot);
+		}
+
+		foreach (var itemSlot in playerScript.DynamicItemStorage.GetNamedItemSlots(NamedSlot.rightHand))
+		{
+			Inventory.ServerDrop(itemSlot);
+		}
+	}
```
NoooneyDude: Would be useful to move to `DynamicItemStorage` if it doesn't already have one.

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterPlayer.cs
```diff
@@ -291,13 +291,34 @@ public void ServerSlip(bool slipWhileWalking = false)
...
 		// Let go of pulled items.
 		playerScript.objectPhysics.StopPulling(false);
 	}
 
+	private void ServerSlip()
+	{
+		ServerCheckStandingChange(true);
+		OnSlipChangeServer.Invoke(IsSlippingServer, true);
+		ServerDropItemsInHand();
+	}
```
NoooneyDude: Wouldn't you need line `350`, `IsSlippingServer = true` here too?

MaxIsJoe: slip and stun are the same thing because this code needs to be reworked
we need to know whats the state of `IsSlippingServer` in `ServerSlip()` so if the player was already slipping or slipped the game knows what to do or not do with that infromation.

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterPlayer.cs
MaxIsJoe: Done

corp-0 (approved):

PR #9069, opened 2022-08-07
Staging fixes6

Bod9001:
and made the spark light affect better
### Changelog:

CL: [Fix] Shuttle console now updates position properly
CL: [Fix] foam doesn't go into space matrix randomly


UnityProject/Assets/Scripts/Objects/Consoles/ShuttleConsole.cs
```diff
@@ -16,7 +16,7 @@ namespace Objects.Shuttles
 	public class ShuttleConsole : NetworkBehaviour, ICheckedInteractable<HandApply>, IServerSpawn
 	{
 		public MatrixMove ShuttleMatrixMove;
-		private RegisterTile registerTile;
+		public RegisterTile registerTile;
```
ThatDan123:
```suggestion
		[NonSerialized]
		public RegisterTile registerTile;
```

ThatDan123: or make it a property with private setter

ThatDan123:
Can you add a test for the offset?

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/ReactionManager.cs
```diff
@@ -186,10 +186,17 @@ private void ProcessWindNodes(MetaDataNode windyNode)
...
+					pushable.NewtonianPush(registerTile.transform.rotation * (Vector2)windyNode.WindDirection, Random.Range((float)(correctedForce * 0.8), correctedForce),  spinFactor: Random.Range(1, 150));
+				}
+
+
 				if (pushable.stickyMovement && windyNode.WindForce > 3)
 				{
-					if (pushable is MovementSynchronisation && windyNode.WindForce < (int)WindStrength.STRONG) continue;
+					var Playermove = pushable as MovementSynchronisation;
+					if ( windyNode.WindForce < (int)WindStrength.STRONG && Playermove == null) continue;
+					if (Playermove.CanBeWindPushed == false) continue;
```
ThatDan123:
```suggestion
					var playerMove = pushable as MovementSynchronisation;
					if ( windyNode.WindForce < (int)WindStrength.STRONG && playerMove == null) continue;
					if (playerMove.CanBeWindPushed == false) continue;
```

ThatDan123: And the ```&& playerMove == null``` is that in the right place? or is it meant to be ```||```

Bod9001: null check is more expensive than magnitude

ThatDan123: yeah, not asking about sequence. You are checking playerMove after that If and you are not returning if it is null before. Is that supposed to be an ```||```?

Bod9001: well the as operation just casts, and then you have to check it to make sure it is not null

PR #9073, opened 2022-08-08
Move SingletonManager to shared assembly

Copern:
### Purpose
Move SingletonManager to the shared assembly as it doesn't rely on anything UnityStation specific. 
Clean up some imports.
Clean up a test.


corp-0 (approved):

corp-0:
Brilliant

PR #9097, opened 2022-08-08
Prevents objects  being pushed by wind and stuff when being draagged

Bod9001:


ThatDan123 (dismissed):

ThatDan123:
This would also prevent someone from walking through something being pulled by someone else?

Bod9001:
"This would also prevent someone from walking through something being pulled by someone else?"
That would be a normal tile push, This is only Newtonian push E,g  stuff bouncing off it and applying momentum And wind

ThatDan123 (dismissed):

Bod9001:
why tests why

Bod9001:
should be fixed?

corp-0:
rebase and push to pass the tests

corp-0 (approved):

PR #9103, opened 2022-08-10
Blob Hit ClientSide Only

ThatDan123:
-Doesnt need to be a networked object


NoooneyDude (dismissed):

UnityProject/Assets/Scripts/Items/Traits/CommonTraits.cs
```diff
@@ -99,5 +99,8 @@ public class CommonTraits : SingletonScriptableObject<CommonTraits>
 	public ItemTrait PowerControlBoard;
 	public ItemTrait Beaker;
 
+	/// <summary>
+	/// Do not use this list to get references to traits, locally reference them in your scripts instead!
+	/// </summary>
```
corp-0, line 105: make it ``[Obsolete]``

corp-0 (approved):

PR #9105, opened 2022-08-10
Minor Electrical Optimisations

ThatDan123:
-Removed unnecessary dictionary

-Removed resistance property (properties do have a performance difference from just using a public variable)

Would be nice in the future to rewrite the whole system to be burst compatible

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Engineering/APCPoweredDevice.cs
+++ b/UnityProject/Assets/Scripts/Objects/Engineering/APCPoweredDevice.cs
@@ -47,11 +47,13 @@ namespace Systems.Electricity
 		[Tooltip("Watts consumed per update when running at 240v")]
 		private float wattusage = 0.01f;
 
-		public float Wattusage {
+		public float Wattusage
+		{
 			get => wattusage;
-			set {
+			set
+			{
 				wattusage = value;
-				resistance = 240 / (value / 240);
+				Resistance = 240 / (value / 240);
 			}
 		}
 
@@ -60,14 +62,8 @@ namespace Systems.Electricity
 		[FormerlySerializedAs("resistance")]
 		private float InitialResistance = 99999999;
 
-
-		public float resistance = 99999999;
-
-
-		// public float Resistance {
-		// 	get => resistance;
-		// 	set => resistance = value;
-		// }
+		[NonSerialized]
+		public float Resistance = 99999999;
 
 		[HideInInspector] public APC RelatedAPC;
 		private IAPCPowerable Powered;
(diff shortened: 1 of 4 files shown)


NoooneyDude (dismissed):

NoooneyDude:
> -Removed resistance property (properties do have a performance difference from just using a public variable)

Don't suppose you could share a link where I can learn more about this? Just curious and little surprised, surely the compiler w/ optimisations enabled would handle it.

Edit: just read #general-dev. Right, we're running dev mode on the server. All good 👍

UnityProject/Assets/Scripts/Objects/Engineering/APCPoweredDevice.cs
```diff
@@ -60,14 +62,8 @@ public class APCPoweredDevice : NetworkBehaviour, IServerDespawn, IEmpAble, IMul
 		[FormerlySerializedAs("resistance")]
 		private float InitialResistance = 99999999;
 
-
-		private float resistance = 99999999;
-
-
-		public float Resistance {
```
Bod9001, line 67: I think this was meant to trigger resistance change?

UnityProject/Assets/Scripts/Systems/Electricity/FunctionsAndClasses/ElectricityFunctions.cs
```diff
@@ -185,76 +185,49 @@ public static float WorkOutResistance(Dictionary<IntrinsicElectronicData, VIRRes
...
-							AnInterestingDictionary[CurrentItem.Key] = (float) -CurrentItem.Value.Current();
-						}
-					}
+					current += (float) currentItem.Value.Current();
 				}
 
-				foreach (var CurrentItem in AnInterestingDictionary)
+				foreach (var currentItem in currentIDItem.Value.CurrentGoingTo)
 				{
-					if (CurrentItem.Value > 0)
```
Bod9001, line 234: was there any scenario where this was false?  might be worth putting in else and then putting a breakpoint to see if anything triggers it

ThatDan123: Added a check for it

UnityProject/Assets/Scripts/Objects/Engineering/APCPoweredDevice.cs
ThatDan123: Can't find any function in the class for a resistance change?

Bod9001 (approved):

PR #9110, opened 2022-08-11
Sickness stuff idk what to title this

### LINKED ISSUE #5905: Extend the functionality of Sickness system
### Description
While I was trying to make radiation intoxication a sickness I found myself with a couple problems.

- [x] Make Sickness stages able to have more than one symptom per stage and choose a random symptom to perform. This is somewhat already possible with the custom messages symptom, but it would be cool to simply set one of the standard symptoms which also can have more specific behavior, like vomiting.
- [x] Make sickness  able to freeze progression or disable auto progression. Also make sickness manager able to set the current stage of a sickness to a particular stage.
- [x] Make any mob with health able to get sick, instead of just players. We will need to handle exceptions, but we can see those later.

### LINKED ISSUE #9109: Sicknesses get the wrong cure
Currently, sicknesses get assigned the wrong cure when randomizing them so the ingredients and their mixes always produce a reagent that isn't the same as this :
image

MaxIsJoe:
fixes #9109 
closes #5905

nothing major here
just some backend stuff and a fix

Technically, everything in #5905 can be done now with everything that we have now, we don't need new classes and components to get stuff like symptoms going because everything can be `SymptompFeedback()` and `SicknessBehavior()` functions.

## Changelog:

CL: [Fix] - Sickness gets the right cure now.

### DIFF
--- a/UnityProject/Assets/Scripts/Health/Sickness/Sickness.cs
+++ b/UnityProject/Assets/Scripts/Health/Sickness/Sickness.cs
@@ -41,7 +41,7 @@ namespace Health.Sickness
 			{
 				if (value > NumberOfStages)
 				{
-					Logger.LogWarning("[Sickness] - Attempted setting stage for sickness that was bigger than the identified stages.");
+					Logger.LogError("[Sickness] - Attempted setting stage for sickness that was bigger than the identified stages.");
 					return;
 				}
 				currentStage = value;
--- a/UnityProject/Assets/Scripts/Health/Sickness/Sickness.cs
+++ b/UnityProject/Assets/Scripts/Health/Sickness/Sickness.cs
@@ -7,7 +7,6 @@ using System.Collections.Generic;
 using System.Linq;
 using Chemistry;
 using UnityEngine;
-using UnityEngine.ResourceManagement.ResourceProviders.Simulation;
 using UnityEngine.Serialization;
 using Random = UnityEngine.Random;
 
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Health/Sickness/Sickness.cs
```diff
@@ -33,7 +34,20 @@ public class Sickness : MonoBehaviour
...
 		private int currentStage = 1;
-		public int CurrentStage => currentStage;
+		public int CurrentStage
+		{
+			get { return currentStage; }
+			set
+			{
+				if (value > NumberOfStages)
+				{
+					Logger.LogWarning("[Sickness] - Attempted setting stage for sickness that was bigger than the identified stages.");
```
corp-0: Warnings are way too easy to ignore, make it an error instead.

corp-0 (changes requested):

UnityProject/Assets/Scripts/Health/Sickness/Sickness.cs
MaxIsJoe: Done

UnityProject/Assets/Scripts/Health/Sickness/Sickness.cs
```diff
@@ -7,6 +7,7 @@
 using System.Linq;
 using Chemistry;
 using UnityEngine;
+using UnityEngine.ResourceManagement.ResourceProviders.Simulation;
```
corp-0: what's this namespace?

MaxIsJoe: probably some rider accidental autocompletion left over, removing.

corp-0 (approved):
I don't see anything else wrong with this one. I don't know if it solves the problems I faced when I was trying to work with sickness and wrote the issue, but we will see.

PR #9112, opened 2022-08-12
feat: adds Behaviour tree library by TheKiwiCoder

corp-0:
### Purpose
Adds the library without making any other addition of importance.

Use the commits to review the changes.

### Notes:
Added a namespace to our ``Orientation`` enum, to avoid a name conflict. This makes the PR have a lot of changes because ``Orientation`` was used all around the codebase. For the same reason, I ask you guys to prioritize this pr so we can start working with the BT library and avoid unnecessary conflicts.


corp-0:
Why is the new namespace ``Core.UsTransform``? First I didn't want enum to be in the broad scope of ``Core`` and second, I can't use the name of the folder it is sitting at, ``Transform``, because it conflicts with the type ``UnityEngine.Transform``.

corp-0:
codacy is complaining about the library code, should we ignore?

UnityProject/Assets/Scripts/Core/Transform/Orientation.cs
```diff
@@ -1,284 +1,286 @@
...
-using System.Collections.Generic;
 using UnityEngine;
 
-/// <summary>
-/// Represents an absolute facing. Can only be in 4 cardinal directions.
-///
-/// Defined in terms of euler angle (rotation about the z axis, right is 0 and up is 90).
-/// </summary>
-public struct Orientation : IEquatable<Orientation>
+namespace Core.UsTransform
```
NoooneyDude: There's `Core.Directionals` already in use, maybe slap it in there? I dunno, `UsTransform` just seems a little odd to me : ) All good if you think otherwise / can't be bothered though

~~I think even just `Core` would be fine.~~ Just read your comment

corp-0: I will look into it after work. It sounds a lot better tho, at least for this enum in particular. Does it apply for the rest of files in the folder?

NoooneyDude:
~~Don't we normally put third party assets under `Standard Assets` folder (which I think is ignored by Codacy?)~~ lol, nevermind me

UnityProject/Assets/Scripts/Core/Transform/Orientation.cs
NoooneyDude: Nah, not really. Hmm, I think it's fine to deviate from the folder name tbh. Well, up to you.

NoooneyDude: Could do `Core.Transforms`?

corp-0:
We probably ignored it at some point, but codacy has acted weird for us in the past and maybe the setting was lost after a reinstall

UnityProject/Assets/Scripts/Core/Transform/Orientation.cs
corp-0: That might work

corp-0: done

NoooneyDude: Hmm, what's done? : )

corp-0: lmao I never pushed

NoooneyDude (approved):

PR #9126, opened 2022-08-14
Server side UI actions 

Bod9001:
This should hopefully prevent client randomly losing their UI actions when re-logging

 
TODO requires a bit more testing

### Changelog:

CL: [Improvement]  UI should now be a bit more resilient when re-logging


corp-0:
chemistry tests are failing? Also, please try to solve as many codacy issues as you can.

ThatDan123:
From glancing over the code it seems fine and i'll do a proper build test just to check it all works.

ThatDan123:
Respawning yourself as an alien in the editor doesnt give you any actions

ThatDan123:
I know the reason why. It's because the interface where they are set assumes mind is set before the transfer, when actually the mind is set after the transfer to a new body so is null.

ThatDan123:
Also in the Mind script there ResendSpellActions() is that still needed?

ThatDan123 (approved):

PR #9127, opened 2022-08-14
ci: add new workflow to check format on PR's

corp-0:
### Purpose
As suggested by Alainx. The check will fail if the incoming code isn't properly formated so we avoid merging wrongly formatted code.


corp-0:
pr is missing the config file

corp-0:
nevermind, we already had it

ThatDan123 (approved):

PR #9132, opened 2022-08-16
Adds Implanters

AtnerNT:
Adds implanters, implanters can be loaded with any small (or smaller) implantable objects, an implanter can be activated by clicking on it in your hand (or using the activate key) at which point it can be used on a player, by default implanters take 5 seconds and implant the implant into the targeted body part. Implanters have also been added to the syndicate uplink to compliment implant explosives and future syndi implants.

### Changelog:
 CL: [New] Adds Implanters
CL: [Balance] Implant explosives will now blow off legs and arms if placed in them


corp-0:
what objects are implantables?

NoooneyDude:
I imagine we'll want some implants to be installed in specific locations e.g. head or something, perhaps an accompanying component on the implant itself that the implanter can check? Thinking ~~a flagged enum~~ an enum of the target body slot, which overrides the performer's target slot.

AtnerNT:
> what objects are implantables?

rn anything with the implantable trait, though it has to be small or tiny to be used with the implanter

AtnerNT:
> I imagine we'll want some implants to be installed in specific locations e.g. head or something, perhaps an accompanying component on the implant itself that the implanter can check? Thinking ~a flagged enum~ an enum of the target body slot, which overrides the performer's target slot.

Hmmm right now implants dont have their own script, they just use the BodyPartScript, also its an inventory transfer, maybe the better solution is that each bodypart type has its own unique slotCapacity that maybe blacklists traits for implants to other parts in the body so only implants that can be implanted in anything or the head specifically can go in there

corp-0:
conceptually, implanters don't add functionality to the target, right? It is just a fancy way to store items inside the body? Otherwise, it might be better to make those items a ``BodyPartFunctionality``

AtnerNT:
> 

They have BodyPart and BodyPart functionality, as BodyPart allows for surgery related stuff as well as a few other nice things. Implant explosives are the only implants we have rn and they use both BodyPart and a child of BodyPartFunctionality to work.
And to answer your question, yeah they are just a quick and fancy way to add something to the bodies storage

NoooneyDude:
> Hmmm right now implants dont have their own script, they just use the BodyPartScript, also its an inventory transfer, maybe the better solution is that each bodypart type has its own unique slotCapacity that maybe blacklists traits for implants to other parts in the body so only implants that can be implanted in anything or the head specifically can go in there

Seems workable, only concern is it sounds a little harder to maintain which body parts are accepted/rejected by an implant as opposed to the implant itself directly having this information.

AtnerNT:
> Seems workable, only concern is it sounds a little harder to keep track of which body parts allow which implants, as opposed to which body part is accepted by each implant.

Dont think itll be too bad, For example lets take the headCapacity, it will whitelist HeadImplant traits, and black list leg arm and chest implant traits. And if you want an implant to only go into head, you just give it the HeadImplant Trait, if you dont want it to care, you just give it the implantable trait and nothing else

NoooneyDude:
Too many traits : P Up to the implementer in the end, I suppose.

AtnerNT:
> Too many traits : P Up to the implementer in the end, I suppose.

Gilles is here too, Ill tryh get his opinion then jsut go with the majority

corp-0:
if it is just BodyPartTarget what we care about, it would be better using the enum. If you want it to be able to be implanted in more than one possible target, same enum but flags.

Sharkygal42:
Unprimed and Primed injectors with microbomb inside @AtnerNT
[Injector.zip](https://github.com/unitystation/unitystation/files/9351655/Injector.zip)

UnityProject/Assets/Scripts/Items/Medical/Implanter.cs
```diff
@@ -0,0 +1,137 @@
...
+	{
+		var LHB = interaction.TargetObject.GetComponent<LivingHealthMasterBase>();
+
+		foreach (var bodyPart in LHB.SurfaceBodyParts)
+		{
+			bool selected = false;
+
+			if (bodyPart.BodyPartType == interaction.TargetBodyPart)
+			{
+				ItemSlot toSlot = bodyPart.OrganStorage.GetNextFreeIndexedSlot();
```
Bod9001, line 66: the only issue with this is you can technically implant into anything, as long as its surface body part

corp-0: I get that isn't intended. Why not?

Bod9001: being able to implant stuff in people's arms for example and then not being able to remove it

Bod9001: because I don't think there's any open surgery option on stuff like arms

corp-0: so what's your suggestion? To move this pr forward

AtnerNT: haven't been on github in awhile, but as for the concern with legs and arms and such, that will only be the case if an implant is added that can be implanted into arms and legs, now there is a change i need to make because I am now realising i set something up wrong so ill need to do another commit. so yes techincally its possible to implant into arms and not be able to remove it, but thatll oinly happen if someone adds an arm implant without a remove procedure

Bod9001: cool sounds good

UnityProject/Assets/Scripts/Items/Medical/Implanter.cs
```diff
@@ -0,0 +1,137 @@
+using UnityEngine;
+using HealthV2;
+using Items;
+
+public class Implanter : MonoBehaviour, ICheckedInteractable<HandApply>, ICheckedInteractable<InventoryApply>, ICheckedInteractable<HandActivate>
+{
+	[SerializeField]
+	private GameObject implantObject = null;
```
corp-0, line 8: add ``[Tooltip]`` explaining the intended use for this serialized field

UnityProject/Assets/Scripts/Items/Medical/Implanter.cs
```diff
@@ -0,0 +1,137 @@
...
+
+	[SerializeField]
+	private int timeToImplant = 5;
+
+	[SerializeField]
+	private ItemTrait ImplantableTrait = null;
+
+	[SerializeField]
+	private SpriteHandler spriteHandler;
```
corp-0, line 16: is it needed to manually set the spritehandler here? Can't we get it same we are getting ``itemStorage`` in ``Awake``?

UnityProject/Assets/Scripts/Items/Medical/Implanter.cs
```diff
@@ -0,0 +1,137 @@
...
+	{
+		if (implantObject != null)
+		{
+			Inventory.ServerSpawnPrefab(implantObject, itemStorage.GetIndexedItemSlot(0));
+		}
+	}
+
+	public bool WillInteract(HandApply interaction, NetworkSide side)
+	{
+		if (!DefaultWillInteract.Default(interaction, side)) return false;
```
corp-0:
```suggestion
		if (DefaultWillInteract.Default(interaction, side) == false) return false;
```

UnityProject/Assets/Scripts/Items/Medical/Implanter.cs
```diff
@@ -0,0 +1,137 @@
...
+		//can only be applied to LHB
+		if (!Validations.HasComponent<LivingHealthMasterBase>(interaction.TargetObject)) return false;
```
corp-0:
```suggestion
		if (Validations.HasComponent<LivingHealthMasterBase>(interaction.TargetObject) == false) return false;
```

UnityProject/Assets/Scripts/Items/Medical/Implanter.cs
```diff
@@ -0,0 +1,137 @@
...
+
+		if (interaction.Intent != Intent.Help || primed == false) return false;
+
+		if(itemStorage.GetIndexedItemSlot(0).IsEmpty)
```
corp-0: replace the magical number with a constant variable

UnityProject/Assets/Scripts/Items/Medical/Implanter.cs
```diff
@@ -0,0 +1,137 @@
...
+			Chat.AddExamineMsgFromServer(interaction.Performer, "No implant present in implanter!");
+			return false;
+		}
+
+		return true;
+	}
+
+	public void ServerPerformInteraction(HandApply interaction)
+	{
+		var LHB = interaction.TargetObject.GetComponent<LivingHealthMasterBase>();
```
corp-0: even if it is an acronym, respect the pascalCase rule for private variables. I would personally just call it ``lhb`` or ``health``.

UnityProject/Assets/Scripts/Items/Medical/Implanter.cs
```diff
@@ -0,0 +1,137 @@
...
+
+			if (bodyPart.BodyPartType == interaction.TargetBodyPart)
+			{
+				ItemSlot toSlot = bodyPart.OrganStorage.GetNextFreeIndexedSlot();
+				ItemSlot fromSlot = itemStorage.GetIndexedItemSlot(0);
+
+				if (toSlot != null && selected == false)
+				{
+					selected = true;
+
```
corp-0: I don't understand the use for this boolean. Can you explain the intention?

UnityProject/Assets/Scripts/Items/Medical/Implanter.cs
```diff
@@ -0,0 +1,137 @@
...
+	public void ServerPerformInteraction(HandActivate interaction)
+	{
+		if (interaction.IsAltClick)
+		{
+			itemStorage.ServerDropAll();
+		}
+		else
+		{
+			if (itemStorage.GetIndexedItemSlot(0).IsEmpty) Chat.AddExamineMsg(interaction.Performer, "Cannot prime without implant");
```
corp-0: separate the body of the if into its own line and enclose it with curly brackets.

UnityProject/Assets/Scripts/Items/Medical/Implanter.cs
```diff
@@ -0,0 +1,137 @@
...
+			else TogglePrimed();
```
corp-0: same here, don't do the body on the same line as the conditional statement

UnityProject/Assets/Scripts/Items/Weapons/ImplantExplosive.cs
```diff
@@ -27,11 +24,15 @@ protected override void Detonate()
...
 				}
-				else if (bodyPart.ContainedIn.BodyPartType == BodyPartType.Head)
+				else if (bodyPart.ContainedIn.BodyPartType != BodyPartType.Chest)
 				{
-					//Decapitates if explosive in head
+					//Removes limb it is implanted in
 					bodyPart.HealthMaster.DismemberBodyPart(bodyPart.ContainedIn);
 				}
+
+				//This prevents those with bomb proof armour from just tanking an explosion thats supposed to me inside them
```
corp-0:
```suggestion
				//This prevents those with bomb proof armour from just tanking an explosion thats supposed to be inside them
```

UnityProject/Assets/Scripts/Items/Weapons/ImplantExplosive.cs
```diff
@@ -27,11 +24,15 @@ protected override void Detonate()
...
+				bodyPart.HealthMaster.ApplyDamageToBodyPart(gameObject, 200, AttackType.Bomb, DamageType.Burn, bodyPart.ContainedIn.BodyPartType);
```
corp-0: you can use the internal attack type to bypass any armor. It also exists for everything that happens inside the body.

corp-0 (changes requested):

AtnerNT:
Okay applied suggested changes,m fixed bugs and added a feedback message for if an implant fails which i forgot to do before, should all be ready to merge for playtest

Bod9001 (approved):

PR #9135, opened 2022-08-17
[Draft] .3.8f1 LTS upgrade

MaxIsJoe:
Needs more testing before getting ready to update

### DIFF
--- a/UnityProject/Packages/packages-lock.json
+++ b/UnityProject/Packages/packages-lock.json
@@ -106,7 +106,7 @@
       "url": "https://packages.unity.com"
     },
     "com.unity.scriptablebuildpipeline": {
-      "version": "1.19.6",
+      "version": "1.20.1",
       "depth": 1,
       "source": "registry",
       "dependencies": {},
@@ -201,7 +201,7 @@
       "depth": 0,
       "source": "git",
       "dependencies": {},
-      "hash": ""
+      "hash": "ecddb4bbbd6ef23f96a6f1c2f090615f996ab043"
     },
     "com.unity.modules.ai": {
       "version": "1.0.0",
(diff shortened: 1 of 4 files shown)


NoooneyDude:
You'll need to update things like actions and documentation too. grep for the current version should be enough to find 'em all.

corp-0:
test will always fail if the version in the workflow isn't also updated.

MaxIsJoe:
I've been testing this version for a while now, and it seems to not have any issues yet, I'll test out builds made with this version again and if everything goes smoothly I'll update the workflow and make this ready to be merged.

MaxIsJoe:
I have found only one issue this whole time, and it's related to right-clicking. I'm not sure how to solve it but once I figure that out I'll push the fix and mark this as ready for review.

NoooneyDude:
What's the issue? I've been running on .8 for a while now, all good here.

NoooneyDude:
Generally we're more concerned with major/minor updates and not patches.

MaxIsJoe:
> What's the issue? I've been running on .8 for a while now, all good here.

Right click menu doesn't work, you can't click on it or hover over it

Copern:
> > What's the issue? I've been running on .8 for a while now, all good here.
> 
> Right click menu doesn't work, you can't click on it or hover over it

This happened in 3.6 as well. I believe it had to do with Unity changing the input system and how it handles events.

It might be the issue described [here](https://stackoverflow.com/questions/70719928/onmouseenter-doesnt-work-when-hovering-over-an-object-in-unity). You might have to change the input system on the new version.

Edit: Or it might be the issue described [here](https://forum.unity.com/threads/problem-with-ipointerenterhandler-ipointerexithandler-and-setactive.1235596/).

MaxIsJoe:
closes for #9291

PR #9139, opened 2022-08-18
Alien Health Fix

ThatDan123:
-Aliens should have correct health now


Bod9001:
what do you think about having higher armour values on Alien body parts ? So they take less damage?

Bod9001:
I suppose it would be harder to balance

corp-0 (approved):

PR #9142, opened 2022-08-19
Modular MiniGame system (Abandoned Crates)

MaxIsJoe:
## Purpose

Jack has proposed an idea for me that caught me interest and while working on it, I've decided to make a modular mini-game system that can be re-used anywhere. The idea of this is that mini-games should not just be tied to crates and we can add all sorts of mini-games in the future without any restrictions.

## Changelog:

CL: [New] - Added goon-style abandoned crates that can be unlocked by playing minigames.
CL: [New] - Added a "voice-activated guess-the-number" minigame.

### DIFF
--- a/UnityProject/Assets/Scripts/MiniGames/IMiniGame.cs
+++ b/UnityProject/Assets/Scripts/MiniGames/IMiniGame.cs
@@ -1,9 +1,0 @@
-namespace MiniGames
-{
-	public interface IMiniGame
-	{
-		public void StartGame();
-		public void GameWin();
-		public void GameLoss();
-	}
-}
--- a/UnityProject/Assets/Scripts/MiniGames/IMiniGame.cs
+++ b/UnityProject/Assets/Scripts/MiniGames/IMiniGame.cs
@@ -1,6 +1,3 @@
-using System.Collections;
-using UnityEngine.Events;
-
 namespace MiniGames
 {
 	public interface IMiniGame
--- a/UnityProject/Assets/Scripts/MiniGames/IMiniGame.cs
+++ b/UnityProject/Assets/Scripts/MiniGames/IMiniGame.cs
@@ -1,0 +1,11 @@
+using System.Collections;
+using UnityEngine.Events;
+
+namespace MiniGames
+{
+	public interface IMiniGame
+	{
+		public void StartGame();
+		public UnityAction<bool> HasWon();
+	}
+}
--- a/UnityProject/Assets/Scripts/MiniGames/IMiniGame.cs
+++ b/UnityProject/Assets/Scripts/MiniGames/IMiniGame.cs
@@ -6,6 +6,6 @@ namespace MiniGames
 	public interface IMiniGame
 	{
 		public void StartGame();
-		public UnityAction<bool> HasWon();
+		public void GameEnd(bool hasWon);
 	}
 }
--- a/UnityProject/Assets/Scripts/MiniGames/IMiniGame.cs
+++ b/UnityProject/Assets/Scripts/MiniGames/IMiniGame.cs
@@ -3,6 +3,7 @@ namespace MiniGames
 	public interface IMiniGame
 	{
 		public void StartGame();
-		public void GameEnd(bool hasWon);
+		public void GameWin();
+		public void GameLoss();
 	}
 }
--- a/UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
+++ b/UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
@@ -22,10 +22,24 @@ namespace MiniGames
 			Tracker.OnStartGame.AddListener(StartMiniGame);
 		}
 
-		//Add your game setup/reset/ui show/etc here
+		/// <summary>
+		/// Add your game setup/reset/ui show/etc here. (Don't forget about OnGameDone() because it's not abstract and
+		/// C# can't implament default behaviors for abstract voids.)
+		/// </summary>
+		/// <returns></returns>
 		public abstract void StartMiniGame();
 
-		//Add your logic here when a player has won or lost a game or has left the game
-		protected abstract void OnGameDone(bool t);
+		/// <summary>
+		/// Add your logic here when a player has won or lost a game or has left the game.
+		/// </summary>
+		protected virtual void OnGameDone(bool t)
+		{
+			if (Tracker == null)
+			{
+				Logger.LogError($"[Minigames] - Tracker missing on {gameObject}!");
+				return;
+			}
+			Tracker.OnGameEnd(t);
+		}
 	}
 }
--- a/UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
+++ b/UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
@@ -10,13 +10,15 @@ namespace MiniGames
 	public class MiniGameModule : MonoBehaviour
 	{
 		protected MiniGameResultTracker Tracker;
+		protected GameObject MiniGameParent;
 
 		/// <summary>
 		/// Use the Setup() function only when you're assigned events programatically and not from the inspector.
 		/// </summary>
-		public virtual void Setup(MiniGameResultTracker tracker)
+		public virtual void Setup(MiniGameResultTracker tracker, GameObject parent)
 		{
 			Tracker = tracker;
+			MiniGameParent = parent;
 			Tracker.OnStartGame.AddListener(StartMiniGame);
 			Tracker.OnGameDone.AddListener(OnGameDone);
 		}
--- a/UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
+++ b/UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
@@ -7,7 +7,7 @@ namespace MiniGames
 	/// They are meant to be used when an object can house more than one minigame or if the minigame's behavior is too complex
 	/// to be stored in just the StartMiniGame() method.
 	/// </summary>
-	public class MiniGameModule : MonoBehaviour
+	public abstract class MiniGameModule : MonoBehaviour
 	{
 		protected MiniGameResultTracker Tracker;
 		protected GameObject MiniGameParent;
--- a/UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
+++ b/UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
@@ -30,5 +30,10 @@ namespace MiniGames
 		{
 			//Add your logic here when a player has won or lost a game or has left the game
 		}
+
+		public struct MiniGameResultData
+		{
+			//When you want to send data to the server or to an object use this struct to move it around
+		}
 	}
 }
--- a/UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
+++ b/UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
@@ -20,7 +20,6 @@ namespace MiniGames
 			Tracker = tracker;
 			MiniGameParent = parent;
 			Tracker.OnStartGame.AddListener(StartMiniGame);
-			Tracker.OnGameDone.AddListener(OnGameDone);
 		}
 
 		public virtual void StartMiniGame()
--- a/UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
+++ b/UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
@@ -9,7 +9,7 @@ namespace MiniGames
 	/// </summary>
 	public class MiniGameModule : MonoBehaviour
 	{
-		private MiniGameResultTracker Tracker;
+		protected MiniGameResultTracker Tracker;
 
 		/// <summary>
 		/// Use the Setup() function only when you're assigned events programatically and not from the inspector.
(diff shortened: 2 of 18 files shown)


UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
```diff
@@ -0,0 +1,34 @@
+using UnityEngine;
+
+namespace MiniGames
+{
+	/// <summary>
+	/// MiniGameModules are gameObjects that sit on the gameObject that has the IMiniGame Interface.
+	/// They are meant to be used when an object can house more than one minigame or if the minigame's behavior is too complex
+	/// to be stored in just the StartMiniGame() method.
+	/// </summary>
+	public class MiniGameModule : MonoBehaviour
```
corp-0: make abstract

UnityProject/Assets/Scripts/Objects/Closets/ClosetControl.cs
```diff
@@ -5,14 +5,18 @@
...
+using UnityEngine.Events;
+using Random = UnityEngine.Random;
 
 namespace Objects
 {
 	/// <summary>
 	/// Allows closet to be opened / closed / locked
 	/// </summary>
-	public class ClosetControl : NetworkBehaviour, IServerSpawn, ICheckedInteractable<PositionalHandApply>, IRightClickable, IExaminable, IEscapable
+	public class ClosetControl : NetworkBehaviour, IServerSpawn, ICheckedInteractable<PositionalHandApply>, IRightClickable, IExaminable, IEscapable, IMiniGame
```
corp-0: I would rather make a new component with the minigame that uses ``ClosetControl``. This minigame also won't be used in most of crates/lockers, so it is better to keep the responsibility of this class scoped. The abandoned crates would be their own prefab which add this component and the minigame component.

TLDR: invert the references so a component that implements ``IMinigame`` uses ``ClosetControl`` and not the other way.

corp-0 (changes requested):

UnityProject/Assets/Scripts/Objects/Closets/ClosetControl.cs
MaxIsJoe: On it when I jump on rider 👍

UnityProject/Assets/Scripts/Objects/Closets/AbandonedCrates.cs
```diff
@@ -0,0 +1,87 @@
+using System.Collections.Generic;
+using MiniGames;
+using Mirror;
+using UnityEngine;
+
+namespace Objects.Closets
+{
+	public class AbandonedCrates : ClosetControl, IMiniGame
```
NoooneyDude: It doesn't make much sense for a minigame to _be_ a closet. ~~Rather, a minigame can _have_ a closet~~ I see, we can attach any minigame to this closet. But the minigame delegator shouldn't extend closets.

`GetComponent<ClosetControl>.DoThingsToClosetControl()` would be more appropriate. ~~This would also allow this class to inherit from `MiniGameModule`, if that makes more sense to you.~~

MaxIsJoe: The minigame is not the closet, AbandonedCrate is a closet that has it's own behavior.

NoooneyDude: All of this code is about minigames, not a closet.

MaxIsJoe: Like I said, Abandoned Crates are closets with their own behavior. Their behavior is that they use minigames to be opened which in this context makes sense because we want the closets' behavior for storing objects and closing/opening them. The minigames take place the open/close interactions and that's the whole point of those crates, minigames.

Am I misreading something?

NoooneyDude: The problem is this can't be easily used for any type of `ClosetControl`. Sure, we might only have one `ClosetControl` script for now, but if we add others, then we could no longer use this script and we'd need to refactor it, if we wanted to add minigames to them. However, if we use _composition_, we don't need to worry about what the `ClosetControl` looks like, just that we trust it to do its job.

So, I am indicating a design issue. But I won't fight you on it.

MaxIsJoe: You actually make a really good point, but:
The problem with separating this into two scripts is that Unity has a problem with component hierarchies, which have caused me headaches multiples times before while working on stuff that involves interactions. So, while I do agree with you, I don't want this to randomly break in the future because Unity decided to shuffle component positions around.

corp-0: I still believe this could very well just be its own component and use a reference to Closet control. Don't see why it should extend it.

Alainx277: I agree with Gilles here.

UnityProject/Assets/Scripts/Objects/Closets/AbandonedCrates.cs
```diff
@@ -0,0 +1,67 @@
...
+			{
+				Logger.LogError("[ClosetControl/MiniGames] - Found MiniGame tracker but no minigames found!");
+				return;
+			}
+			currentMiniGameIndex = Random.Range(0, miniGameModules.Count - 1);
+			miniGameModules[currentMiniGameIndex].Setup(miniGameTracker, gameObject);
+			SetLock(Lock.Locked);
+		}
+
+		protected override void InteractionChecks(PositionalHandApply interaction)
```
Alainx277: I feel like it would make more sense to override `TryToggleLock`? I'd prefer an event approach, but overriding a more specific method is also fine.

UnityProject/Assets/Scripts/MiniGames/IMiniGame.cs
```diff
@@ -0,0 +1,9 @@
+namespace MiniGames
+{
+	public interface IMiniGame
```
Alainx277: I don't see this interface used anywhere?

UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
```diff
@@ -0,0 +1,35 @@
...
+		/// Use the Setup() function only when you're assigned events programatically and not from the inspector.
+		/// </summary>
+		public virtual void Setup(MiniGameResultTracker tracker, GameObject parent)
+		{
+			Tracker = tracker;
+			MiniGameParent = parent;
+			Tracker.OnStartGame.AddListener(StartMiniGame);
+		}
+
+		public virtual void StartMiniGame()
```
Alainx277: If there's no default implementation it should be abstract.

UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
```diff
@@ -0,0 +1,35 @@
...
+		{
+			//Add your game setup/reset/ui show/etc here
+		}
+
+		protected virtual void OnGameDone(bool t)
```
Alainx277, line 35: This doesn't seem to be used in the context of the base class. Can you remove it from here?

UnityProject/Assets/Scripts/Objects/Closets/AbandonedCrates.cs
Alainx277: I feel like almost everything here belongs into the minigame tracker.

UnityProject/Assets/Scripts/Objects/Closets/AbandonedCrates.cs
```diff
@@ -0,0 +1,67 @@
...
+using UnityEngine;
+
+namespace Objects.Closets
+{
+	public class AbandonedCrates : ClosetControl, IMiniGame
+	{
+
+		[SerializeField] private MiniGameResultTracker miniGameTracker;
+		[SerializeField] private List<MiniGameModule> miniGameModules;
+		[SyncVar] private int currentMiniGameIndex = -1;
```
Alainx277: Why is this synced to the client?

Alainx277 (changes requested):

UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
MaxIsJoe: abstracc

UnityProject/Assets/Scripts/Objects/Closets/AbandonedCrates.cs
MaxIsJoe: bi-product of old code while testing, removing.

UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
MaxIsJoe: it is being used

UnityProject/Assets/Scripts/MiniGames/IMiniGame.cs
MaxIsJoe: It is being used.

UnityProject/Assets/Scripts/Objects/Closets/AbandonedCrates.cs
MaxIsJoe: Why would you "Try" to do something that should be a forced state?
This feels like it's going to result in bugs, so I'm not taking risks.

UnityProject/Assets/Scripts/Objects/Closets/AbandonedCrates.cs
MaxIsJoe: can you explain to me how I can move this to the minigame tracker? I don't see how It makes sense to be moved there.

MaxIsJoe: also, again, read what I said about interactions. How do you want to trigger the minigame if you guys didn't like me putting it inside the closet code and I have told you already how having two interactable interfaces causes one to be priortized over the other.

UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
Alainx277: It's only being used in the child class, never in the context of `MiniGameModule`. So why have it on the base class?

UnityProject/Assets/Scripts/Objects/Closets/AbandonedCrates.cs
Alainx277: I thought a minigame was supposed to be generic enough to be used for multiple things. I reckon the minigame tracker (maybe call it a manager) would receive requests for a minigame, start the minigame and handle the minigame success/fail.
Why does the abandoned crate check for multiple minigame modules? It should not need to care about that.
About the interaction: We want to react when a player tries to open the locked crate. Why not create an event that fires if it's locked, which is listened to by a minigame tracker, which starts a random minigame. That way the crate code doesn't care about minigames (as it should be imo).

MaxIsJoe: Moved out the functionality out of the closet control code

However, expect this to randomly break sometime in the future when Unity decides to scramble the position of components again.

MaxIsJoe:
this should be ready for review again

MaxIsJoe:
Fixed some codacy issues, ignore the ones related to events because it thinks they're not being used but they are in the inspector.

UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
```diff
@@ -0,0 +1,31 @@
...
+			Tracker = tracker;
+			MiniGameParent = parent;
+			Tracker.OnStartGame.AddListener(StartMiniGame);
+		}
+
+		//Add your game setup/reset/ui show/etc here
+		public abstract void StartMiniGame();
+
+		//Add your logic here when a player has won or lost a game or has left the game
+		protected abstract void OnGameDone(bool t);
```
Alainx277: Would it make sense to have this call `Tracker.OnGameEnd` so it's not forgotten?

UnityProject/Assets/Scripts/Objects/Closets/AbandonedCrates.cs
```diff
@@ -0,0 +1,78 @@
...
+using UnityEngine;
+using Random = UnityEngine.Random;
+
+namespace Objects.Closets
+{
+	public class AbandonedCrates : MonoBehaviour, IServerSpawn, ICheckedInteractable<HandApply>
+	{
+
+		[SerializeField] private MiniGameResultTracker miniGameTracker;
+		[SerializeField] private List<MiniGameModule> miniGameModules;
```
Alainx277, line 12: I see that the crate is still linked directly to minigames. It's fine for now, just keep in mind that everything that has a minigame will need to do this too.

Alainx277 (changes requested):
Looks much better. Just a small nit.

UnityProject/Assets/Scripts/MiniGames/MiniGameModule.cs
MaxIsJoe: what do you mean

corp-0: I think he means giving the method a default implementation and call Tracker.OnGameEnd() inside this method

MaxIsJoe: I don't like the idea of turning this into a virtual void but since this PR has been taking forever.. eh, whatever. done.

corp-0 (approved):

PR #9145, opened 2022-08-20
Singularity fix

ThatDan123:
-Made the scale change only go to the current stage (was confusing for players and looked like it was escaping the field even if it couldnt when moving)

-Can no longer be pushed by wind and is now sticky movement

-Fix issue where PA setting 0 would increase singularity points

-Collider now bigger for larger singularities so the PA projectiles will all hit


Q4ik:
We did a singi on AE and it ate the lower left corner of that room. Emitter and rad collector taken as well. Singi stayed in that corner until it fizzed out. The field was intact, though we could clip in and out of it

Q4ik:
Lower right*

ThatDan123:
> We did a singi on AE and it ate the lower left corner of that room. Emitter and rad collector taken as well. Singi stayed in that corner until it fizzed out. The field was intact, though we could clip in and out of it

Yeah i think thats the movement/wind issue

corp-0 (approved):

PR #9146, opened 2022-08-20
Ghost should always spawn when players spawn 

Bod9001:
### Purpose
starting work towards mind rework

### DIFF
--- a/UnityProject/Assets/Scripts/Player/GhostMove.cs
+++ b/UnityProject/Assets/Scripts/Player/GhostMove.cs
@@ -13,11 +13,11 @@ public class GhostMove : NetworkBehaviour, IPlayerControllable
 
 	public bool Moving = false;
 
-	private RegisterTile registerTile;
+	public RegisterTile registerTile;
 
-	private PlayerScript playerScript;
+	public PlayerScript playerScript;
 
-	private Rotatable rotatable;
+	public Rotatable rotatable;
 
 	public Vector3 LocalTargetPosition;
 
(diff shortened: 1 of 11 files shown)


ThatDan123:
Codacy stuff

UnityProject/Assets/Scripts/Player/GhostMove.cs
```diff
@@ -13,11 +13,11 @@ public class GhostMove : NetworkBehaviour, IPlayerControllable
 
 	public bool Moving = false;
 
-	private RegisterTile registerTile;
+	public RegisterTile registerTile;
```
ThatDan123:
```suggestion
	public RegisterTile RegisterTile {get; private set;};
```

Can you do that for all of these

Bod9001: I was  Settingit through the editor

ThatDan123:
```suggestion
	public RegisterTile RegisterTile {get; set;};
```
Do that then, it just doesnt need to be shown in the inspector.

ThatDan123: oh ok, i cant read

ThatDan123: Cant you just [SerialiseField] then? I don't see why they need to be public?

corp-0:
Try this
[field: SerializeField]
public int Value
{
    get;
    private set;
}

UnityProject/Assets/Scripts/Player/GhostMove.cs
Bod9001: ok

Bod9001:
does that work of the unity editor I thought it was incapable of doing getters and setters?

Bod9001:
'GhostMove.registerTile.set' accessor must be more restrictive than the property or indexer 'GhostMove.registerTile'

UnityProject/Assets/Scripts/Player/GhostMove.cs
```diff
@@ -13,11 +13,14 @@ public class GhostMove : NetworkBehaviour, IPlayerControllable
 
 	public bool Moving = false;
 
-	public RegisterTile registerTile;
+	[SerializeField] private RegisterTile registerTile {
```
ThatDan123: Would need to be public to get rid of that error

PR #9147, opened 2022-08-21
Optimise MetaDataNode Memory Usage

ThatDan123:
-Foam, smoke and radiation nodes will only be created when needed (In a further PR might add checks for some access whether they actually need to create or whether checking for existence is all they need)

-Clients won't have the memory usage for the gasMix

-Fixed some case names in Foam and Smoke scripts

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/SmokeAndFoam/FoamNode.cs
+++ b/UnityProject/Assets/Scripts/Systems/SmokeAndFoam/FoamNode.cs
@@ -36,13 +36,14 @@ public class FoamNode : SpreadNode
 						SmokeAndFoamManager.Instance.WallFoam);
 				}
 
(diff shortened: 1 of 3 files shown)


MaxIsJoe:
why about we pool nodes?

ThatDan123:
> why about we pool nodes?

Maybe in the future, but this PR only makes it so they are created when first needed (they're not destroyed after that).

When you think about it 99% of the time they are never really needed. Theres around 8000 MetaDataNodes just for SquareStation and in a round barely any of them will ever get smoke or foam. Then think of Lavaland probably 100x the amount of nodes. Just wasted memory.

Also since bod made slipper sync to clients using the nodes and clients don't need anything else, its sorta just bad.

Might want to make a separate script like ClientMetaDataNode just for the stuff we want to sync.

NoooneyDude (approved):

PR #9151, opened 2022-08-21
Makes All Alerts Animate

ThatDan123:
-Makes all alerts animated using sprite handler

-Removed Animate Image script (didnt realise spritehandler could do Images)


corp-0:
nice, I often missed the alerts because they weren't obvious enough in my peripheral vision (ultra wide monitor). Movement will make them a lot more noticeable.

Bod9001 (approved):

PR #9152, opened 2022-08-21
Tile List Fix

ThatDan123:
-Forgot i changed it then didnt test rip.

-Adds check for welder (isnt what is causing them to break)

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Tool/Welder.cs
+++ b/UnityProject/Assets/Scripts/Items/Tool/Welder.cs
@@ -8,7 +8,7 @@ using Chemistry.Components;
 using Items;
 
 [RequireComponent(typeof(Pickupable))]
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Items/Tool/Welder.cs
```diff
@@ -102,6 +102,11 @@ public void OnSpawnServer(SpawnInfo info)
 		SyncIsOn(isOn, false);
 	}
 
+	public bool WillInteract(HandActivate interaction, NetworkSide side)
+	{
+		return DefaultWillInteract.Default(interaction, side);
```
corp-0, line 107: is this really the only check needed for this interaction?

Bod9001: isn't it implemented by default if it's not specified?

PR #9153, opened 2022-08-21
Staging fixes7

Bod9001:
### Changelog:

CL: [Improvement] Smaller hollow projector 
CL: [Improvement] Stop throwing stuff on yourself


ThatDan123 (approved):

ThatDan123 (changes requested):
Codacy stuff

Bod9001:
"The Cyclomatic Complexity of this method is 11 which is greater than 10 authorized. " 
RemovePlayerbyUserId

it's not even that complicated , stupid thing

UnityProject/Assets/Scripts/Managers/PlayerList.cs
```diff
@@ -512,6 +512,46 @@ public GameObject TakeLoggedOffPlayerbyUserId(string userId)
 		return null;
 	}
 
+	[Server]
+	public PlayerInfo RemovePlayerbyUserId(string userId, PlayerInfo newPlayer)
```
corp-0, line 516: could this instead be a ``TryRemovePlayerByUserId`` and return a bool instead? It can ``out``the ``PlayerInfo``

ThatDan123:
just this codacy one that needs to be looked at:

image

If Equals isnt static then it could NRE?

UnityProject/Assets/Scripts/Managers/PlayerList.cs
Bod9001: why does it have to be a try?

Bod9001:
var Setstate = default(MatrixState);
		if (state != null)
		{
			Setstate = state.Value;
		}

		if (Setstate.Equals(default(MatrixState)))
		{
		looks like it should be fine?

ThatDan123:
default(MatrixState) is null though, right? So if state is null then Setstate  will be null, so will NRE during the equals?

Bod9001:
I don't think so, because it's a 
		MatrixState Setstate = default(MatrixState);
		this is fine it can't physically be null

ThatDan123:
Humm yeah is a struct so should actually be fine

ThatDan123 (dismissed):

UnityProject/Assets/Scripts/Managers/PlayerList.cs
NoooneyDude: To indicate that there's a possibility it won't succeed e.g. userId is invalid. Might be useful for something.

ThatDan123 (approved):

UnityProject/Assets/Scripts/Managers/PlayerList.cs
corp-0: also because you are returning null in case it fails, so this is exactly the kind of methods that are perfect for that pattern. I find it better for the rest of the flux to deal with a bool and act accordingly than handling a possible null.

PR #9161, opened 2022-08-24
VV Fix

ThatDan123:
-Fix vv opening


UnityProject/Assets/Scripts/UI/Systems/AdminTabButtons.cs
```diff
@@ -84,7 +84,6 @@ private void DisableAllGUI()
 			devCloner.gameObject.SetActive(false);
 			devDestroyer.gameObject.SetActive(false);
 			devSelectTile.gameObject.SetActive(false);
-			vv.SetActive(false);
```
NoooneyDude, line 87: Sorry, can you explain what happened here?

ThatDan123: The vv script is actually on a sub object of the VV prefab, so setting active false breaks it.

I added that in the tile placer PR, as i thought it was just missing.

NoooneyDude (approved):

PR #9162, opened 2022-08-24
Tile Changer Area

ThatDan123:
-Click drag to fill areas

-Graffiti tiles

-Colour mode to set colour without changing tile type

-Fix bug with pressing ecs to close


UnityProject/Assets/Scripts/UI/Systems/AdminTools/DevTools/GUI_DevTileChanger.cs
```diff
@@ -272,47 +295,70 @@ public void OnActionButtonClick(int buttonType)
...
+			if (CommonInput.GetMouseButtonUp(0) == false) return;
+
+			if(dragStartPos == MouseInputController.MouseWorldPosition.RoundToInt()) return;
+
+			//End drag
+			switch (currentAction)
+			{
+				case ActionType.Place:
+					//Also remove if shift is pressed when placing for quick remove
+					if (Input.GetKey(KeyCode.LeftShift) || Input.GetKey(KeyCode.RightShift))
```
NoooneyDude, line 357: Would it be sensible to do the inverse if in remove mode? Or do you see shift as to guarantee remove mode.

A little hint in the UI might be nice too if it doesn't have one already.

NoooneyDude (dismissed):

NoooneyDude:
Could consider sending a struct instead, might be a bit tidier => help shut up Codacy.

UnityProject/Assets/Scripts/UI/Systems/AdminTools/DevTools/GUI_DevTileChanger.cs
AtnerNT: On the note of remove mode, I still find it funny how the remove tool for admins actually works correctly yet the unity editor version is still broke

ThatDan123:
> Could consider sending a struct instead, might be a bit tidier => help shut up Codacy.

Will do that now

UnityProject/Assets/Scripts/UI/Systems/AdminTools/DevTools/GUI_DevTileChanger.cs
ThatDan123: > Would it be sensible to do the inverse if in remove mode? Or do you see shift as to guarantee remove mode.
> 
> A little hint in the UI might be nice too if it doesn't have one already.

Have made it just do the reverse, not sure what UI thing i could add?

NoooneyDude: Could replace the mode buttons with w/ a toggle group and change the current mode string to a hint label. Up to you though.

corp-0:
was it complaining because the methods had too many parameters? That is a fair concern, most of the time that is a symptom of a function just doing way too many things.

NoooneyDude:
> was it complaining because the methods had too many parameters? That is a fair concern, most of the time that is a symptom of a function just doing way too many things.

In this case it was the entry point for client => server, so a struct is sensible.

NoooneyDude (approved):

PR #9166, opened 2022-08-25
Rest

ThatDan123:
-Clean up some of the rest functions

-Rename the registerTile variable in PlayerScript to RegisterPlayer (since that is actually what it is)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Objects/RegisterPlayer.cs
```diff
@@ -104,6 +104,33 @@ public override bool IsPassableFromOutside(Vector3Int from, bool isServer, GameO
 		return IsPassable(isServer);
 	}
 
+	[Command]
+	public void CmdToggleRest(bool layingDown)
```
corp-0: if this is a toggle, why do we also have to tell it if we are laying down or not?

corp-0: also the name can be a little misleading, since it is more about changing the standing/laying down position than it is resting.

PR #9167, opened 2022-08-26
MOTD window works again + improvements

### LINKED ISSUE #9122: MOTD window appears to be printing new lines forever?
There's a lot of weird issues with the MOTD window

for starters, the scrollbar keeps moving on it's own to the top as if something keeps getting printed

Rules and Events buttons appear in the ESC menu when they shouldn't because there are no rules on staging currently nor is it christmas yet.

windows disappear randomly when moved

stuff needs to be investigated to see why they happen because code wise there are no issues it seems when I went over it multiple times.

image

MaxIsJoe:
Fixes #9122 + I've worked on stuff that has been annoying me for some time like how there were 2 different versions of the server info UI which made things pretty hard to keep track of and fix bugs for.

most of these improvements are on a prefab level so for reviewers :

- There is now only one serverinfo prefab in the UI manager that is independent from any other UI windows.
- removed `SetActive` checks from some functions because the `ClientSetValues` already does this for you
- fixed some issues in the load manager that prevented the MOTD window from getting it's values set on the server plus removed missing scripts
- Something was wrong with the manager prefab so I had to manually re-import it then fix the issues it kept popping up (something related to missing scripts)
- Added "serverDesc.txt" and "serverRules.txt" to streaming assets because they should have been there from the start

## Changelog:

CL: [Fix] MOTD window works again


ThatDan123 (dismissed):

ThatDan123:
Oh tests failing

NoooneyDude:
😌 was worried for a second there we were modifying the same prefab 🙂

MaxIsJoe:
should be fixed now

corp-0:
> Added "serverDesc.txt" and "serverRules.txt" to streaming assets because they should have been there from the start

remind me to update docker mounting points accordingly to reflect this change after this is merged.

UnityProject/Assets/Scripts/UI/Systems/ServerInfoUILobby.cs
```diff
@@ -29,7 +29,8 @@ public class ServerInfoUILobby : MonoBehaviour, IInitialise
 
         public static string serverDesc;
 
-        public static string serverDiscordID;
+        private static string serverDiscordID;
+        public string ServerDiscordID => serverDiscordID;
```
corp-0:
```suggestion
        public string ServerDiscordID { get; private set;}
```

That way you don't need the private field. You would of course need to correct all references to the private field to point now to the property.

UnityProject/Assets/Scripts/UI/Systems/ServerInfoUILobby.cs
```diff
@@ -199,7 +205,7 @@ public override void Process(NetMessage msg)
 		{
 			ServerInfoLobbyMessageServer.Send(
 					SentByPlayer.Connection, ServerData.ServerConfig.ServerName,
-					ServerInfoUILobby.serverDesc, ServerInfoUILobby.serverDiscordID, ServerInfoUILobby.serverRules);
+					UIManager.Instance.ServerInfoUILobby.ServerDesc.text, UIManager.Instance.ServerInfoUILobby.ServerDiscordID, UIManager.Instance.ServerInfoUILobby.ServerRules.text);
```
corp-0: separate each argument into its own line.

I thought this was an insanely long chain of stuff at first glance.

corp-0 (changes requested):
Other than my comments (and assuming this works), I see no other issue worth mentioning.

MaxIsJoe:
Forgot to push this yesterday, should be done now.

corp-0 (approved):

PR #9172, opened 2022-08-27
Mind is now a mono behaviour

Bod9001:
### Purpose
as part of the mind rework


corp-0:
you left the constructor for ``Mind``, that's why codacy is complaining. Maybe move it to an ``Awake``?

corp-0:
I personally don't see any issue with this, maybe @ThatDan123 can catch something?

corp-0 (approved):

PR #9187, opened 2022-08-31
Admin Tile Changer fixes

### LINKED ISSUE #9158: Admin Tile Placer Electrical
-Currently placing electrical tiles doesnt hook them up to the electrical grid, Bod will have to take a look at it as i couldnt understand the code in CableCoil.cs in order to get it working.

-Just needs to be hooked in here in the AdminCommandsManager for placing tiles
image

AtnerNT:
Fixes #9158

### Changelog:

CL: [Fix] Admin tile placer now correctly places electrical cables
CL: [Fix] Fixes an issue in the tile placer UI were all sprites were rotated incorrectly, leading to much confusion with asymmetrical tiles

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/AdminCommandsManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/AdminCommandsManager.cs
@@ -1057,7 +1057,7 @@ namespace AdminCommands
 
 			if (tile is ElectricalCableTile electricalCableTile)
 			{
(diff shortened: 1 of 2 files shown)


AtnerNT:
getting codacy fail cause magic number. However I don't think 90 is a magic number in this case, 90 degrees is pretty universal, like 0, 1 and -1. I don't think its worth making into constant

ThatDan123 (approved):

ThatDan123:
Might have a compilation error?

I've rerun the tests just in case it was a fluke

PR #9188, opened 2022-08-31
Fixes randomised hunger

Bod9001:
technically work before but  is to fix it for non  1 values


corp-0:
Wait, why is hunger random now? What does it solve?

MaxIsJoe:
To be truly random, you should apply randomness to each body part.
Also, I still don't like that it's random in the first place.

NoooneyDude:
> Wait, why is hunger random now? What does it solve?

Perhaps it solves the issue of all the roundstart crew bumrushing the vending machines at the same time. Initial hunger on spawn should be random, at least.

NoooneyDude (approved):

Bod9001:
It's only the starting amount of hunger

PR #9217, opened 2022-09-11
Fixes protips and finally adds tip queuing

MaxIsJoe:
sorry this took a while
part of #5316

### Notes:

Tips might appear to be still not functional but they're not; they are saved on your machine as interacted with before.
If you want to test, check out the super matter as it can be triggered over and over again.

### Changelog:
CL: [New] Tips are now queued.
CL: [Fix] Tips now function again when picking up items or approaching ranged tips.

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Inventory/Pickupable.cs
+++ b/UnityProject/Assets/Scripts/Systems/Inventory/Pickupable.cs
@@ -43,9 +43,9 @@ public class Pickupable : NetworkBehaviour, IPredictedCheckedInteractable<HandAp
 
 	public ItemAttributesV2 ItemAttributesV2;
 
-	public event Action OnMoveToPlayerInventory;
+	public event Action<GameObject> OnMoveToPlayerInventory;
 
-	
+
 	#region Lifecycle
 
 	private void Awake()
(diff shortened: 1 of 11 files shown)


PerfectTangent:
rather than automatically marking tips as shown, I think I'd prefer the Rimworld way where they show you it a few times, but giving you the choice to permanently mark as read. There are some tips that don't need to be shown repetitively, but there are others that would benefit from it a lot, like how to use internals when dying.

corp-0:
it depends on the tip currently. Like it says in the description, the super matter triggers all the time while others are just not that important to be shown repeatedly

MaxIsJoe:
> rather than automatically marking tips as shown, I think I'd prefer the Rimworld way where they show you it a few times, but giving you the choice to permanently mark as read. There are some tips that don't need to be shown repetitively, but there are others that would benefit from it a lot, like how to use internals when dying.

That's actually a very good idea.
Will add it in the next PRs

corp-0:
C# has a queue data structure https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.queue-1?view=net-6.0

UnityProject/Assets/Scripts/Systems/Inventory/Pickupable.cs
```diff
@@ -213,7 +213,7 @@ private IEnumerator ServerPerformInteractionLogic(HandApply interaction)
 
 	private void PickupAnim(GameObject interactor)
 	{
-		OnMoveToPlayerInventory?.Invoke();
+		OnMoveToPlayerInventory?.Invoke(interactor);
```
corp-0: I get it works, but the event should be invoked in the coroutine that calls this method instead of the method that should be only responsible of playing the picking up animation.

MaxIsJoe: done

MaxIsJoe:
> C# has a queue data structure https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.queue-1?view=net-6.0

I don't get how we can use this here

corp-0:
Instead of the list

corp-0 (dismissed):

corp-0 (approved):

PR #9219, opened 2022-09-14
Fixes inventory not being visible in Sub- inventories

Bod9001:
### Changelog:

CL: [Fix] Fixes inventory not being visible in Sub- inventories


corp-0:
you fixed the bug with just deleting code?

Bod9001:
yeah, it looks like there's already a system for handling observing inventories, that's on the object itself so this is unnecessary

corp-0:
less code is always good. Fix the tests failing and we are good. Also register the chemical addition in changelogs.

corp-0 (approved):

PR #9220, opened 2022-09-14
Hopefully fixs and A ghost blob

Bod9001:
please review 

### Changelog:

CL: [Fix] Hopefully fixs blob and  A ghost


corp-0:
changelog detection wasn't workin because there was an space before ``CL:``

Bod9001:
might be worth to check for that

corp-0 (approved):

PR #9232, opened 2022-09-19
Staging fixes17

Bod9001:
### Changelog:

CL: [Fix] Fixed Rolledposters 
CL: [Fix] fixes paper alignment
CL: [Fix] Fixes dice not rolling properly

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Dice/RollDie.cs
+++ b/UnityProject/Assets/Scripts/Items/Dice/RollDie.cs
@@ -61,7 +61,7 @@ namespace Items.Dice
 		{
 			ObjectPhysics.OnThrowStart.AddListener(ThrowStart);
 			ObjectPhysics.OnImpact.AddListener(ThrowEnd);
-			ObjectPhysics.OnThrowEnd.AddListener(ThrowStartEndold);
+			ObjectPhysics.OnThrowEnd.AddListener(ThrowEndOld);
 			if (cookable != null && isRiggable)
 			{
 				cookable.OnCooked += Cook;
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/Items/Dice/RollDie.cs
```diff
@@ -134,6 +134,11 @@ private void ThrowStart(UniversalObjectPhysics throwInfo)
 			Chat.AddActionMsgToChat(throwInfo.thrownBy, $"You throw the {dieName}...", $"{throwInfo.thrownBy.ExpensiveName()} throws the {dieName}...");
 		}
 
+		private void ThrowStartEndold(UniversalObjectPhysics throwInfo)
+		{
+			this.RestartCoroutine(WaitForSide(), ref waitForSide);
```
corp-0, line 139: I don't understand why is this method needed now. Did something break after the change to new physics?

I also can't read the name of the method.

Bod9001: throwing works differently now so you can't use the old system exactly

corp-0: ok so why does it need a new method instead of using the one we have now? Is it too hard to make rolling dice use the new signature instead of creating a ``ThrowEndOld`` method?

Bod9001: well, you can't reuse it because it expects there to be momentum difference but, implicitly by the name it saying that it's finished moving, so therefore there is no momentum change

corp-0: momentum 0 ?

Bod9001: yeah, because it stopped moving, Because it's reached the end of the throw

corp-0: yes, so why not call ``OnThrowEnd`` with ``0`` as the argument

corp-0: actually, why does ThrowEnd need a force if it is called when it has no force?

Bod9001: That's what uses the old method because it doesn't have a force, and also loads of other stuff still uses it

PR #9241, opened 2022-09-26
More scores for the score machine

MaxIsJoe:
# Changelog:

CL: [New] Station getting nuked has a massive score penalty now.
CL: [New] Hugs are now recorded in the score machine.
CL: [New] Electrified doors from an AI are now recorded in the score machine.
CL: [New] When the big service bell Easter egg triggers, it will show up in the round end score screen.
CL: [Improvement] "Most hurt crewmember" score will only appear if there is a crew member below 50 overall health.

### DIFF
--- a/UnityProject/Assets/Scripts/Items/ServiceBell.cs
+++ b/UnityProject/Assets/Scripts/Items/ServiceBell.cs
@@ -1,5 +1,6 @@
 ﻿using UnityEngine;
 using AddressableReferences;
+using Systems.Score;
 
 namespace Objects
 {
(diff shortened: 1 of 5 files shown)


MaxIsJoe:
CODACY, IM GOING TO KILL YOOOOUUUUU
image

corp-0:
those look more like constant to me though

MaxIsJoe:
> those look more like constant to me though

static because you can just do `RoundEndScoreBuilder.COMMON_SCORE_SOMETHING` instead of `RoundEndScoreBuilder.instance.COMMON_SCORE_SOMETHING`

corp-0:
you can access constant members of a class without the singleton pattern

MaxIsJoe:
oh shoot, you actually can

why cant there be a const for functions then

corp-0:
because functions aren't values, they perform an operation and would, optionally, return a value form that operation. If you want to call a method from a class without instantiating it, you can make the function static but then you can't really access the class members that aren't also static.

corp-0 (approved):

PR #9254, opened 2022-10-03
Remove some on validate

Bod9001:
hopefully this might stop the editor randomly opening prefabs and doing stuff, though On validate was still Triggering for some random stuff, so I think that is within unity land


UnityProject/Assets/Scripts/Tilemaps/Tiles/ObjectTile.cs
```diff
@@ -29,46 +29,46 @@ public class ObjectTile : BasicTile
...
-			}
-
-			objectCurrent = Object;
-
-			if (objectCurrent != null && objectCurrent.GetComponentInChildren<RegisterItem>() != null)
-			{
-				IsItem = true;
-			}
-		}
+		// private void OnValidate()
```
corp-0: if this really isn't needed, just remove it. Don't comment it.

Bod9001: it is weirdobject tile system we don't use any more

corp-0: more reason to nuke the entire script then!

corp-0 (dismissed):

Fogapod (dismissed):

PR #9255, opened 2022-10-04
Protips are remembered in json now instead of your OS's registry

MaxIsJoe:
CL: [New] - You can forget all protips to make them appear again from the help menu -> settings now.
CL: [Fix] - Fixed an issue where the "Remember" button would only appear on widescreen resolutions.

### DIFF
--- a/UnityProject/Assets/Scripts/Learning/ProtipManager.cs
+++ b/UnityProject/Assets/Scripts/Learning/ProtipManager.cs
@@ -12,8 +12,9 @@ namespace Learning
 		public ProtipListUI ListUI;
 		public ExperienceLevel PlayerExperienceLevel;
 		public List<ProtipSO> RecordedProtips;
+		public Dictionary<string, bool> ProtipSaveStates { private set; get; } = new Dictionary<string, bool>();
 
-		public List<ProTipSaveData> ProtipSaveStates { private set; get; } = new List<ProTipSaveData>();
+
 		private readonly Queue<ProtipSO> queuedTips = new Queue<ProtipSO>();
 		private string jsonPath;
 		private string jsonFileName = "/protips.json";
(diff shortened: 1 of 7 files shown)


UnityProject/Assets/Scripts/Learning/ProtipManager.cs
```diff
@@ -39,13 +45,53 @@ public override void Awake()
...
+		{
+			UpdateRecordedTips();
+			var found = false;
+			foreach (var tip in ProtipSaveStates)
+			{
+				if (tip.ID != ID) continue;
+				tip.Remembered = dontShowAgain;
+				found = true;
+				break;
+			}
```
NoooneyDude: Why not serialize as a dictionary?

UnityProject/Assets/Scripts/Learning/ProtipManager.cs
```diff
@@ -39,13 +45,53 @@ public override void Awake()
...
+
+			SaveProtipSaveStates();
+		}
+
+		private void UpdateRecordedTips()
+		{
+			if (File.Exists(jsonPath) == false || File.ReadAllText(jsonPath).Length <= JSON_EMPTY_LIST) return;
+			var newList = JsonConvert.DeserializeObject<List<ProTipSaveData>>(File.ReadAllText(jsonPath));
+			ProtipSaveStates = newList;
```
NoooneyDude: As this comes from IO we can't guarantee the string is always deserializable.  `DeserializeObject()` might return null or throw (check documentation).

At the least, null check it:
```suggestion
			var newList = JsonConvert.DeserializeObject<List<ProTipSaveData>>(File.ReadAllText(jsonPath));
			ProtipSaveStates = newList ?? new List<ProTipSaveData>();
```

UnityProject/Assets/Scripts/Learning/UI/ProtipListUI.cs
```diff
@@ -1,14 +1,14 @@
...
-using UnityEngine.UI;
 
 namespace Learning
 {
 	public class ProtipListUI : MonoBehaviour
 	{
 		[SerializeField] private GameObject entryToSpawn;
 		[SerializeField] private Transform entryList;
+		[SerializeField] private Transform mainPage;
+		[SerializeField] private Transform settingsPage;
```
NoooneyDude, line 11: I don't like this - settings consumers shouldn't communicate directly with the settings UI - leave that up to the settings scripts.

UnityProject/Assets/Scripts/Learning/ProtipManager.cs
MaxIsJoe: this causes an NRE when Rider suggested it

corp-0: handle it wherever you use ``ProtipSaveStates``.

UnityProject/Assets/Scripts/Learning/UI/ProtipListUI.cs
MaxIsJoe: what do you mean

NoooneyDude: Apologies, I thought this was linking to the escape-menu settings page 🙂

NoooneyDude (approved):

PR #9258, opened 2022-10-05
All textures from TGMC (up to 2022/10/01)

PerfectTangent:
### Purpose

I tried to comb through the .dmis to make sure weren't taking entire .dmis that were just stuff we already had. I think I was pretty thorough. However, there's going to be duplicates from .dmis that were a mix of old and new. It'd be nice to trim those out, but it's litterally tens of thousands of images to sort, so it will take time to do them all. Not a priority atm.


PerfectTangent:
lmao, wrong repository. My mistake.

PR #9269, opened 2022-10-16
Welder fixes

MaxIsJoe:
CL: [Fix] Welders now have proper stun immunity
CL: [Fix] Fixed an issue where stun immunity only works on clothes and not the body part's armor.
CL: [Fix] Fixed an issue where spells wouldn't work on non-wizards. (The welder can now teleport)

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Spells/Spell.cs
+++ b/UnityProject/Assets/Scripts/Systems/Spells/Spell.cs
@@ -18,7 +18,6 @@ namespace Systems.Spells
 	[DisallowMultipleComponent]
 	public class Spell : NetworkBehaviour, IActionGUI
 	{
-		[SerializeField] private Antagonist wizardData;
 		private SpellData spellData = null;
 		public SpellData SpellData {
 			get => spellData;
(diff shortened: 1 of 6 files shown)


Bod9001 (dismissed):

UnityProject/Assets/Scripts/Systems/Spells/Spell.cs
```diff
@@ -17,6 +18,7 @@ namespace Systems.Spells
 	[DisallowMultipleComponent]
 	public class Spell : NetworkBehaviour, IActionGUI
 	{
+		[SerializeField] private Antagonist wizardData;
```
ThatDan123: Not sure if i like this, can we just put this on some singleton somewhere?

ThatDan123: Actually sorry i thought this was a SO and didnt want to have to add this to all of them.

Might be fine, but you did forget to change the prefab this Spell monobehaviour is on and add the antag data

UnityProject/Assets/Scripts/Systems/Spells/Spell.cs
```diff
@@ -248,6 +250,9 @@ public virtual bool ValidateCast(PlayerInfo caster)
 
 		private bool CheckWizardGarb(Equipment casterEquipment)
 		{
+			// none-wizards shouldn't need a check for robes.
+			if (casterEquipment.ItemStorage.registerPlayer.PlayerScript.mind.GetAntag().Antagonist !=
```
ThatDan123: can GetAntag() return null?

corp-0: it can. You could add null propagation to the statement or use ``Mind.IsAntag`` before trying to access the antag. It is also weird how we get it with a method like this was old school java.

For these cases I think the better solution is using the ``Try`` pattern and return bool + out the wanted value if found.

MaxIsJoe:
Done. Review pls.

ThatDan123 (approved):

PR #9284, opened 2022-10-31
the option between classic and new hover tooltip

MaxIsJoe:
CL: [New] Classic tooltip can now be re-enabled via the options menu. (Though it's enabled by default from now on for the 13th of november playtest)
CL: [Improvement] The default tooltip is now more bound to the top of the screen.
CL: [Improvement] changed the font of the default tooltip.
CL: [Improvement] The classic tooltip's color is now bolder for people to be able to notice it more if they wish to enable the classic tooltip.

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Core/OptionsMenu/ThemeOptions/ThemeManager.cs
+++ b/UnityProject/Assets/Scripts/UI/Core/OptionsMenu/ThemeOptions/ThemeManager.cs
@@ -79,7 +79,7 @@ namespace Unitystation.Options
 
 	        if (PlayerPrefs.HasKey(PlayerPrefKeys.EnableClassicHoverTooltip) == false)
 	        {
-		        ClassicHoverTooltipToggle(false);
+		        ClassicHoverTooltipToggle(true);
 	        }
 
 	        ChatHighlight = PlayerPrefs.GetInt(PlayerPrefKeys.HighlightChat) == 1;
--- a/UnityProject/Assets/Scripts/UI/Core/OptionsMenu/ThemeOptions/ThemeManager.cs
+++ b/UnityProject/Assets/Scripts/UI/Core/OptionsMenu/ThemeOptions/ThemeManager.cs
@@ -7,6 +7,7 @@ using AddressableReferences;
 using Initialisation;
 using Shared.Util;
 using UnityEngine;
+using UnityEngine.UI;
 using Util;
 using YamlDotNet.RepresentationModel;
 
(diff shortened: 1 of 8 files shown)


Bod9001:
can you have the option to have the classic tooltip and the new one at the same time?

MaxIsJoe:
> can you have the option to have the classic tooltip and the new one at the same time?

no. Though I have to question, why would you need both at the same time?

Fr-Dae:
> 

why not ?
Redundancy is good. 
especially since we still don't know what new functionality is coming with the tooltips on top.

Bod9001:
it seems fine to have both

MaxIsJoe:
well this PR does not intend to add it, so maybe later with the rest of the customization options.

UnityProject/Assets/Scripts/UI/Core/OptionsMenu/ThemeOptions/ThemeManager.cs
```diff
@@ -74,6 +77,11 @@ void IInitialise.Initialise()
 		        PlayerPrefs.SetInt(PlayerPrefKeys.MentionSoundIndex, 0);
 	        }
 
+	        if (PlayerPrefs.HasKey(PlayerPrefKeys.EnableClassicHoverTooltip) == false)
+	        {
+		        ClassicHoverTooltipToggle(false);
```
corp-0: couldn't this just be
```cs
ClassicHoverTooltipToggle(PlayerPrefs.HasKey(PlayerPrefKeys.EnableClassicHoverTooltip))
```
No need for the if

MaxIsJoe: I'm enabling this by default for now for next playtest

UnityProject/Assets/Scripts/UI/Core/OptionsMenu/ThemeOptions/ThemeManager.cs
```diff
@@ -79,7 +79,7 @@ void IInitialise.Initialise()
 
 	        if (PlayerPrefs.HasKey(PlayerPrefKeys.EnableClassicHoverTooltip) == false)
 	        {
-		        ClassicHoverTooltipToggle(false);
+		        ClassicHoverTooltipToggle(true);
```
corp-0, line 82: this is way too odd. What if I see the option in my settings and enable it? If you want to force it enabled, just get rid of the if or something like that.

MaxIsJoe: ah, good point. we aren't updating `IsOn` directly so it wouldn't be updated on the user's end in the options menu

MaxIsJoe: it's now enabled by default in the options menu as well

Bod9001 (approved):

PR #9287, opened 2022-10-31
Fixes status display exploit

### LINKED ISSUE #9286: Status Display Concerns
### Environment
- Build Version: 22102800
- OS Version: Windows 10 19044.2130

Status displays do not show steps to finish building when right clicked, but more importantly can be used to open/close any working airlock or shutter from any range regardless of clearance level.
 
### Steps to reproduce

1. Use two metal to construct status display screen then mount it to a wall.
2. Right click the status display to see the next construction step.
3. Nothing can be observed.
4. To complete constructed status display add five cables, two sheets of glass, and use screwdriver.
5. To exploit the power of a status display simply sync it to a functioning airlock or shutter with a multitool, right click the status display, and use the cell timer. 
6. Synced items will close when timer is running and open when timer has stopped.
7. I have yet to find a functioning airlock/shutter it wouldn't open or a range limit.

MaxIsJoe:
Fixes #9286

CL: [Fix] Fixed exploit that would allow players to open doors without access cards via status displays.
CL: [New] Status displays now can have their access restriction setup upon creation. You cannot override already existing restrictions.

### DIFF
--- a/UnityProject/Assets/Scripts/Items/IDCard.cs
+++ b/UnityProject/Assets/Scripts/Items/IDCard.cs
@@ -359,6 +359,16 @@ public class IDCard : NetworkBehaviour, IServerInventoryMove, IServerSpawn, IInt
 		SyncJobTitle(jobTitle, newJobTitle);
 	}
 
+	public Clearance GetHighestClearance()
+	{
+		var highestClearance = Clearance.MaintTunnels;
+		foreach (var variabClearance in GetClearance)
+		{
+			if (variabClearance > highestClearance) highestClearance = variabClearance;
+		}
+		return highestClearance;
+	}
+
 	public string Examine(Vector3 pos)
 	{
 		return $"The account linked to the ID belongs to '{registeredName}' and reports a balance of {currencies[(int)CurrencyType.Credits]} credits. " +
--- a/UnityProject/Assets/Scripts/Items/IDCard.cs
+++ b/UnityProject/Assets/Scripts/Items/IDCard.cs
@@ -359,16 +359,6 @@ public class IDCard : NetworkBehaviour, IServerInventoryMove, IServerSpawn, IInt
 		SyncJobTitle(jobTitle, newJobTitle);
 	}
 
-	public Clearance GetHighestClearance()
-	{
-		var highestClearance = Clearance.MaintTunnels;
-		foreach (var variabClearance in GetClearance)
-		{
-			if (variabClearance > highestClearance) highestClearance = variabClearance;
-		}
-		return highestClearance;
-	}
-
 	public string Examine(Vector3 pos)
 	{
 		return $"The account linked to the ID belongs to '{registeredName}' and reports a balance of {currencies[(int)CurrencyType.Credits]} credits. " +
--- a/UnityProject/Assets/Scripts/Systems/Clearance/Clearance.cs
+++ b/UnityProject/Assets/Scripts/Systems/Clearance/Clearance.cs
@@ -1,7 +1,11 @@
 ﻿namespace Systems.Clearance
 {
+	/// <summary>
+	/// TODO: ORGANISE THESE FROM LOWEST TO HIGHEST NUMBER BASED ON CLEARANCE LEVEL.
+	/// </summary>
 	public enum Clearance
 	{
+		BasicPublicAccess = 0,
 		// Security equipment, security records, gulag item storage, secbots
 		Security = 1,
 		// Brig cells+timers, permabrig, gulag+gulag shuttle, prisoner management console
--- a/UnityProject/Assets/Scripts/Systems/Clearance/Clearance.cs
+++ b/UnityProject/Assets/Scripts/Systems/Clearance/Clearance.cs
@@ -1,8 +1,5 @@
 ﻿namespace Systems.Clearance
 {
-	/// <summary>
-	/// TODO: ORGANISE THESE FROM LOWEST TO HIGHEST NUMBER BASED ON CLEARANCE LEVEL.
-	/// </summary>
 	public enum Clearance
 	{
 		BasicPublicAccess = 0,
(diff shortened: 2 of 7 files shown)


UnityProject/Assets/Scripts/Items/IDCard.cs
```diff
@@ -359,6 +359,16 @@ public void ServerSetJobTitle(string newJobTitle)
 		SyncJobTitle(jobTitle, newJobTitle);
 	}
 
+	public Clearance GetHighestClearance()
+	{
+		var highestClearance = Clearance.MaintTunnels;
+		foreach (var variabClearance in GetClearance)
+		{
+			if (variabClearance > highestClearance) highestClearance = variabClearance;
```
corp-0: this assumes the order at which they are declared matters, it actually doesn't.

UnityProject/Assets/Scripts/Systems/Clearance/Clearance.cs
```diff
@@ -1,7 +1,11 @@
 ﻿namespace Systems.Clearance
 {
+	/// <summary>
+	/// TODO: ORGANISE THESE FROM LOWEST TO HIGHEST NUMBER BASED ON CLEARANCE LEVEL.
```
corp-0: ah just noticed this comment. Yeah, it isn't a good idea to shuffle the numeric values. Literally everything will break if you do it.

corp-0 (changes requested):

UnityProject/Assets/Scripts/Systems/Clearance/Clearance.cs
MaxIsJoe: yeah that's why I left a todo for now, we can deal with these later.

UnityProject/Assets/Scripts/Items/IDCard.cs
MaxIsJoe: left as a todo for later

UnityProject/Assets/Scripts/Systems/Clearance/Clearance.cs
corp-0: is it even a thing in ss13? Can they compare access like that?

UnityProject/Assets/Scripts/Items/IDCard.cs
corp-0: what is the intention of adding this method?

MaxIsJoe: Returns the highest clearance you have, right now it won't work well with security but for other roles I think it will work fine

UnityProject/Assets/Scripts/Systems/Clearance/Clearance.cs
MaxIsJoe: I mean.. It should? Honestly, I'm not sure.

MaxIsJoe: We still need to distinguish between more important levels of clearance so this is a thing that we'll need to work on at some point

corp-0: if it were part of the requirements, I would have designed a different system for clearance to make it comparable. What do we need to distinguish levels of clearance, as you put it, for? When do we need it?

MaxIsJoe: My suggestion for a new system would honestly involve reworking the entire clearance system again, not just making a new one. Clearance should be divided into groups, the higher the group in level the more access you are given to around the station rather than giving crew members individual access to each and every door.
so lets say group 00 doesn't let you access anything, group 01 lets you access public areas on the station, group 02 lets you access service job areas, group 03 lets you access medical areas, group 04 lets you access science areas, and the list goes on.
"but how are you going to prevent people from different jobs not access other people's workplaces just because they have a higher clearance?" Answer, we shouldn't. In the higher your clearance, the easier it should for you to access other areas to do your job or take over someone else's shift if a department is lacking manpower. 

This would solve our problem of not knowing when someone has a higher clearance than another player from a gameplay perspective, and we can further push this forward around many different mechanics.

From a gameplay perspective, this is a change that must be done due to how lowpop our servers are and it's easier to comprehend for newer players. From a technical perspective, it's easier to work with just numbers to assign access to people rather than long list of enums.

corp-0: interesting suggestion then but it has nothing to do in this PR. We should discuss it further in discord.

MaxIsJoe: Technically it does, it's just a bigger suggestion than just re-ordering the enum around then fixing all the prefabs / SOs.

MaxIsJoe: Both time-consuming still.

corp-0: remove from the PR for now.

MaxIsJoe: the comment?

corp-0: the method as well. It is inventing a new system we don't have. To fix the issue, it should communicate the door the same credentials that were presented to the display at the moment of interaction.

MaxIsJoe: hardcoded security clearance into this until we figure out something else.

Bod9001 (approved):

corp-0:
this wasn't ready to be merged, crap

MaxIsJoe:
wot?

corp-0:
I said I was going to make some changes, I was waiting for the weekend so I can have some time for it.

corp-0:
it is fine I guess. It is just that putting it on hold gave it more urgency for me.

PR #9291, opened 2022-11-03
Updates to Unity 2021.3.12f1 LTS

MaxIsJoe:
### Main Purpose
Updating the project to this version seems to slightly fix some issues with performance and import times. There have also been a lot of crash fixes that Unity has made that make the editor much more stable, especially when editing maps.

### Updated Packages

Some packages have been updated.

### New Packages

Unity at some point finally de-coupled the graphics shader package and the Post Processing package out of URP, meaning that we can finally install those without URP and fully utilise them in unitystation without upgraing our whole projects to be URP compatible.
I've also installed some of the new profiling and multiplayer tools Unity have finally fully released. 

### Removed Unity Visual Scripting

For security reasons, Visual Scripting is completely removed from this project.

### (Experimental) Upped the asset importer count from 2 to 4

There have been some issues that were reportedly fixed on Unity's end that should make these play better with older hardware. If you feel like this has slowed down unity instead of improve import times for you, please report once this has merged.

### Burst settings

Burst has had it's setting configured to favor speed for compiling and performance for server builds, this should improve things massively.


AtnerNT:
Does this perhaps fix the issue with editing lists in the inspector?

MaxIsJoe:
nope

Bod9001:
are you going to try out new Unity version or is good to merge?

MaxIsJoe:
I'd say make a build for tomorrow first after character attributes is merged then merge this

PR #9297, opened 2022-11-06
Adds shaders for x-ray, colourblind mode, blindness, fuzzy vision

Bod9001:
### Changelog:

CL: [New] Adds shaders for x-ray, colourblind mode, blindness, fuzzy vision
CL: [New] adds mutations for the corresponding  x-ray, colourblind mode, blindness, fuzzy vision
CL: [Improvement] implemented new system for tracking client side shaders and if they should be active or not


ThatDan123:
I just hope these shaders don't completely break on Mac and linux

Bod9001:
they shouldn't do, since I've done them a different way to prevent the flipping issue

PR #9312, opened 2022-11-12
May be good NAN prevention?

Bod9001:
May be good NAN prevention

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/GasMix.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/GasMix.cs
@@ -29,10 +29,10 @@ namespace Systems.Atmospherics
 		public float Temperature
 		{
 			get { return temperature; }
-			set { temperature = Math.Clamp(value, 0, Single.MaxValue); }
+			set { temperature = Math.Clamp(value, 0.001f, Single.MaxValue); }
 		}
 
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/GasMix.cs
```diff
@@ -29,10 +29,10 @@ public class GasMix
 		public float Temperature
 		{
 			get { return temperature; }
-			set { temperature = Math.Clamp(value, 0, Single.MaxValue); }
+			set { temperature = Math.Clamp(value, 0.001f, Single.MaxValue); }
```
corp-0, line 32: explain

Bod9001: I'm thinking maybe something tries to go 
n / temperature

AtnerNT:
A big issue with the NaN and inf exploits is gases behave very strangely if you have both water and a gas in the same pipe. Water gets in pipes from turbines and I'm assuming pipes dont have a way to handle both liquid and gas so that can be leveraged for the exploits. While this will solve temp / 0s I dont think it will fix the NaN exploits using water.
Might wanna talk to Butch for clarification how the NaN exploit is done

Bod9001:
yeah
I think I did a while ago, forgot though xD

PR #9321, opened 2022-11-21
Improvements, script refactoring, tooltip and bug fixes for Night vision

MaxIsJoe:
https://user-images.githubusercontent.com/34368774/203039690-e5d52032-a11a-42ed-b09f-8e745ec51e0c.mp4

Cleaned up the NVG script a lot and documented some stuff

CL: [New] Added tooltip data for night vision goggles.
CL: [Improvement] Improved visibility for the night vision goggles.
CL: [Improvement] The Night Vision shader now has minimal loss in detail in darker areas.
CL: [Fix] Fixed bug with night vision goggles where they would not disable correctly after removing them from your eyes.

### DIFF
--- a/UnityProject/Assets/Scripts/Clothing/Eyewear/NightVisionGoggles.cs
+++ b/UnityProject/Assets/Scripts/Clothing/Eyewear/NightVisionGoggles.cs
@@ -57,11 +57,16 @@ namespace Clothing
 		{
 			if (player == null) return false;
 			// Checks if it's not null and checks if NamedSlot == NamedSlot.eyes
-			if (player != null && player.PlayerScript.RegisterPlayer == pickupable.ItemSlot.Player &&
-			    pickupable.ItemSlot is { NamedSlot: NamedSlot.eyes })
-				return true;
+			return player.PlayerScript.RegisterPlayer == pickupable.ItemSlot.Player && IsInCorrectNamedSlot();
+		}
 
-			return false;
+		/// <summary>
+		/// Checks if the item is in the correct ItemSlot which is the eyes.
+		/// Automatically returns false if null because of the "is" keyword and null propagation.
+		/// </summary>
+		private bool IsInCorrectNamedSlot()
+		{
+			return pickupable.ItemSlot is { NamedSlot: NamedSlot.eyes };
 		}
 
 		void IItemInOutMovedPlayer.ChangingPlayer(RegisterPlayer HideForPlayer, RegisterPlayer ShowForPlayer)
@@ -140,6 +145,8 @@ namespace Clothing
 			// Don't do anything.
 			if (Camera.main == null ||
 			    Camera.main.TryGetComponent<CameraEffectControlScript>(out var effects) == false) return;
+			// If the item is not in the correct slot, ensure the effect is disabled.
+			if (IsInCorrectNamedSlot() == false) state = false;
 			// Visibility is updated based on the on/off state of the goggles.
 			// True means its on and will show an expanded view in the dark by changing the player's light view.
 			// False will revert it to default.
--- a/UnityProject/Assets/Scripts/Clothing/Eyewear/NightVisionGoggles.cs
+++ b/UnityProject/Assets/Scripts/Clothing/Eyewear/NightVisionGoggles.cs
@@ -14,7 +14,6 @@ namespace Clothing
 		private static readonly float defaultvisibilityAnimationSpeed = 0.85f;
 		private static readonly float revertvisibilityAnimationSpeed = 0.2f;
 		private static readonly Vector3 expandedNightVisionVisibility = new(25, 25, 42);
-		private static readonly Vector3 normalNightVisionVisibility = new(3.5f, 3.5f, 8);
 
 		private IClientSynchronisedEffect Preimplemented => this;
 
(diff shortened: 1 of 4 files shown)


MaxIsJoe:
myprisfinishedplsrev

PerfectTangent:
hm, what's up with the character sprites losing most of it's detail? Can we tweak it so that doesn't happen?

MaxIsJoe:
> hm, what's up with the character sprites losing most of it's detail? Can we tweak it so that doesn't happen?

what do you mean?

PerfectTangent:
Screenshot 2022-11-21 234221

The shader is nuking most of the details on the character sprite, making it much less legible.

UnityProject/Assets/Scripts/Clothing/Eyewear/NightVisionGoggles.cs
```diff
@@ -1,186 +1,200 @@
...
+			Preimplemented.ImplementationSyncOnPlayer(PreviouslyOn, CurrentlyOn);
+		}
 
-	private IClientSynchronisedEffect Preimplemented => (IClientSynchronisedEffect) this;
+		public void ApplyDefaultOrCurrentValues(bool def)
+		{
+			// (Max): I have no idea what "def" or "Default" means because the person who wrote this code
+			// didn't bother to document their shit or give this value a proper name that's easy to understand,
+			// I'm not obligated to do that job for them especially after they left this script in such a horrible state.
+			ApplyEffects(def);
```
Bod9001: I thought it would be quite self-explanatory that it would apply the default values For an effect e.g , when you ghost you don't want nightvision to be present because that's the default value, so that's what the 
ApplyEffects(Default ? DefaultVisionData : VisionData); 
does, 

ApplyDefaultOrCurrentValues 
this triggers when you enter your body and leave your body

Bod9001:
fix nightvision effect being active when ghosting
fix night vision effect being deactivated when returning to body

when wearing night vision goggles

MaxIsJoe:
> fix nightvision effect being active when ghosting fix night vision effect being deactivated when returning to body
> 
> when wearing night vision goggles

Done 👍

MaxIsJoe:
> Screenshot 2022-11-21 234221
> 
> The shader is nuking most of the details on the character sprite, making it much less legible.

Since this is written in hlsl.. no, i can't change it.

MaxIsJoe:
Actually wait, I could.

Whoever left the shader actually left some documentation explaining what it does, so I managed to change the values easily without testing and breaking it.

MaxIsJoe:
image

the loss of detail is now smaller, unfortunately you can't have a 100% lossless final image because the shader directly changes the colors of everything; so whiter colors are going to suffer regardless of what you do.

MaxIsJoe:
pr2changelog failed for some reason

UnityProject/Assets/Scripts/Clothing/Eyewear/NightVisionGoggles.cs
corp-0: this comment is such an example of bad comments that help no one

corp-0:
> pr2changelog failed for some reason

you can read the output in the check status. It failed because you have a mistype in one of your changes. The category is invalid.

UnityProject/Assets/Scripts/Core/Camera/CameraEffectControlScript.cs
```diff
@@ -27,9 +28,16 @@ public class CameraEffectControlScript : MonoBehaviour
...
+		
 		public void Awake()
 		{
 			LightingSystem = this.GetComponent<LightingSystem>();
+			if (minimalVisibilitySprite != null)
+			{
+				MinimalVisibilityScale = minimalVisibilitySprite.transform.localScale;
+				return;
+			}
+			Logger.LogWarning("[CameraEffectControlScript] - visibilitySprite is null! please set it from the inspector.");
```
corp-0, line 40: I think you can add the ``[Required]`` attribute from NaughtyAttributes and have a big red warning in inspector.

Bod9001 (approved):
the function still rubs me the wrong way but, it's simplified to a bool so idk

UnityProject/Assets/Scripts/Core/Camera/CameraEffectControlScript.cs
MaxIsJoe: I think some things like xenos don't actually require it because it messes up some of their effects, so I don't think it should be a requirement for now.

MaxIsJoe:
> the function still rubs me the wrong way but, it's simplified to a bool so idk

which one?

UnityProject/Assets/Scripts/Clothing/Eyewear/NightVisionGoggles.cs
MaxIsJoe: Environmental story telling

Bod9001:
ApplyEffects(!Default);
vs
ApplyEffects(Default ? defaultValueForShaders : OnValueForShaders );
doesn't really matter is same both ways

MaxIsJoe:
> ApplyEffects(!Default); vs ApplyEffects(Default ? defaultValueForShaders : OnValueForShaders ); doesn't really matter is same both ways

Just using pure syncvars is cleaner imo

Bod9001:
sadly you can't use synchvars,  
the problem is if you have synchvars, by the time the hook fires on the synchvar for it turning off, it doesn't know what body it is supposed to be on so, which client to turn off night vision for

the solution I thought of was,  you look through everything the client owns on the client and told that you've left the body so don't apply the values any more

MaxIsJoe:
> sadly you can't use synchvars, the problem is if you have synchvars, by the time the hook fires on the synchvar for it turning off, it doesn't know what body it is supposed to be on so, which client to turn off night vision for
> 
> the solution I thought of was, you look through everything the client owns on the client and told that you've left the body so don't apply the values any more

As far as I know, from my own testing; there is nothing wrong with this approach.

Bod9001:
with the synchvar?
it's also a bit of anti-pattern since the synchvar is getting used for is it a valid set up , But then also has the player left or entered?

MaxIsJoe:
> But then also has the player left or entered?

Isn't that the point of your effect sync system?

Issue #9178, opened 2022-08-29
[Rework Needed] - Gamify the health system and modularize it


MaxIsJoe:
After the last few health simplification PRs we have gained a boost in performance that has managed to reduce the load on the servers finally but there is still a lot more to be desired from a development and game design standpoint. As of the time of writing this issue the health script is over 1k lines of code and houses a lot of systems and features that are still too complicated or too demanding/complex for anyone to work on and some of these things are absolutely not desired from the perspective of the player or developer currently.

We need to sit down and rethink the entire system so that it is approached in a gamified manner and not a system that attempts to simulate how the biological bodies work in real life. Gamifying our systems allow developers to approach the code more easily and reduces the likelyhood of bugs happening because of the simplified approach that we should take.

This rework can be looked at after the mind rework is done, until then we must discuss and document how we want this rework to be laid out on discord in the #health channel.

Once we start development on this rework we should follow the same philosophy that we're using with Signals and the Mind Rework and make sure that everything we change is done in small chunks under multiple PRs as it lets us quickly discover game breaking bugs and take care of it before working on anything else. This is important so we don't repeat the HealthV2 incident.
Also, this rework should not be called HealthV3 and we should remove the old V1 code. One Health, Works everywhere.

## What is needed for this rework :

- Health must be divided into multiple systems that only affect general info on the main health controller script. (Modularization)
- Systems must be independent of each other and can be added/removed whenever needed. (Modularization)
- Systems must be simple and easy to approach from both the player's and developer's view. (Simple code)
- Systems must not attempt to simulate anything. (No complex systems or calculations/state checks via periodic updates)
- Systems must only be intractable via events. (Modularization)
- Systems must not contain any features that do not have a general purpose and must have a good gameplay reason for its existence.
- Health and its systems must be universal, so it can be attached to any mob, item or object.
- Document all code thoroughly and update the developer docs to include a page dedicated to how health works, how each system operates and an example for how to properly add a new system for a specie(s).

Bod9001:
"Systems must not attempt to simulate anything. (No complex systems or calculations/state checks via periodic updates)"

that's actually the entirety of SS13 you removed there, might as well just remove atmospherics just make it so" hole in station = player get Pushed to it"


"Systems must only be intractable via events. (Modularization)"  I do get what this means?

"Health must be divided into multiple systems that only affect general info on the main health controller script. (Modularization)"

 like there is issue with Respiratory system and circulatory system, 
 
[...]

PR #8830, opened 2022-06-14
movement documentation

Bod9001:
can someone have a look over to see if there's anything I missed

Issue #8529, opened 2022-03-30
[Refactor Needed] Player Body sprites need to stop using strings for referencing


MaxIsJoe:
### Environment
- Game Version: 30-3-2022 Upstream/Develop
- OS Version: Win 10

### Detailed description of bug

Currently, character settings use strings to reference body sprites for hair, beards, underwear, tails, wings, etc. and the only way you can set those strings in a readable way is via the customization screen only.
This is prone to breaking if we ever change anything about how these strings are read/written and they're not entirely human readable either and it also limits programmers in a million ways; from not being able to randomly generate these sprites for dummy and NPC characters and not being able to change these sprites in real time (i.e : using something like the mirror from ss13 that lets you change your hair style for example).

The problem mainly hits these two the most 
![image](https://user-images.githubusercontent.com/34368774/160854732-b2cbfdc0-6bd8-408b-a34b-3a304452a417.png)

and they use data that can't be manually set because it requires something like the player character customization UI to set it.

![image](https://user-images.githubusercontent.com/34368774/160889488-12d5d7b8-050a-470f-aaae-eba7072a9211.png)

### Possible solution 

Like SpriteSOs, we need a catalogue of customizations that we can easily reference in the inspector and we need an API that lets us change/add/remove sprites using these SOs on a body part.

Bod9001:
the problem is you to specify what body part your editing, 

 for example let's say you can modify your left and right arm to be cybernetic,
 
 how do you tell the difference between left and right, While keeping the same code for them?

MaxIsJoe:
what does that have to do with the fact that the system currently is near impossible to do anything with unless it's through the player's character customizer because these strings cannot be human writable and we have zero clue how the game even gets them?

MaxIsJoe:
> how do you tell the difference between left and right, While keeping the same code for them?

Like I have explained in the bottom of this issue, we can create a catalogue for body parts and we can assign these sort of details to them.
But please still explain to me what do you mean by this so that i or whoever in the future works on this knows what exactly you're thinking or whats your understanding of this current system.

Bod9001:
so for example let's say you have a customisable body part on each hand, 

and the hand is within the arm, 

 from the point of view of the code both hands are technically a same, so to differentiate them you say the path of where the hand is located within the body
 
 So 
 Left arm > Hand {data}
 Right arm > Hand {data}

PR #9364, opened 2022-11-22
Expands upon development standards doc

MaxIsJoe:
Title.

A lot of the new items has explanations for why we should and shouldn't do a thing, and it comes with videos that explain how to properly do stuff in detail without over-bloating the doc.

### DIFF
--- a/docs/contribution-guides/Development-Standards-Guide.md
+++ b/docs/contribution-guides/Development-Standards-Guide.md
@@ -250,13 +250,13 @@ Unity has a great talk on how to do just that in this video: [How to adopt a mod
 
 Below are good practices and explainations on what to do, so you can avoid having your code refactored/reworked 5 months later and make the lives of other contributors much easier.
 
-## 0. In-code documentation
+## 0. In-code documentation and clarity
 
 * You can avoid confusing others (and even yourself) by naming your functions and variables in a clear and descriptive way.
 
 ```cs
 //bad code
-void doThing(bool b)
+private void DoThing(bool b)
 {
     state = b;
 }
@@ -264,41 +264,44 @@ void doThing(bool b)
 
 ```cs
 //good code
-private void ThisFunctionWillDoA_B_And_C(bool newState)
+private void UpdateState(bool newState)
 {
     state = newState;
 }
 ```
 
-* Explain what your function does step by step whenever possible. Your code might seem obvious to you, but for other people; that may not be the case.
+* Avoid using extremely shortened names whenever possible.
 
 ```cs
-//Example from NightVisionGoggle.cs
+//bad name.
+var DIS = GetComponent<DynamicInventorySystem>();
+```
 
-/// <summary>
-/// Checks if the item is in the correct ItemSlot which is the eyes.
-/// Automatically returns false if null because of the "is" keyword and null propagation.
-/// </summary>
-private bool IsInCorrectNamedSlot()
-{
-	return pickupable.ItemSlot is { NamedSlot: NamedSlot.eyes };
-}
+```cs
+//good name.
+var inventory = GetComponent<DynamicInventorySystem>();
+```
 
-private void ApplyEffects(bool state)
-{
-	var finalState = state;
-	// If for whatever reason unity is unable to catch the correct main camera that has the CameraEffectControlScript
-	// Don't do anything.
-	if (Camera.main == null ||
-		Camera.main.TryGetComponent<CameraEffectControlScript>(out var effects) == false) return;
-	// If the item is not in the correct slot, ensure the effect is disabled.
-	if (IsInCorrectNamedSlot() == false) finalState = false;
-	// Visibility is updated based on the on/off state of the goggles.
-	// True means its on and will show an expanded view in the dark by changing the player's light view.
-	// False will revert it to default.
-	effects.AdjustPlayerVisibility(finalState);
+* Explain things like long LINQ expressions, uncommon C# syntax and RegEx.
+
+LINQ and Regex are extremely useful but they can slow down everyone sometimes whenever they have to work on/with it. Having comments in code that explains what it does can save a developer minutes (if not hours) of their life to be productive elsewhere.
+
+Reminder, just because your code might seem obvious to you; doesn't mean that it's obvious to everyone else.
+
+```cs
+// Linq expression that handles grabbing multiple names from a prefab.
+// it grabs all prefabs in documents then loops through all prefabs and grabs all searchable names.
+// if the searchable name contains a substring that the user is searching it will return it.
+var docs = (from doc in documents from prefabNames in doc.SearchableName
+	where prefabNames.Contains(standardizedSearch) select doc).ToList();
 }
 ```
+
+* Make use of the [Unitystation developer docs](https://unitystation.github.io/unitystation/Documentation-Guide/) and document features and systems there.
+
+It's widely agreed everywhere that excessive commenting inside of code can make it look ugly. As a result, we move all or in-depth explanations of how things work to the developer docs.
+
+Whenever you finish working on a new feature or system, always try your best to write/update a page that explains what your changes does in detail.
 
 
 * Make proprer use of `Logger`
@@ -318,9 +321,9 @@ Logger.Log("something happened");
 ```cs
 //good use of logger.
 
-Logger.Log("[Manager/OnRoundEnd/Cleanup] - Manager cleaned references succesfully.");
-Logger.LogError($"[Item/ServerPerformInteraction] - Tried applying item on {gameObject.name} but it was missing X!");
-Logger.LogWarning("[Object/OnEnable] - Integer values appear to be too high, unintended behavior may occur.");
+Logger.Log("[Manager/OnRoundEnd/Cleanup] - Manager cleaned references succesfully.", Category.DebugConsole);
+Logger.LogError($"[Item/ServerPerformInteraction] - Tried applying item on {gameObject.name} but it was missing X!", Category.Interactions);
+Logger.LogWarning("[Object/OnEnable] - Integer values appear to be too high, unintended behavior may occur.", Category.Atmospherics);
 ```
 
 
@@ -414,17 +417,17 @@ class PlayerInventory : Inventory
 
 _Note: Events can leak! Always remember to unsubscribe your events on `OnDisable()`._
 
-* Minimise Polymorphism, Embrace Interfaces.
+* Minimise inheritance, Embrace Interfaces.
 
-[Interface](https://www.w3schools.com/cs/cs_interface.php) in C# is a blueprint of a class. It is like abstract class because all the methods which are declared inside the interface are abstract methods. It cannot have method body and cannot be instantiated. It is used to achieve multiple inheritance which can't be achieved by class.
+[Interface](https://www.w3schools.com/cs/cs_interface.php) in C# is a blueprint of a class. It is like abstract class because all the methods which are declared inside the interface are abstract methods. It cannot be instantiated, but it can hold optional default implamentations. It is used to achieve multiple inheritance which can't be achieved by class.
 
-Interfaces allow us to easily communicate features between objects and managers by keeping design/implamentation contained in predictable states and architecture while allowing us to easily expand, maintain and iterate upon them.
+Interfaces allows to share behaviour with classes in a horizontal manner, contrary to inheritance where the sharing is vertical (from parent to children). In horizontal, there is no need for a parental relationship between the classes. 
 
 _CodeMonkey has a great video on how to utilise interfaces for a modular workflow: [Modular Character System in Unity](https://youtu.be/mJRc9kLxFSk)_
 
-That being said, [Polymorphism](https://www.w3schools.com/cs/cs_polymorphism.php) is not something to be afraid of. Polymorphism makes sense to be used while in the right context, which is mostly going to be in ScriptableObjects and scanerios where you want to work with common data that will be shared around. 
+That being said, Concepts such as [Polymorphism](https://www.w3schools.com/cs/cs_polymorphism.php) is not something to be afraid of. Inheritance makes sense to be used while in the right context, which is mostly going to be in ScriptableObjects and scanerios where you want to work with common data that will be shared around. 
 
-So, use Polymorphism for data driven scenarios while keep systems and functionality bound to C# Interfaces.
+So, use inheritance for data driven scenarios while keep systems and functionality bound to C# Interfaces.
 
 
 * Reduce duplicate code.
@@ -434,16 +437,19 @@ If you've seen something that has been done more than twice in the project, writ
 ```cs
 // bad code.
 
-classA : MonoBehavior{
-    void DoX_InTransform();
+class ClassA : MonoBehavior
+{
+    public void DoX_InTransform();
 }
 
-classB : MonoBehavior{
-    void DoX_InTransform();
+class ClassB : MonoBehavior
+{
+    private void DoX_InTransform();
 }
 
-classC : MonoBehavior{
-    void DoX_InTransform();
+class ClassC : MonoBehavior
+{
+    public void DoX_InTransform();
 }
 ```
 
--- a/docs/contribution-guides/Development-Standards-Guide.md
+++ b/docs/contribution-guides/Development-Standards-Guide.md
@@ -425,7 +425,7 @@ Interfaces allows to share behaviour with classes in a horizontal manner, contra
 
 _CodeMonkey has a great video on how to utilise interfaces for a modular workflow: [Modular Character System in Unity](https://youtu.be/mJRc9kLxFSk)_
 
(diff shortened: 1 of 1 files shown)


MaxIsJoe:
I think I've gotten all spelling errors and pacing issues, should be a lot cleaner now.

review pls.

docs/contribution-guides/Development-Standards-Guide.md
```diff
@@ -226,3 +242,301 @@ to just leave it as a public field without a property. As always, make sure to d
...
+```cs
+//Example from NightVisionGoggle.cs
+
+/// <summary>
+/// Checks if the item is in the correct ItemSlot which is the eyes.
+/// Automatically returns false if null because of the "is" keyword and null propagation.
+/// </summary>
+private bool IsInCorrectNamedSlot()
+{
+	return pickupable.ItemSlot is { NamedSlot: NamedSlot.eyes };
```
Bod9001: is stuff might be confusing for new devs

MaxIsJoe: I don't think it can be confusing because the whole point of this section is to demonstrate that comments allow everyone to understand what exactly the code is doing.

Bod9001: Could work, but also brings up a good point try not to use code that is too confusing in the first place

MaxIsJoe: But this code isn't confusing to start with, and the quirks of it are explained in the docs + the whole point of this section is to tell people the usefulness of code docs.

Bod, respectfully. You're the last one who should have a say or talk about any of this, since the whole reason I've added this was because of **you** specifically.

*Edit: and old doobly code.. but still. my point stands. You guys shouldn't really have a say on this when you've given me such a hard time with code/communication for the past two years*

docs/contribution-guides/Development-Standards-Guide.md
```diff
@@ -226,3 +242,301 @@ to just leave it as a public field without a property. As always, make sure to d
...
+//bad code
+void doThing(bool b)
+{
+    state = b;
+}
+```
+
+```cs
+//good code
+private void ThisFunctionWillDoA_B_And_C(bool newState)
```
corp-0: I dont think this is a great example of a good name. Given the body of the method, call it something like "UpdateState" I guess.

docs/contribution-guides/Development-Standards-Guide.md
```diff
@@ -226,3 +242,301 @@ to just leave it as a public field without a property. As always, make sure to d
...
+{
+    state = newState;
+}
+```
+
+* Explain what your function does step by step whenever possible. Your code might seem obvious to you, but for other people; that may not be the case.
```
corp-0: Jesus no, we don't need a comment every step. In the summary it is fine, but only required for public methods. A private method could have a summary if the thing it does is particularly obscure

docs/contribution-guides/Development-Standards-Guide.md
```diff
@@ -226,3 +242,301 @@ to just leave it as a public field without a property. As always, make sure to d
...
+To make our lives easier as well, add the class/gameObject name and function in the log message so it's easier to tell ourselves what went wrong from a surface.
+
+_Note: Avoid using unity's `Debug` class for logging as it doesn't store logs where we need them to be._
+```cs
+//bad use of logger.
+
+Logger.Log("something happened");
+```
+```cs
+//good use of logger.
```
corp-0, line 322: The examples of good use arent using the logger categories

docs/contribution-guides/Development-Standards-Guide.md
```diff
@@ -226,3 +242,301 @@ to just leave it as a public field without a property. As always, make sure to d
...
+
+class PlayerInventory : Inventory
+{
+    //same thing as in DeathUI, just subscribe to the OnDeath event.
+}
+```
+
+_Note: Events can leak! Always remember to unsubscribe your events on `OnDisable()`._
+
+* Minimise Polymorphism, Embrace Interfaces.
```
corp-0: You mean inheritance, not polymorphism.

docs/contribution-guides/Development-Standards-Guide.md
```diff
@@ -226,3 +242,301 @@ to just leave it as a public field without a property. As always, make sure to d
...
+
+[Interface](https://www.w3schools.com/cs/cs_interface.php) in C# is a blueprint of a class. It is like abstract class because all the methods which are declared inside the interface are abstract methods. It cannot have method body and cannot be instantiated. It is used to achieve multiple inheritance which can't be achieved by class.
```
corp-0: Interface can have default method implementation

docs/contribution-guides/Development-Standards-Guide.md
```diff
@@ -226,3 +242,301 @@ to just leave it as a public field without a property. As always, make sure to d
...
+
+Interfaces allow us to easily communicate features between objects and managers by keeping design/implamentation contained in predictable states and architecture while allowing us to easily expand, maintain and iterate upon them.
```
corp-0: Allows to share behaviour with classes in a horizontal manner, contrary to inheritance where the sharing is vertical (from parent to children). In horizontal, there is no need for a parental relationship between the classes. Not sure how you got to the "predictable state" part.

docs/contribution-guides/Development-Standards-Guide.md
```diff
@@ -226,3 +242,301 @@ to just leave it as a public field without a property. As always, make sure to d
...
+
+_CodeMonkey has a great video on how to utilise interfaces for a modular workflow: [Modular Character System in Unity](https://youtu.be/mJRc9kLxFSk)_
+
+That being said, [Polymorphism](https://www.w3schools.com/cs/cs_polymorphism.php) is not something to be afraid of. Polymorphism makes sense to be used while in the right context, which is mostly going to be in ScriptableObjects and scanerios where you want to work with common data that will be shared around.
```
corp-0: Again, you meann inheritance. Classes composed by interfaces are polymorphic as well.

docs/contribution-guides/Development-Standards-Guide.md
```diff
@@ -226,3 +242,301 @@ to just leave it as a public field without a property. As always, make sure to d
...
+
+* Reduce duplicate code.
+
+If you've seen something that has been done more than twice in the project, write a common class or class extension that allows us to reduce the likelyhood of having to maintain duplicate code that all do the same thing in multiple areas.
+
+```cs
+// bad code.
+
+classA : MonoBehavior{
+    void DoX_InTransform();
```
corp-0: Avoid improper naming convention in examples, even if it is the bad example when we are not talking about naming conventions.

docs/contribution-guides/Development-Standards-Guide.md
```diff
@@ -226,3 +242,301 @@ to just leave it as a public field without a property. As always, make sure to d
...
+* Use Time-Splitting whenever your iterators/loops are attempting to go through a large amount of objects/tasks.
+```cs
+//How to time split.
+
+var currentIndex = 0;
+var maximumIndexesPerFrame = 20;
+
+foreach(var item in LargeList)
+{
+    if(currentIndex >= maximumIndexesPerFrame)
```
corp-0, line 502: We have an extension for enumerables to work in chunks

docs/contribution-guides/Development-Standards-Guide.md
MaxIsJoe: I honestly don't even know whats the point of categories, it just seems like a way to silently hide logs since most of the logs under categories aren't even visible to the client.

docs/contribution-guides/Development-Standards-Guide.md
MaxIsJoe: True. I only thought about polymorphism during that instance so my brain defaulted to that. Changing.

docs/contribution-guides/Development-Standards-Guide.md
corp-0: Don't they get written to the log file? Categories aren't meant for the client.

MaxIsJoe: Logger prints to the console and discord, if it's hidden behind a category it wouldn't alarm anyone about it, and we'd need to go through long log files to find what we need.
Ask yourself this, when was the last time we, as the developers, looked at a log file directly or asked one of our players to send their entire log file?

docs/contribution-guides/Development-Standards-Guide.md
MaxIsJoe: what is it called?

docs/contribution-guides/Development-Standards-Guide.md
corp-0: When you are only interested in reading a category, it is good to have the option.

Reading logs on discord is a misbehaviour of us, that really needs to change.

MaxIsJoe: Alright then, will update to change it.

docs/contribution-guides/Development-Standards-Guide.md
MaxIsJoe: I actually agree, changing.

docs/contribution-guides/Development-Standards-Guide.md
corp-0: I will look for it after lunch

docs/contribution-guides/Development-Standards-Guide.md
MaxIsJoe: I'm sorry, but this one I will have to fight for it to stay.

Step-by-step documentation should be mandatory, since I don't want to constantly scratch my head whenever I want to work on anything.
Not only is it good practice that can make the maintainer's life easier when reviewing PRs, but it's also good for when someone wants to touch a new script that's alien to them in the future. Especially if that script is in a complicated system like Electricity, Movement, Atmos, etc.

Like I've said, your code might seem obvious to **you**, but for other people that may not be always the case.

This is also important when you have people like Bod or Dae who do not possess the best communication skills and also forces you, the person making the changes, to do a Rubber duck debugging session with yourself to catch on any problems you might have missed while working on your code.

corp-0: I will try to be precise and concise:

### I . **Too many comments in code is actually a detriment to understanding the code**.

If you document literally every step, like you are asking here, I will expend more time reading comments about how something is supposed to work instead of how it is actually working. I could ignore the logic presented and believe the comment, when that might not actually be the truth. And I should also mention the clutter.

### II . **The need for a lot of comments is actually hiding bad code**.

If you feel the need to document every step, it could be that the algorithm you have written is simply hard to understand and isn't using proper names for variables/methods. When code is well written, you can understand it by simply reading it. Naming methods is a powerful way to let people know what the algorithm is supposed to do. As a general rule, you should let the code itself say the ``how`` and if it is really necessary, you can add a ``why`` in comments.

### III . **Unitystation as a learning experience**.

Unitystation is indeed a powerful learning experience, but it shouldn't be the learning experience for someone who doesn't know programming at all. We shouldn't add comments like
```cs
// returns early if something is null
if (something == null) return;
```
If that explanation was needed, then the person should first try in smaller projects of their own to get a grasp on programming before trying to contribute to Unitysation.

Also take in cosideration the **source code** is for developers, people who aren't writing code shouldn't need to read it. We have other instances for those cases.

### IV . **If extensive explanation of a system is needed, the source code isn't the place for that explanation**

In some cases, the systems we are building can be complex enough that a thorough explanation is required. You won't understand the system by reading literally every comment we wrote line by line of code in the system. We have a proper development wiki for that pupose where a well thought documentation can be written so the reader gets a birdeye view of the system and then a dense explanation of all its parts, including examples, etc.

### V . **Literally no one ever has ever needed commenting every step. We would be the first codebase to do that**.

And I don't mind being pioneer, but in this case it is simply because it is a wrong move. If you lurk around, you will find that most people when talking about "too many comments" are referring to summaries of algorithms, commenting literally every step would be even worse than "too many comments".

I will leave you some random articles I found on google so you can get a better written explanation on the concept:

[Portrait of a n00b](https://steve-yegge.blogspot.com/2008/02/portrait-of-n00b.html)
[Fighting Evil in Your Code: Comments on Comments](https://www.red-gate.com/simple-talk/opinion/opinion-pieces/fighting-evil-code-comments-comments/)
[You’re commenting your code too much (and other controversial thoughts on documentation)](https://levelup.gitconnected.com/youre-commenting-your-code-too-much-and-other-controversial-thoughts-on-documentation-1ee617ed46af)

MaxIsJoe: > I . Too many comments in code is actually a detriment to understanding the code.

I can make a case and say that we don't need to do it all the time, but we still should enforce it in a lot of cases.

> II . The need for a lot of comments is actually hiding bad code.

Yeah, no shit? 

This has been a thing that I've been complaining about earlier than #8529 and has gotten worse with #9178 when our lead developer literally **threatened** to veto contributions that aim to **improve** bad code because he's stubborn and does not want anyone to challenge things he poorly implemented, and he still continues to push code that is difficult to understand and work with until as recently as effect syncs with #9321 .

Since no one wanted to make an effort to improve the code quality recently after all my begging to do so and I still have to re-write scripts to improve its general use and understanding; then at least explain to me what the hell your bad code does in the first place.

~~I don't understand why this is such a difficult thing to ask, either treat Bod's PRs with the same level of annoying semi-strictness as me when it comes to my PRs or make it mandatory for him to **at least** make the effort into explaining what his bad code does.~~

> Unitystation as a learning experience.

Like I've stated above, I'm willing to reduce how mandatory it is to document steps that don't require docs to start with; but we have to still tell people to explain what their code does when it starts to get technical or large.

a good example:
```cs
/// <summary>
/// Checks if the item is in the correct ItemSlot which is the eyes.
/// Automatically returns false if null because of the "is" keyword and null propagation.
/// </summary>
private bool IsInCorrectNamedSlot()
{
	return pickupable.ItemSlot is { NamedSlot: NamedSlot.eyes };
}
```

> IV . If extensive explanation of a system is needed, the source code isn't the place for that explanation

In any other circumstance I would have agreed with this because even I complained to SS14 devs about them not having any sort of extensive developer docs like [we do](https://unitystation.github.io/unitystation/development/Helpful-Functions-Cheatsheet/) 

But all I have to point to is #8830 to show you how this is going to be a failed thing to push.

"So, you want us to do [bad thing] instead??"

Yes, since there is no viable alternative; I'd rather that we right a wrong with another wrong.

> Literally no one ever has ever needed commenting every step. We would be the first codebase to do that.

I agree with this, but I still want to push this standard in some form of way or another because this is important to the codebase's quality in the future since we shouldn't have only one person who understands how everything works in this project and does not want to cooperate with you fully to add even the most basic of features sometimes.

If you want to discuss the best middle ground we can have, we can discuss that on discord.

_Edit: I think I may have exaggerated the strictness part a little bit, but my point still stands that we need more quality control over the things we merge._

docs/contribution-guides/Development-Standards-Guide.md
corp-0: https://github.com/unitystation/unitystation/blob/d5679c962287120d80dbe627452903d78ade4628/UnityProject/Assets/Scripts/Util/SweetExtensions.cs#L389

MaxIsJoe: I don't think it awaits a frame like I've written here.

docs/contribution-guides/Development-Standards-Guide.md
MaxIsJoe: Done, review that again please.

docs/contribution-guides/Development-Standards-Guide.md
MaxIsJoe: did i miss anything else?

docs/contribution-guides/Development-Standards-Guide.md
```diff
@@ -226,3 +242,307 @@ to just leave it as a public field without a property. As always, make sure to d
...
+
+* Minimise inheritance, Embrace Interfaces.
+
+[Interface](https://www.w3schools.com/cs/cs_interface.php) in C# is a blueprint of a class. It is like abstract class because all the methods which are declared inside the interface are abstract methods. It cannot be instantiated, but it can hold optional default implamentations. It is used to achieve multiple inheritance which can't be achieved by class.
+
+Interfaces allows to share behaviour with classes in a horizontal manner, contrary to inheritance where the sharing is vertical (from parent to children). In horizontal, there is no need for a parental relationship between the classes. 
+
+_CodeMonkey has a great video on how to utilise interfaces for a modular workflow: [Modular Character System in Unity](https://youtu.be/mJRc9kLxFSk)_
+
+That being said, Concepts such as [Polymorphism](https://www.w3schools.com/cs/cs_polymorphism.php) is not something to be afraid of. Inheritance makes sense to be used while in the right context, which is mostly going to be in ScriptableObjects and scanerios where you want to work with common data that will be shared around.
```
corp-0: you are still linking to polymorphism here. The concept you should not be afraid of is inheritance. Both interfaces and class inheritance achieve polymorphism.

MaxIsJoe: Done 👍

corp-0 (approved):

PR #9373, opened 2022-11-23
Server Password

### LINKED ISSUE #9371: graphic interface to host private servers
in the idea of creating a private server, it would be interesting to have an option where, to join, a password, defined in the configuration of the server (thus with the ingame admin interface, and not a file to be edited)

This way, even if an IP is broadcasted by mistake, only the people with the IP + Password will be able to join.

ThatDan123:
CL: [New] Servers can now be password protected, can be set in config or in admin UI

Fixes #9371


Fr-Dae:
Have you a screenshot of the UI please ?
and thanks you for tthis

corp-0 (approved):

PR #9385, opened 2022-11-27
Protip behavior improvements and options placement change

MaxIsJoe:
CL: [New] Added "ShowFirstTimeExpScreen" and "clearProtips" as console commands.
CL: [Improvement] The player experience dropdown has been moved from the Help menu to the Options Screen.
CL: [Improvement] To avoid further confusion and protips spam, players with the lowest experience level will no longer have a special exception to the "Trigger Once" rule.
CL: [Fix] Fixed the pro-tips related to item traits not spawning correctly at random due to inconsistent spawning initialization times. 
CL: [Fix] The protips list is now fixed finally and will display correctly.


UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
```diff
@@ -74,6 +75,18 @@ public static void CloneSelf()
 			playerBody.ApplyDamageAll(null, 2, AttackType.Internal, DamageType.Clone, false);
 		}
 
+		[ConsoleMethod("clearProtips", "Clears all saved states of protips.")]
+		public static void ClearAllProtips()
+		{
+			ProtipManager.Instance.ClearSaveState();
+		}
+
+		[ConsoleMethod("ShowFirstTimeExpScreen", "Shows the player experience screen.")]
```
corp-0: We have to decide on a naming convention for the commands. 
Currently there seems to be all of the following coexisting:
- ``kebab-case``
- ``PascalCase``
- ``lowercase``
- ``camelCase`` 

This is only confusing and users will have to remember which naming convention each command is using.

MaxIsJoe: > We have to decide on a naming convention for the commands.

no, we do not.

MaxIsJoe: we do need it. all lower case seems better imo

corp-0: ``alllowercase`` can get messy when there are many words. I think ``kebab-case`` looks good and is also the convention you used in the other pr.

MaxIsJoe: done

UnityProject/Assets/Scripts/Learning/ProtipObject.cs
```diff
@@ -16,21 +16,31 @@ public class ProtipObject : MonoBehaviour
...
 			}
-			if(CheckSaveStatus()) Destroy(this);
+
+			if (CheckSaveStatus()) RemoveThisComponent();
+		}
+
+		/// <summary>
+		/// Ensures that we don't remove any components from the player gameObject.
+		/// </summary>
+		private void RemoveThisComponent()
```
corp-0, line 39: I don't get it, from where is it destroyed? Also what issues did the previous flow cause?

UnityProject/Assets/Scripts/Learning/ProtipObjectTypes/ProtipObjectOnPickupTrait.cs
```diff
@@ -11,8 +11,11 @@ public class ProtipObjectOnPickupTrait : ProtipObject
 		[FormerlySerializedAs("ProtipsForTraits"), SerializeField]
 		private SerializableDictionary<ItemTrait, ProtipSO> protipsForTraits = new SerializableDictionary<ItemTrait, ProtipSO>();
 
-		private void OnEnable()
+		private IEnumerator OnEnable()
 		{
+			//(Max): For some reason, the spawning times will become inconsistent between frames.
+			//To play it say, we intentionally await 2 full seconds until everything has run its course.
+			yield return WaitFor.Seconds(2f);
```
corp-0, line 18: is this an initialisation order issue? Because we have that initialsiation interface to make sure things work a set order.

UnityProject/Assets/Scripts/Learning/ProtipObject.cs
MaxIsJoe: > where is it destroyed

It's literally written in the comment, we don't want to destroy any protip components while they're on the player.

UnityProject/Assets/Scripts/Learning/ProtipObjectTypes/ProtipObjectOnPickupTrait.cs
MaxIsJoe: not order, timing.

MaxIsJoe: This is one of those mind rework things that I don't get why it happens, and Bod caused it and didn't bother fixing one of his issues that he created. This wasn't a hidden thing either; this is an error you always see at the bottom of the editor whenever you start the game when this fix isn't here.

This fix works well, so I see no point in trying to correct it.

UnityProject/Assets/Scripts/Learning/ProtipObject.cs
MaxIsJoe: Protips can be applied on basically any prefab
but some protips are on the player object, all the player ones have multiple tips on them so to ensure that the protip doesn't delete itself while the "triggerOnce" variable is on we check if we're on a player before deleting.

corp-0: > Protips can be applied on basically any prefab but some protips are on the player object

this is the piece of context I was missing.

corp-0 (approved):

PR #9402, opened 2022-11-29
chore: moved issue templates to github forms

corp-0:
### Purpose
Replaces our current issue templates with github form templates so they are less likely to be ignored and ensure issue quality.

### Notes:
The fields I added can be discussed further with the engineers if needed.


Bod9001:
so, how does it work?

corp-0:
sample-issue-form
they are fields in a form to fill instead of just a text area where you can ignore the template.

Bod9001:
Game Version could be date?

OS I don't think is required? , 
It should be there but I don't think it needs to be required

What happened? and 
What should have happened instead?

I feel like could be in scenarios where a bug wouldn't need them but they are required


blank_issues_enabled: false?
I assume this still allows admins to write more detailed stuff e.g stuff like security White Paper

   options:
    - BoxStation
    -ect
maybe have other option Like lava land and asteroids and any new stations that get added

corp-0:
> Game Version could be date?

Date is ambiguous because you could just input current date and be playing an older version.

> OS I don't think is required? ,
It should be there but I don't think it needs to be required

I disagree. It should be required. It also has a nice dropdown so it isn't like it is a huge amount of work to choose your OS.

> I feel like could be in scenarios where a bug wouldn't need them but they are required

what scenario? Can't think of any.

> I assume this still allows admins to write more detailed stuff e.g stuff like security White Pap

Yeah, admins can bypass.

Bod9001:
I don't really find date/build version to be that useful as is good for getting rid of duplicate issues, 
I still don't understand what needs to be required, It's not like they have different clients for different OSs with their own code?

Fr-Dae:
could you, too add 
- The map used, 
- The role played. 
- and if it is a local server (solo) or the EU/USA server

corp-0:
> I don't really find date/build version to be that useful as is good for getting rid of duplicate issues, I still don't understand what needs to be required, It's not like they have different clients for different OSs with their own code?

different OS does make a difference. For example, a common issue with windows is when it doesn't have permissions to write config files and you have to run it as administrator. The symptom is being unable to login.

Another thing that did happen in the past, addressables being written to ``/`` folder on linux, so the application needed sudo to run otherwise it would just crash.

> I don't really find date/build version to be that useful 

knowing the version where the issue occured is the bare mininum you need to debug, in conjuction with the description of the issue itself. Say someone was running an older version of the game and is reporting an issue we already fixed in latest.

> could you, too add
The map used,
The role played.
and if it is a local server (solo) or the EU/USA server

No, that data is already included in the other fields.

Bod9001:
but the majority of the time it's completely OS agnostic

Bod9001:
as for the version thing I see it being an issue when folks start appearing, but currently there is just the staging server and we know the version that

corp-0:
> but the majority of the time it's completely OS agnostic

We can't tell if it is a factor before the fact. That's why it was a field in the old issue template as well.

> as for the version thing I see it being an issue when folks start appearing, but currently there is just the staging server and we know the version that

Doesn't need to be a fork. All it takes is someone trying stuff from a local instance. If you insist this much on this one, I can concede I guess, although I don't agree at all.

.github/ISSUE_TEMPLATE/mapping_issue_report.yml
```diff
@@ -0,0 +1,47 @@
...
+    - PogStation
+    - FallStation
+    - MiniStation
+    - TestStation
+    - SquareStation
+    - UnrealStation
+    - OutpostStation
+    - Lavaland
+    - Asteroid
+  validations:
```
Bod9001, line 31: aren't we going to have "another" option?

corp-0: sure, I can add that. Although it would be preferred to update the template if we add more scenes in the future.

corp-0: done

Bod9001 (approved):

PR #9417, opened 2022-12-02
Wind optimization and cleanup

MaxIsJoe:
CL: [Fix] Wind has received some optimizations which will improve server performance.


ThatDan123:
Can you try doing a build as i still think Span shouldnt be able to do that.

Bod9001:
build works fine

Bod9001 (approved):

PR #9422, opened 2022-12-03
Another attempt at fixing name not showing 

Bod9001:
let's get the name from the mind instead

### Changelog:

CL: [Fix]Another attempt at fixing name not showing


MaxIsJoe:
I've said this before on discord but grabbing the name from the mind is a terrible idea.
This should be on the playerscript or object/item attributes, not the mind.

MaxIsJoe:
This also means that you're going to break `visableName` with this fix

PR #9426, opened 2022-12-04
Design philosophy ( discussion )

Bod9001:
Adds design philosophy for unitystation, 

feel free to suggest and add stuff

( as for stuff that we split over such as simulation versus gameplay, let's just keep this to things we all can agree on )

 feel free to discuss

### DIFF
--- a/docs/GameplayBalancing/DesignPhilosophy.md
+++ b/docs/GameplayBalancing/DesignPhilosophy.md
@@ -1,0 +1,10 @@
+# Design Philosophies For Unitystation
+
+If a feature is badly coded or hard to maintain, if you can remove it, you can rewrite it.
+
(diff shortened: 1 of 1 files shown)


Bod9001:
Gilles mention the good point that we don't have too much philosophy currently so, there's not really much point in writing it down

Fr-Dae:
for each department (or role)
the room should contain a certain number of minimum objects, and others simply recommended.

an example, in the kitchen, a microwave (mandatory)
have a stray cleaner, (optional)

this would serve as a checklist for the mappers, and so each department would be guaranteed a minimum number of items.
This would reduce the fluctuation in comfort from one station to another.

Fr-Dae:
in the same idea, define that all X tiles are linear.
a player should find a fire extinguisher.
as well as a "firewall" button

PR #9449, opened 2022-12-18
Singulo Improvements

### LINKED ISSUE #8790: Singularities sprites dont overlap properly
### Environment
- Game Version: 
- OS Version: 

image

### Steps to reproduce
spawn two singulooses near each other

AtnerNT:
Fixes #8790

About the shader changes:
I left a comment in the shader script but Ill say it again here because I think it is of relevance to mention.
So the shader not working before has to do with GrabPasses. As it was only doing one grab pass per frame the singularities were drawing over the top of one another. This change means each singulo does its own grab pass, allowing them to distort the distortion of other singulos. This allows the singulos to render correctly but does come at the cost of performance. I opted to do this anyways as it is very rare that players will even have 5 singulos on screen at once so the performace impact of doing the multiple passes will be neglible most of the time. However in the scenarios where an admin spawns a whole bunch of singulos all at the same place, it could result in a client side performance impact for people looking at the singulos.
See also: Comment in warpShader.shader

image

### Changelog:
CL: [Improvement] Singularities will now follow more distinct lines as opposed to hovering about in one spot.
CL: [Fix] Fixed the visual bug where multiple singularities next to one another would overlap and cut each other off.

### DIFF
--- a/UnityProject/Assets/Prefabs/Objects/Electricity/Engines/SingularityEngine/SingularityStages/Singularity.prefab
+++ b/UnityProject/Assets/Prefabs/Objects/Electricity/Engines/SingularityEngine/SingularityStages/Singularity.prefab
@@ -102,6 +102,7 @@ GameObject:
   serializedVersion: 6
   m_Component:
   - component: {fileID: 5124450014596669819}
+  - component: {fileID: 6281549647186351087}
   - component: {fileID: 6604085803539985251}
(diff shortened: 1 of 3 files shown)


Bod9001 (approved):

jamessimo:
Great job @AtnerNT ! !!!

PR #9450, opened 2022-12-18
Syncs character settings

### LINKED ISSUE #9423: [BUG / Mind Rework] gender is no longer being set properly.

### Game Version

22120318

### I was

Client

### OS

Linux

### What happened?

Trying to use features that reference gender show all characters as male or with he/him pronouns even though players might have set their preferences to something else.

Try this with the hover tooltips or with the laughing emotes, you'll always get male laughs and "he/him" in the detailed mode of the hover tooltip.

While laughing as a female character, you should make a female voice laugh.
While hovering over another player in detailed mode, you should display their correct pronouns.

1. Type *laugh in chat.
2. Observe if the sound made is the same as your chosen gender's
3. Press and hold shift while hovering over players.
4. Ask if their characters pronouns that they set while making this character is the same as the one you're seeing.

### Additional context

Players should have their characters genders and pronouns set correctly so that systems can correctly use and display the correct values.

MaxIsJoe:
closes #9423 

CL: [Fix] Character screams, gender, etc. will now correctly reflect the character's sheet.


UnityProject/Assets/Scripts/Player/Mind.cs
```diff
@@ -48,7 +48,7 @@ public class Mind : NetworkBehaviour, IActionGUI
 	public PlayerInfo ControlledBy;
 
 
-	public CharacterSheet CurrentCharacterSettings;
+	[SyncVar] public CharacterSheet CurrentCharacterSettings;
```
Bod9001, line 51: does it really grab the he and she stuff from the client?

MaxIsJoe: according to Atner's tests, yes it does.

Bod9001 (approved):

PR #9455, opened 2022-12-20
Some balancing and fixes

AtnerNT:
To accomodate the maintroom chance, quantum pads have a new field that influences the chance of being sent to the maintrooms. Square station quantum pads are 10x less likely to sent someone to the maintrooms due to their nessacity to station movement (1 in 200 instead of 1 in 20), With the exception of the two quantum pads in maint which are 10x MORE likely, for the fun and cause they in maint.

### Changelog:
CL: [Balance] Increased price of implanter in syndicate uplink from 2 TC to 6 TC.
CL: [Balance] Increased price of all implant explosives in syndicate uplink by 1 TC
CL: [Balance] Chance of randomly being teleported to Maintrooms increased from 1 in 1000 to 1 in 20 if Maintrooms is loaded.
CL: [Fix] Fixed sprite issue where fairy grass had the sprite of regular grass.

### DIFF
--- a/UnityProject/Assets/Scenes/AwaySites/Backrooms.unity
+++ b/UnityProject/Assets/Scenes/AwaySites/Backrooms.unity
@@ -139,7 +139,7 @@ PrefabInstance:
     - target: {fileID: 3383979214839746496, guid: c10109559e87d924bb10ee31d3bd33cb,
         type: 3}
       propertyPath: m_RootOrder
-      value: 20
+      value: 23
       objectReference: {fileID: 0}
     - target: {fileID: 3383979214839746496, guid: c10109559e87d924bb10ee31d3bd33cb,
         type: 3}
@@ -924,7 +924,7 @@ PrefabInstance:
     - target: {fileID: 7645387831379206111, guid: bc87c38bbcd94b24598001367ccfcebc,
         type: 3}
       propertyPath: m_RootOrder
-      value: 61
+      value: 1
       objectReference: {fileID: 0}
     - target: {fileID: 7645387831379206111, guid: bc87c38bbcd94b24598001367ccfcebc,
         type: 3}
@@ -1002,7 +1002,7 @@ PrefabInstance:
       objectReference: {fileID: 0}
     - target: {fileID: 4935453910410470, guid: a58eb8439b40f4edcb454576873ef674, type: 3}
       propertyPath: m_RootOrder
-      value: 37
+      value: 40
       objectReference: {fileID: 0}
     - target: {fileID: 4935453910410470, guid: a58eb8439b40f4edcb454576873ef674, type: 3}
       propertyPath: m_LocalPosition.x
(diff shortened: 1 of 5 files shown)


MaxIsJoe:
I know this is going to sound nitpicky, but we should follow the S.O.L.I.D principle and make these things open for extension and closed for modifcation.

Meaning that the chance of you teleporting to the maint-rooms should not be handled by the transport utility itself but rather by another thing like a scene manager that handles that chance of teleportation.
So have something like an event that the scene manager can hook onto and whenever something teleports that event gets fired and the chance function checks if that person should go there or not.

AtnerNT:
So just moving the check from TeleportUtility to SubScene manager? Subscene manager is already a very bloated class having like 6 partial classes. I see where you coming from but I think it is better suited to remaining on teleport utility, I think A) Its a better fit to have teleport related logic in the teleportUtility even if it realtes to a scene as it is after all part of the teleport utility. And B) I dont see the benefit of the propsed structure. Having every teleport action invoke an event just so the subscene manager can just perform a probability check does not seem ideal. Espcially since what this will end up meaning is that the player will probably get teleported twice instead of just changing the destination

AtnerNT:
```if (SubSceneManager.Instance.IsMaintRooms && Random.Range(0, MAINTROOM_CHANCE/maintRoomChanceModifier) <= 1) dest = MaintRoomLocations.PickRandom().RegisterTile().WorldPositionServer;```
Creating all this extra stuff to save one line in teleport utility I just dont see the benefit of

MaxIsJoe:
Doesn't necessarily have to be the scene manager, could be a new component that subscribes to the event on the teleporter and disable itself/remove itself when it's not needed

AtnerNT:
But having an event system for this means players will end up getting teleported multiple times. Its far cleaner to just having one if statement that changes the destination prior to a teleport action taking place

MaxIsJoe:
Why would it teleport you twice?

AtnerNT:
Well its not a nesscacity, but returning data from a event can be a bit dirty. bAsically I dont see a clean way you can implement your suggestion. And I think it is just far nicer in general to just have the singular if in the telepot utility as the effect is a teleport utility

Bod9001 (approved):

MaxIsJoe:
Well, it was a suggestion. This can be merged.

eurtod:
1 in 20 seems WAY too high in my opinion. You gotta think about it from the perspective of a new player joining for the first time, joining and instantly ending up in some far-off complicated place where you can't get help isn't very good.

AtnerNT:
Gotta keep in mind how rarely players actually teleport. The hand tele's/Teleport hub and bluespace banana peels are basically the only way players can teleport. The hub of which is only for captain (and RD i think??). And on average it will take 20 banana peel slips on a round where maintrooms is active in order to end up there. The only exception to this is square station, which is why the quantum pads have an override to 1 in 200.
Maintrooms are a 1 in 5 or 6 chance, and on the chance it spawns, on average you will have to teleport 20 times that round

eurtod:
Ah okay, I didn't know quantum pads had an override. Overall I think it's a good change then. :)

PR #9468, opened 2023-01-02
Temperature fixes

Bod9001:
hopefully catches a few more cases of Sad numbers getting into Atmospherics

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Objects/Atmospherics/Canister/NumberSpinner.cs
+++ b/UnityProject/Assets/Scripts/UI/Objects/Atmospherics/Canister/NumberSpinner.cs
@@ -138,7 +138,7 @@ namespace UI.Core.NetUI
 					Category.Atmos, newValue, MaxValue);
 			}
 
-			newValue = Mathf.Clamp(newValue, MaxValue, 0);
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/UI/Objects/Atmospherics/Canister/NumberSpinner.cs
```diff
@@ -134,10 +134,12 @@ public void DisplaySpinTo(int newValue)
 
 			if (newValue > MaxValue || newValue < 0)
 			{
-				Logger.LogErrorFormat("New value {0} is out of range, should be between 0 and {1} inclusive",
+				Logger.LogWarningFormat("New value {0} is out of range, should be between 0 and {1} inclusive",
 					Category.Atmos, newValue, MaxValue);
 			}
 
+			newValue = Mathf.Clamp(newValue, MaxValue, 0);
```
MaxIsJoe: Is this reversed? Clamp should take in a (value, minimum, maximum) but this looks like its doing it in reverse with the max value

Bod9001: thanks,  they need to make it more clear some way

corp-0 (approved):

PR #9469, opened 2023-01-02
Adds hot and cold thermal resistances

Bod9001:
should I combine some of them because is quite a lot should I just make thermal resistance and thermal weakness 
instead of cold resistance hot resistance, cold weakness hot weakness

### Changelog:
CL: [New] Added new temperature resistance and weakness mutations
CL: [New] Aded new pressure resistance and weakness mutations


corp-0:
> should I combine some of them

I don't think so. They are just instances of the class you created.

PR #9532, opened 2023-01-18
Make fastened station intercoms so they can't be picked up or pulled by players

CorruptComputer:
### Purpose
Currently players can pull and pick up wall mounted station intercoms. This PR does 2 things:
1. Makes them so they are not able to be picked up
2. Makes them so they cannot be pushed while fastened

### Notes:
Had to edit the `Pickupable.cs` script since I was not able to just remove this script from the wall mounted intercoms, Unity said it was required. So I made the `canPickup` variable able to be changed in the Unity UI for the objects that use it and set this to false for the fastened prefab.

image

image

image

### Changelog:
CL: [Fix] Station intercoms can no longer be pulled or picked up when fastened to walls

### DIFF
--- a/UnityProject/Assets/Prefabs/Objects/Wallmounts/StationIntercom.prefab
+++ b/UnityProject/Assets/Prefabs/Objects/Wallmounts/StationIntercom.prefab
@@ -33,6 +33,16 @@ PrefabInstance:
       propertyPath: foreverID
       value: 37832ac0101fb324aa588e3b400229fe
       objectReference: {fileID: 0}
+    - target: {fileID: 4553068655538650608, guid: 23a0b694284152342b2ec0817fade0d6,
+        type: 3}
+      propertyPath: isNotPushable
+      value: 1
+      objectReference: {fileID: 0}
+    - target: {fileID: 4553068655538650608, guid: 23a0b694284152342b2ec0817fade0d6,
+        type: 3}
+      propertyPath: CanBeWindPushed
+      value: 0
+      objectReference: {fileID: 0}
     - target: {fileID: 8279022589428523628, guid: 23a0b694284152342b2ec0817fade0d6,
         type: 3}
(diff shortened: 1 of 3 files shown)


Bod9001:
The idea that you assemble it by wrenching it down? 
actually why is it an item? 
Anyone know why?
maybe it's to do with the construction steps

CorruptComputer:
On tgstation at least it looks like it is screwed and wrenched to the wall: https://github.com/tgstation/tgstation/blob/4d6eadabadec42779771aa8fb544df4841fc00e3/code/game/objects/items/devices/radio/intercom.dm#L58

The wiki has almost no information on this so I had to go digging into the code to find it.

I'm not actually sure if you can pick it up after its detached, or if its a drag only. I can test that out sometime and get back on it.

corp-0:
can you see if after wrenching it, it is still the same object? Sometimes we have different prefabs representing states of an object.

CorruptComputer:
I looked into how this works in tgstation a bit closer and it seems that when unmounting it from the wall the intercom turns into an intercom frame item rather than just an unfastened intercom you can pick up and continue to use. Assuming that UnityStation wants it to work in a similar way to that, I am going to close this PR for now while I work on some changes to it.

My current plan is:
- Check if the unfastened prefab is used anywhere, if not just remove it. If it is I'll just leave it in place for now and that can be updated later since that would require scene changes
- Instead of changing the pickupable script, remove the item attributes and pickupable components from the constructed intercom
- Check that construction/deconstruction works fine, if not try to get those working
  - `Assets/Prefabs/Items/Construction/Wallframes/IntercomFrame.prefab` (deconstructed)
  - `Assets/Prefabs/Objects/Wallmounts/StationIntercom.prefab` (constructed)
- Perhaps a different sprite for the intercom frame to make it easier for players to differentiate between them? Currently they both use the same sprite

PR #9533, opened 2023-01-18
Fix the running griddle sound being able to play twice

### LINKED ISSUE #9530: [BUG] Griddle sound effect can have multiple instances playing

### Game Version

a2f70e207768c7bacaca4c16ba755abbd9632172

### I was

Client

### OS

Linux

### What happened?

If you spam click on the griddle to rapidly turn it on and off the sizzling sound effect will play more than once, if you then turn it off only one of the sound effect will get stopped. This leads to the sound effect getting stuck on forever. 

Looking into it, this seems to be an issue with threading in this method: 
https://github.com/unitystation/unitystation/blob/develop/UnityProject/Assets/Scripts/Objects/Machines/Griddle.cs#L206

Only one sound effect should be playing at a time and the sounds should stop when it is turned off.

1. Go to the kitchen
2. Either rapidly click or use an auto-clicker turn the griddle on and off
3. Listen to the sounds it makes

### Additional context

I have this fixed in my local and it will be included in a PR soon, just figured I should open an issue for the sake of completeness on this.

CorruptComputer:
### Purpose
Fixes a threading related issue with the griddle running sound effect.
Fixes #9530

### Notes:
If a thread gets to the point where it waits for 250ms and a second thread requesting the griddle be on comes in, that second thread can get to the wait and the first one continues. Tried just adding the extra IsNullOrEmpty check in DelayGriddleRunningSfx, but the problem still happened with that there. So I ended up just wrapping the whole sound effect request in a Mutex to guarantee only one thread can change its state at a time.

Also renamed some of the related variables to make it more clear what they are used for.

### Changelog:
CL: [Fix] Fix the griddle running sound effect playing twice when you rapidly turn it on and off

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Machines/Griddle.cs
+++ b/UnityProject/Assets/Scripts/Objects/Machines/Griddle.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using UnityEngine;
 using Mirror;
 using NaughtyAttributes;
@@ -36,8 +37,6 @@ namespace Objects.Kitchen
 		[Tooltip("The looped audio source to play while the griddle is running.")]
 		private AddressableAudioSource RunningAudio = default;
 
-		private string runLoopGUID = "";
-
 		[SerializeField, Foldout("Power Usages")]
 		[Tooltip("Wattage of the griddle's circuitry and display.")]
 		private int circuitWattage = 5;
@@ -52,8 +51,11 @@ namespace Objects.Kitchen
 		private SpriteHandler spriteHandler;
 		private APCPoweredDevice poweredDevice;
 
+		// Audio loop related vars
 		[SyncVar(hook = nameof(OnSyncPlayAudioLoop))]
 		private bool playAudioLoop;
+		private Mutex audioLoopLock = new Mutex();
+		private string audioLoopGUID = string.Empty;
 
 		public bool IsOperating => CurrentState is GriddleRunning;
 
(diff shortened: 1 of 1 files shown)


CorruptComputer:
Still need to be tested on a server/client connection, but running locally it works fine.

Bod9001 (approved):

NoooneyDude (approved):

UnityProject/Assets/Scripts/Objects/Machines/Griddle.cs
```diff
@@ -185,21 +187,29 @@ private void CheckCooked(float cookTime)
 
 		private void OnSyncPlayAudioLoop(bool oldState, bool newState)
 		{
+
+			// Only one thread should be able to try to play the sound at a time.
+			// Otherwise causes issues where while one thread is waiting another thread
+			// can get in here to request it a second time and the sound plays twice.
+			audioLoopLock.WaitOne();
```
ThatDan123, line 194: I'm slightly confused. To my knowledge none of this is threaded. 

@Bod9001 You didn't thread Mirror for this sort of stuff right?

CorruptComputer: I'm not certain but I think it is something built into Unity with the `[SyncVar(hook = nameof(OnSyncPlayAudioLoop))]` on line 55 above. Their documentation on it is sparse and basically just says it exists so it is not very helpful.

ThatDan123: > I'm not certain but I think it is something built into Unity with the `[SyncVar(hook = nameof(OnSyncPlayAudioLoop))]` on line 55 above. Their documentation on it is sparse and basically just says it exists so it is not very helpful.

It's not from Unity. It's our networking package Mirror. And hooks should all run on the mainthread.

https://mirror-networking.gitbook.io/docs/guides/synchronization/syncvar-hooks

Bod9001: could be further is receiving synchvar updates , but I doubt it

PR #9545, opened 2023-01-23
Update to 2021.3.16, update packages

NoooneyDude:
- Updates Unity editor from 2021.3.12f1 to 2021.3.16f1.
- Updates packages.


UnityProject/Packages/packages-lock.json
```diff
@@ -252,7 +252,7 @@
       "depth": 0,
       "source": "git",
       "dependencies": {},
-      "hash": ""
```
Bod9001: Can you make sure this is empty since if it generates a hash it always pulls the same version of that, and you have to empty it out to get the newest version

Bod9001 (approved):

PR #9554, opened 2023-01-26
Eating food can't be spammed now

MaxIsJoe:
CL: [New] Different foods have different consumption times now. Making most foods not instantly consumed upon using them.


corp-0:
that was quick, but I don't know man. I like the idea of food lasting more bites. In this case, if you start eating something and cancel at the very last moment, you ate nothing.

MaxIsJoe:
SS13 has this as well so its fine until we add multiple bites. Besides, most food are still less than 0.5 seconds of consumption, so they're pretty quick to eat still except for bigger stuff like the Big Bite Burger item.

Consider it a quick introduction and parity feature to ss13 until the consumable component is reworked to support multiple bites.

corp-0 (approved):
alright

PR #9561, opened 2023-01-29
Staging fixes

Bod9001:
### Changelog:

CL: [Fix] Fixed error from walking on splats with no shoes
CL: [Fix] Fixed Brain not unregistering itself when being destroyed from mind
CL: [Balance] Halfed the likelihood of being sent to the back rooms,  And re-does the RNG calculation
CL: [Fix] Fixed soap not working

### DIFF
--- a/UnityProject/Assets/Scripts/Effects/FloorEffect/LeaveFootprints.cs
+++ b/UnityProject/Assets/Scripts/Effects/FloorEffect/LeaveFootprints.cs
@@ -63,6 +63,7 @@ namespace Objects.Other
 				foreach (var feetSlot in playerStorage.GetNamedItemSlots(NamedSlot.feet))
 				{
 
+					if (feetSlot.ItemObject == null) continue;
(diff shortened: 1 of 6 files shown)


UnityProject/Assets/Scripts/Effects/FloorEffect/LeaveFootprints.cs
```diff
@@ -63,6 +63,7 @@ public override bool WillAffectPlayer(PlayerScript eventData)
 				foreach (var feetSlot in playerStorage.GetNamedItemSlots(NamedSlot.feet))
 				{
 
+					if (feetSlot.ItemObject == null) continue;
```
corp-0, line 66: barefoot don't leave footprints?

Bod9001: yeah, there's no reagent container to do it since the script that leaves footprints lives on the shoes

PR #9566, opened 2023-02-01
Fix Blob Hit Effect

ThatDan123:
CL: [Fix] Blob hit effect will now disappear


corp-0:
image
seeing the history of the file, it doesn't look like it has changed a lot. Why is it necessary to do this way now?

ThatDan123:
I think i broke it in this PR #9103

It worked fine as host, i probably just assumed it would work fine as client.

corp-0 (approved):

PR #9567, opened 2023-02-01
Fixes client spawning in space

Bod9001:
### Changelog:

CL: [Fix] Fixed client spawning in space when spawning as nuclear operative or Wizard


Bod9001:
TODO record which client did it so client can't spam multiple messages causing it to load too fast on clients

corp-0:
seems like it doesn't compile

Bod9001:
TODO is done, wasn't getting Compile errors on my end, will run tests

Bod9001:
all good

PR #9594, opened 2023-02-08
Gives Ministation its own syndicate base

AtnerNT:
Capture

Gives Ministation its own syndicate base to match pog and fallstation

### Changelog:
CL: [New] Added new syndicate base to ministation


PerfectTangent:
Cool, very rad!

I have a couple thoughts; First, I like the flavor, cool shuttle design, it all feels unique. Thanks for reminding me that syndicate secret docs should be in each syndicate base. You should try to fit a tool room in the base somewhere with the parts vendor, can't remember what it's called. That is so nukies have the opportunity to craft IEDs and other gizmos if they want. An airlock separating the bridge of the shuttle from the rest also would be a good addition, I think.

Also, for sake of balance and not prescribing how the operatives play, maybe it would be better to replace the sniper with a randomized contraband gun spawner. It keeps replayability just a bit fresher and doesn't upset the balance as much as an always available sniper rifle would. I think one artifact spawner is sufficient.


The shuttle looks like a space invader.

AtnerNT:
Suggestions applied

MaxIsJoe:
Tests are failing, check your SOs and build settings.

corp-0:
> Ministation Syndicate: "LightSwitch (1)" at (-3008.00, -244.00, 0.00) has no light sources.

AtnerNT:
Will finish this, just been very busy irl as of late

AtnerNT:
Aight, Got round to fixing tests, good to merge

NoooneyDude (approved):

NoooneyDude:
Nice one 👍

PR #9595, opened 2023-02-08
Janitor Cart improvements

MaxIsJoe:
### Known issues:

- Water sprites don't immediately update and requires a second interaction without a mop or bucket to update correctly.

### Changelog:

CL: [New] Janitor carts can now hold mops.
CL: [Improvement] Streamlined janitor cart interactions. Making them much more obvious and quicker to interact with.
CL: [Improvement] Added hover tooltips to janitorial carts. Giving you extra context on what you can do with carts based on the item you're holding.
CL: [Fix] Fixed janitorial carts having their water sprites not update at all.
CL: [Fix] Fixed janitorial carts not playing sounds when transferring liquids between mops, buckets and the carts.

PR #9597, opened 2023-02-08
chore: rescuing janitor cart PR

corp-0:
same as #9595


UnityProject/Assets/Scripts/ChemistryComponents/ReagentContainer.TransferInteraction.cs
```diff
@@ -1,5 +1,7 @@
 ﻿using System.Collections;
 using System.Collections.Generic;
+using AddressableReferences;
+using Castle.Core.Internal;
```
corp-0, line 4:
```suggestion
```

PR #9605, opened 2023-02-10
Cloning has a small chance to give you random mutations

Bod9001:
2.5% for each body part that is damaged 
because 25 damages applied to each body part

### Changelog:
CL: [New] Cloning has a small chance to give you random mutations negative and positive

### DIFF
--- a/UnityProject/Assets/Scripts/HealthV2/Living/BodyPartMutations.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/BodyPartMutations.cs
@@ -41,7 +41,7 @@ public class BodyPartMutations : BodyPartFunctionality
 	public override void Awake()
 	{
 		base.Awake();
-		RelatedPart.OnDamageChangeBy += OnDMGMutationCheck;
+		RelatedPart.OnDamageTaken += OnDMGMutationCheck;
 	}
 
 
--- a/UnityProject/Assets/Scripts/HealthV2/Living/BodyPartMutations.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/BodyPartMutations.cs
@@ -38,6 +38,43 @@ public class BodyPartMutations : BodyPartFunctionality
 		return MutationVariants[Mutation];
 	}
 
+	public override void Awake()
+	{
+		base.Awake();
+		RelatedPart.OnDamageChangeBy += OnDMGMutationCheck;
+	}
+
+
+	public void OnDMGMutationCheck(AttackType attackType,DamageType damageType, float amount)
+	{
+		if (damageType == DamageType.Clone)
+		{
+			//Range = -999999 to 9999999
+			if (amount > 0)
+			{
+				//Range = 0 to 9999999
+				amount = Mathf.Clamp(amount, 0, 100);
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartDamage.cs
```diff
@@ -183,6 +183,12 @@ public float TotalDamage
 			}
 		}
 
+		/// <summary>
+		/// So functions can subscribe and do there Funky stuff when damage is taken
+		/// </summary>
+		public event Action<AttackType,DamageType, float> OnDamageChangeBy;
```
corp-0: I can't believe we didn't have this event already. I have some suggestions tho.
```suggestion
		public event Action<AttackType,DamageType, float> OnDamageTaken;
```
Current name suggests you can get who did the damage or so I understand.

UnityProject/Assets/Scripts/HealthV2/Living/BodyPartMutations.cs
```diff
@@ -38,6 +38,43 @@ public static MutationRoundData GetMutationRoundData(MutationSO Mutation)
...
+			//Range = -999999 to 9999999
+			if (amount > 0)
+			{
+				//Range = 0 to 9999999
+				amount = Mathf.Clamp(amount, 0, 100);
+				//Range = 0 to 100
+				//Percentage 100 = 10
+
+				var RNG= Random.Range(0, 1000);
+				if (amount >= RNG)
```
corp-0, line 61: amount can only be ``100`` at most. Why make the range from 0 to 1000?

UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartDamage.cs
Bod9001: it describes the change in damage?

UnityProject/Assets/Scripts/HealthV2/Living/BodyPartMutations.cs
Bod9001: so 100 = 10% chance of happening

UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartDamage.cs
corp-0: maybe change isn't the right word, uhmm

UnityProject/Assets/Scripts/HealthV2/Living/Damage/BodyPartDamage.cs
```diff
@@ -183,6 +183,12 @@ public float TotalDamage
 			}
 		}
 
+		/// <summary>
+		/// So functions can subscribe and do there Funky stuff when damage is taken
```
corp-0:
```suggestion
		/// Triggers when a body part receives damage. 
		///It has the attack type, damage type and the amount of damage as parameters for the callback
```

corp-0 (approved):

PR #9616, opened 2023-02-16
Fix Codacy issues for `Classes with "IDisposable" members should implement "IDisposable"`

CorruptComputer:
### Purpose
This might fix a memory leak or two, not sure. They are labeled as critical on Codacy, so they are probably important to fix.

### Notes:
See: #7389

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Kitchen/InteractableGriddle.cs
+++ b/UnityProject/Assets/Scripts/Objects/Kitchen/InteractableGriddle.cs
@@ -10,7 +10,7 @@ namespace Objects.Kitchen
 	/// </summary>
 	[RequireComponent(typeof(Griddle))]
 	public class InteractableGriddle : MonoBehaviour, ICheckedInteractable<PositionalHandApply>,
-			IRightClickable, ICheckedInteractable<ContextMenuApply>
+			IRightClickable, ICheckedInteractable<ContextMenuApply>, IDisposable
 	{
 
 		[SerializeField]
@@ -86,5 +86,10 @@ namespace Objects.Kitchen
 		}
 
 		#endregion Interaction-ContextMenu
+
+		public void Dispose()
+		{
+			griddle?.Dispose();
+		}
 	}
 }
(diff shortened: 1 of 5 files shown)


UnityProject/Assets/Scripts/Objects/Machines/Griddle.cs
```diff
@@ -351,5 +351,10 @@ public override void PowerStateUpdate(PowerState state)
 		}
 
 		#endregion
+
+		public void Dispose()
+		{
+			audioLoopLock?.Dispose();
```
CorruptComputer, line 362: This was added by me in #9533

Is there not a scan happening on new PRs to catch when issues are introduced?

MaxIsJoe: We had codacy scanning enabled before, but it kept reporting false positives.

corp-0: > We had codacy scanning enabled before, but it kept reporting false positives.

not really. I disabled it because I was done trying to convince @Bod9001 to comply with it.

MaxIsJoe: Im his new codacy system now.

corp-0:
lgtm, I don't know if Unity has some weird behaviour with disposables or something like that.

UnityProject/Assets/Scripts/Objects/Kitchen/InteractableGriddle.cs
```diff
@@ -86,5 +86,10 @@ public void ServerPerformInteraction(ContextMenuApply interaction)
 		}
 
 		#endregion Interaction-ContextMenu
+
+		public void Dispose()
```
Bod9001, line 95: on destroy might be better

Bod9001: since that's the Unity way of saying hey this is being destroyed

CorruptComputer: I think I could add something like this to all of these classes:
```csharp
void OnDestroy()
{
    this.Dispose();
}
```

But the Dispose() method is a requirement of having the IDisposable interface, so I can't remove that.

CorruptComputer: Just pushed that change, does that seem better?

corp-0 (approved):

PR #9618, opened 2023-02-16
permissions iteration 1

corp-0:
### Purpose
Adds the new permissions system. It will allow server owners and administrators to set arbirtrary ranks in a config file and grant permissions to said ranks.

In this iteration, this is still not integrated into the game, but I wanted to PR as soon as possible because I don't know when will I have time to complete it.

### Notes
Since this is still not visible by players, I'm not including a changelog record.

### DIFF
--- a/UnityProject/Assets/StreamingAssets/admin/permissions.toml
+++ b/UnityProject/Assets/StreamingAssets/admin/permissions.toml
@@ -1,0 +1,29 @@
+﻿[ranks.host]
+show_in_chat = false
+abbreviation = "H"
+color = "red"
+permissions = ["perma_promote", "promote", "demote", "kick", "ban", "fun"]
+
+[ranks.admin]
+show_in_chat = true
+abbreviation = "A"
+color = "blue"
+permissions = ["promote", "kick", "ban", "fun"]
+
+[ranks.normal]
+show_in_chat = false
+abbreviation = "N"
+color = "white"
+permissions = []
+
+[[players]]
+identifier = "hostPlayer"
+rank = "host"
+
+[[players]]
+identifier = "adminPlayer"
+rank = "admin"
+
+[[players]]
+identifier = "normalPlayer"
+rank = "normal"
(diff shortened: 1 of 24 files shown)


corp-0:
Don't merge yet, I have to make some changes in the build script

corp-0:
done. Let's see if this works for tests.

UnityProject/Assets/StreamingAssets/admin/permissions.toml
```diff
@@ -0,0 +1,29 @@
...
+identifier = "hostPlayer"
+rank = "host"
+
+[[players]]
+identifier = "adminPlayer"
+rank = "admin"
+
+[[players]]
+identifier = "normalPlayer"
+rank = "normal"
```
ThatDan123: Maybe add an example comment of how to add a user to one of these?

Or is this just a file to setup the ranks and perms and then in another you'll give users that rank?

corp-0: In this file you create the ranks and assign them to users.

To add a new record you just add the [[players]] list and under it, the identifier (ugly ass firebase string) and the rank you created above.

I thought it was self explanatory.

ThatDan123:
```suggestion
[[players]]
identifier = "hostPlayerUserID"
rank = "host"

[[players]]
identifier = "adminPlayerUserID"
rank = "admin"

[[players]]
identifier = "normalPlayerUserID"
rank = "normal"
```

Maybe that then?

corp-0: Sure, the ranks I created there are merely for testing purposes. I didn't intend to do the actual ranks we will have nor the permissions.

The normal user on the bottom only represents a record without permissions. Of course we don't need it in the game.

corp-0: I separated the example config file form the test one. Now the example is a better representation of reality. Take a look.

ThatDan123 (approved):

UnityProject/Assets/StreamingAssets/admin/permissions.toml
ThatDan123: Yeah, i think that explains it well.

PR #9619, opened 2023-02-16
Staging fixes22.2

Bod9001:
adds a new component that animates light sprites

### Changelog:

CL: [Fix] Fixes eggs sometimes not hatching
CL: [Balance] syringe has text when starting injection
CL: [Balance] baby dinosaurs are bigger
CL: [Fix]  Body parts now get Thrown out If they can't Be accessed in the Parent that's being morphed
CL: [Fix] fixes cyborgs having body parts
CL: [Fix] fixes cyborgs having mutations


UnityProject/Assets/Scripts/Core/Sprite Handler/LightSpriteHandler.cs
```diff
@@ -0,0 +1,99 @@
+using Light2D;
+using UnityEngine;
+
```
corp-0, line 3: adding a new class without namespace

UnityProject/Assets/Scripts/Messages/Client/RequestCableCut.cs
```diff
@@ -0,0 +1,37 @@
+using Messages.Client;
+using Mirror;
+using UnityEngine;
+
```
corp-0, line 3: another new class without namespace

corp-0 (changes requested):
other than wondering why the interaction message on tiles was removed, I don't have much to say.

Add the namespaces and we are done.

Bod9001:
Because it had a register handler And I'm wondering is it that's causing the weird mirror issue, 
And also it's not like any other Interaction messages so is better to make it standard

PR #9622, opened 2023-02-19
CMO hyper  Spray can no longer kill instantly

Bod9001:
### Changelog:

CL: [Balance] CMO hyper  Spray can no longer kill instantly


corp-0:
I think you should also increase the dangerous salt amount. It doesn't take much to apply two injections anyways.

corp-0 (approved):

PR #9639, opened 2023-02-23
Fix server-client auth rejection when not headless

NoooneyDude:
CL: [Fix] Fixes the server's client authentication rejection handling, when self-hosting and not headless. The network server will now correctly be shut down instead of leaving the client stuck in the server lobby.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Networking/Authenticator.cs
+++ b/UnityProject/Assets/Scripts/Core/Networking/Authenticator.cs
@@ -382,9 +382,24 @@ namespace Core.Networking
 				return;
 			}
 
-			Logger.Log($"Disconnected from server. Reason: {msg.Code}.");
+			Logger.Log($"Disconnecting from server. Reason: {msg.Code}.");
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Core/Networking/Authenticator.cs
```diff
@@ -382,9 +382,24 @@ public void OnAuthResponse(AuthResponseMessage msg)
 				return;
 			}
 
-			Logger.Log($"Disconnected from server. Reason: {msg.Code}.");
+			Logger.Log($"Disconnecting from server. Reason: {msg.Code}.");
+			ClientReject(); // Gracefully handle rejection by disconnecting.
 
+			// If this client is also the host but not headless (i.e. not handled by OnServerClientAuthRequest()),
```
NoooneyDude, line 389:
```suggestion
			// Then shut down the client to return to the main menu.
			// If this client is also the host but not headless (i.e. not handled by OnServerClientAuthRequest()),
```

Bod9001 (approved):

PR #9649, opened 2023-02-25
Memory leak fixing : Adding foundation

havietisov:
### Purpose
Adds foundation that will be used later in all future PR's related to fixing memory leaks 

### Notes:
Should not break anything on it's own but double checking would always be nice

### Changelog:
Added ability to log typenames and, sometimes, names of leaked objects 
Replaced Debug.Log with Logger.Log
Added special logger category specifically for memory cleanup debug messages
Got rid of snake case
Added docstrings

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Lifecycle/CleanupUtil.cs
+++ b/UnityProject/Assets/Scripts/Core/Lifecycle/CleanupUtil.cs
@@ -8,7 +8,12 @@ using Audio.Containers;
 
 public static class CleanupUtil
 {
-
+	/// <summary>
+	/// this methods scans the provided list and removes from it all elements, that are not marked as DontDestroyOnLoad. Will keep elements that are equal to null
+	/// <paramref name="targetExtractor"/> is just used to get the MonoBehaviour that is then being used to figure out whether or not element should be removed
+	/// </summary>
+	/// <param name="listInQuestion"> is a list that should be altered according to the rules provided by the method</param>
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/Core/Lifecycle/CleanupUtil.cs
```diff
@@ -0,0 +1,194 @@
...
+using UnityEngine;
+using UnityEditor;
+using System;
+using UI.Action;
+using Audio.Containers;
+
+public static class CleanupUtil
+{
+
+	public static int RidListOfSoonToBeDeadElements<T>(IList<T> list_in_question, Func<T, UnityEngine.MonoBehaviour> target_extractor)
```
Bod9001: could you add an option, to make it complain and log everything  so people can work out what leaking if debugging  ?

havietisov: Done in c5c93cf411cf9888714a492bc12d7820d9f0f465

Bod9001 (dismissed):

corp-0 (changes requested):
please use C# naming convention, not snake_case. 
Remove all instances of ``Debug.Log`` and replace it with ``Logger.Log``
Althoug the naming for the methods is descriptive enough, consider adding docstrings to the public static methods, so other developers know how to use it.

Bod9001 (approved):

PR #9651, opened 2023-02-25
Memory leak fix : cleanup of tiles

havietisov:
### Purpose
Adds cleanup routines into tiles and whatnot. 
Also removes fields that do no good and only create rogue references that keep certain objects alive in memory. 
### Notes:
Could actually break something so please do some checks.
Should only be merged after https://github.com/unitystation/unitystation/pull/9649

### Changelog:
Cleaned up log a bit
Removed junk methods that sneaked into PR
Replaced Debug.Log's with Logger.Log's

### DIFF
--- a/UnityProject/Assets/Scripts/Items/PlaceableTile.cs
+++ b/UnityProject/Assets/Scripts/Items/PlaceableTile.cs
@@ -61,6 +61,21 @@ namespace Tiles
 			Cooldowns.TryStartClient(interaction, CommonCooldowns.Instance.Melee);
 		}
 
+		public void Clear()
+		{
+			int i = 0;
+			foreach (var entry in waysToPlace)
+			{
+				if (entry.layerTile is ConnectedTile)
+				{
+					(entry.layerTile as ConnectedTile).Clear();
+					i++;
+				}
+			}
+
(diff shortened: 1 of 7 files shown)


UnityProject/Assets/Scripts/Items/PlaceableTile.cs
```diff
@@ -61,6 +61,21 @@ public void ClientPredictInteraction(PositionalHandApply interaction)
...
+			foreach (var entry in waysToPlace)
+			{
+				if (entry.layerTile is ConnectedTile)
+				{
+					(entry.layerTile as ConnectedTile).Clear();
+					i++;
+				}
+			}
+
+			Debug.Log("Called clear" + i + " times on a connected tile");
```
Bod9001: Might be worth cleaning up the log?

Bod9001: actually just  use logger.log  instead  and make a custom category called cleaning

havietisov: Done

Bod9001 (approved):

PR #9654, opened 2023-02-27
Initial tidy-up of CharacterCustomization

NoooneyDude:
Baby steps to split business from UI in the daunting monolith that is `CharacterCustomization`.
- Creates `CharacterManager` and moves character list and offline serialisation/deserialisation.
- Fixes probably harmless NRE during lobby startup.
- Creates new `MathUtils` class with new util, `Mod()`, for calculating mathematical modulo.
- Poaches `IsEqual()` float comparison util from `CodeUtilities` so that `Mod()` in `MathUtils` doesn't get lonely.

>Spot-tested and only in the editor, but changes are straightforward enough.


UnityProject/Assets/Scripts/UI/Systems/Lobby/CharacterCustomization.cs
```diff
@@ -329,16 +329,9 @@ private void CheckIfCharacterListIsEmpty()
...
 		{
-			if (currentCharacterIndex != 0)
-			{
-				currentCharacterIndex--;
-			}
-			else
-			{
-				currentCharacterIndex = PlayerCharacters.Count();
-			}
+			currentCharacterIndex = Math.Clamp(index, 0, CharacterManager.Characters.Count());
```
corp-0: this doesn't keep the wrapping around functionality, so if you go left when you are at index 0 you will stay at index 0.

If you use modulo operator you can keep the functionality.

```cs
//changeAmount is +1 or -1 depending on if is called by right or left.
public void ChangeIndex(int changeAmount)
{
    currentCharacterIndex = (currentCharaterIndex + changeAmount) % (PlayerCharacters.Count() + 1);
}
```

if you are at index 0 and the count is 4, when you left the current index will be 4. If you are at 4 and go right, current index will be 0. Rest work as expected.

UnityProject/Assets/Scripts/Systems/Lobby/CharacterManager.cs
```diff
@@ -0,0 +1,93 @@
...
+		public CharacterManager Init()
+		{
+			LoadOfflineCharacters();
+
+			return this;
+		}
+
+		public int GetCurrentCharacterIndex()
+		{
+			int lastCharacterIndex = PlayerPrefs.GetInt(PlayerPrefKeys.LastCharacterIndex);
```
corp-0, line 25: perhaps this could be moved to the init and add a property to get the current selected index to the class. That way you can make it a ``BehaviourSubject``(https://github.com/unitystation/unitystation/blob/develop/UnityProject/Assets/Scripts/Util/Rx/BehaviourSubject.cs). You would then subscribe to it from the GUI and react to when it is changed.

Bod9001:
I presume it's compatible with the old Saves?

UnityProject/Assets/Scripts/Systems/Lobby/CharacterManager.cs
```diff
@@ -0,0 +1,93 @@
...
+using UnityEngine;
+
+namespace Systems.Character
+{
+	/// <summary>Manage a player's characters. Intended for the local player client.</summary>
+	public class CharacterManager
+	{
+		public List<CharacterSheet> Characters { get; } = new();
+
+		private string OfflineStoragePath => $"{Application.persistentDataPath}characters.json";
```
MaxIsJoe, line 14: does this return the path correctly or is there a missing /

UnityProject/Assets/Scripts/Systems/Lobby/CharacterManager.cs
```diff
@@ -0,0 +1,93 @@
...
+		{
+			int lastCharacterIndex = PlayerPrefs.GetInt(PlayerPrefKeys.LastCharacterIndex);
+
+			return Math.Clamp(lastCharacterIndex, 0, Characters.Count);
+		}
+
+		/// <summary>Load characters that have been saved to the cloud.</summary>
+		public void LoadCharacters()
+		{
+			throw new NotImplementedException();
```
MaxIsJoe, line 33: not being used?

UnityProject/Assets/Scripts/Systems/Lobby/CharacterManager.cs
```diff
@@ -0,0 +1,93 @@
...
+		}
+
+		/// <summary>Load characters that are saved to Unity's persistent data folder.</summary>
+		public void LoadOfflineCharacters()
+		{
+			Characters.Clear();
+
+			if (!File.Exists(OfflineStoragePath)) return;
```
MaxIsJoe: `if (File.Exists(OfflineStoragePath) == false) return;`

UnityProject/Assets/Scripts/Systems/Lobby/CharacterManager.cs
```diff
@@ -0,0 +1,93 @@
...
+
+			string json = File.ReadAllText(OfflineStoragePath);
+
+			var characters = JsonConvert.DeserializeObject<List<CharacterSheet>>(json);
+
+			if (characters == null) return;
+
+			foreach (var character in characters)
+			{
+				if (ValidateCharacterSheet(character))
```
MaxIsJoe, line 51: nesting can be reduced 
```cs
if (ValidateCharacterSheet(character) == false) continue;
Characters.Add(character);
```

UnityProject/Assets/Scripts/Systems/Lobby/CharacterManager.cs
```diff
@@ -0,0 +1,93 @@
...
+			{
+				File.Delete(OfflineStoragePath);
+			}
+
+			File.WriteAllText(OfflineStoragePath, json);
+		}
+
+		public bool ValidateCharacterSheet(CharacterSheet character)
+		{
+			// TODO: not implemented
```
MaxIsJoe, line 89: can't copy the validation that exists elsewhere and put it here?

UnityProject/Assets/Scripts/UI/Systems/Lobby/CharacterCustomization.cs
NoooneyDude: Thanks Gilles! I hadn't thought of the circular index function. Will fix.

UnityProject/Assets/Scripts/Systems/Lobby/CharacterManager.cs
NoooneyDude: I don't like the current indexing and how this is set up will probably change later, but will keep this cool class in mind 👍

UnityProject/Assets/Scripts/Systems/Lobby/CharacterManager.cs
NoooneyDude: This was how the path was formed before, it just uses string interpolation now.

UnityProject/Assets/Scripts/Systems/Lobby/CharacterManager.cs
NoooneyDude: No, the Firebase storage of characters hasn't been used since the character creator overhaul. It's a future task. As mentioned, this PR is just baby steps.

UnityProject/Assets/Scripts/Systems/Lobby/CharacterManager.cs
NoooneyDude: Forgot about this rule... thanks

UnityProject/Assets/Scripts/Systems/Lobby/CharacterManager.cs
NoooneyDude: Thanks but I'd rather leave it as it is.

UnityProject/Assets/Scripts/Systems/Lobby/CharacterManager.cs
NoooneyDude: That would be a new behaviour. That's the idea eventually, though. To validate on load. Just not yet.

NoooneyDude:
> I presume it's compatible with the old Saves?

Yep that's right and will be true for the followup PRs.

NoooneyDude:
Thanks guys, fixes applied 👍

corp-0 (approved):
Looks good

PR #9671, opened 2023-03-04
Trauma cleanup

MaxIsJoe:
# Purpose

This PR aims to clean the `BodyPart` class by moving the trauma partial class out of it, and making trauma its own class and component(s) that can easily be expanded upon in the future. This also cleans up trauma significantly in terms of quality and functionality to re-enable and improve older behaviors.

# Changelog

CL: [New] Progressing through the different stages/tiers of trauma now shows different messages that correspond with which trauma the user has received.
CL: [New] The first and second burn trauma stages are now visually visible, instead of just the last stage.
CL: [New] Bullets will now do piercing damage.
CL: [Improvement] Slashing and Piercing traumas cause bleeding once more.
CL: [Improvement] The gauze now shows messages when getting applied, as well as a progress bar to show some feedback to the player.
CL: [Improvement] Trauma armor calculation and rules has been significantly simplified and are much more straight forward now compared to how it used to be back then. Slash, Pierce, Blunt and Piercing traumas now rely on the Total Damage a body part has in order to progress through different stages.
CL: [Improvement] The effects of Bomb Trauma should be much more consistent now. Armour that has 100% chance to block any and all effects of bomb explosion traumas will always block it regardless of what state the body part or over-all health of the creature is.
CL: [Improvement] Slash and Piercing Traumas now have natural healing on the first and second stage of their traumas. With the first stage being the quickest to revert itself and the second the longest.
CL: [Improvement] Blunt force trauma has had its internal damage logic simplified. Organ damage is now much more consistent, making the damage much more predictable. Each trauma stage will now do damage on random internal organs inside the affected body part.
CL: [Improvement] All traumas now have a "Deadly One Hit" check that ignores armor chances. If you receive damage higher than 50-55 damage; all requirement checks are bypassed, and you'll progress your trauma to the next stage. This change is mainly there to allow for things such as door/airlock crushing to be re-implemented in the future, and to allow for "stamina shoe bombs" to return once more as an admin meme weapon.
CL: [Improvement] Simplified trauma reporting in the advance health scanner. It will now only show relavent information when you need it.
CL: [Balance] The gauze will no longer instantly heal bleeding; Instead, it will half the amount of bleeding stacks a player has. However, if the player's bleeding stacks is equal to or less than 8, it will instantly heal all stacks.
CL: [Balance] Blunt force trauma damage has been nerfed. It will now cause 3 damage every 45 seconds on the first stage, 6 damage every 25 seconds on the second stage, and 8 damage every 18 seconds on the last stage.
CL: [Balance] Rebalanced the damage numbers that are required to progress through trauma stages. Piercing requires 16, Blunt is 14, Slash 16 and Bomb is 30.
CL: [Balance] Removed custom internal bleeding logic for the heart and lungs as they did not have any treatments at the time, this of course led to infurating moments where doctors were unable to stop patients from drowning in their own blood or having an infinite amount of heart attacks that can never be healed. The removal of these effects are only temporary until proper trauma healing is added.
CL: [Balance] Disemboweling, the act of removing organs from a body using Pierce trauma hits, has been removed until further notice.


Bod9001:
could you check surgery quickly, removing and adding The same body part, to check the change with remove body part

MaxIsJoe:
> could you check surgery quickly, removing and adding The same body part, to check the change with remove body part

seems ok?

UnityProject/Assets/Scripts/Core/Chat/Chat.cs
```diff
@@ -513,6 +513,31 @@ public static void AddGameWideSystemMsgToChat(string message)
...
+	/// Do not use this method for combat messages
+	/// Remember to only use this server side
+	/// </summary>
+	/// <param name="originator"> The player who caused the action</param>
+	/// <param name="everyoneMessage"> The message that everyone (including the orignator) will see</param>
+	public static void AddActionMsgToChat(GameObject originator, string everyoneMessage)
+	{
+		if (!IsServer()) return;
+
+		//dont send message if originator message is blank
```
corp-0: this comment is both redudant and a lie

MaxIsJoe: done

UnityProject/Assets/Scripts/HealthV2/Living/Damage/Trauma/BodyPartTrauma.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System.Collections.Generic;
+using UnityEngine;
+
+namespace HealthV2
+{
+	public class BodyPartTrauma : MonoBehaviour
```
Bod9001: maybe should inherit from body part functionality?

MaxIsJoe: no

UnityProject/Assets/Scripts/HealthV2/Living/Damage/Trauma/CreatureTraumaManager.cs
```diff
@@ -0,0 +1,61 @@
+﻿using System;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace HealthV2
+{
+	public class CreatureTraumaManager : MonoBehaviour
```
Bod9001, line 6: how should we console this with the changes I'm doing with the factory approach?

UnityProject/Assets/Scripts/HealthV2/Living/Damage/Trauma/BodyPartTrauma.cs
Bod9001: why is this a separate class from TraumaLogic?

UnityProject/Assets/Scripts/HealthV2/Living/Damage/Trauma/TraumaTypes/TraumaBlunt.cs
```diff
@@ -0,0 +1,94 @@
+﻿using System;
+using UnityEngine;
+
+namespace HealthV2.TraumaTypes
+{
+	public class TraumaBlunt : TraumaLogic
```
Bod9001, line 6: why  are they individual classes?

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -316,6 +319,8 @@ public BleedingState CalculateBleedingState()
 			MultiInterestBool.RegisterBehaviour.RegisterFalse,
 			MultiInterestBool.BoolBehaviour.ReturnOnFalse);
 
+		[SerializeField, Range(1,60f)] private float updateTime = 1.25f;
```
Bod9001: why did you change the time?
and also there's systems that calculate presuming each update is one sec

Bod9001: this also would completely change the rate of how much you would take space damage and stuff, plus fire damage and all that stuff

UnityProject/Assets/Scripts/HealthV2/Living/Damage/Trauma/CreatureTraumaManager.cs
MaxIsJoe: I don't know, CreatureTraumaManager usually sits on the mob prefab next to livinghealthmasterbase and only serves an API for healing stuff right now.

UnityProject/Assets/Scripts/HealthV2/Living/Damage/Trauma/TraumaTypes/TraumaBlunt.cs
MaxIsJoe: because its cleaner and easy to manage in the inspector, especially if there are any traumas in the future that will have more than one component to handle itself.

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -954,7 +959,7 @@ private void SetConsciousState(ConsciousState NewConsciousState)
 		/// <param name="damageSplit">Should the damage be divided by number of body parts or applied to each body part separately</param>
 		[Server]
 		public void ApplyDamageAll(GameObject damagedBy, float damage, AttackType attackType, DamageType damageType,
-			bool damageSplit = true)
+			bool damageSplit = true, TraumaticDamageTypes traumaticDamageTypes = TraumaticDamageTypes.NONE, double traumaChance = 50)
```
Bod9001, line 1041: I thought the trauma damage was dependent on the attack type, 
e.g attack type sharp pointy thing = likelihood of getting bleeding, 
 what benefits from having the  chance of causing a trauma?

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
MaxIsJoe: I didn't mean to change it, was just trying to move away the magic number into a variable.

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
MaxIsJoe: It never worked that way, this has always been the way we detected traumas and their chances.

UnityProject/Assets/Scripts/HealthV2/Living/Damage/Trauma/BodyPartTrauma.cs
MaxIsJoe: BodyPartTrauma is the root that identfies traumas a body part can have.
TraumaLogic is in the name, its logic a trauma has.

Bod9001:
got to discuss if it's okay to do   sub game objects with functionality

MaxIsJoe:
annnd, done. It now uses body part functionality.

MaxIsJoe:
hewwo?

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
Bod9001: What's the benefit of having an independent Trauma dmaged type?

MaxIsJoe: You get to decide what each weapon and damage source can do in terms of trauma damage without being bound to existing damage types

UnityProject/Assets/Scripts/HealthV2/Living/Damage/Trauma/BodyPartTrauma.cs
```diff
@@ -0,0 +1,62 @@
...
+		}
+
+		public override void RemovedFromBody(LivingHealthMasterBase livingHealth)
+		{
+			var creatureTraumaAPI = livingHealth.GetComponent<CreatureTraumaManager>();
+			if (creatureTraumaAPI == null) return;
+			creatureTraumaAPI.Traumas.Remove(RelatedPart);
+		}
+
+		private void OnDamageTaken(BodyPartDamageData data)
```
Bod9001: Might be worth putting this inside of  body part functionality , So other stuff can plug into it easily

UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
Bod9001: Can't you define that by attack type though?

MaxIsJoe: no, you can't.

Bod9001: well The attack type Has blunt and sharp and stuff?

MaxIsJoe: You can only define burn via energy. But that itself is very limiting; what if you get burnt from other sources that aren't just energy? Traumas are better defined separately because everyone gets finer control over what traumas happen.

MaxIsJoe: > well The attack type Has blunt and sharp and stuff?

No, It doesn't. It only has "Brute" which is very hard to define in terms of slashing, piercing or blunt force traumas.

Bod9001: Since we agreed that it's too limited with non-flag , To do is investigate flag attack types

Bod9001: Also another good point is maybe We want a system where trauma isn't always present

Bod9001 (dismissed):

PR #9680, opened 2023-03-05
Memory leak fix : PlayerList.Admin

havietisov:
### Purpose
Fixes admin list having dangling references coming from those event handlers. Funniest stuff so far 

### Notes:
Shouldn't break anything. Check if it's as funny as I think it is.

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
+++ b/UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
@@ -113,81 +113,81 @@ public partial class PlayerList
 		LoadJobBanList();
 	}
 
-	void LoadBanList()
+	static void LoadBanList()
 	{
-		StartCoroutine(LoadBans());
+		Instance.StartCoroutine(LoadBans());
 	}
-	void LoadWhiteList(object source, FileSystemEventArgs e)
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
```diff
@@ -113,81 +113,81 @@ void InitAdminController()
 		LoadJobBanList();
 	}
 
-	void LoadBanList()
+	static void LoadBanList()
 	{
-		StartCoroutine(LoadBans());
+		Instance.StartCoroutine(LoadBans());
```
Bod9001, line 118: isn't the class not destroyed between rounds?

havietisov: Yes, it being destroyed, but ... 
This change is kinda old and, IIRC, filesystemwatcher is not being re-built between rounds, instead, the one from very first round being used. And that's the problem because event handlers keep closures on instance of PlayerList. By making these methods static, I'm making sure there's no closures on PlayerList instance that will be destroyed. Instead, it will poll for new static instance of PlayerList.

Bod9001 (approved):

PR #9682, opened 2023-03-05
Memory leak fix : sound manager

havietisov:
### Purpose
Removes dangling references to sound manager

### Notes:
If something breaks because of it, I'l be impressed

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/SoundManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/SoundManager.cs
@@ -52,17 +52,7 @@ public class SoundManager : MonoBehaviour
 		layerMask = LayerMask.GetMask("Walls", "Door Closed");
 	}
 
-	private void OnEnable()
-	{
-		SceneManager.activeSceneChanged += OnSceneChange;
-	}
-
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Managers/SoundManager.cs
```diff
@@ -52,17 +52,7 @@ private void Init()
...
-		SceneManager.activeSceneChanged += OnSceneChange;
-	}
-
-	private void OnDisable()
-	{
-		SceneManager.activeSceneChanged -= OnSceneChange;
-	}
-
-	private void OnSceneChange(Scene oldScene, Scene newScene)
+	public void Clear()
```
Bod9001, line 55: I presume this is called manually somewhere else?

havietisov: Yes. all "Clear" calls are later called in CleanupUtil. 
I just want all of this to be revieved and merged before I add calls one by one so we see could what gets messed up by what.

Bod9001 (approved):

corp-0 (approved):

PR #9686, opened 2023-03-05
Memory leak fix : PlayerManager

havietisov:
### Purpose
By subscribing to this unity event PlayerManager was causing nothing but trouble.
This change fixes it. 

### Changelog:


corp-0:
did you forget to push something? This seems to have compiling errors and since you removed a property from the class, I imagine you have to update the class that used that property.

havietisov:
Ah, seems like class was updated it while I was messing with it
Wait for fix

havietisov:
@corp-0 please check

corp-0 (approved):

PR #9688, opened 2023-03-05
Memory leak fix : RoundEndScoreBuilder

havietisov:
### Purpose
Pretty straightforward : allows RoundEndScoreBuilder to remove itself from places necessary in order to not cause any memory oppsie-doopsie

### Changelog:


corp-0:
during the countdown of the next round, the score should still be visible. Does this work like that?

havietisov:
@corp-0 Theoretically, yes, it should still be visible. However, it's probably nice to check.

corp-0:
well, did you at least test it in play mode in editor? It should be easy to start a round, finish it and see what happens with the score window

havietisov:
@corp-0 well, I decided to double-check and after round restart, I still see score.

Bod9001 (approved):

corp-0 (approved):

PR #9697, opened 2023-03-06
PogstationREDUX Fixes & Additions

LastCaptainBonk:
This Pr is meant to fix the error on PogstationREDUX regarding tiles and tables along with adding some vendors for the crew to experiment with.

Notes:
There will be another commit with the tile fixes. The problem was that I set the floor layer and table layer to "Ginger Amaretto" instead of the "Sprite - PixelSnap" it was supposed to be set as.

Changelog:

CL: [New] Added some vendors to PogStation.


corp-0:
I will wait for tests and pull it later today to see if it works now locally

corp-0:
image
PinkyCuteStation

corp-0:
Good news, Pog is fixed

PR #9698, opened 2023-03-06
Memory leak fix : StatsTab

havietisov:
### Purpose
Allowing StatsTab to clean stuff after itself to prevent some memory leaks

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Systems/MainHUD/StatsTab.cs
+++ b/UnityProject/Assets/Scripts/UI/Systems/MainHUD/StatsTab.cs
@@ -14,11 +14,23 @@ public class StatsTab : SingletonManager<StatsTab>
 
 	void OnEnable()
 	{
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/UI/Systems/MainHUD/StatsTab.cs
```diff
@@ -14,13 +14,25 @@ public class StatsTab : SingletonManager<StatsTab>
 
 	void OnEnable()
 	{
+		Invoke("SetScrollToTop",0.1f);
```
Bod9001: can you remove this, since its unneeded?

havietisov: Dunno why you needed this specifically here, but ok

Bod9001 (approved):

PR #9709, opened 2023-03-08
Fixes client and editor with the new Memory changes

Bod9001:
### Changelog:
CL: [Fix] Fixer client not being able to interact after Around Ending


UnityProject/Assets/Scripts/Managers/GameManager.cs
```diff
@@ -233,14 +233,25 @@ private void OnEnable()
 	{
 		SceneManager.activeSceneChanged += OnSceneChange;
 		UpdateManager.Add(CallbackType.UPDATE, UpdateMe);
+		EventManager.AddHandler(Event.Cleanup,ClientCleanupInbetweenScenes );
```
corp-0: can you separate the arguments with a space? They are hard to read like this.

NoooneyDude (approved):

PR #9711, opened 2023-03-09
CharacterManager expansion

NoooneyDude:
Continues #9654 and #9703.

- More responsibility handed to `CharacterManager`.
- Fixed a TODO or two.

> Spot checked in the editor.


Bod9001 (approved):

corp-0:
I only gave it a quick read, does getting the active character support having a discardable character? For example, resoawned by admin and having a whole new identity or picking s ghost role

NoooneyDude:
> I only gave it a quick read, does getting the active character support having a discardable character? For example, resoawned by admin and having a whole new identity or picking s ghost role

The character sheet actually being used in a game is set on `PlayerScript`. Client sends server the ActiveCharacter, server then decides what happens with it or if player gets something else entirely.

corp-0:
Aqesome

PR #9714, opened 2023-03-10
Bugpatching

### LINKED ISSUE #9650: [BUG] Janitor can use holosign to force doors open.

### Game Version

23022319

### I was

Client

### OS

Windows 8, 10, 11

### What happened?

By trying to open a door you don't have access to you can force it open using a janitor holosign.

Holosign should not force doors open.

1. Get the janitor holo sign maker thing
2. Spam tries to open door while space clicking the holosign maker
3. Opens door

### Additional context

only seen someone else do this dunno how it really works

Lankercool:
### Purpose:

Fixes #9650 by implementing an obstacle check to the holoprojector.

### Notes:

I've added a condition to the holoprojector in which you cannot project a holosign to an impassable object (walls, closed doors, etc.) which takes care of the ability to open doors with holosigns.

### Changelog:

CL: [Improvement] Holosigns cannot occupy the same space as an impassable object.
CL: [Fix] Holosigns can no longer be used to force open doors.


MaxIsJoe (approved):
LGTM

corp-0 (approved):
You are so cool, Lankercool

PR #9719, opened 2023-03-11
Fixes not being able to throw stuff into disposals

### LINKED ISSUE #9718: [BUG] Can't throw anything into disposal bins anymore.

### Game Version

23031117

### I was

Client

### OS

Windows 8, 10, 11

### What happened?

Throwing objects at a disposal bin no longer has them go in at all, make a sound, or say anything about whether it bounced off or made it in.

Throwing an object at a disposal bin should have it make a sound, make it say something in chat if you made it in or not, and be able to go in.

1. Throw object at disposal bin

### Additional context

_No response_

Bod9001:
### Changelog:

 CL: [Fix] Fixes not being able to throw stuff into disposals

### DIFF
--- a/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
+++ b/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
@@ -1519,7 +1519,7 @@ public class UniversalObjectPhysics : NetworkBehaviour, IRightClickable, IRegist
 					{
 						if (isServer)
 						{
-							if ((UniversalObjectPhysics) bump == this) continue;
+							if (bump as UniversalObjectPhysics == this) continue;
 							bump.OnBump(this.gameObject, null);
 						}
 					}


UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
```diff
@@ -1519,7 +1519,7 @@ public void FlyingUpdateMe()
 					{
 						if (isServer)
 						{
-							if ((UniversalObjectPhysics) bump == this) continue;
+							if (bump as UniversalObjectPhysics == this) continue;
```
corp-0, line 1522: why is this different?

MaxIsJoe: it does some automatic null check in the background i believe

corp-0: ah just remembered, the first one will throw, the second doesn't throw but leaves the value as null

Bod9001: apparently throws an error if it can't cast with the other one

corp-0 (approved):

PR #9724, opened 2023-03-12
refactor: lowpop clearance now works as extra clearance

corp-0:
### Purpose
Before we had to set individually every clearance level granted for low pop, now when the round is lowpop it will fetch the normal clearance and lowpop clearance as an extra.

### Notes:
* Added tests for the new behaviour.
* Tested something that @FuntimeDudeski was trying to do and didn't work on his local. It works perfectly for me now.
* Corrected the method that was adding lowpop clearance. It follows the Mirror guidelines now and modifies the synced list instead of the actual list.

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Clearance/IClearanceSource.cs
+++ b/UnityProject/Assets/Scripts/Systems/Clearance/IClearanceSource.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Linq;
 
 namespace Systems.Clearance
 {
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Systems/Clearance/IClearanceSource.cs
```diff
@@ -12,15 +13,15 @@ public interface IClearanceSource
...
 		/// <summary>
 		/// Issued clearance for this clearance source. This list is consulted when the current round has a normal amount of population.
 		/// </summary>
 		IEnumerable<Clearance> IssuedClearance { get; }
 
 		/// <summary>
-		/// Issued clearance for this source. This list is consulted when the current round has low population.
+		/// Extra issued clearance for this source. When the round is lowpop, this list is consulted in addition to the normal
 		/// </summary>
 		IEnumerable<Clearance> LowPopIssuedClearance { get; }
```
Bod9001, line 26: wouldn't you haveto synchronise this as well?

corp-0: The property returns the synced fields

Bod9001 (approved):

PR #9725, opened 2023-03-13
Fixes some extra sync issues with the ALR

MaxIsJoe:
CL: [Fix] Fixed some issues with the Advanced Light Replacer not syncing clients between the client and server.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Lighting/LightSource.cs
+++ b/UnityProject/Assets/Scripts/Core/Lighting/LightSource.cs
@@ -18,18 +18,8 @@ namespace Objects.Lighting
 	public class LightSource : ObjectTrigger, ICheckedInteractable<HandApply>, IAPCPowerable, IServerLifecycle,
 		IMultitoolSlaveable
 	{
-		[SyncVar, SerializeField, FormerlySerializedAs("ONColour")]
-		private Color currentOnColor;
-
-		public Color ONColour
-		{
-			get => currentOnColor;
-			set
-			{
-				currentOnColor = value;
-				SetAnimation();
-			}
-		}
+		[SyncVar(hook = nameof(SetAnimation)), SerializeField, FormerlySerializedAs("ONColour")]
+		public Color CurrentOnColor;
 
 		public Color EmergencyColour;
 
@@ -99,7 +89,11 @@ namespace Objects.Lighting
 			ChangeCurrentState(InitialState);
 			traitRequired = currentState.TraitRequired;
 			RefreshBoxCollider();
-			SetAnimation();
+		}
+
+		private void Start()
+		{
+			lightSprite.Color = CurrentOnColor;
 		}
 
 		private void OnEnable()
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Core/Lighting/LightSource.cs
```diff
@@ -253,6 +253,7 @@ private void SetSprites()
 			RefreshBoxCollider();
 		}
 
+		[ClientRpc]
 		private void SetAnimation()
```
Bod9001: isn't this called by other stuff?

MaxIsJoe: Yes, and it should always sync when getting called.

Bod9001: but the problem is then it Wouldn't be called on server?

MaxIsJoe: ClientRpc calls on everything, including the server. And it should so values can be synced properly

corp-0: > but the problem is then it Wouldn't be called on server?

That's how it works. ClientRpc is called on server to update clients. From mirror's docs:

> ClientRpc calls are sent from objects on the server to objects on clients. They can be sent from any server object with a NetworkIdentity that has been spawned.

Bod9001: But this won't run on a self hosted build?

UnityProject/Assets/Scripts/Items/Tool/AdvancedLightReplacer.cs
```diff
@@ -84,9 +84,11 @@ private void AddLightToFixture(ItemSlot target, LightSource source, HandApply in
 		[TargetRpc]
 		private void LightTunerWindowOpen(NetworkConnection target)
 		{
+			UIManager.Instance.GlobalColorPicker.CurrentColor = currentColor;
 			UIManager.Instance.GlobalColorPicker.EnablePicker(SetColorToTune);
 		}
 
+		[Command(requiresAuthority = false)]
```
Bod9001, line 91: could you Put validation checks in this,  I think if you put a field for connection it automatically pick it up which client requested it, then it's easy to validate if they contract with the item

UnityProject/Assets/Scripts/Core/Lighting/LightSource.cs
MaxIsJoe: it will run on a self hosted build

UnityProject/Assets/Scripts/Items/Tool/AdvancedLightReplacer.cs
MaxIsJoe: I don't think this is a big security problem to require validation

Bod9001: I think it's always good practice

MaxIsJoe:
review again pls

Bod9001 (approved):

PR #9739, opened 2023-03-19
Explosion damaged balancing

Bod9001:
### Changelog:

CL: [Balance] balances explosion damag, heads should take more damage to be remove from explosions
CL: [Balance] Small explosions have a radius buff ( unless manually Overwritten)

### DIFF
--- a/UnityProject/Assets/Prefabs/Items/Implants/Organs/HumanHead.prefab
+++ b/UnityProject/Assets/Prefabs/Items/Implants/Organs/HumanHead.prefab
@@ -312,6 +312,11 @@ PrefabInstance:
       objectReference: {fileID: 0}
     - target: {fileID: 2780246240171570206, guid: e241039a7933eb347bd1a9bb67a9078a,
         type: 3}
+      propertyPath: _assetId
+      value: 4279644326
+      objectReference: {fileID: 0}
+    - target: {fileID: 2780246240171570206, guid: e241039a7933eb347bd1a9bb67a9078a,
+        type: 3}
       propertyPath: m_AssetId
       value: 78253771c2047f94c9ae40161a6be7ee
       objectReference: {fileID: 0}
@@ -374,6 +379,11 @@ PrefabInstance:
         type: 3}
       propertyPath: m_LocalEulerAnglesHint.z
       value: 0
+      objectReference: {fileID: 0}
+    - target: {fileID: 7822575272286095514, guid: e241039a7933eb347bd1a9bb67a9078a,
+        type: 3}
+      propertyPath: deadlyDamageInOneHit
+      value: 100
       objectReference: {fileID: 0}
     - target: {fileID: 8234622556944347693, guid: e241039a7933eb347bd1a9bb67a9078a,
         type: 3}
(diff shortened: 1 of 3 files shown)


AtnerNT:
Min radius of 5 seems like alot of that's in tiles.

Bod9001:
if it's a small explosion it does no damage anyway so it's not too bad

AtnerNT:
We have alot of low power explosives that a design to be like control charges. Like C4 and implant explosives, they are designed to only be 1-2tiles big. Instead of adding a linear +5, maybe just alter the formula so radius increases more with yield

UnityProject/Assets/Prefabs/Items/Implants/Organs/HumanHead.prefab
```diff
@@ -375,6 +380,11 @@ PrefabInstance:
       propertyPath: m_LocalEulerAnglesHint.z
       value: 0
       objectReference: {fileID: 0}
+    - target: {fileID: 7822575272286095514, guid: e241039a7933eb347bd1a9bb67a9078a,
+        type: 3}
+      propertyPath: deadlyDamageInOneHit
+      value: 100
```
MaxIsJoe, line 386: this doesn't fix the issue with explosions, this just hides it.
Smaller explosions shouldn't be doing over 30 damage

MaxIsJoe (changes requested):

Bod9001:
problem is most of the explosions have hardcoded Radius values

UnityProject/Assets/Prefabs/Items/Implants/Organs/HumanHead.prefab
Bod9001: well a C4 , For example would need enough to go through a wall, and a wall has more than a hundred health

corp-0: now that you mention it. We gave walls that kind of health because we didn't have damage deflection and people kept punching walls until they fell down. Maybe we need to go back and re-evaluate those values.

Bod9001: It still means that you need to do like a hundred damage on tile , To break the wall, and that's too high in terms of health

FuntimeDudeski: a shaped charge designed to destroy a wall would need to have much more focused damage at the center point than, say, a fragementation grenade, which could do lower but more even damage in a wider area. at least for non-reinforced walls, hitting or shooting them enough times with a damaging weapon should damage/ultimately destroy them

FuntimeDudeski: health granularity is important for simulating damage to tiles. not sure what the problem in principle is with a player having 100 health and a plastic explosive charge doing 100/150/200 if that level of damage can be localized to the tile it detonates on. being within a meter of most kinds of explosive detonations would gib a person.

FuntimeDudeski: it also wouldn't feel right if walls were immune to lower-yield explosives/frag grenades. a sufficient number of grenades should add up to an x4 as far as damage to (most) tiles is concerned.

PR #9743, opened 2023-03-20
Help menu navigation is now clearer

MaxIsJoe:
The help menu now has three buttons 
image

The Protips button will lead you to a list of protips only now
The Wiki button will open the [unitystation wiki](https://wiki.unitystation.org/index.php/Welcome_to_Unitystation) that we use currently
And lastly, "CALL MENTOR" will open the "OOC HELP" menu which has some great text that can help new-players understand what to use 
image

This is a continued effort to wrap up the protips PR, and this should be the final thing. After this is merged I'll contact Cross to see if the bounty is finished or requires some extra work to be done.

### Changelog:

CL: [Improvement] Improved help menu navigation.


corp-0:
If Call Mentor button now shows that other window, it should be renamed to something else. I also find "OOC Help" to be a weird name, why not just "Help"?

MaxIsJoe:
> If Call Mentor button now shows that other window, it should be renamed to something else. I also find "OOC Help" to be a weird name, why not just "Help"?

You know what? I agree. Changing it.

corp-0:
awesome, can you change the text on the button itself as well? From ``Call Metor`` to ``Call for help``

MaxIsJoe:
done

corp-0 (approved):

PR #9747, opened 2023-03-23
Artifact Random Chance

FuntimeDudeski:
### Purpose

Adds 25%, 50%, and 75% artifact spawners and a corresponding icon for mappers. Replaces all asteroid artifact locations with spawners--75% for Asteroid8 (wall-enclosed), 50% for Asteroid12 (igneous rock enclosed), 25% for all others.

### Changelog:

CL: [New] %chance artifact spawners w/ icon added for mappers to use.
CL: [Improvement] Artifact locations on asteroids now have a % chance to spawn an artifact instead of one always appearing.

image


Bod9001:
Why do they need to be separate prefabs? 
Can you just place it down and modify the chance yourself and have it as a scene override?

FuntimeDudeski:
> Why do they need to be separate prefabs? Can you just place it down and modify the chance yourself and have it as a scene override?

Based on the documentation and convo with Giles, you can get a specified % chance of spawning an item by setting the pool to 100 and the item spawn chance to the % chance you want. It wasn't clear to me how exactly the probability in the spawner prefab asset interacted with that value.

https://discord.com/channels/273774715741667329/285240893160685570/1078803279531737138
https://unitystation.github.io/unitystation/development/How-to-Random-Loot/

_If you set low probabilities in your pools, the spawner will do 5 attempts before interrupting everything and it won't spawn anything. Try group low prob pools with a 100 pool, so if the low prob fails it defaults to the 100 one._

_If you want your spawner to spawn something with low probs or nothing, use a 100 weight on its pool but make the items itself have low prob. An example of this is the Xeno Egg spawner, with 100 weight on its only pool but 2% chances of spawning the egg._


If setting the item spawn to 100% and the pool value to X works, it makes it sound like you need to multiply X by some figure to target the probability the item will spawn based on the 5 attempts. I guess I could do that, but am not 100% clear it works that way and testing might be tough since it'd be observational.

i.e. I think what it's describing is: if you set the item chance at 100% and the pool value at 50, you'd have a 96.875% chance of spawning the item (5 attempts @ 50%). With the pool value at 5, you'd have a 22.622% chance of spawning the item. With pool at 1, 4.9%... But I'm not totally sure. I can do it either way if that's indeed how it works. Mappers might find the flat % spawners more convenient to place and interpret when viewing others' maps.

Bod9001:
this seems more like an issue with the, UI display of percentages?, I could add a little inspector item to tell you the percentage chance? If that would be good

FuntimeDudeski:
> this seems more like an issue with the, UI display of percentages?, I could add a little inspector item to tell you the percentage chance? If that would be good

It's still a bit awkward given mappers are then manipulating two variables across two objects if they want a % chance in-between values (e.g. pool chance 1 = 4.9% , pool chance 2 = 9.6%, pool chance 3 = 14.1%), but if you think it's worth it to save on prefabs, sure.

I guess it's a matter of how much the spawner prefab list is likely to bloat if we're slightly more profligate with them. Having multiple simple spawners with different flat% would be too much for most object types but it made sense to me in the context of artifacts as a mapping convenience.

Bod9001:
True

Bod9001 (approved):

PR #9752, opened 2023-03-26
Fixes staging

Bod9001:
### Purpose
 <br>

### Changelog:

CL: [Balance] Default brightness for light replacer Has been corrected
CL: [Fix] Fixed hunger taking forever to be felt.
CL: [Fix] Fixed SMES not acting properly when turbine would produce intermittent power
CL: [Balance] Turbine now has rolling average of output, for scenarios where there is no SMES to handle the desynchronised nature of atmospherics and electrical


corp-0:
What happened with hunger?

Bod9001:
there is some initialisation code to,  set up values to make sure the default player is 1u nutriment = one minute of hunger, it was not updating the values properly so it was using the value set by the prefabs that were really low so, it would take forever for someone to get hungry

corp-0 (approved):

PR #9459, opened 2022-12-23
[Addressables] Part 1 of engine/content separation

MaxIsJoe:
# Purpose:
 
This PR is meant to get our feet wet and get things setup for the upcoming PRs that aims to rework how our assets get loaded and detected. Right now, there hasn't been any major changes that allow us to magically apply load in extra maps due to our bad over-use of ScriptableObjects and static references, but we have opened ourselves the path to load in scenes via addressables, and we can now experiment with updating the game **without having to re-build our executables every time we need to deploy and update**.

# Performance Improvements:

This PR introduces massive performance improvements inside the editor and on the client, as entering play-mode and loading scenes have dramatically had their load times reduced to upwards of 53.8% on average.

| Scenario       | Before Addressables (Seconds)          |  After Addressables (Seconds)   | Percentage Improvement |
| ------------- |:-------------:| -----:|  -----:|
| No Quick Load     | 21 | 16 | 23.8% |
| Quick Load      | 12 | 7 |  41.6% |
| PlayMode enter | 7 | 4 | 50% |
| Client Initialization | 6 | Instant | 100%
| Build Time | 1960~ | 800~ | 59.1%~

_* Tested on Windows 10 with an I7-4790k (@3.77GHz) and 16GBs of RAM.
Results are averaged between 10 tests
Numbers may vary on your own hardware._

# Notes:
Due to the major changes of this PR, here's a list of things that were removed or considered low priority currently and can be fixed in later PRs:

- Scene dependence for things like syndicate ships were removed due to them being underutilized and just being an obscure feature that basically has no use aside for one map.
- Away sites do not have human-readable names in admin tools.
- SubScene unit test has been removed due to its incompatibility with the addressables.
- (Untested) Discord bot is likely broken or will report incorrect names for scenes.
- Quick Build will not work as intended and will still build the entire asset bundle.
- Map pop checks needs to be reimplemented, as right now maps are completely chosen by random.
- The game will no longer force your computer to remember whats the next gamemode or map using PlayerPrefs and when the server gets rebooted again it will load the intial data from the gameConfig instead.

Also, the SubScene code is just.. bleh; I cleaned it up as much as I can.

# Changelog:

CL: [New] First stages of engine content separation are in effect. Expect bugs and some missing features.
CL: [New] All unitystation assets are now packaged in addressable bundles, meaning that basic mods for replacing the UI and textures are currently possible now. (Though it will require you to replace the entire bundle and build it from scratch)

PR #9754, opened 2023-03-28
Porting some of Addressables PR changes

### LINKED ISSUE #9514: [Rework Needed] Subsystems should be initialized as part of the server loading process, not outside of it.

### Game Version

14/1/2023

### I was

Server

### OS

Windows 8, 10, 11

### What happened?

Subsystems are initialized after all scenes are loaded from an event that developers may or may not forget to subscribe to manually and can be initialized more than once accidentally.

There should be a high level manager that handles all the initialization processes of our subsystems that also handles adding all behaviors automatically on Awake to the manager.

Just open an IDE and try adding any new behaviors to loading scenes.

The easiest way to cause issues with this is to set a scene as inactive via code when it gets spawned, then enable it. This will cause all subsystems on every map to get confused and break itself, including the newly activated scene.

### Additional context

This is bad design and does not allow for easy modability for scenes.

MaxIsJoe:
fixes #9514

This PR ports some of the cleaning I've done in #9459 since that PR is a bit stuck in rebase/conflict merges hell. All addressable stuff is not enabled, but some of the settings are there.

The reasoning behind this is:
1) In case I get stuck again for long periods of time and someone decides to touch these files again or any scene assets, they don't get in my way again and force me to re-do everything from scratch like now.
2) to close #9514
3) to reduce the gigantic size of the addressables PR changes, and split into chunks to detect any funny mistakes that may or may not lurk in the sea of changes that I've made.
4) To keep the important settings that are hard to configure on upstream/develop in-case I go on a two-month hiatus (again) away from addressables.

In the following PR after this is merged I'll work again on converting scene code to use `AssetReference` and load scenes from bundles... then its off to "game-breaking bug fixing land"


UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.cs
```diff
@@ -112,6 +112,12 @@ IEnumerator LoadSubScene(string sceneName, SubsceneLoadTimer loadTimer = null, b
 				RequestObserverRefresh.Send(sceneName);
 			}
 		}
+
+		loadedScenesList.Add(new SceneInfo
```
Bod9001, line 116: Make sure to set is dirty,  if it's a Synch  list

UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.SceneList.cs
```diff
@@ -172,7 +148,7 @@ IEnumerator ServerLoadAdditionalScenes(SubsceneLoadTimer loadTimer)
 	{
 		if (GameManager.Instance.QuickLoad)
 		{
-			yield return null;
+			yield break;
```
Bod9001, line 146: doesn't this just cause it to  get stuck?

MaxIsJoe: wdym? `break` should exit the function, not make get stuck?

UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.cs
MaxIsJoe: it already does in the function thats calling it... which good thing you reminded me; this can be further cleaned by moving that over to the load function

UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.SceneList.cs
Bod9001: but the UI gets stuck thinking it's loading, if I'm reading it right?

UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.cs
Bod9001: I think the load function gets called on client?

MaxIsJoe: ?

UnityProject/Assets/Scripts/Managers/SubSceneManager/SubSceneManager.SceneList.cs
```diff
@@ -246,13 +231,6 @@ public IEnumerator LoadWizard()
 
 		yield return StartCoroutine(LoadSubScene(pickedScene));
 
-		loadedScenesList.Add(new SceneInfo
```
Bod9001, line 289: does this get added automatically?

MaxIsJoe: yes

Bod9001 (approved):

Issue #5061, opened 2020-09-05
Dragging/pulling doesn't persist through gateways/portals


AvenDonn:
## Description
Introduced with #5007 

### Detailed Bug Report
If the host tries dragging through a teleporter, it works as intended. However, it seems that clients notice that the conditions for dragging are no longer met, and cause the server to disconnect the dragging. 

### Steps to Reproduce
Easiest to reproduce on SquareStation due to the Quantum Pads. The code is the same for both Gateways and Quantum Pads in this regard. (I know because I wrote it)
However, it _might_ be relevant to distinguish between pulling "things" and "people" as there are slight differences. 

1.  Start pulling someone/something
2.  Walk into a Quantum Pad, or a Station/World gateway (The big portals that open up 5-20 minutes into the round)
3.  Continue walking after being teleported

### Expected result:
You continue dragging whatever you were dragging

### Actual result:
You have to let go Jack.

### Demo:
![Sad conga noises](https://user-images.githubusercontent.com/1438444/91669918-72299f80-eb21-11ea-8bd8-14001ddc3deb.gif)

MrAsbestos:
I think this is already fixed

pyrosshade:
> 
> 
> I think this is already fixed

It is not

MaxIsJoe:
The code suggests that this is intentional behavior, not a bug.

PR #9763, opened 2023-03-31
fix for #5061

MaxIsJoe:
I'm marking this as draft to see why Bod did this in the first place
but this fixes teleporters dropping pulled objects when teleporting

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Gateway/TransportUtility.cs
+++ b/UnityProject/Assets/Scripts/Objects/Gateway/TransportUtility.cs
@@ -71,22 +71,13 @@ namespace Gateway
 			for (var node = linkedList.First; node != null; node = node.Next)
 			{
 				var currentObj = node.Value;
-				var previous = node.Previous?.Value;
-
-				//Disconnect pulling to make it not be a problem
-				currentObj.PullSet(null, false); //TODO Test without
 
 				//Transport current
 				TransportObject(currentObj, transportTo, doTileStep, maintRoomChanceModifier);
 
-				if (previous != null && currentObj.gameObject != null)
-				{
-					//There was another object before this one, pulling it. Re-establish pulling. (But only if the current object's gameObject is not null)
-					previous.CmdPullObject(currentObj.gameObject);
-				}
-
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Objects/Gateway/TransportUtility.cs
```diff
@@ -71,22 +71,13 @@ public static void TransportObject(UniversalObjectPhysics objectPhysics, Vector3
...
 				//Transport current
 				TransportObject(currentObj, transportTo, doTileStep, maintRoomChanceModifier);
 
-				if (previous != null && currentObj.gameObject != null)
-				{
-					//There was another object before this one, pulling it. Re-establish pulling. (But only if the current object's gameObject is not null)
-					previous.CmdPullObject(currentObj.gameObject);
-				}
-
 				//TODO: Make pulling acutally continue working across teleporters for clients, not just server
```
corp-0, line 78: when you pull someone and teleport, you stop pulling on client it seems. That's what I think he meant.

UnityProject/Assets/Scripts/Objects/Gateway/TransportUtility.cs
```diff
@@ -71,22 +71,13 @@ public static void TransportObject(UniversalObjectPhysics objectPhysics, Vector3
...
 				//TODO: Find a way to make the teleporter the teleport not all be on the same tile.
```
corp-0, line 79: when A and B teleport and A is pulling B, both of them occupy the same tile after teleporting.

MaxIsJoe: ...but that's exactly what happens now?

MaxIsJoe: nevermind, i double checked; it doesn't.
though i do have to ask, why does it matter? ss13 has similar behavior to what I've made in this PR, and it seems to be fine?

corp-0: I'm merely explaining  Bod's comments. I do think it is nice not to break the conga line when teleporting

PR #9765, opened 2023-03-31
fixes-cubees-in3D-Not-being-destroyed

Bod9001:


corp-0:
removed the changelog from the description since it isn't really relevant to the players

PR #9766, opened 2023-03-31
Skyboxes for 3D camera

MaxIsJoe:
image
ss14 skybox

image
Skyrat skybox

image
lavaland skybox

image
unitystation skybox

CL: [New] Odd.. When was the ceiling so colorful?
CL: [New] Since when was the roof so close to our heads?


corp-0:
can we repeat the ground as ceiling?

MaxIsJoe:
idk, it probably is possible but I'll have to spend some time figuring out how to do it because currently the floor is tied to the tilemap system thats a bit hard to manipulate in such cases

MaxIsJoe:
almost everything that is bill-boarded is an actual `gameObject` that can be manipulated
tiles in tilemaps don't work that way because they're not treated as gameObjects, so there is a very severe engine limitation to what you're allowed to do with them.

Bod9001 (approved):

Issue #6774, opened 2021-06-10
Non Latin Characters do not display correctly in the chatbox or on paper


Elijahrane:
### Environment
- Game Version: 21060821
- OS Version: Linux standalone

### Detailed description of bug
Non Latin characters (Cyrillic, CJK, etc) do not render correctly in the chatbox or on paper or in the Centcom Announcements. They seem to work in the bottom left mouseover examine, the PDA, ahelp, mhelp, admin chat, notes, and the input preview before you hit enter.

### Steps to reproduce

1. Start saying something.
2. Type something in non latin characters.
3. Notice how it displays as boxes (i.e. the glyph is missing) in the chatbox when you hit enter.

ThatDan123:
Due to the Font used as the chat and paper are TMP text so use different fonts

Bod9001:
need image for the, Bug Reporting contest

MaxIsJoe:
I think this has been fixed already

badbon:
I can input Russian but not UTF-8 letters. To be fair I haven't seen much games that support UTF-8, so that's probably intended

craftxbox:
Reproducible as of 25032122 with Japanese characters

PR #9773, opened 2023-04-02
fix Cyrillic letters turning into ☐☐☐

DetectiveEnotsky:
### Purpose
 Basically, most of the Cyrillic symbols are turning into ☐☐☐ when typing them into chat. This should fix it, by adding a fallback font asset with cyrillic symbols. Partially fixеs #6774

### Notes:
 ATTENTION. This code may not work. I haven't tested it, yet(and i'm not going to), because my PC explodes every time I press Build and Run in my UnityEditor. 
UPD: I tested it. Worked as intended.  <br>
image

### Changelog:

CL: [Fix] Fixes Cyrillic letterns turning into  ☐


UnityProject/Packages/packages-lock.json
```diff
@@ -48,7 +48,7 @@
       "dependencies": {}
     },
     "com.unity.2d.tilemap.extras": {
-      "version": "2.2.4",
+      "version": "2.2.5",
```
corp-0: I think you shouldn't push this file

DetectiveEnotsky: Yeah, probably. I'll try and do it now.

DetectiveEnotsky: Okie dokie, fixed in the commit 7362e4426c635b30629ec37d210a48a3f654fb37...

DetectiveEnotsky:
Okie dokie, now it's ready. I think.

Bod9001 (approved):

PR #9776, opened 2023-04-02
Mod tool for max

Bod9001:
max most properly wants to work on this more


corp-0:
If this is the same mod tool I suggested time ago, then it has a github repo and can be installed using the unity package manager

Bod9001:
This was installed via a unity package

corp-0:
But this pr is pushing the downloaded files

MaxIsJoe:
There;s one important question:
Currently, the ModTool is incredibly slow because Unity takes forever to create the Exporter/Mods with how big our project is right now and can take hours on older hardware; which makes it very inefficient and unintuitive to use.

Should we fork our own repo of ModTool and fix this by zipping the essential contents instead of relying on unity's very slow importer?
We still have to modify ModTool either way to use whitelists instead of blacklists, so a fork would make more sense for us.

Bod9001:
a Forking sounds like a good idea

corp-0:
yes, forking was in my plans

PR #9783, opened 2023-04-08
Dynamic alerts

Bod9001:
### Changelog:

CL: [Improvement] new system for alerts, allowing for dynamic alerts


UnityProject/Assets/Scripts/HealthV2/Living/LivingHealthMasterBase.cs
```diff
@@ -1912,8 +1971,74 @@ public void ExposePressureTemperature(float EnvironmentalPressure, float Environ
 				}
 			}
 
-			healthStateController.SetTemperature(ExtremistTemperature);
-			healthStateController.SetPressure(ExtremistPressure);
+			if (ExtremistPressure != ExtremistPressureCash)
```
corp-0, line 1974: this might be polluting the class unnecesarilly. Can move the alerts to a dedicated class and make LHB use it (or even maybe the other way around). LHB has way too many concerns.

Bod9001: humm, Where would be the best place to put it?

corp-0: it probably deserves to be its own class somewhere inside the UI namespace

Bod9001: to be fair this is living on the player, and doesn't care if there's a player Connected or not

PR #9784, opened 2023-04-09
add Ukrainian letters support and fix terrible font size

DetectiveEnotsky:
### Purpose
To this moment, Ukrainian letters Ї ї, Є є, Ґ ґ were not supported and have been turned into ☐. This should fix it.
I also increased PTSheriff (the fontmesh used to support cyrillic letters) font scale from 1 to 1.3, because it looked small, terrible and ulmost unreadable.

### Notes:
**Only 0b027e2 commit needs to be pushed. Other commits are unrelevant(so please, squash them) and are there just because I'm a hecking retard who doesn't know how to work with Git.**

### Changelog:

CL: [Improvement] Increased font size for Cyrillic letters fontmesh

CL: [Fix] Ukrainian letters not being supported


DetectiveEnotsky:
download

MaxIsJoe:
Can you share screenshots of the font size change? If i recall, changing that file directly affects all of the game.

DetectiveEnotsky:
> Can you share screenshots of the font size change? If i recall, changing that file directly affects all of the game.

It shouldn't. I changed it in the fontmesh PTSheriff so it kinda should.. I dunno, change the scale of only PTSheriff?

I'll make screenshots later

Bod9001 (approved):

DetectiveEnotsky:
Ok. I tested it in the game. It doesn't change other fonts scales, but it breaks existing things, so i'll just redo it in better way without polluting everything with my commits.

PR #9790, opened 2023-04-12
 Fix for  #8789

### LINKED ISSUE #8789: Sliding glass doors need to be manually opened from their back side
### Environment
- Game Version: 22060401
- OS Version: win10

Walking through a sliding glass door from its front causes it to open automatically as intended. However, walking into it from the other side does not open it automatically, having you open it manually by clicking on it

### Steps to reproduce

1. Go behind a sliding glass door
2. Walk into it
3. Go to medical to patch up your head bruise damage

coolcat1711:
### Purpose
Fixes #8789

### Notes:
Adds a second layer of bump checking for tiles that have directional occlusion. This allows you to trigger bumps on non-creature objects on the same tile.

### Changelog:

CL: [Fix] Allows creatures to trigger bump actions for directional objects in the same tile (Windoors).

### DIFF
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/ObjectLayer.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/ObjectLayer.cs
@@ -139,7 +139,7 @@ public class ObjectLayer : Layer
 			}
 
 			//If you can't bump anything on an adjacent tile, you may be blocked by a bumpable object on your current tile (probably a windoor)
-			if (!Bumps.Any())
+			if (Bumps.Any() == false)
 			{
 				foreach (var objectOnTile in Matrix.Get<UniversalObjectPhysics>(originalFrom, CustomNetworkManager.IsServer))
 				{
--- a/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/ObjectLayer.cs
+++ b/UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/ObjectLayer.cs
@@ -138,6 +138,23 @@ public class ObjectLayer : Layer
 				}
 			}
 
+			//If you can't bump anything on an adjacent tile, you may be blocked by a bumpable object on your current tile (probably a windoor)
+			if (!Bumps.Any())
+			{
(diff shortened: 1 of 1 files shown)


corp-0:
looks good to me but I will wait on Bod's review since he designed the physics and collisions code. @Bod9001

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/ObjectLayer.cs
```diff
@@ -138,6 +138,23 @@ public float GetObjectResistanceAt(Vector3Int position, bool isServer)
 				}
 			}
 
+			//If you can't bump anything on an adjacent tile, you may be blocked by a bumpable object on your current tile (probably a windoor)
+			if (!Bumps.Any())
```
Bod9001: Bumps.Any() == false

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/ObjectLayer.cs
```diff
@@ -138,6 +138,23 @@ public float GetObjectResistanceAt(Vector3Int position, bool isServer)
...
+			{
+				foreach (var objectOnTile in Matrix.Get<UniversalObjectPhysics>(originalFrom, CustomNetworkManager.IsServer))
+				{
+					if (objectOnTile.Intangible) continue;
+					//Prevents living creatures from bumping themselves (or other living creatures) if on the same tile.
+					if (objectOnTile.GetComponent<HealthV2.HealthStateController>() != null) continue;
```
Bod9001, line 148: Might be better to use context instead

Bod9001:
otherwise looks good

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/ObjectLayer.cs
coolcat1711: Which piece would context replace? objectOnTile and context are both "UniversalObjectPhysics" classes and I think context only accounts for the current player/creature moving instead of any additional creatures occupying a tile (which would want to ignore being bumped if they're on the same tile)

Bod9001: ok, fair enough

Bod9001 (approved):

PR #9795, opened 2023-04-13
Fixes #9694

### LINKED ISSUE #9694: Clicking a vomit/paint splat with a reagent container transfers reagents from the container to the splat

### Game Version

23030314

### I was

Client

### OS

Windows 8, 10, 11

### What happened?

If you have a reagent container, such as a spray bottle with space cleaner, and click the vomit/paint splat it will transfer space cleaner from the bottle to the splat.

Splats should be out only regaents containers. This interaction should have ended with the splat being cleaned by the space cleaner instead.

1. Have a bottle with space cleaner
2. Find a vomit/paint splat
3. Click it directly on its sprite

### Additional context

This should be a good first issue. Just finding the splat prefab and changing parameters in the ReagentContainer component should be enough.

MaxIsJoe:
Fixes #9694

CL: [Fix] You can no longer accidentally dump reagents into floor splats

### DIFF
--- a/UnityProject/Assets/Scripts/ChemistryComponents/ReagentContainer.TransferInteraction.cs
+++ b/UnityProject/Assets/Scripts/ChemistryComponents/ReagentContainer.TransferInteraction.cs
@@ -27,6 +27,8 @@ namespace Chemistry.Components
 	{
 		[Header("Transfer settings")]
 
+		[SerializeField] private bool isInteractable = true;
+
 		[Tooltip("If not empty, another container should have one of this traits to interact")]
 		[FormerlySerializedAs("TraitWhitelist")]
 		[FormerlySerializedAs("AcceptedTraits")]
@@ -74,6 +76,7 @@ namespace Chemistry.Components
 
 		public bool WillInteract(HandApply interaction, NetworkSide side)
 		{
+			if (isInteractable == false) return false;
 			if (!DefaultWillInteract.Default(interaction, side)) return false;
 
 			var playerScript = interaction.Performer.GetComponent<PlayerScript>();
--- a/UnityProject/Assets/Prefabs/Decal/Resources/ChemSplat.prefab
+++ b/UnityProject/Assets/Prefabs/Decal/Resources/ChemSplat.prefab
@@ -52,6 +52,7 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: c77080d7d322e0d46b715644eb4b0ab9, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
+  syncDirection: 0
   syncMode: 0
   syncInterval: 0.1
   Cleanable: 1
@@ -74,6 +75,7 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 322493e979844b028b86387c7a0aa50c, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
+  syncDirection: 0
   syncMode: 0
   syncInterval: 0.1
   matrixDebugLogging: 0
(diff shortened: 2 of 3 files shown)


UnityProject/Assets/Scripts/ChemistryComponents/ReagentContainer.TransferInteraction.cs
```diff
@@ -27,6 +27,8 @@ public partial class ReagentContainer : ICheckedInteractable<HandApply>, //Trans
 	{
 		[Header("Transfer settings")]
 
+		[SerializeField] private bool isInteractable = true;
```
Bod9001, line 30: I don't think you need at this as an option to make it not intractable with Reagent containers already?

MaxIsJoe: you kinda do

Bod9001: image

MaxIsJoe: No transfer might be doing other stuff in the background, this is why we disable interactions completely on `WillInteract()` rather than using that enum in wherever it is being used.

corp-0: that's an odd answer. The transfer mode is what you need to change to avoid transfering reagents into the splats. That's the issue you are trying to fix.

MaxIsJoe: yet the no transfer mode doesn't work, but my fix does; curious.

corp-0: if it doesn't work, maybe see why it doesn't? Instead of adding another bool to the component.

MaxIsJoe: This is a symptom of how our interaction system works, it doesn't care about the order of things, so you'll have instances where it will do things that its not supposed to do or have some interactions blocked completely unless you tell `WillInteract()` to return false.

This isn't the first time I fix this issue, and you'll recall a similar fix I did for an older issue that you reviewed regarding hand lablers.
The proper solution is to add both a priority system and the ability for us to explicitly specify how interactions should interact with each other (pass to the next interaction, block next interaction, can both interactors work with each other and how, etc)
but that requires a massive rework that will take forever because we have hundreds of interactions in this game that we'll have to go through for such a rework. Will probably leave it as a task for the simplification of the intent system when we actually do it.

MaxIsJoe: the closest example for how an interaction rework should look like is actually in this same reagent system as well; take a look at `TransferResult` as a concept and see how we can apply this bad boy to interactions.
It would give us finer control over things.

corp-0: well, regardless of all you said, we can't just make them non-interactable because we sometimes want to interact with them. Grabbing samples from blood splats or vomit and analyse their content. Grabbing sample from a whatever chemical splat to see what is there.

corp-0: the issue is not that they are interactable but that the container is allowed to receive transfer when it should be an out only operation.

MaxIsJoe: > we can't just make them non-interactable because we sometimes want to interact with them. Grabbing samples from blood splats or vomit and analyse their content. 

image

...good point.
I'll revise my solution.

PR #9816, opened 2023-04-28
Pipe fixes and Improvements

MaxIsJoe:
part of #4654 

CL: [New] Pipes were missing an interaction for humanoids to crawl into them without the need of a disposal bin or such. This has been added.
CL: [Improvement] Slightly Improved some pipe crawling behaviors to make them less jank.
CL: [Improvement] Lifted a restriction on pipe crawl movement that restricted the player's movement towards a pipe they've previously crawled from.
CL: [Fix] Fixed an issue where the traversal system could not detect valid pipes to crawl out of.
CL: [Fix] Fixed an issue where players would get stuck on one tile after crawling out of a pipe.
CL: [Fix] Fixed a massive issue with virtual containers not getting cleared after they were no longer needed. This issue may have caused unnecessary CPU usage on servers, as hundreds of virtual containers were still being processed due to them not getting cleared out of the instance list.

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Disposals/DisposalTraversal.cs
+++ b/UnityProject/Assets/Scripts/Systems/Disposals/DisposalTraversal.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using UnityEngine;
 using Objects.Disposals;
@@ -110,7 +111,27 @@ namespace Systems.Disposals
 
 			if (nextPipe == null)
 			{
-				if (virtualContainer.SelfControlled == false) EjectViaPipeEnd();
+				if (virtualContainer.SelfControlled)
+				{
+					var directionsToCheck = new List<Vector3Int>
+						{
+							new Vector3Int(-1, 0, 0),
+							new Vector3Int(1, 0, 0),
+							new Vector3Int(0, -1, 0),
+							new Vector3Int(0, 1, 0)
+						};
+
+					foreach (var direction in directionsToCheck)
+					{
+						if (GetPipeAt(out var pipe, currentPipeLocalPos + direction,
+							    requiredSide: nextPipeRequiredSide) == null) continue;
+						if (pipe.PipeType != DisposalPipeType.Terminal) continue;
+						EjectViaPipeEnd();
+						break;
+					}
+					return;
+				}
+				EjectViaPipeEnd();
 				return;
 			}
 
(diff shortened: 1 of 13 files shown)


UnityProject/Assets/Scripts/Systems/Disposals/DisposalTraversal.cs
```diff
@@ -262,8 +299,10 @@ private void TryDamageTileFromEjection(Vector3Int localPosition)
 		/// </summary>
 		private void DespawnContainerAndFinish()
 		{
-			_ = Despawn.ServerSingle(virtualContainer.gameObject);
+			virtualContainer.SelfControlled = false;
 			TraversalFinished = true;
+			if (virtualContainer.SelfControlled) DisposalsManager.Instance.FinishDisposal(this);
+			_ = Despawn.ServerSingle(virtualContainer.gameObject);
```
NoooneyDude: No point for the if condition, right?

Also, disposals manager is responsible for removing finished instances from its own list, so it would now be unnecessarily removing the same instance again.

Edit: ah I see, the manager was since updated and removal was being short-circuited. Still, I think an individual traversal instance shouldn't be concerned with the manager's list.

I can't seem to insert a suggestion 😢 but yeah, just need to remove line 304 👍

NoooneyDude:
Nice to see we have pipe crawling now 💪 have you considered adding the new interaction to all pipes? Would be nice to be able to crawl into any pipe end, even exploded ones provided it has a connection without an adjacent pipe. Haven't looked into it much though, not sure if worth the effort

UnityProject/Assets/Scripts/Systems/Disposals/DisposalTraversal.cs
MaxIsJoe: > No point for the if condition, right?

No, because it skips virtual containers that are self controlled when processing. self controlled ones use movement events to update, instead of sitting on the disposals manager's update loop.

MaxIsJoe:
> Nice to see we have pipe crawling now 💪 have you considered adding the new interaction to all pipes? Would be nice to be able to crawl into any pipe end, even exploded ones provided it has a connection without an adjacent pipe. Haven't looked into it much though, not sure if worth the effort

I've added it to all terminals, not sure what other pipes require this.

UnityProject/Assets/Scripts/Systems/Disposals/DisposalTraversal.cs
NoooneyDude: Alright, but SelfControlled will always be false at that point.

Won't it? Surprised your IDE didn't highlight it or Codacy..

And from what I can tell the self-controlled ones are still in the manager's loop, just being short-circuited. Setting self-controlled false will mean the TraversalFinished property is checked and so will be removed from its list. So all you need is line 302.

MaxIsJoe: oh, you're right.

MaxIsJoe: fixed

NoooneyDude (approved):

PR #9818, opened 2023-04-28
Adds a debug command to resetting movement

MaxIsJoe:
CL: [New] Added a debug command to resetting movement. This will allow players to help identify issues related to movement in the future, and let them possibly fix problems by themselves without admin intervention so they can continue playing the game. To use the new command, use "reset-movement" in the F5 console.

### DIFF
--- a/UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
+++ b/UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
@@ -775,6 +775,8 @@ namespace IngameDebugConsole
 				return;
 			}
 			PlayerManager.LocalPlayerScript.PlayerNetworkActions.CmdResetMovementForSelf();
+			Logger.Log("[Console Command] - Movement Reset Successfully. " +
+			           "If you're still stuck, please report this and any errors you might find in the console on github/discord.", Category.DebugConsole);
 		}
 	}
 }
--- a/UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
+++ b/UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
@@ -764,5 +764,17 @@ namespace IngameDebugConsole
 			}
 
 		}
+
+		[ConsoleMethod("reset-movement", "Resets all movement values. Helpful if you get stuck for no reason.")]
+		public static void ResetMovementStats()
+		{
+			if(IsAdmin() == false) return;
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
```diff
@@ -764,5 +764,19 @@ private static void GenerateCyborg()
 			}
 
 		}
+
+		[ConsoleMethod("reset-movement", "Resets all movement values. Helpful if you get stuck for no reason.")]
+		public static void ResetMovementStats()
+		{
+			if(IsAdmin() == false) return;
```
corp-0: PR description says you don't need an admin

UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
```diff
@@ -990,4 +990,11 @@ public void CmdDoEmote(string emoteName)
 	{
 		EmoteActionManager.DoEmote(emoteName, playerScript.gameObject);
 	}
+
+	[Command]
+	public void CmdResetMovementForSelf()
+	{
+		playerScript.playerMove.ResetEverything();
```
corp-0, line 997: what does this do?

eurtod:
HUGE if true. this will save A LOT of relogging. praise be!

MaxIsJoe:
> HUGE if true. this will save A LOT of relogging. praise be!

This doesn't mean that it will solve everything, it just means we get to narrow down a list of suspect problems that could be causing movement issues.

UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
MaxIsJoe: Basically gets rid of things like sliding, flying, etc and resets movement to a non-moving state. An issue I found while working on #9816 is that when the player gets launched out, they are still considered flying/sliding which causes the game to prevent the player from moving. 
If other issues follow a similar pattern, we will now know that the issue is from how we handle all these variables that are in the movement script and narrow it down there.

UnityProject/Assets/IngameDebugConsole/Scripts/DebugLogUnitystationCommands.cs
MaxIsJoe: oh shoot, you're right.

MaxIsJoe: fixed.

corp-0 (approved):

PR #9819, opened 2023-04-28
Suicide command fix

### LINKED ISSUE #9817: [BUG] Suicide command does not work.

### Game Version

23042416

### I was

Client

### OS

Windows 8, 10, 11

### What happened?

Pressing f5 and typing "suicide" does not kill you. It displayed the _falls limp eyes dead and lifeless_ message but does not kill the player.

It should kill the player.

1. Press f5
2. Type suicide in the command prompt
3. Press enter
4. Keep living ):

### Additional context

_No response_

MaxIsJoe:
fixes #9817 

CL: [Fix] Fixed an issue where the suicide console command did not work.


UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
```diff
@@ -985,6 +987,34 @@ public void CmdServerReplaceItemInInventory(GameObject gameObjectSent, uint id,
...
+	{
+		if (playerScript.TryGetComponent<AiPlayer>(out var aiPlayer))
+		{
+			aiPlayer.Suicide();
+			return;
+		}
+
+		if (playerScript.TryGetComponent<BlobPlayer>(out var blobPlayer))
+		{
+			blobPlayer.Death();
```
corp-0, line 1001: not telling you to do it, but it would be great if we moved this stuff to interfaces like IGIbbable, so we have ISuicidable, IDeath and whatever. We would just get the interface then intstead of checking types.

MaxIsJoe: It would be even better as well if both the AI and Blob used `LivingHealthMasterBase`. but alas, they don't, because our current system did not support that. Thankfully, thanks to Bod's changes, we can move them to use that class soon and maybe even all other mobs as well.

corp-0 (approved):

PR #9833, opened 2023-05-04
fixes an NRE and bug with trauma

MaxIsJoe:
CL: [Fix] Fixed newly imported body parts not having trauma enabled on them by default.
CL: [Fix] Fixed an NRE with trauma trying to access it's related body part when it was destroyed or removed.

### DIFF
--- a/UnityProject/Assets/Scripts/HealthV2/Living/Damage/Trauma/TraumaLogic.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/Damage/Trauma/TraumaLogic.cs
@@ -46,6 +46,13 @@ namespace HealthV2
 		/// </summary>
 		protected void GenericStageProgression()
 		{
+			if (gameObject == null || bodyPart == null)
+			{
+				Logger.LogWarning(
+					"[TraumaLogic/GenericStageProgression] - bodyPart might have been destroyed during other stage progressions. Skipping..");
+				return;
+			}
+			if (stages == null || stages.ContainsKey(currentStage + 1) == false) return;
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/HealthV2/Living/Damage/Trauma/TraumaLogic.cs
```diff
@@ -46,6 +46,13 @@ public virtual void OnTakeDamage(BodyPartDamageData data)
 		/// </summary>
 		protected void GenericStageProgression()
 		{
+			if (gameObject == null || bodyPart == null)
+			{
+				Logger.LogWarning(
```
Bod9001, line 51: this should be errors since it Shouldn't be happening in the first place

MaxIsJoe: no, this happens **after** the second stage that turns these into null.

Bod9001: shouldn't it be removed From the update if it's null

MaxIsJoe: its an event, not an update.
It will go through the entire list regardless of what happens, so we tell it that if "hey, these two are missing or the stage is missing; assume that something has changed and return."

Bod9001: ok

Bod9001 (approved):

PR #9837, opened 2023-05-08
Remove Toxiin generation When dead

Bod9001:
### Changelog:

CL: [Balance] Remove Toxiin generation When dead
CL: [fix] Fixes face huggers not implanting eggs Properly
CL: [fix] Fixes Suit fabricators not Working

### DIFF
--- a/UnityProject/Assets/Scripts/HealthV2/Living/PolymorphicSystems/NaturalChemicalReleaseSystem.cs
+++ b/UnityProject/Assets/Scripts/HealthV2/Living/PolymorphicSystems/NaturalChemicalReleaseSystem.cs
@@ -134,7 +134,7 @@ namespace HealthV2.Living.PolymorphicSystems
 			float Multiplier = HeartEfficiency;
 			if (HeartEfficiency == 0)
 			{
-				Multiplier = 0.0025f;
+				return;
 			}
 
 			foreach (var KVP in Toxicity)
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/HealthV2/Living/PolymorphicSystems/NaturalChemicalReleaseSystem.cs
```diff
@@ -134,7 +134,7 @@ public void ToxinGeneration(float HeartEfficiency)
 			float Multiplier = HeartEfficiency;
 			if (HeartEfficiency == 0)
```
corp-0, line 135: this doesn't say "dead" specifically. Is there any situation where you have a bad heart but are still alive?

Bod9001: You can be alive with no heart but None of the systems work such as Oxygenation

Bod9001:
Scene changes are to add the Suit fabricators to APC Since they powered now

UnityProject/Assets/Scripts/HealthV2/Living/PolymorphicSystems/NaturalChemicalReleaseSystem.cs
corp-0: would it be better to be more specific and check if the creature is dead instead?

PR #9838, opened 2023-05-09
New hub fields for playercount & real round time

craftxbox:
### Purpose
Reports three new fields to the current hub api for future use, 
RoundTime: Mirror of GameManager.Instance.RoundTimeInMinutes, Gives an accurate display of how long a round has gone on for (that doesn't include janky math on the existing IngameTime variable). This field is OPTIONAL and may not appear on the API if it's not being reported.
PlayerCountMax: Should be pretty self-explanatory. If it's not reported, this value is 0 through the api.
Passworded: ditto. If a connection password is set, this is true. Else false (also if not reported)

### Changelog:
Not userfacing, No CL.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Database/ServerData.ServerStatus.cs
+++ b/UnityProject/Assets/Scripts/Core/Database/ServerData.ServerStatus.cs
@@ -161,6 +161,10 @@ namespace DatabaseAPI
 				status.CurrentMap = SubSceneManager.ServerChosenMainStation;
 			}
 
+			status.Passworded = !string.IsNullOrEmpty(config.ConnectionPassword);
+            status.RoundTime = GameManager.Instance.RoundTimeInMinutes.ToString();
+            status.PlayerCountMax = GameManager.Instance.PlayerLimit;
+
 			status.GameMode = GameManager.Instance.GetGameModeName();
 			status.IngameTime = GameManager.Instance.roundTimer.text;
 			if (PlayerList.Instance != null)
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Core/Database/ServerData.ServerStatus.cs
```diff
@@ -161,6 +161,10 @@ IEnumerator SendServerStatus()
 				status.CurrentMap = SubSceneManager.ServerChosenMainStation;
 			}
 
+			status.Passworded = !string.IsNullOrEmpty(config.ConnectionPassword);
```
Bod9001: == false

Bod9001:
does this need any other PRs to be Merged first?

UnityProject/Assets/Scripts/Core/Database/ServerData.ServerStatus.cs
craftxbox: I can do that but that doesn't seem very easy to understand in a variable declaration

craftxbox:
> does this need any other PRs to be Merged first?

Nope, This makes fields available for stationhub to read should it want to.

UnityProject/Assets/Scripts/Core/Database/ServerData.ServerStatus.cs
Bod9001: thouggh it is coding practice to do == false, 
I think it's fine with stuff like !NoPassword, 
where it is a direct inverse of a bool that is already there

Bod9001 (approved):

PR #9863, opened 2023-05-21
Improved laying down behavior

MaxIsJoe:
This PR aims to take the first steps into fixing a long-standing issue with player rotations not being consistent.

We're moving away from whole gameObject rotations and are only rotating the sprites from now on.
There's also a timed safe-guard for when things go wrong, and we need to ensure that the sprite is shown correctly.

### Changelog:

CL: [Improvement] Improved laying down behavior to ensure that rotations are always consistent and added a way for them to attempt to self-correct every 5 seconds.
CL: [Fix] Fixed dead players standing up again when teleporting.
CL: [Fix] Fixed bug where pulled players would suddenly stand up after getting teleported while being dragged by another player.
CL: [Fix] Fixed players randomly losing their correct laying down state in the event of them disappearing and then reappearing again. 
CL: [Fix] Fixed chat bubble always being shifted downwards after laying down once.
CL: [Fix] Fixed an issue where sound would not play consistently on quantum pads.

### DIFF
--- a/UnityProject/Assets/Scripts/Player/LayDown.cs
+++ b/UnityProject/Assets/Scripts/Player/LayDown.cs
@@ -1,4 +1,5 @@
-﻿using Mirror;
+﻿using HealthV2;
+using Mirror;
 using Player.Movement;
 using UnityEngine;
 
@@ -7,6 +8,7 @@ namespace Player
 	public class LayDown : NetworkBehaviour
 	{
 		[SerializeField] private Transform sprites;
+		[SerializeField] private LivingHealthMasterBase health;
 		[SerializeField] private Rotatable playerDirectional;
 		[SerializeField] private PlayerScript playerScript;
 		[SerializeField] private Util.NetworkedLeanTween networkedLean;
@@ -20,6 +22,7 @@ namespace Player
 		{
 			playerScript ??= GetComponent<PlayerScript>();
 			playerDirectional ??= GetComponent<Rotatable>();
+			health ??= GetComponent<LivingHealthMasterBase>();
 		}
 
 		private void OnEnable()
@@ -40,13 +43,12 @@ namespace Player
 
 		public void EnsureCorrectState()
 		{
-			if (IsLayingDown) { LayingDownLogic(true); }
-			else { UpLogic(true); }
+			if (IsLayingDown || health.IsDead) { LayingDownLogic(true); } else { UpLogic(true); }
 		}
 
 		public void OnLayDown(bool oldValue, bool newValue)
 		{
-			if (newValue)
+			if (newValue || health.IsDead)
 			{
 				LayingDownLogic();
 			}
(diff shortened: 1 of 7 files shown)


corp-0:
> We're moving away from whole gameObject rotations and are only rotating the sprites from now on.

This should have been the case from the very beginning and it was like that before, dunno when we changed to move the game object again. Anyways, thanks for doing this.

UnityProject/Assets/Scripts/Player/LayDown.cs
```diff
@@ -0,0 +1,98 @@
...
+		{
+			if (CustomNetworkManager.IsServer == false)
+			{
+				UpdateManager.Remove(CallbackType.PERIODIC_UPDATE, EnsureCorrectState);
+			}
+		}
+
+		public void EnsureCorrectState()
+		{
+			if (IsLayingDown || health.IsDead) { LayingDownLogic(true); } else { UpLogic(true); }
```
corp-0: please do a single line per block. The body of each if and else should have their own line to live on.

UnityProject/Assets/Scripts/Player/LayDown.cs
```diff
@@ -0,0 +1,98 @@
...
+		private void LayingDownLogic(bool forceState = false)
+		{
+			if (forceState) sprites.localRotation = layingDownRotation;
+			foreach (SpriteRenderer spriteRenderer in GetComponentsInChildren<SpriteRenderer>())
+			{
+				spriteRenderer.sortingLayerName = "Bodies";
+			}
+			playerScript.PlayerSync.CurrentMovementType  = MovementType.Crawling;
+			if (CustomNetworkManager.IsServer) playerDirectional.LockDirectionTo(true, playerDirectional.CurrentDirection);
+			if (CustomNetworkManager.IsServer) playerScript.OnLayDown?.Invoke();
```
corp-0: codacy is right, you just checked the very same thing. Merge it to that if statement. Make a body for the if and wrap it with brackets.

UnityProject/Assets/Scripts/Player/LayDown.cs
MaxIsJoe: > > We're moving away from whole gameObject rotations and are only rotating the sprites from now on.
> 
> This should have been the case from the very beginning and it was like that before, dunno when we changed to move the game object again. Anyways, thanks for doing this.

It was changed when Bod was working on the new movement to support players having these funny rotation moments when getting thrown around

MaxIsJoe: done

UnityProject/Assets/Scripts/Player/LayDown.cs
MaxIsJoe: fixed

UnityProject/Assets/Scripts/Player/LayDown.cs
```diff
@@ -0,0 +1,108 @@
...
+
+		private void OnDisable()
+		{
+			if (CustomNetworkManager.IsServer == false)
+			{
+				UpdateManager.Remove(CallbackType.PERIODIC_UPDATE, EnsureCorrectState);
+			}
+		}
+
+		public void EnsureCorrectState()
```
Bod9001, line 28: why have update loop?

MaxIsJoe: safe guard, ensures the correct state every 5 seconds

Bod9001: should be confident enough in code To not need a safeguard

MaxIsJoe: True, but until I curbstomb every mistake we make in code, this will do.
Also, it's not an uncommon thing for things to have safeguards, check any triple A game in the past 15 years, and you'll see the most common safeguard that is used everywhere; The anti-falling into the void safeguard.

Bod9001 (approved):

PR #9878, opened 2023-05-29
Try catchiing this

Bod9001:
Adds a bunch of try catch players Spawn code, don't want a rogue error stopping people from spawning

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Lifecycle/PlayerSpawn.cs
+++ b/UnityProject/Assets/Scripts/Core/Lifecycle/PlayerSpawn.cs
@@ -353,6 +353,8 @@ public static class PlayerSpawn
 		catch (Exception e)
 		{
 			Logger.LogError(e.ToString());
+
+			physics.AppearAtWorldPositionServer(SpawnPoint.GetRandomPointForLateSpawn().transform.position);
 		}
 
 		switch (spawnType)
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Core/Lifecycle/PlayerSpawn.cs
```diff
@@ -298,26 +320,40 @@ private static void ApplyNewSpawnRoleToBody( GameObject body, Occupation request
...
+		{
+			Logger.LogError(e.ToString());
+		}
 
-		//determine where to spawn them
-		physics.AppearAtWorldPositionServer(GetSpawnPointForOccupation(requestedOccupation));
+		try
+		{
+			//determine where to spawn them
+			physics.AppearAtWorldPositionServer(GetSpawnPointForOccupation(requestedOccupation));
```
MaxIsJoe, line 351: It would be better if there was also some fallback behavior so players dont get stuck half-way through

Bod9001: yeah,  What happens if that Errors though?

corp-0: if that happens, you make it delete the game

MaxIsJoe: you log the error first, then try the fallback logic.
the game shouldn't error if you're trying to spawn someone at arrivals since that's technically the default always; but if it does fail then rip i guess, means there's a bigger issue.

PR #9883, opened 2023-05-31
Fixing444

Bod9001:
### Changelog:

CL: [Fix] Fixed research collectors not being able to be built.
CL: [Fix] Fixed laser projection not bouncing off mirrors after hitting them.
CL: [Fix] Fixed pushing containers in an line being buggy.
CL: [Fix] Fixed objects staying in the middle of tiles after pulling.


corp-0:
I'm re running tests. If they fail again you might have a compiling error.

PR #9889, opened 2023-06-03
UNITY_EDITOR Debug

Bod9001:


corp-0:
Is distance a float? Perhaps your problem is you are comparing a float to zero without approximation

MaxIsJoe:
> Is distance a float? Perhaps your problem is you are comparing a float to zero without approximation

At this point I say we reject any PR that doesn't use Math.Approx because Bod is stubborn about this

PR #9892, opened 2023-06-07
[critical fix] Fixed infinite teleportation loop for portals

MaxIsJoe:
CL: [Fix] Fixed portals causing the server to crash due to players and objects teleporting back and forth infinitely

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Research/Portal.cs
+++ b/UnityProject/Assets/Scripts/Objects/Research/Portal.cs
@@ -91,7 +91,8 @@ namespace Objects.Research
 
 		public override void OnPlayerStep(PlayerScript playerScript)
 		{
-			Teleport(playerScript.gameObject);
+			if (isOnCooldown) return;
+			_ = Teleport(playerScript.gameObject);
 		}
 
 		public override bool WillAffectObject(GameObject eventData)
@@ -108,19 +109,20 @@ namespace Objects.Research
 
 		public override void OnObjectEnter(GameObject eventData)
 		{
+			if (isOnCooldown) return;
 			_ = Teleport(eventData);
 		}
 
 		private async Task Teleport(GameObject eventData)
 		{
 			if (connectedPortal == null || isOnCooldown) return;
-			if (eventData.HasComponent<PlayerScript>() == false || eventData.GetComponent<SparkEffect>() != null) return;
+			if (eventData.HasComponent<PlayerScript>() == false || eventData.HasComponent<SparkEffect>()) return;
 			connectedPortal.isOnCooldown = true;
 			isOnCooldown = true;
 			TransportUtility.TeleportToObject(eventData, connectedPortal.gameObject,
 				connectedPortal.ObjectPhysics.OfficialPosition.RandomOnOneAxis(-1, 1), true, false);
 			SparkUtil.TrySpark(gameObject, expose: false);
-			await Task.Delay(750);
+			await Task.Delay(850);
 			isOnCooldown = false;
 			connectedPortal.isOnCooldown = false;
 		}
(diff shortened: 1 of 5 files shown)


corp-0:
Something that could be annoying with this is if I am standing in the portal while on cooldown, I have to step down and in again in order to teleport after the cooldown is done. I think the way we solved this in the past (and I don't know why it isn't in effect anymore) was teleporting people one tile away from the teleporter

FuntimeDudeski:
yeah, currently pads can be set to send to the 4 directions adjacent to the receiving pad as well as 'on top', which is the default but appears hardcoded to send things 1 tile above instead (this applies to both 'passive detect' and qpads that have to be clicked). it's a bit aesthetically annoying that you can't teleport directly onto a destination pad.

rather than polling every few frames to see if something is on a passive detect qpad, perhaps the passive detect setting should make the check only 1) when a mob or object moves onto the teleporter (treat walking onto and reaching the center of the qpad's tile the same as having clicked it with passive detect off) or 2) at any time when it goes off cooldown and an object/mob is on it

i'm a bit interested in this matter cause valley has a rather large number of quantum pads.

MaxIsJoe:
> quantum pads

this is the third time someone talks about or mentions them without even looking at what issue is this related to.
It's not quantum pads, this is portals for things like the hand tele.

UnityProject/Assets/Scripts/Objects/Research/Portal.cs
```diff
@@ -102,19 +109,26 @@ public override bool WillAffectObject(GameObject eventData)
...
-		private void Teleport(GameObject eventData)
+		private async Task Teleport(GameObject eventData)
 		{
-			if(connectedPortal == null) return;
-
-			SparkUtil.TrySpark(gameObject, expose: false);
-
-			TransportUtility.TeleportToObject(eventData, connectedPortal.gameObject,
-				connectedPortal.ObjectPhysics.OfficialPosition, true, false);
+			await Task.Delay(200);
```
Bod9001: didn't actually fix the issue reeee

MaxIsJoe: actually it did, I fixed the infinte loop that crashes the server; never promised to fix another unrelated issue.

Bod9001: I will fix  it then

UnityProject/Assets/Scripts/Objects/Research/Portal.cs
```diff
@@ -102,19 +109,26 @@ public override bool WillAffectObject(GameObject eventData)
 
 		public override void OnObjectEnter(GameObject eventData)
 		{
-			Teleport(eventData);
+			if (isOnCooldown) return;
+			_ = Teleport(eventData);
 		}
 
-		private void Teleport(GameObject eventData)
+		private async Task Teleport(GameObject eventData)
```
Bod9001, line 116: also I think puts it on a different thread so no accessing unity functions + put a try captain for logging errors because they won't Logged to the consul

FuntimeDudeski:
mb, i always forget to call quantum pads quantum pads vs. teleporters and didn't know what they're named in the code/whether they shared any of the same code with other teleportation functions.

it sounds like the same or comparable issue with infinite loops between two portals and two passive detect qpads.

FuntimeDudeski:
iow, my comment/question was potentially pertinent to teleportation of mobs/objects in general and future improvements of this function regardless of which feature might be calling it.

MaxIsJoe:
I'll take a look at your ideas, you might be right about a lot of these, and we can definitely improve a lot of these or create a few functions that make this not only safer/developer friendly but also user friendly.

PR #9894, opened 2023-06-09
Determinastic randomness

### LINKED ISSUE #9803: [BUG] Maintrooms chances are way too high.

### Game Version

Current

### I was

Client

### OS

Windows 8, 10, 11

### What happened?

I got sent to Maintrooms twice in thirty minutes on the same round.

Should be a 0.1% or so.

1. Literally just play on squarestation

### Additional context

_No response_

MaxIsJoe:
Read [Random13.md](https://github.com/unitystation/unitystation/compare/develop...MaxIsJoe:unitystationrandom?expand=1#diff-74d00d5554a0de5f19320dde4f370cd6b4063866c76943edb37533f9d8724a27) and [Random13.cs](https://github.com/unitystation/unitystation/compare/develop...MaxIsJoe:unitystationrandom?expand=1#diff-148d2d2f09bab0a0dc1b494ca56f15b272725bbecf69e6d6ddc8a1e0fe656b49) for more info

fixes #9803

CL: [Improvement] Updated some random behaviors to a more deterministic model. 
CL: [Improvement] Reduced maintroom chances.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Random13.cs
+++ b/UnityProject/Assets/Scripts/Core/Random13.cs
@@ -16,7 +16,7 @@ namespace Core
 		public static Random13 Instance;
 
 		//https://github.com/id-Software/DOOM/blob/77735c3ff0772609e9c8d29e3ce2ab42ff54d20b/linuxdoom-1.10/m_random.c#LL31C1-L31C1
-		private readonly int[] randomTable = new[]
+		private readonly byte[] randomTable = new byte[]
 		{
 			0, 8, 109, 220, 222, 241, 149, 107, 75, 248, 254, 140, 16, 66,
 			74, 21, 211, 47, 80, 242, 154, 27, 205, 128, 161, 89, 77, 36,
--- a/UnityProject/Assets/Scripts/Core/Random13.cs
+++ b/UnityProject/Assets/Scripts/Core/Random13.cs
@@ -47,7 +47,7 @@ namespace Core
 		{
 			CurrentRandomTableIndex++;
 			if (CurrentRandomTableIndex >= randomTable.Length - 1) CurrentRandomTableIndex = 0;
-			OnRandomTableIndexChanged?.Invoke(CurrentRandomTableIndex);
+			OnRandomTableIndexChanged?.Invoke(randomTable[CurrentRandomTableIndex]);
 		}
 
 		/// <summary>
--- a/UnityProject/Assets/Scripts/Core/Random13.cs
+++ b/UnityProject/Assets/Scripts/Core/Random13.cs
@@ -58,11 +58,44 @@ namespace Core
 		{
 			if (Instance == null)
 			{
-				throw new NotImplementedException("Random13 was never created and stored anywhere.");
+				throw new Exception("Random13 was never created and stored anywhere.");
 			}
 			Instance.ProgressIndex();
-			if (Instance.RandomNumber % 2 == 0) return true;
-			return false;
+			return Instance.RandomNumber % 2 == 0;
+		}
+
+		/// <summary>
+		/// Consistent random numbers for things like damage.
+		/// </summary>
+		public static float RandomNumberWave(float baseNumber)
+		{
+			if (Instance == null)
+			{
+				throw new Exception("Random13 was never created and stored anywhere.");
+			}
+			Instance.ProgressIndex();
+			//Random number between 1 and 8 then multiplied by base damage.
+			return (Instance.RandomNumber % 8 + 1) * baseNumber;
+		}
+
+		/// <summary>
+		/// Random deterministic chance. Try your luck!
+		/// </summary>
+		/// <param name="chance">Chance between 0 and 1 (0.5 is 50%, etc.)</param>
+		public static bool ProbChance(double chance)
+		{
+			if (Instance == null)
+			{
+				throw new Exception("Random13 was never created and stored anywhere.");
+			}
+
+			if (chance is > 1 or < 0)
+			{
+				throw new ArgumentException("Chance must be between 0 and 1 as a double!!");
+			}
+			Instance.ProgressIndex();
+			var random = new Random(Instance.RandomNumber); //uses the table as a seed
+			return random.NextDouble() < chance;
 		}
 
 		/// <summary>
--- a/UnityProject/Assets/Scripts/Core/Random13.cs
+++ b/UnityProject/Assets/Scripts/Core/Random13.cs
@@ -1,0 +1,78 @@
+﻿using System;
+
+namespace Core
+{
+	/// <summary>
+	/// An extremely fast, efficient and deterministic class for randomness.
+	/// </summary>
+	public class Random13
+	{
+		public Random13()
+		{
+			Instance = this;
+		}
+
+		public static Random13 Instance;
+
+		//https://github.com/id-Software/DOOM/blob/77735c3ff0772609e9c8d29e3ce2ab42ff54d20b/linuxdoom-1.10/m_random.c#LL31C1-L31C1
+		private readonly int[] randomTable = new[]
+		{
+			0, 8, 109, 220, 222, 241, 149, 107, 75, 248, 254, 140, 16, 66,
+			74, 21, 211, 47, 80, 242, 154, 27, 205, 128, 161, 89, 77, 36,
+			95, 110, 85, 48, 212, 140, 211, 249, 22, 79, 200, 50, 28, 188,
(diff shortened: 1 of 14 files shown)


UnityProject/Assets/Scripts/Core/Random13.cs
```diff
@@ -0,0 +1,80 @@
...
+
+		/// <summary>
+		/// Quick way to randomly decide if something is true or false. Perfect for deterministic behavior as it relies on the
+		/// random table and its current used index.
+		/// </summary>
+		public static bool Prob()
+		{
+			if (Instance == null)
+			{
+				throw new NotImplementedException("Random13 was never created and stored anywhere.");
```
corp-0: Not the right exception for this case. I would have to look what other exceptions are there, but the not implemented exception is for when you haven't done the body of a method you intend to implement from an interface/abstract

UnityProject/Assets/Scripts/Objects/Singularity.cs
```diff
@@ -427,7 +428,7 @@ private void DestroyObjects(IEnumerable<Vector3Int> coords)
 						if (health.RegisterPlayer.PlayerScript.Mind?.occupation == OccupationList.Instance.Get(JobType.CLOWN))
 						{
 							health.OnGib();
-							ChangePoints(DMMath.Prob(50) ? -1000 : 1000);
+							ChangePoints(Random13.ProbFromTime(GameManager.Instance.RoundTime) ? -1000 : 1000);
```
corp-0, line 431: I don't understand when to use the Prob and ProbFromTime.

In the original code this was 50% chance. What's the new chance of occurrence?

MaxIsJoe: https://www.youtube.com/watch?v=pq3x1Jy8pYM&t
I think the first few minutes of this video can shed some light into why ProbFromTime is the better for these kinds of things.

Prob has 255 predictable states, ProbFromTime has an infinite amount of states that is always changing all the time.
255 always changing states is pretty hard to brute force in a game like Unitystation where there are a lot of changing elements that cause the random table to constantly move all the time, but it's still possible to predict a desired result if someone finds the current index from a pattern that the server is on via observation.
ProbFromTime is much harder to predict and use to your advantage while you're playing, but is still very accurate and predictable for when you want to replicate events by providing the server's DateTime that it only has access to. So it makes it the perfect fit for when you want to make sure that players don't influence things like scores or events.

UnityProject/Assets/Scripts/Core/Random13.cs
```diff
@@ -0,0 +1,80 @@
...
+	{
+		public Random13()
+		{
+			Instance = this;
+		}
+
+		public static Random13 Instance;
+
+		//https://github.com/id-Software/DOOM/blob/77735c3ff0772609e9c8d29e3ce2ab42ff54d20b/linuxdoom-1.10/m_random.c#LL31C1-L31C1
+		private readonly byte[] randomTable = new byte[]
```
Bod9001, line 19: Isn't there seed options for this stuff?

MaxIsJoe: nope

MaxIsJoe: You don't need a seed to change the results of table that will never change

UnityProject/Assets/Scripts/Objects/Singularity.cs
corp-0: so how do you make something happen 5% of the time?

UnityProject/Assets/Scripts/Core/Random13.cs
MaxIsJoe: changed it

UnityProject/Assets/Scripts/Objects/Singularity.cs
MaxIsJoe: Use `ProbChance()` 
I just commited it

UnityProject/Assets/Scripts/Core/Random13.cs
```diff
@@ -58,11 +58,44 @@ public static bool Prob()
...
-			return false;
+			return Instance.RandomNumber % 2 == 0;
+		}
+
+		/// <summary>
+		/// Consistent random numbers for things like damage.
+		/// </summary>
+		public static float RandomNumberWave(float baseNumber)
+		{
+			if (Instance == null)
```
corp-0, line 72: since you are doing thischeck and raise the exception in all usages, it would be better to move it to the property declaration

MaxIsJoe: wdym?

corp-0:
```cs
public static Random13 Instance => instance ?? throw new Exception("you didn't blah blah");

private static Random13 instance;
```

And you would assign the private field in the constructor. That would make it so any time someone tries to access the public property, it will throw if it is null and you don't need to check it in every method.

MaxIsJoe: ah

corp-0: I messed up the C# syntax, long time without writing any dotnet. But you get the gist of what I tried to do.

UnityProject/Assets/Scripts/Core/Random13.cs
```diff
@@ -0,0 +1,113 @@
...
+		/// Extremely fast and much more predictable.
+		/// </summary>
+		/// <param name="time">Usually DateTime.Now but a specific time can be given to replay expected events.</param>
+		/// <returns>50/50 chance for true or false based on current time given.</returns>
+		public static bool ProbFromTime(DateTime time)
+		{
+			if (time.Ticks % 2 == 0) return true;
+			return false;
+		}
+	}
```
Alainx277, line 112: Why reinvent the wheel with random numbers? There's a perfectly fine stdlib class.
```suggestion
	/// <summary>
	/// Deterministic class for randomness.
	/// Best reserved for server logic, while client logic can use DMMath.
	/// </summary>
	public class Random13 : Random
	{
		// Use a specific seed for determinism (could also be unique per round)
		public Random13() : base(13)
		{
			Instance = this;
		}

		public static Random13 Instance;

		/// <summary>
		/// Randomly decide if something is true or false.
		/// </summary>
		public static bool Prob()
		{
			if (Instance == null)
			{
				throw new Exception("Random13 was never created and stored anywhere.");
			}

			return Instance.Next(2) == 0;
		}

		/// <summary>
		/// Consistent random numbers for things like damage.
		/// </summary>
		public static float RandomNumberWave(float baseNumber)
		{
			if (Instance == null)
			{
				throw new Exception("Random13 was never created and stored anywhere.");
			}

			//Random number between 1 and 8 then multiplied by base damage.
			return (Instance.Next(8) + 1) * baseNumber;
		}

		/// <summary>
		/// Random deterministic chance. Try your luck!
		/// </summary>
		/// <param name="chance">Chance between 0 and 1 (0.5 is 50%, etc.)</param>
		public static bool ProbChance(double chance)
		{
			if (Instance == null)
			{
				throw new Exception("Random13 was never created and stored anywhere.");
			}

			if (chance is > 1 or < 0)
			{
				throw new ArgumentException("Chance must be between 0 and 1 as a double!!");
			}
			return Instance.NextDouble() < chance;
		}

		/// <summary>
		/// Uses a given time to determine the probability of having a 50/50 chance for something to happen.
		/// Extremely fast and much more predictable.
		/// </summary>
		/// <param name="time">Usually DateTime.Now but a specific time can be given to replay expected events.</param>
		/// <returns>50/50 chance for true or false based on current time given.</returns>
		public static bool ProbFromTime(DateTime time)
		{
			if (time.Ticks % 2 == 0) return true;
			return false;
		}
	}
```

UnityProject/Assets/Scripts/Core/Random13.cs
PJB3005: That's not how PRNG seeding works. You're still "seeding" the PRNG. It just turns out the seed you picked is a `0`. The exact same can be done with other PRNGs.

UnityProject/Assets/Scripts/Core/Random13.cs
PJB3005: `System.Random` doesn't guarantee stability across runtimes/versions. I'd recommend you google an algorithm online (they're really simple) instead, so you know it's consistent.

PR #9895, opened 2023-06-09
Fixes blob not spawning #9884

### LINKED ISSUE #9884: [BUG] Blob doesn't spawn correctly

### Game Version

23053022

### I was

Client

### OS

Windows 8, 10, 11

### What happened?

I spawned as CE and rolled a blob. When I was supposed to die and become a blob I died and become a ghost. My body was still alive but I was a ghost and there was no blob core. I tried to re-enter my body and suddenly I actually became a blob. There was still no core but I could move around the station (see screenshot). Jumping to core didn't do anything.
image

I should've died and become a blob with a core.

1. Become blob antagonist 
2. Wait for you to die

### Additional context

_No response_

MaxIsJoe:
fixes #9884

CL: [Fix] Fixed an issue with blobs not spawning due to outdated information when moving the player's mind to the blob.

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Antagonists/Antags/Blob/BlobStarter.cs
+++ b/UnityProject/Assets/Scripts/Systems/Antagonists/Antags/Blob/BlobStarter.cs
@@ -83,12 +83,7 @@ namespace Blob
 
 		private void OnEnable()
 		{
-#if UNITY_EDITOR
-			UpdateManager.Add(PeriodicUpdate, 0.15f);
-			Logger.Log("Periodic Update adjusted for blob in editor for faster testing.");
-#else
 			UpdateManager.Add(PeriodicUpdate, updateTime);
-#endif
 		}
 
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Systems/Antagonists/Antags/Blob/BlobStarter.cs
```diff
@@ -83,7 +83,12 @@ private void Start()
 
 		private void OnEnable()
 		{
+#if UNITY_EDITOR
+			UpdateManager.Add(PeriodicUpdate, 0.15f);
```
Bod9001: might be better just to have a debug bool on it, that sskips the updateTime stuff?

MaxIsJoe: no need

Bod9001 (approved):

PR #9900, opened 2023-06-11
Continues improvements for laying down behavior

### LINKED ISSUE #9753: [BUG] People randomly being sideways.

### Game Version

23032711

### I was

Client

### OS

Windows 8, 10, 11

### What happened?

Sometimes people will randomly appear sideways but otherwise walk, talk, interact, etc. the same. Throwing them typically fixes the issue. It's just a very annoying visual bug.

People should only be sideways if they are lying down.

1. Don't know it seems to be random when playing.

### Additional context

_No response_

MaxIsJoe:
cleaned up some of the movement code while I'm at it
fixes #9753

CL: [Improvement] Flying and sliding animations now target the sprites instead of the Mob's root gameobject. Making the behavior much more consistent. 
CL: [Improvement] Burning animations and body overlays will now correctly follow the body sprites's rotation.
CL: [Fix] Whenever the server asks to ensure mob's laying down state, it will notify all clients respectively as well.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
+++ b/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
@@ -1385,10 +1385,7 @@ public class UniversalObjectPhysics : NetworkBehaviour, IRightClickable, IRegist
 			return;
 		}
 
-		if (IsMoving)
-		{
-			return;
-		}
+		if (IsMoving) return;
 
 		isFlyingSliding = true;
 		MoveIsWalking = false;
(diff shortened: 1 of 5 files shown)


UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
```diff
@@ -310,6 +274,24 @@ public virtual void Awake()
 		registerTile = GetComponent<RegisterTile>();
 		rotatable = GetComponent<Rotatable>();
 		pickupable.DirectSetComponent(GetComponent<Pickupable>());
+		SetRotationTarget();
+	}
+
+	private void SetRotationTarget()
+	{
+		if (rotationTarget != null) return;
+		if (this is MovementSynchronisation c && c.playerScript.RegisterPlayer.LayDownBehavior != null)
```
corp-0, line 283: base class shouldn't handle cases for a child class. Can you make this method protected virtual and do the second part of the if on the subclass?

corp-0: nevermind, I just took a look at this class and the subclass. I wouldn't want to touch any of that either.

corp-0 (approved):

PR #9902, opened 2023-06-13
feat: add initial name and job title to ID cards for story telling

corp-0:
### Purpose
Mappers are now able to add an initial name and job title to ID cards so they can do some story telling with their level design.

### Changelog:
__NOCL__


MaxIsJoe:
hell yea

Bod9001 (approved):

PR #9911, opened 2023-06-17
Fixes #4746.

SerheyKr:
### Purpose
PR fixes bug #4746. That fake walls very noticeable becase they don`t connects with other walls.

### Changelog
CL: [Fix] Fixed fake walls not blending properly with neighbour walls.

PR #9918, opened 2023-06-20
Fixes #9911

SerheyKr:
###Purpose
PR fixes bug in #9911. That fake walls has still empty wall tile when they removed.


corp-0:
I removed the changelog from the PR description as the last one already mentions a fix. Could you test this locally? Are you sure this one works?

SerheyKr:
Tested localy, found one bug.

Bod9001 (approved):

PR #9921, opened 2023-06-21
Ministation improvements + Filth Gen reduction

MaxIsJoe:
image

image

image

image

image

image

### Changelog:

CL: [New] Added sustenance vendor to arrivals on Ministation.
CL: [New] Added a security checkpoint to arrivals on Ministation.
CL: [Improvement] Spiced up lighting in multiple areas around Ministation to give a moody vibe.
CL: [Improvement] The science department now has access to the lavaland teleporter.
CL: [Improvement] Reduced the chance for reagent filth to spawn from 50% to 35%.
CL: [Improvement] Subsystems will now report how many milliseconds they've taken to finish loading instead of seconds.
CL: [Fix] Miners will no longer get stuck in a wall when getting teleported back from lavaland to the station.
CL: [Fix] Fixed the filth generator subsystem running twice which resulted in the station being more filthy than it should.


CorruptComputer:
It might be worth adding another air vent and scrubber to the area below the security checkpoint here:
image

It doesn't look like there is any way to get air down there when the doors are closed

MaxIsJoe:
Good point

CorruptComputer:
Does this room have them? I can't tell from the screenshot if they are below the tables.
image

MaxIsJoe:
Nope, Adding them to it as well

PerfectTangent:
the security checkpoint in evac is very _not secure._ It needs glass windows and locked sliding doors.

MaxIsJoe:
image

added to arrivals and lavaland teleporter

MaxIsJoe:
> the security checkpoint in evac is very _not secure._ It needs glass windows and locked sliding doors.

image

Added windows and setup clearance to require security or higher to open them

UnityProject/Assets/Scripts/Systems/FilthGenerator/FilthGenerator.cs
```diff
@@ -13,33 +14,38 @@ namespace Systems.FilthGenerator
...
-		private float FilthDensityPercentage;
+		private float filthReagentChance = 35f;
 
 		[SerializeField] private List<GameObject> filthDecalsAndObjects = new List<GameObject>();
-		[SerializeField] private bool generateFilthReagent = true;
-		
+
 		public override void Initialize()
 		{
+			if (Initialized) return;
```
Bod9001: I feel like it would be better just to fix the issue in the first place?

MaxIsJoe: Doesn't matter if we fix the issue or not, this kind of logic is intended still.

Bod9001: how so? Seems like a better fix would just be it not happening in the first place

MaxIsJoe: Lets say that we add a function to allow for mappers to test subsystems during runtime and some subsystems should not be run twice 
or a debug command for admins to re-run some subsystems incase something goes wrong

You'll need to remind the game that some subsystems should not be run twice when initialized

Bod9001: then that should be baked inside the SubsystemBehaviour , the issue is not fixing the original problem that it runs twice

MaxIsJoe: I've solved the issue for filth gen. Have fun looking at the comments because I'm not touching that with a ten-foot pole again

UnityProject/Assets/Scripts/Managers/SubsystemBehaviourQueueInit.cs
```diff
@@ -25,6 +25,11 @@ private static void ClearSubsystems()
 
 		public static void Queue(SubsystemBehaviour behaviour)
 		{
+			if (Instance.behaviours.Contains(behaviour))
```
corp-0: anticipating to what Bod might say, do we know now why it can be added twice?

MaxIsJoe: It's not getting added twice to this system, but I just added it there just incase.
What is actually happening is that its adding it to two systems, the new one and the old one; the old one cannot be removed for the time being because the entire game implodes if you try using the new system.. even if the new system does the same exact thing.

Bod9001:
Clients get stuck after round restart

Bod9001:
testing and building now

PR #9927, opened 2023-06-27
Fixes light not getting updated when dropped

MaxIsJoe:
CL: [Fix] Fixed an issue where lights would not get removed off of players when dropped/thrown.
CL: [Improvement] Light rotations are now animated on the player.


Bod9001 (dismissed):
has a similar issue with the snap back to container, might be worth adding that as well

MaxIsJoe:
> has a similar issue with the snap back to container, might be worth adding that as well

Can't replicate issue for paddle containers. Works fine.

MaxIsJoe:
@Bod9001

corp-0 (approved):

PR #9930, opened 2023-06-28
Staging thee fixes

Bod9001:
### Changelog:
CL: [New] Cyborgs can now "wear" hats
CL: [Fix] You can no longer drag powder
CL: [Balance] it is not longer possible to sell brains in cargo shuttle
CL: [Fix] Fixed objects being registered to the wrong register tile
CL: [Fix] Fixes cyborgs taking damage when walking on glass shardds
CL: [Fix] Fixes lying down and  standing bypassing critical movement restriction
CL: [Fix] fixes a server crash


corp-0:
what did the last thing in the changelog cause to the players?

Bod9001:
"what did the last thing in the changelog cause to the players?" 
oh? it might affect Players but it was more noticeable with spawning in glass shards on light fixtures breaking

PR #9936, opened 2023-07-04
Slime improvements

Bod9001:
I think it might be ready to map in a different PR

### Changelog:

CL: [Improvement] Syringes now can change the mode from injecting to pulling by activating it in hand
CL: [Fix] fixed the Slime processor not working

### DIFF
--- a/UnityProject/Assets/Scripts/ChemistryComponents/ReagentContainer.TransferInteraction.cs
+++ b/UnityProject/Assets/Scripts/ChemistryComponents/ReagentContainer.TransferInteraction.cs
@@ -40,6 +40,8 @@ namespace Chemistry.Components
 		[FormerlySerializedAs("TransferMode")]
 		[SerializeField] private TransferMode transferMode = TransferMode.Normal;
 
+		[SerializeField] public bool SyringePulling;
+
 		public TransferMode TransferMode => transferMode;
 
 		[FormerlySerializedAs("PossibleTransferAmounts")]
@@ -190,23 +192,49 @@ namespace Chemistry.Components
 		{
 			if (DefaultWillInteract.Default(interaction, side) == false) return false;
 
-			return possibleTransferAmounts.Count != 0;
+			return possibleTransferAmounts.Count != 0 || transferMode == TransferMode.Syringe;
 		}
 
 		public void ServerPerformInteraction(HandActivate interaction)
 		{
-			var currentIndex = possibleTransferAmounts.IndexOf(TransferAmount);
-			if (currentIndex != -1)
+			if (transferMode == TransferMode.Syringe)
 			{
-				TransferAmount = possibleTransferAmounts.Wrap(currentIndex + 1);
+				SyringePulling = !SyringePulling;
+				if (IsFull && SyringePulling)
+				{
+					SyringePulling = false;
+					Chat.AddExamineMsg(interaction.Performer,
+						$"The {gameObject.ExpensiveName()} Is full you can't pull any more.");
+					return;
+				}
+
+				if (IsEmpty && SyringePulling == false)
+				{
+					SyringePulling = true;
+					Chat.AddExamineMsg(interaction.Performer,
+						$"The {gameObject.ExpensiveName()} Is empty you can't inject any more.");
+					return;
+				}
+
+				var pullstreing = SyringePulling ? "Pulling mode" : "Injecting mode";
+				Chat.AddExamineMsg(interaction.Performer,
+					$"you change {gameObject.ExpensiveName()} to {pullstreing}.");
 			}
 			else
 			{
-				TransferAmount = possibleTransferAmounts[0];
-			}
+				var currentIndex = possibleTransferAmounts.IndexOf(TransferAmount);
+				if (currentIndex != -1)
+				{
+					TransferAmount = possibleTransferAmounts.Wrap(currentIndex + 1);
+				}
+				else
+				{
+					TransferAmount = possibleTransferAmounts[0];
+				}
 
-			Chat.AddExamineMsg(interaction.Performer,
-				$"The {gameObject.ExpensiveName()}'s transfer amount is now {TransferAmount} units.");
+				Chat.AddExamineMsg(interaction.Performer,
+					$"The {gameObject.ExpensiveName()}'s transfer amount is now {TransferAmount} units.");
+			}
 		}
 
 		/// <summary>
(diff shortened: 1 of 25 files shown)


MaxIsJoe (dismissed):

UnityProject/Assets/Scripts/ChemistryComponents/ReagentContainer.cs
```diff
@@ -134,6 +135,16 @@ public ReagentMix CurrentReagentMix
...
 		}
 
+		//[RightClickMethod]
+		public void PrintContents()
+		{
+			foreach (var keyd in currentReagentMix.reagents.m_dict)
+			{
+				Logger.LogError(keyd.Key.Name + " > " + keyd.Value);
+			}
+			Logger.LogError("=====================================");
```
AtnerNT: Why an error log for a print function? I am 75.6% sure we have something like just a normal Logger.Log() that's not as an error.

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
```diff
@@ -346,16 +346,20 @@ public void RemoveInventoryAndBody(Vector3 AppearAtWorld)
 		/// </summary>
 		public void TryRemoveFromBody(bool beingGibbed = false, bool CausesBleed = true, bool Destroy = false, bool PreventGibb_Death = false) //TODO It should do the stuff automatically when removed from inventory
 		{
+			if ( HealthMaster == null ) return;
 			bool alreadyBleeding = false;
 			if (CausesBleed)
 			{
-				foreach (var bodyPart in HealthMaster.BodyPartList)
+				if (HealthMaster != null)
```
AtnerNT: if(CausesBleed == true && HealthMaster != null)

Can just be combined.

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
```diff
@@ -346,16 +346,20 @@ public void RemoveInventoryAndBody(Vector3 AppearAtWorld)
...
-					if (bodyPart.BodyPartType == BodyPartType.Chest && alreadyBleeding == false)
+					foreach (var bodyPart in HealthMaster.BodyPartList)
 					{
-						bodyPart.IsBleeding = true;
-						alreadyBleeding = true;
-						HealthMaster.ChangeBleedStacks(limbLossBleedingValue);
+						if (bodyPart.BodyPartType == BodyPartType.Chest && alreadyBleeding == false)
+						{
+							bodyPart.IsBleeding = true;
+							alreadyBleeding = true;
```
AtnerNT: As far as I can see, alreadyBleeding is only used to prevent this block from triggering for multiple body parts - in that case this variable can be removed as just be replaced with a break after HealthMaster.ChangeBleedStacks().

UnityProject/Assets/Scripts/ChemistryComponents/ReagentContainer.TransferInteraction.cs
```diff
@@ -40,6 +40,8 @@ public partial class ReagentContainer : ICheckedInteractable<HandApply>, //Trans
 		[FormerlySerializedAs("TransferMode")]
 		[SerializeField] private TransferMode transferMode = TransferMode.Normal;
 
+		[SerializeField] public bool SyringePulling;
```
AtnerNT, line 43: Would be better to write this as:
[field: SerializeField] public bool SyringePulling { get; private set; }
as I don't think is intended for this to be able to be edited elsewhere.

MaxIsJoe: Have fun trying to convince Bod to do this

Bod9001: some of you might say is it might automatically change when you fill up the syringe in the future

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
Bod9001: idk, I just added some stuff so it won't error   if health Master is null

UnityProject/Assets/Scripts/ChemistryComponents/ReagentContainer.cs
Bod9001: is a debug, also easy to see in the Consul

corp-0: It isn't an error. Don't log non errors as errors. You can do it while you debug locally, but don't push it to prod

UnityProject/Assets/Scripts/ChemistryComponents/ReagentContainer.TransferInteraction.cs
Bod9001: *something I will say it might be changed in the future when you fill up your syringe it swaps to inject mode

AtnerNT: > *something I will say it might be changed in the future when you fill up your syringe it swaps to inject mode

If that's the case it would be better to either
- move the syringe pulling bool to the syringe script if there is one
- have it call a function on reagentcontainer as opposed to directly changing the bool

corp-0: I have to agree with Atner here. ReagentContainer shouldn't need to know anything about syringes. The syringe itself is the one with changing modes and can interact with ReagentContainer API to manage it.

Bod9001: problem is reagent container needs to know if it's a syringe

Bod9001: because one of the interactions checks if it's going to be adding or removing From it

corp-0: move that interaction to the syringe compoenent(?). The syringe should be the one that handles that and makes use of the ReagentContainer component to transfer in or out depending on the state of the syringe.

corp-0: this is like explosion code having bools to know if the grenade is a frag, EMP or bang and doing all the interaction in the explosion code.

Bod9001: it's literally just a small fix to make it so you can inject slime cores easier

corp-0: it's literally tech debt. But I don't know why am I even trying, you are going to merge this anyways

Issue #6156, opened 2021-03-01
Pyrotechnic Chemistry Bounty


PerfectTangent:
## Description

This bounty calls for the implementation of the explosive and Pyrotechnic side of chemstry, including grenade crafting, chemical foaming, chemical smoking, and some reagents from the misc chems category. 

Chemicals should be able to be released as foam either when the react or because foaming agent was included in the reaction. Foam spreads the chemicals out and coats items within the area of effect, causing topical application to any floors, objects, or mobs in the area of the foam. The higher proportion of foaming agent used in a reaction, the greater the area of effect, but the more diluted the effects or on affected targets.

Chems should also be able to be aerosolized into chemical smoke either as part of the reaction or because smoking agent was included in the reaction. When a smoke reaction occurs, the smoke will consume any other reagent in its original containers. Depending on the reagents included in the reaction, smoke topically applies to any floors, objects, and mobs in the smoke cloud. In addition, mobs that do not have gas masks or internals equipped will breathe in the smoke, causing them to ingest it; a good way to apply toxins. if it is thick enough, smoke also blocks line of sight.

In addition this should come with a tool tool that lets the chemist do such a thing; the smoke machine, a portable device that lets off flavoured chemical smoke. with the merging of #5882, the smoke machine should be able to be upgraded with improved stock parts in its first implmentation. Also, foam spreads slower than smoke and usually makes the floor slippery as it is reacting.

Chemists can also manufacture customized grenades which when triggered, mix the chemical reagents stored inside, causing an immediate reaction. A grenade that contains smoke or foam, or chemicals that produce smoke or foam when they react will produce such effects when used. A grenade that contains explosives with explode. A grenade that doesn't include any will instead splash the contents in all directions for a short distance. grenades also come in a variety of sizes with the ability to store variable amounts of chems.

Reagents should have a touch component that makes them apply their effects when they splash, foam, or engulf a target in smoke. There should also be a inhale component for chems released as smoke.

Lastly, this bounty needs a variety of reagents to use in the new mechanics. All chemical reactions and reagents classed as pyrotechnics should be included except EMP, Teslium, Tesla shock, Bee explosion, and holy explosion. the list can be seen here: https://tgstation13.org/wiki/Guide_to_chemistry#Pyrotechnics

In addition, a few non-pyrotechnic chems should be included so this isn't just an exclusively destructive update. Glitter, metal foam, and acid should be added with this bounty.

## What is needed to fulfill the bounty:

[...]

PR #9938, opened 2023-07-05
Adds in Thermite chemical reaction

MaxIsJoe:
part of #6156 
image

### Changelog:

CL: [New] Adds in a chemical reaction for Thermite. It will react strongly to fire/heat and will burn through floors and walls.

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Chemistry/Effect.cs
+++ b/UnityProject/Assets/Scripts/Systems/Chemistry/Effect.cs
@@ -7,5 +7,6 @@ namespace Chemistry
 	public abstract class Effect : ScriptableObject
 	{
 		public abstract void Apply(GameObject sender, float amount);
+		public abstract void HeatExposure(GameObject sender, float heat, ReagentMix inMix);
 	}
 }
--- a/UnityProject/Assets/Scripts/Systems/Chemistry/Effect.cs
+++ b/UnityProject/Assets/Scripts/Systems/Chemistry/Effect.cs
@@ -6,6 +6,6 @@ namespace Chemistry
 {
 	public abstract class Effect : ScriptableObject
 	{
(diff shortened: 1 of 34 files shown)


UnityProject/Assets/Scripts/Systems/Chemistry/Reaction.cs
```diff
@@ -34,11 +34,20 @@ public virtual bool Apply(object sender, ReagentMix reagentMix)
 		{
 			if (IsReactionValid(reagentMix) == false) return false;
 
-			ApplyReaction(sender as MonoBehaviour, reagentMix);
+			ApplyReaction(sender as GameObject, reagentMix);
 
 			return true;
 		}
 
+		public virtual void OnExposedToHotSpot(GameObject sender, float heat, ReagentMix inMix)
```
Bod9001, line 42: surely be better to raise the temperature, in the reagent mix and then check if there's any valid reaction?

UnityProject/Assets/Scripts/Systems/Chemistry/Effect.cs
```diff
@@ -6,6 +6,7 @@ namespace Chemistry
 {
 	public abstract class Effect : ScriptableObject
 	{
-		public abstract void Apply(MonoBehaviour sender, float amount);
+		public abstract void Apply(GameObject sender, float amount);
+		public abstract void HeatExposure(GameObject sender, float heat, ReagentMix inMix);
```
corp-0, line 10: I see why Alain suggested to make it virtual on discord. Since not every ReactionEffect uses the heat exposure, it is odd having them obliged to implement this. Making it a virtual with an empty body (or perhaps leaving a message stating that the reaction was exposed to heat, but the implementation wasn't overriden) is fair in my opinion.

corp-0:
You have two review to address here, just in case you didn't notice

PR #9939, opened 2023-07-05
Unity 2023 Upgrade

MaxIsJoe:
This is somehow more stable and faster than 2022 LTS
but still requires more testing.


MaxIsJoe:
I'll put this on hold right now because making builds is broken right now, understandably because this is a tech stream. But once we get a stable-ish build, we should make the move asap due to how good the improvements are in this build.

PR #9944, opened 2023-07-07
Sound Expanion P1 - The PR where space becomes less quite

MaxIsJoe:
I know, I know. I am about 7+ months late since I promised this PR; but this PR had to be split and redone twice due to how uncooperative and buggy unity is with sounds, even in this smaller chunk; there are still issues that I couldn't escape like sounds not playing at all despite them being carbon copies of other sounds that work.

This PR is a first parter of a project I've been slowly working on, and this part focuses first on adding more noise to players.
That being said, there has been sneak peaks to other things you can expect, like pickup/inventory sounds and object ambience.

### Changelog:

CL: [New] Vending machines now have a humming noise while turned on.
CL: [New] Light Fixtures emit sounds now.
CL: [New] Cyborgs will emit sounds when dying or being revived.
CL: [New] Monkeys now have their own scream noise.
CL: [New] Several emotes (such as gag, clap, giggle, cough, etc) have received sounds.
CL: [New] Toolbelts now have an inventory move sound.


corp-0:
I'm merging this just so we have the monkey screeches during the playtest

corp-0 (approved):

PR #9949, opened 2023-07-08
quick staging fixes

MaxIsJoe:
CL: [Fix] Fixed some sounds not playing on clients
CL: [Fix] Fixed a few instances of the laydown behavior attempting to call server code from the client.
CL: [Fix] Reduced the size of text in chat templates. People will no longer yell now.

### DIFF
--- a/UnityProject/Assets/Scripts/Strings/ChatTemplates.cs
+++ b/UnityProject/Assets/Scripts/Strings/ChatTemplates.cs
@@ -5,11 +5,11 @@ namespace Strings
 {
 	public static class ChatTemplates
 	{
-		public const int SmallText = 4;
-		public const int NormalText = 8;
-		public const int LargeText = 14;
-		public const int VeryLargeText = 22;
-		public const int ExtremelyLargeText = 32;
+		public const int SmallText = -2;
+		public const int NormalText = 1;
+		public const int LargeText = 8;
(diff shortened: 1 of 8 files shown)


Bod9001 (dismissed):

MaxIsJoe:
wait for a second, still trying to fix chat sizes

UnityProject/Assets/Scripts/Strings/ChatTemplates.cs
```diff
@@ -5,12 +5,12 @@ namespace Strings
 {
 	public static class ChatTemplates
 	{
-		public const int SmallText = -2;
-		public const int NormalText = 1;
-		public const int LargeText = 8;
-		public const int VeryLargeText = 16;
-		public const int ExtremelyLargeText = 26;
-		public const int AnnoyingText = 48;
+		public const int SmallText = 12;
```
corp-0: if I understand this correctly, size used to be ``-2`` but now it is ``12 + FONTSCALE``. So it still won't be the same it used to be. Are you sure you don't want to multiply instead?

MaxIsJoe: ignore this change

corp-0:
speed merging is my passion

MaxIsJoe:
am SPEEDDDD

PR #9950, opened 2023-07-09
Stating fixs

Bod9001:
### Changelog:
CL: [New] Made the possess right click option work only if you are the server or an admin
CL: [Improvement] Slime cores now have their colour in their names.
CL: [Fix] Fixed Space damage being too strong
CL: [Fix] Fixed release gas reaction when done in inventory not releasing gas at all.
CL: [Fix] Slime processor can now handle multiple cores.
CL: [Fix] Prevents Slimes from double eating something
CL: [Fix] Hopefully fixed the wizard Teleport kick

### DIFF
--- a/UnityProject/Assets/Prefabs/Items/Implants/Slime/SlimeCore -  Adamantine.prefab
+++ b/UnityProject/Assets/Prefabs/Items/Implants/Slime/SlimeCore -  Adamantine.prefab
@@ -110,6 +110,11 @@ PrefabInstance:
       propertyPath: m_Name
       value: SlimeCore -  Adamantine
       objectReference: {fileID: 0}
+    - target: {fileID: 4418078389270877362, guid: 789d74731ecbf8e4a813fc024afaf831,
+        type: 3}
+      propertyPath: initialName
+      value: Adamantine SlimeCore
+      objectReference: {fileID: 0}
     - target: {fileID: 7135682799884072893, guid: 789d74731ecbf8e4a813fc024afaf831,
         type: 3}
       propertyPath: foreverID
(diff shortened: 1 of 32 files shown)


corp-0:
> Fixed release gas reaction when done in inventory

What was wrong with it?

corp-0:
Can you fix the slimes being called "Cuban Pete"? I imagine it is just changing the name in their character sheets when you spawn them. Same with monkeys.

Bod9001:
what should they be called?

Bod9001:
"What was wrong with it?"
it would use the local position of the object that would be hidden pos, now it should actually grab the Official position then convert it to local

corp-0:
> what should they be called?

Slime, Monkey. Could be fun to give them some more personality by randomly picking adjectives so you have Courageous Monkey... Maybe. But just Slime and Monkey is fine I think.

> it would use the local position of the object that would be hidden pos, now it should actually grab the Official position then convert it to loca

in terms of what the players experienced. What was wrong? What did a player see when this bug happened.

Bod9001:
it wouldn't release gas at all

corp-0:
Will you do the name fix for monkeys and slimes? 

Also, your last commit will conflict with your other PR

Bod9001:
I can do if you want to?

corp-0:
> I can do if you want to?

Please

Bod9001:
done

corp-0:
You change it in the prefab? I would have thought it was changed in the CharacterSheet

corp-0 (approved):

corp-0:
nvm I saw the CharacterSheet is exposd to the prefab so that's perfect.

PR #9953, opened 2023-07-10
Xenobio remap and quick fix

### LINKED ISSUE #9951: [BUG] Mobs can attack while inside containers

### Game Version

23070821

### I was

Client

### OS

Windows 8, 10, 11

### What happened?

When placing a mob such as a xenomorph queen inside a locker, the mob will still attack and harm players next to the locker, even if the locker is closed and welded.

Mobs should not be able to attack players when they are inside a closed container. 
Should be as simple as adding a check to hostile mob attack logic checking if they are inside a container first.

1. Spawn a hostile mob with a melee attack such as a space carp or xenomorph queen.
2. Open a locker a lure the creature onto the lockers tile.
3. Close the locker on the mob, trapping it inside.
4. Stand next to the creature trapped inside the locker.
5. The mob will continue attacking you despite being trapped.

### Additional context

Known issue for awhile, just making sure their is an issue for it.

AtnerNT:
Closes #9951 

Capture

### Changelog:
CL: [Improvement] Remaps ministation's xenobiology lab
CL: [Fix] Fixed mobs being able to attack from within lockers and containers

### DIFF
--- a/UnityProject/Assets/Scripts/NPC/AI/MobMeleeAction.cs
+++ b/UnityProject/Assets/Scripts/NPC/AI/MobMeleeAction.cs
@@ -75,6 +75,7 @@ namespace Systems.MobAIs
 		public override void DoAction()
 		{
 			if(isOnCooldown) return;
+			if (objectPhysics.ContainedInObjectContainer != null) return;
 			base.DoAction();
 			var hitInfo = ValidateTarget();
 			if (hitInfo.ItHit == false)


MaxIsJoe:
I'll wait for @Bod9001 approval because I'm not entirely sure yet how the requirements of xenobio here are met or not

Bod9001:
I would say use usual disposal bins, since you might accidentally walk into them, and make sure the ACU has no access requirement

AtnerNT:
> I would say use usual disposal bins, since you might accidentally walk into them, and make sure the ACU has no access requirement

applied

Bod9001 (approved):

PR #9957, opened 2023-07-11
Just some quick fixes and improvements

MaxIsJoe:
Some things I did before going back to study

### Changelog:

CL: [New] Added interaction tooltips to some storage containers.
CL: [Fix] Fixed some characters being forced upright even though they should be laying down.
CL: [Fix] Players will no longer have their sprites randomly stand up when crossing a different Matrix.

### DIFF
--- a/UnityProject/Assets/Scripts/Player/LayDown.cs
+++ b/UnityProject/Assets/Scripts/Player/LayDown.cs
@@ -52,7 +52,10 @@ namespace Player
 			var state = IsLayingDown;
 			if (health != null)
 			{
-				if (health.IsDead || health.IsCrit) state = true;
+				if (health.IsDead || health.IsCrit)
+				{
+					state = true;
+				}
 			}
 			if (state)
 			{
--- a/UnityProject/Assets/Scripts/Player/LayDown.cs
+++ b/UnityProject/Assets/Scripts/Player/LayDown.cs
@@ -103,7 +103,7 @@ namespace Player
 			{
 				spriteRenderer.sortingLayerName = "Players";
 			}
-			if (CustomNetworkManager.IsServer == false) return;
+			if (CustomNetworkManager.IsServer == false || playerScript == null || playerScript.PlayerSync == null) return;
 			playerDirectional.LockDirectionTo(false, playerDirectional.CurrentDirection);
 			playerScript.PlayerSync.CurrentMovementType = MovementType.Running;
 		}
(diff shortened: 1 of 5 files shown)


UnityProject/Assets/Scripts/Player/LayDown.cs
```diff
@@ -42,9 +48,13 @@ public void ServerEnsureCorrectState()
 
 		private void CorrectState()
 		{
-			gameObject.transform.rotation = Quaternion.Euler(0, 0, 0);
-			if (disabled || health == null) return;
-			if (IsLayingDown || health.IsCrit || health.IsDead)
+			if (disabled) return;
+			var state = IsLayingDown;
```
corp-0, line 52: why create a new variable for this?

MaxIsJoe: to check for health, if health is different; affect the variable

corp-0: am I missing something? If you just used ``IsLayingDown`` seems to be exactly the same.

MaxIsJoe: can you read the lines below that

MaxIsJoe:
```cs
			if (health != null)
			{
				if (health.IsDead || health.IsCrit) state = true;
			}
```

corp-0: can you not set ``IsLayingDown`` there instead of ``state``? Is it a readonly property?

corp-0: also give the if a body. Don't do things in the same line as the condition, it is confusing.

MaxIsJoe: > can you not set `IsLayingDown` there instead of `state`? Is it a readonly property?

I can, but you shouldn't change IsLayingDown because you should only use it if you're laying down. Also you'd probably crash the game by calling the server hook for the syncvar over and over again because you keep changing the variable inside the hook.

MaxIsJoe: > also give the if a body. Don't do things in the same line as the condition, it is confusing.

no, bad nesting.

corp-0: no, what you are doing is bad and is against our code guidelines.

Bad nesting is when you have multiple levels of nesting, making it confusing to grasp. This isn't bad nesting, this is just what it should be.

MaxIsJoe: It looks ugly imo, but w/e

corp-0 (approved):

PR #9970, opened 2023-07-15
fix: temporarily undo rotation changes from #9957

corp-0:
### Purpose
Undoes rotation changes from #9957 so players aren't always in a weird angle during this week's playtest. Everything should be working just like it did before that PR which means some dead and uncon people will appear standing but that is probably better than everyone being at weird rotation angles.

### Changelog:
__nocl__

### DIFF
--- a/UnityProject/Assets/Scripts/Player/LayDown.cs
+++ b/UnityProject/Assets/Scripts/Player/LayDown.cs
@@ -48,6 +48,7 @@ namespace Player
 
 		private void CorrectState()
 		{
+			gameObject.transform.rotation = Quaternion.Euler(0, 0, 0);
 			if (disabled) return;
 			var state = IsLayingDown;
 			if (health != null)


MaxIsJoe:
I think you can just revert this one line only:
`gameObject.transform.rotation = Quaternion.Euler(0, 0, 0);`

Bod9001:
I thought the rid rotation was before that PR?

MaxIsJoe:
> I thought the rid rotation was before that PR?

There is something somewhere that still rotates the root gameobject rather than the sprites

corp-0:
> I think you can just revert this one line only: `gameObject.transform.rotation = Quaternion.Euler(0, 0, 0);`

Since I don't completely understand what it is doing, I didn't want to risk it and simply rolled back to the previous state.

MaxIsJoe:
That function ensures that the root object always has its rotation at 0,0 since we only want to rotate the sprites themselves, not the root 
players started being sideways again when I removed that, which means that there are still things messing with the root rotation still that I need to weed out

corp-0:
done. Mergerino fasterino for the playtesterino?

PR #9976, opened 2023-07-19
Several Fixes and improvements

MaxIsJoe:
CL: [Improvement] Several edible/drinkable items utilize the vomit mechanics now.
CL: [Fix] Fixed an error that popped up when the ghost UI would attempt to update itself while inactive.
CL: [Fix] Fixed an NRE when pressing F7 to spawn test characters that caused inputs to misbehave for a few frames.
CL: [Fix] Fixed an issue with edible items that resulted in their bite mechanics to disappear due to an oversight while adding Slimes.

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Food/Edible.cs
+++ b/UnityProject/Assets/Scripts/Items/Food/Edible.cs
@@ -9,6 +9,7 @@ using Messages.Server.SoundMessages;
 using Mirror;
 using Systems.Score;
 using UI.Systems.Tooltips.HoverTooltips;
+using UnityEngine.Serialization;
 
 namespace Items.Food
 {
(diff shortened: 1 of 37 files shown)


Bod9001 (approved):

UnityProject/Assets/Scripts/Items/Food/Edible.cs
```diff
@@ -112,38 +117,34 @@ public override void TryConsume(GameObject feederGO, GameObject eaterGO)
...
 			//TODO: Reimplement metabolism.
-			AudioSourceParameters eatSoundParameters = new AudioSourceParameters(pitch: RandomPitch);
-			SoundManager.PlayNetworkedAtPos(sound, eater.WorldPos, eatSoundParameters, sourceObj: eater.gameObject);
-
-			var Stomachs = eater.playerHealth.GetStomachs();
-			if (Stomachs.Count == 0)
+			var stomachs = eater.playerHealth.GetStomachs();
+			if (stomachs.Count == 0)
 			{
 				//No stomachs?!
```
corp-0, line 139: I feel like there is a message missing here to tell the player they can't eat but I can't think of aynthing to say that would hint at a missing stomach

CorruptComputer: Something like "You try to eat, but the food won't go down!" might work.

Also maybe vomit when you do that? Not sure if that would make sense to show you spitting the food out. Though if this happens you would probably still want to have the chewing sound and take some of the food away for it to make more sense.

corp-0: okay. I will merge this for now. We will have to discuss it further, but I want this merged to test the build candidate for this Saturday and this isn't really a critical issue.

PR #9979, opened 2023-07-20
Fixes an issue that caused players to become sideways

MaxIsJoe:
CL: [Fix] Fixed an issue that caused the root transform of players to become rotated accidentally instead of their sprites, which caused them to appear sideways.


corp-0:
let's merge this and try on staging today so we know it is working as intended for tomorrow

corp-0:
ah, shit. It isn't Friday yet. Still better have stuff tested today and freeze during Friday so we don't get surprise builds during the play test

MaxIsJoe:
all tests passed, sppeeeddd mmeerggeee

corp-0 (approved):

PR #9999, opened 2023-07-28
application open URL Made safev2

Bod9001:
note second PR because I couldn't update the branch of the old one for some bizarre reason
+ moves profile manager into Safe namespace


Bod9001:
@corp-0  so  it already validates the URL with 	if (TrySanitizeURL(URL, out var goodURL ))

				if (Uri.TryCreate(inputURL, UriKind.Absolute, out Uri uriResult) &&
		    uriResult.Scheme == Uri.UriSchemeHttps)
		{

UnityProject/Assets/Assets.asmdef
```diff
@@ -26,7 +26,7 @@
         "Shared",
         "Mirror.Ignorance",
         "Mirror.Transports",
-        "SafeFilesystem"
+        "ConfigurationSaves"
```
corp-0: why?

corp-0 (changes requested):

UnityProject/Assets/Assets.asmdef
Bod9001: it should be renamed to  safe stuff

PR #10012, opened 2023-08-06
Safe API + Cleanup 

Bod9001:
so, this will automatically block any requests leading to internal IPs, 
probably should make it so if it's trusted it bypasses this

the intention is to set up a thing on the hub to validate hosts, 
and if it's unknown then pop-up a prompt Box in the hub to save you want to access this host

also have launch arguments so you can launch a server headless without having hub, and a pop-up saying that the hub is not connected forget the launch arguments

please review


Bod9001:
so, need to think of scenarios where someone makes a DNS that points to an internal IP

MaxIsJoe (approved):
LGTM

PR #10015, opened 2023-08-07
chore: add tests errors to PR comments

corp-0:
### Purpose
This should work for what Max was trying to do, more or less.

### Notes:
The xml from the results are not parsed into human language because I don't know what that structure looks like. It can be parsed, if it isn't a huge pain in the butt, after we see them live.

### Changelog:
__nocl__


corp-0:
lmao it didn't comment shit

PR #10017, opened 2023-08-10
More securrity stuff

Bod9001:
Renames a few name spaces
moves over the rest of the unity requests to the new API
uses json.net to serialise
These are not hard Fast rules, 
just requires time adding them to the white list so if you want them then Recommending them/work on adding them to the white list


corp-0:
I see some changes in the loading process, are you certain tha if I merge this now it won't cause any issue for the Saturday's playtest?

Bod9001:
the never say for certain but tested on build

PR #10024, opened 2023-08-15
Maintrooms Balancing and Improvements

AtnerNT:
### Changelog:
CL: [Improvement] Maintrooms now inhibits teleporters, preventing players from simply teleporting out of the maintrooms.
CL: [Improvement] The exit to the maintrooms is now randomised between 5 different locations.
CL: [Improvement] The maintrooms has been remapped with new areas and different rooms.
CL: [Balance] Maintrooms loot has been rebalanced slightly to prevent players from obtaining gear far above what is normal.
CL: [Fix] Fixed handtele portals being unable to teleport non-player objects.


AtnerNT:
Having some issues with portals right now with the handtele fix, dont merge yet

AtnerNT:
> Having some issues with portals right now with the handtele fix, dont merge yet

Aight this all worked out, will be ready to merge once tests done

UnityProject/Assets/Scripts/Items/Others/HandTeleporter.cs
```diff
@@ -45,6 +48,22 @@ public bool WillInteract(HandActivate interaction, NetworkSide side)
 
 		public void ServerPerformInteraction(HandActivate interaction)
 		{
+			var worldPosEntrance = objectPhysics.OfficialPosition.RoundToInt();
+
+			if (ignoreInhibitors == false)
+			{
+				foreach(TeleportInhibitor inhib in TeleportInhibitor.Inhibitors)
```
Bod9001, line 55: might be worth doing a circle overlap, to see if there's any inhibitors , Though you need to set the Collider to acustom one?

AtnerNT: I did it this way is because each inhibitor will have its own range centered on itself.  Like an inhibitor might have a range of 200 but a second one might only have a range of 10, so a circle overlap wont pick that up unless it's has a ridicously large range, and it also requires like you mentioned, custom colliders. Maybe there's a third method? but I think doing simple range checks here is the best option

AtnerNT: (Also doing a collision check it would still iterate over all the Teleport Inhibitors on the backend, so it doesnt make a performance improvement as is)

Bod9001: ok cool

Bod9001 (approved):

PR #10025, opened 2023-08-15
wrapped in a try catch

Bod9001:
wrapped in a try catch


corp-0:
LESS GOOO IT WORKED, I mean the test results thing. And yeah, you merged a failing PR

PR #10048, opened 2023-09-02
 Security stuff + loggy

Bod9001:
Security stuff + loggy


corp-0:
where is the security stuff?

Bod9001:
Removing file write bytes in wave utility
removing the Conversion for the old character settings
and updating addressable low catalogue method to use the new file access for loading  Addressable catalogue (noting the actual addressable system still needs looking into)

PR #10057, opened 2023-09-08
Fixed bugs and balance with flashes

MaxIsJoe:
CL: [New] Flashbangs will leave a spark before despawning to make their disappearance more visually appealing for those wearing sunglasses.
CL: [Improvement] Flashbangs and handheld flashers will now blind players for shorter amounts of time, but their stun durations have been buffed as a result.
CL: [Fix] Fixed an issue where players would get stunned by flashing entities even when wearing protective gear that would prevent stuns from happening.
CL: [Fix] Fixed an NRE when attempting to flash a HealthV2 character that did not have a `RegisterPlayer` component

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/FlasherBase.cs
+++ b/UnityProject/Assets/Scripts/Objects/FlasherBase.cs
@@ -11,13 +11,15 @@ namespace Objects
 {
 	public class FlasherBase : NetworkBehaviour
 	{
-		[SerializeField] private float maximumDistance = 12f;
+		[SerializeField] protected float maximumDistance = 12f;
 		[SerializeField] protected float flashRadius = 12f;
 		[SerializeField] protected float flashTime = 12f;
-		[SerializeField] protected float weakDuration  = 12f;
+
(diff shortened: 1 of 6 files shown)


MaxIsJoe:
For the record, I hate inheritance and I disapprove of shoving all the flashing properties in one class since not every object will use them. but eh, I have like two other PRs I'm working on, and I don't want to spend more time undoing Bod's changes.

this was meant to be a break for me

UnityProject/Assets/Scripts/Objects/FlasherBase.cs
```diff
@@ -63,24 +64,22 @@ public void FlashInRadius()
 		}
 
 		[Server]
-		public void FlashTarget(GameObject target, float time, float stunnedtime)
+		public void FlashTarget(GameObject target, float time, float stunnedtime, bool checkForProtectiveGear = true)
```
corp-0, line 67: when would you not need to check for protection?

Bod9001: could be stuff that overrides protection, such as a Wizard spell, or super Flash

Bod9001 (approved):

PR #10058, opened 2023-09-08
Makes slime hit detection more reliable while they're eating players

### LINKED ISSUE #9962: [BUG] Can't hit slimes on players off.

### Game Version

23071201

### I was

Client

### OS

Windows 8, 10, 11

### What happened?

NOT ALL THE TIME but something caused it where I had a slime on me and I wasn't able to hit it off. Clicking it with harm intent did nothing but cause the telekinesis sparkles to appear on it. The same happened when other people tried to hit off me.

The player should be able to punch the slime off of them no matter what.

1. Get slime on you
2. ???? something causes bug to happen
3. Try to hit it off and fail

### Additional context

_No response_

MaxIsJoe:
fixes #9962

CL: [Fix] Fixed an issue where the server and client would constantly disagree with each other when deciding who to hit when targeting slimes that are buckled to players. This fix applies to all future mobs and objects that use `MovementSync` as well.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
+++ b/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
@@ -2141,11 +2141,10 @@ public class UniversalObjectPhysics : NetworkBehaviour, IRightClickable, IRegist
 
 	// netid of the game object we are buckled to, NetId.Empty if not buckled
 	[SyncVar(hook = nameof(SyncBuckledToObject))]
-	protected UniversalObjectPhysics ObjectIsBuckling = null;
+	public UniversalObjectPhysics ObjectIsBuckling = null;
 
 	public CheckedComponent<UniversalObjectPhysics> ObjectIsBucklingChecked =
 		new CheckedComponent<UniversalObjectPhysics>();
-
 
 	public UniversalObjectPhysics BuckledToObject;
 
--- a/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
+++ b/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
@@ -2140,8 +2140,8 @@ public class UniversalObjectPhysics : NetworkBehaviour, IRightClickable, IRegist
 	#region Buckling
 
 	// netid of the game object we are buckled to, NetId.Empty if not buckled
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
```diff
@@ -2141,12 +2141,11 @@ public static float SizeToWeight(Size size)
 
 	// netid of the game object we are buckled to, NetId.Empty if not buckled
 	[SyncVar(hook = nameof(SyncBuckledToObject))]
-	protected UniversalObjectPhysics ObjectIsBuckling = null;
+	public UniversalObjectPhysics ObjectIsBuckling = null;
```
corp-0: we shouldn't let anyone change the value of this except inside the sync method, can you make it a property with private setter?

like this:
```suggestion
    [field: SyncVar(hook = nameof(SyncBuckledToObject))]
	public UniversalObjectPhysics? ObjectIsBuckling { get; private set; }
```

UnityProject/Assets/Scripts/Items/Weapons/WeaponNetworkActions.cs
```diff
@@ -84,6 +84,10 @@ public void SetNewDamageValues(float newAttackSpeed, float newAttackDamage, Dama
 	public void ServerPerformMeleeAttack(GameObject victim, Vector2 attackDirection, BodyPartType damageZone, LayerType layerType)
 	{
 		if (victim == null) return;
+		if (playerMove.ObjectIsBuckling.OrNull()?.gameObject != null && playerMove.ObjectIsBuckling is MovementSynchronisation)
```
Bod9001, line 87: This does mean you can't destroy chairs that people are tied to

MaxIsJoe: nope, `MovementSync` is mobs only
so you'll hit chairs still

UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
MaxIsJoe: done

Bod9001 (approved):

PR #10065, opened 2023-09-10
Statting the fixs

Bod9001:
### Changelog:
CL: [New] Added the option to change the throwing behaviour. You can choose between hold to throw or toggle to throw.
CL: [Balance] Player brains are no longer edible
CL: [Fix] Fixed chat bubbles not showing
CL: [Fix] Fixed chat bubble jitter
CL: [Fix] Fixed incorrect fonts in some places

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Systems/MainHUD/UI Bottom/ControlAction.cs
+++ b/UnityProject/Assets/Scripts/UI/Systems/MainHUD/UI Bottom/ControlAction.cs
@@ -12,12 +12,39 @@ namespace UI
 
 		public Image pullImage;
 
+		public static bool ThrowHold = true;
+
 		private void Start()
 		{
 			UIManager.IsThrow = false;
 
 			pullImage.enabled = false;
+			ThrowHold = GetThrowPreference();
 		}
+
+		public static bool GetThrowPreference()
+		{
+			return 1 == PlayerPrefs.GetInt("ThrowHoldPreference", 1);
+		}
+
+
+		public static void SetPreference(bool preference)
+		{
+			if (preference)
+			{
+				PlayerPrefs.SetInt(PlayerPrefKeys.ThrowHoldPreference, 1);
+				ThrowHold = true;
+			}
+			else
+			{
+				PlayerPrefs.SetInt(PlayerPrefKeys.ThrowHoldPreference, 0);
+				ThrowHold = false;
+			}
+
+			PlayerPrefs.Save();
+		}
+
+
 
 		#region Buttons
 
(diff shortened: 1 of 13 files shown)


UnityProject/Assets/Scripts/UI/Systems/MainHUD/UI Bottom/ControlAction.cs
```diff
@@ -12,13 +12,40 @@ public class ControlAction : MonoBehaviour
...
 		{
 			UIManager.IsThrow = false;
 
 			pullImage.enabled = false;
+			ThrowHold = GetThrowPreference();
+		}
+
+		public static bool GetThrowPreference()
+		{
+			return 1 == PlayerPrefs.GetInt("ThrowHoldPreference", 1);
```
corp-0, line 27: ever since alpha launch, the other behaviour has been the default one. Are you sure you want to make hold the deault now?

UnityProject/Assets/Scripts/UI/Systems/MainHUD/UI Bottom/ControlAction.cs
```diff
@@ -12,13 +12,40 @@ public class ControlAction : MonoBehaviour
...
 		}
 
+
+		public static void SetPreference(bool preference)
```
corp-0: change the name to ``SetThrowHoldPreference`` otherwise, the name is way too generic and can cause confusion on other developers when reading it.

UnityProject/Assets/Scripts/UI/Systems/MainHUD/UI Bottom/ControlAction.cs
```diff
@@ -12,13 +12,40 @@ public class ControlAction : MonoBehaviour
...
+
 		private void Start()
 		{
 			UIManager.IsThrow = false;
 
 			pullImage.enabled = false;
+			ThrowHold = GetThrowPreference();
+		}
+
+		public static bool GetThrowPreference()
```
corp-0: naming this ``GetTrowHoldPreference`` makes it clearer than a ``true`` value means the throw behaviour is set to hold.

UnityProject/Assets/Scripts/UI/Systems/MainHUD/UI Bottom/ControlAction.cs
Bod9001: yeah,  anyone can change it back in the preferences if they would like, as I think it's more user-friendly to have it a hold thing,  since a new user could press it accidentally and then wonder why they throw when they click

corp-0 (approved):

PR #10067, opened 2023-09-13
Better ghost teleport

Bod9001:
### Changelog:

CL: [Improvement] Improved ghost teleport by removing unnecessary mobs from the mobs list.

### DIFF
--- a/UnityProject/Assets/Scripts/Player/Mind.cs
+++ b/UnityProject/Assets/Scripts/Player/Mind.cs
@@ -40,6 +40,16 @@ public class Mind : NetworkBehaviour, IActionGUI
 	//Type of Antagonist
 	[field: SyncVar] public JobType NetworkedAntagJob { get; private set; }
 
+	public GameObject ControllingObject
+	{
+		get
+		{
+			if (IsGhosting) return this.gameObject;
+
+			return CustomNetworkManager.Spawned[IDActivelyControlling].gameObject;
+		}
+	}
+
 	public GameObject PossessingObject
 	{
 		get
@@ -100,7 +110,8 @@ public class Mind : NetworkBehaviour, IActionGUI
 	/// </summary>
 	private Dictionary<string, object> properties = new Dictionary<string, object>();
 
-	public bool NonImportantMind = false;
+
+	[SyncVar] public bool NonImportantMind = false;
 
 	public bool IsMute
 	{
--- a/UnityProject/Assets/Scripts/Player/TeleportUtils.cs
+++ b/UnityProject/Assets/Scripts/Player/TeleportUtils.cs
@@ -36,17 +36,10 @@ namespace Systems.Teleport
 			foreach (Mind player in sortedStr)
 			{
 				//Don't add to the list the same player consulting it and ghosts.
-				if (player == PlayerManager.LocalMindScript)
+				if (player == PlayerManager.LocalMindScript || player.NonImportantMind)
 				{
 					continue;
 				}
-
-				if (player.NonImportantMind)
-				{
-					continue;
-				}
-
-
 
 				//Gets Name of Player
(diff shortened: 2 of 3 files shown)


UnityProject/Assets/Scripts/Player/TeleportUtils.cs
```diff
@@ -21,25 +22,32 @@ public static class TeleportUtils
...
+			foreach (Mind player in sortedStr)
 			{
 				//Don't add to the list the same player consulting it and ghosts.
-				if (player == PlayerManager.LocalPlayerScript || player.IsGhost)
+				if (player == PlayerManager.LocalMindScript)
 				{
 					continue;
 				}
 
+				if (player.NonImportantMind)
```
corp-0: add this to the previous check with an OR

UnityProject/Assets/Scripts/Player/TeleportUtils.cs
```diff
@@ -48,17 +56,27 @@ orderby name.name
 					nameOfObject = "Spectator";
 				}
 
-				var status = player.PlayerType switch
+				string status = "";
+
+				if (player.IsGhosting)
+				{
+					status = "(Ghost)";
```
corp-0, line 56: we don't need to see ghosts in the list. They were omitted from that list in this PR https://github.com/unitystation/unitystation/pull/9969

Bod9001: well this now tracks by a mind so this tell someone if someone has ghosted or not

Bod9001: so you won't get duplicate entries from the player and the ghost

corp-0: what use does it have to spectate a ghosted player?

UnityProject/Assets/Scripts/Player/Mind.cs
```diff
@@ -100,7 +110,8 @@ public IPlayerPossessable PlayerPossessable
 	/// </summary>
 	private Dictionary<string, object> properties = new Dictionary<string, object>();
 
-	public bool NonImportantMind = false;
+
+	[SyncVar] public bool NonImportantMind = false;
```
MaxIsJoe, line 114: wouldn't it be better to just have different types of minds so we can add filtering for them in the future? this looks like a quick hack to a single problem that we have elsewhere

UnityProject/Assets/Scripts/Player/TeleportUtils.cs
Bod9001: well, you can look at the list and tell if your favourite antagonist has died

UnityProject/Assets/Scripts/Player/Mind.cs
Bod9001: seems fine to bung all the data into one mind, since any generic behaviour should be defined by antagonist/objectives/body/spells

MaxIsJoe: you misuderstood what I said. I meant that it would be better to tag what kind of mind are we dealing with instead of just saying "is this mind important or not"

Bod9001: The problem is what's the difference between a player monkey and a monkey cube monkey?

UnityProject/Assets/Scripts/Player/TeleportUtils.cs
corp-0: that doesn't sound like a good thing to have

Bod9001: it's not that bad having that little bit of extra information

PR #10069, opened 2023-09-15
Fixes dropdown menus using broken fonts

MaxIsJoe:
This has been getting on my nerves so I fixed it.

CL: [Fix] Fixed the right click dropdown menu entries using broken fonts.


corp-0:
what font was it using? What font does it use now?

MaxIsJoe:
it was using fontawesome and now it uses ubuntu mono

corp-0 (approved):

PR #10071, opened 2023-09-17
Staging improvements

Bod9001:
### Changelog:
CL: [New] added portable scrubbers
CL: [New] Footsteps have a small chance to make floors dirty
CL: [Improvement] Better descriptions for integrity of items
CL: [Balance] Genetic console balancinng, you get rewarded number of points by number of sliders in the DNA generator, there is no cap as well
CL: [Fix] Fixed chat bubbles appearing twice.
CL: [Fix] Fixed chat bubbles always having obscured text when whispering, disregarding distance to the origen of the whisper.
CL: [Fix] Fixed getting stuck when exiting disposals through a pipe exit


corp-0:
since walking can make the tiles dirty now, I think the initial mess needs to be balanced again and reduced in quantity (something I believed even before this change, but now it makes it more relevant)

UnityProject/Assets/Scripts/UI/Objects/Medical/genetics/MutationUnlockMiniGame.cs
```diff
@@ -102,7 +97,7 @@ public void TryGenerateEgg()
 	{
 		if (GUI_DNAConsole.DNAConsole.CurrentDNACharge >= GUI_DNAConsole.DNAConsole.RequiredDNASamples)
 		{
-			GUI_DNAConsole.DNAConsole.CurrentDNACharge = 0;
+			GUI_DNAConsole.DNAConsole.CurrentDNACharge =- GUI_DNAConsole.DNAConsole.RequiredDNASamples;
```
corp-0: like codacy asks, was this intentional? Do you want to asign the negative of ``RequiredDNASamples`` ? If so, then separate the minus symbol from the assignation symbol and attach it to the variable itself.

UnityProject/Assets/Scripts/Player/MovementV2/MovementSynchronisation.cs
```diff
@@ -1047,18 +1044,29 @@ public void CMDTryEscapeContainer(MoveAction moveAction)
 		}
 	}
 
+	public void HandleFootstepLogic()
+	{
+		if (RNG.Random.Next(0, 1000 + 1) == 1000)
```
corp-0: why do chances like this and not using the ``DMMath.Prob`` method instead? That one takes an int so you can directly type the percentage of how likely something is to happen

Pockets-byte:
the item integrity feedback isn't completely correct- it's been a while, but as an example I believe the final state is "it's falling apart"  I haven't checked it directly though, so I can't actually remember the full set

UnityProject/Assets/Scripts/Player/MovementV2/MovementSynchronisation.cs
MaxIsJoe: `Random.Next` is also slightly more expensive because it has to create an array in the background to generate the random number.

UnityProject/Assets/Scripts/UI/Objects/Medical/genetics/MutationUnlockMiniGame.cs
Bod9001: yeah,  remove the required amount from the pool

PR #10072, opened 2023-09-20
Adds in an anti-stuck timer to logins

MaxIsJoe:
CL: [Fix] Temporarily added a safety mechanism to logging in to avoid getting stuck forever without any feedback for new players.

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/LobbyManager/LobbyManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/LobbyManager/LobbyManager.cs
@@ -81,12 +81,13 @@ namespace Lobby
 			lobbyDialogue.ShowLoadingPanel("Signing in...");
 
 			bool isLoginSuccess = false;
+			Loggy.Log("[LobbyManager/TryLogin()] - Executing FirebaseAuth.DefaultInstance.SignInWithEmailAndPasswordAsync()");
 			await FirebaseAuth.DefaultInstance.SignInWithEmailAndPasswordAsync(email, password)
 					.ContinueWithOnMainThread(async task =>
 			{
 				if (task.IsCanceled)
 				{
-					Loggy.LogWarning($"Sign in canceled.", Category.DatabaseAPI);
+					Loggy.LogWarning($"Sign in canceled.");
 					lobbyDialogue.ShowLoginPanel();
 				}
 				else if (task.IsFaulted)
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Managers/LobbyManager/LobbyManager.cs
```diff
@@ -321,6 +353,16 @@ private void DetermineUIScale()
 			}
 		}
 
+		private async void LoginTimer()
+		{
+			await Task.Delay(14000);
+			if (cancelTimer) return;
+			lobbyDialogue.ShowLoadingPanel("This is taking longer than it should..\n\n If it continues, try disabling your VPNs and installing the game in full English path.");
```
Bod9001, line 360: loadingPanelScript.SetActive(true);
Unity will complain about not being on the main game thied

MaxIsJoe: but everything else is also working on async?

Bod9001: Is it?

Bod9001: Yeah but it hasn't hit an awite yet so sill on mine thied

MaxIsJoe: I've tested this multiple times and there doesn't seem to be any issues.
Also we're calling `lobbyDialogue.ShowAlphaPanel();` and `lobbyDialogue.ShowLoadingPanel();` without awaiting so clearly its working just fine.

Bod9001: the point I'm making is that because it hasn't hit an await yet it still on the main thread, but anything after a await  is on a non main thread

MaxIsJoe: why is it not complaining then?

MaxIsJoe: is it perhaps not complaining because we're not accessing the `gameObject` property and have a direct reference to what we want to work with?

Bod9001 (approved):

corp-0 (approved):

PR #10081, opened 2023-10-01
Staging work

Bod9001:
### Changelog:
CL: [New] you can now select the sprite of what the pill dispenser will dispense
CL: [Improvement] pressing escape when you're in a UI tab should close it now
CL: [Fix] Fixed machine connectors being invisible
CL: [New] You can now make machine connectors in the Lathe
CL: [New] Some EVA Helmets have Flash protection


UnityProject/Assets/Scripts/Core/SecureStuff/HubValidation.cs
```diff
@@ -56,8 +76,10 @@ private enum ClientRequest
 
 		private class URLData
 		{
-			public List<string> AllowedOpenHosts = new List<string>();
-			public List<string> AllowedAPIHosts = new List<string>();
+			public HashSet<string> AllowedOpenHosts = new HashSet<string>();
```
corp-0: why do you need the same fields on the inner and outer class? Couldn't you just make a field of type ``URLData``?

Bod9001: oh?

corp-0: these 3 fields with the very same type and name are present in the outter class, ``HubValidation`` which is why codacy is complaining. The inner class shadows the fields of the inner class and that is confusing for everyone.

UnityProject/Assets/Scripts/Core/SecureStuff/HubValidation.cs
```diff
@@ -104,21 +127,36 @@ private static void SaveCashedURLConfiguration(URLData URLData)
 
 		}
 
-		private static void AddTrustedHost(string Host, bool IsAPI)
+		private static void AddTrustedHost(Uri URL, bool IsAPI)
 		{
 			if (IsAPI)
 			{
-				AllowedAPIHosts.Add(Host);
+				if (URL.Host == "raw.githubusercontent.com")
```
corp-0:
```suggestion
				if (URL.Host == "raw.githubusercontent.com")
```
move the string to a constant

UnityProject/Assets/Scripts/Core/SecureStuff/HubValidation.cs
```diff
@@ -104,21 +127,36 @@ private static void SaveCashedURLConfiguration(URLData URLData)
...
+				{
+					var StringURL = URL.ToString();
+
+					var splits = StringURL.Split("/");
+					if (splits.Length < 6)
+					{
+						return;
+					}
+					AllowedGithubRepositories.Add(splits[3].ToLower() + "/" + splits[4].ToLower() + "/" + splits[5].ToLower());
```
corp-0: It is probably better to use the segments of the url, instead of manually splitting the string. I would also advise to name the segments so people who are reading the code can follow it without hassle.

```cs
private static void AddTrustedHost(Uri url, bool isAPI)
{
    if (isAPI)
    {
        if (url.Host == "raw.githubusercontent.com")
        {
            var segments = url.Segments;

            // Expected format: /username/reponame/branchname/...
            if (segments.Length >= 4) 
            {
                var username = segments[1].TrimEnd('/');
                var repoName = segments[2].TrimEnd('/');
                var branchName = segments[3].TrimEnd('/');
                
                AllowedGithubRepositories.Add($"{username.ToLower()}/{repoName.ToLower()}/{branchName.ToLower()}");
            }
        }
        else
        {
            AllowedAPIHosts.Add(url.Host);
        }
    }
    else
    {
        AllowedOpenHosts.Add(url.Host);
    }

    SaveCashedURLConfiguration(new URLData()
    {
        AllowedAPIHosts = AllowedAPIHosts,
        AllowedOpenHosts = AllowedOpenHosts,
        AllowedGithubRepositories = AllowedGithubRepositories
    });
}
```

corp-0 (changes requested):

UnityProject/Assets/Scripts/Core/SecureStuff/HubValidation.cs
Bod9001: idk if you can serialise static class could change the names?

corp-0: What's the use for the fields in the outer class?

Bod9001: Quick access

corp-0: okay, yeah rename them and it's all good but I think it could be just a field of ``URLData`` type

Bod9001: cool

PR #10087, opened 2023-10-07
feat: fluent utility to create Rich texts with TMP

corp-0:
### Purpose
When merged this PR will add the ``RichText`` utility class that will hopefully aid developers when doing fancy text with TMP.

Usage example:

```cs
RichText myText = new()
	.Size(40).Color(Color.Red).Center()
	.ApplyTo("Captain Announcement")
	.AddOnNewLine("Vital announcement for ")
	.Bold().ApplyTo("ALL CREWMEMBERS").Append(".");
```

And it would produce: 
```
<size=40><color="red"><align="center">Captain Announcement</align></color</size>
Vital announcement for <b>ALL CREWMEMBERS</b>.
```

Since ``RichText`` is implicity converted to string, you can use your instance anywhere it takes an string as argument, like the chat API for example.
	
The entire TMP api has been implemented with the following exceptions:
- Sprite is partially implemented. It is missing the tint and color optional properties.
- Page break doesn't have an example of usage, which makes it hard to know how to implement it.

### DIFF
--- a/UnityProject/Assets/Scripts/Util/Independent/FluentRichText/Styles/CommonValidations.cs
+++ b/UnityProject/Assets/Scripts/Util/Independent/FluentRichText/Styles/CommonValidations.cs
@@ -1,0 +1,27 @@
+﻿using System.Text.RegularExpressions;
+
+namespace Util.Independent.FluentRichText.Styles
+{
+	public static class CommonValidations
+	{
+		public static bool IsValidPixelValue(string value)
+		{
+			return Regex.IsMatch(value, @"^-?\+?\d+(\.\d+)?$");
+		}
+
+		public static  bool IsValidFontUnitValue(string value)
+		{
+			return Regex.IsMatch(value, @"^-?\+?\d+(\.\d+)?em$");
+		}
+
+		public static bool IsValidPercentageValue(string value)
+		{
+			return Regex.IsMatch(value, @"^-?\+?\d+(\.\d+)?%$");
+		}
+
+		public static bool IsValidHexColor(string hexColor)
+		{
+			return Regex.IsMatch(hexColor, @"^#([a-fA-F0-9]{6}|[a-fA-F0-9]{8})$");
+		}
+	}
+}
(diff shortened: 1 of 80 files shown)


Bod9001:
I feel like it would be easier if it was something like

String = "all crewmates make their way to the HOP for hat inspection".bold(). Centre(). Size(23);

corp-0:
this is now ready for review

UnityProject/Assets/Scripts/Util/Independent/FluentRichText/Styles/CommonValidations.cs
```diff
@@ -0,0 +1,27 @@
...
+		}
+
+		public static bool IsValidPercentageValue(string value)
+		{
+			return Regex.IsMatch(value, @"^-?\+?\d+(\.\d+)?%$");
+		}
+
+		public static bool IsValidHexColor(string hexColor)
+		{
+			return Regex.IsMatch(hexColor, @"^#([a-fA-F0-9]{6}|[a-fA-F0-9]{8})$");
```
NoooneyDude, line 24: My understanding is three or four characters like #fff should also be considered valid - or does TMP not support that?

NoooneyDude (approved):
Looks good. Some thoughts I had:
- Any consideration for localisation? This seems to be another step in the way of l10n.
- No concerns about instantiating and immediately discarding stateless objects? They look nice and are small enough, I guess..
- Error level might be a bit too aggressive to be uncategorized? A warning might be sufficient or a new category for these.

UnityProject/Assets/Scripts/Util/Independent/FluentRichText/Styles/CommonValidations.cs
corp-0: I'm not sure, I based all the implementation on the one site that has the TMP formatting documentation

MaxIsJoe (approved):
LGTM

PR #10090, opened 2023-10-07
Silly errorr

Bod9001:
it is not being able to join server straightaway without having to login again


corp-0:
delay this so the build doesn't interrupt the ongoing playtest

PR #10092, opened 2023-10-07
attempts to fix #10086

### LINKED ISSUE #10086: [BUG] Clients keep resetting the rotations of laying down bodies.
https://github.com/unitystation/unitystation/assets/34368774/bf211362-5a91-47eb-8fcf-b9e14630eb26

# A bit of history

Unitystation has had a problem with player bodies and their rotation for quite some time now. The issue first appeared after HealthV2, where dead players would constantly appear standing up when they should be laying down instead.
The introduction of the new movement system did not help, as the issue was now more wide spread in a lot of scenarios.

In an attempt to solve this issue, I introduced the `LayDown` component. A component that is designed to control the player's rotation without having to swift through a bunch of duplicate rotation code.
This component was largely successful in fixing a lot of the issues that were introduced with the new movement system, such as player bodies standing up when they're laying down inside of lockers or the burning sprites of players not following the rotation of the player sprites.
However, even after fixing all these issues; the main problem that this component was trying to solve only got worse and was starting to be more disruptive for all players.

# Deep dive into the issue

After debugging this issue for several months, I've narrowed down my observations into these:

- When the player stands up, only their `Sprites` transform is affected.
- Commenting out the sorting layer changes from the `Laydown` Component rules out that its a layering issue.
- Only clients are affected by this, the server always preserves the correct state.
- High ping makes state desyncs worse.

So, with us narrowing down the behavior of this issue; It's time to look at the first culript.

#### Sprites Transform

A quick look into the `LayDown` component on rider, we can see that there is only one reference to the sprites transform used in the entire game.

image

And its use case is this:
image

Odd, right? The whole point of the `LayDown` component was to avoid other components from touching the sprites transform.
But in this case, there is a legitimate use case for referencing this.

```cs

rotationTarget.Rotate(new Vector3(0, 0, spinMagnitude * NewtonianMovement.magnitude * Time.deltaTime));

```

This `Rotate()` function is called within the `FlyingUpdateMe()` function, the one that is responsible for rotating the player/object when they're getting thrown around in space and are spinning.
It would be quite expensive to call `GetComponent<LayDown>()` every frame due to GC, so we cache a reference to the sprites transform that we use to rotate the player around in laydown behavior on `Awake()`, and default it to the root transform if there are no `LayDown` compontents.

image

So now that we've established a reason for why this is referenced here to start with, let us look at other culprits.

#### 10 references with one delete 

image

`rotationTarget` has 10 total references being used, all of them are inside `UniversalObjectPhysics` so we can safely say that the issue is clearly related to movement/physics, **right?**

To quickly answer that question, I did what every big brained programmer would do, and I just nuked all code that affects rotations in UOP. Surely, because of that, we would slowly close the case now as we figure out which reference in the 10 are the ones causing this issue; Right???

Nope, if you delete all code related to rotations in UOP; The issue still persists.
Meaning that there is a component somewhere else that is accessing the `Sprites` transform (somehow??) and is directly manipulating the transform's rotation.

If you are as confused as me, pull up your Jaw; because it gets worse.

#### Unitystation rarely rotates stuff.

Fun fact, unitystation rarely uses `transform.rotation`.

image
image

You might be thinking that 100+ use cases of rotation is a lot, but when you look deeper into what is being rotated; You'll notice that most of these are just boilerblate code for LeanTween, UI or very niche things like 3D mode. 
The only gameplay element that makes heavy use of rotations, is projectiles.

By this time, you might have come to the conclusion that I've already swifted through the entire codebase to find anything that might cause rotations to get messed up for players; and you'd be correct.
There is nothing that I could find that directly breaks the sprite transform, and if you think about it logically; we've only started using the sprites transform for rotations very recently; yet the old issues that were on the Root transform of the player are now applied to the sprites transform instead; so there couldn't be any code anywhere that should have affected this, and it should have affected the root transform instead.

# A possible theory.

A few days ago, Our resident Fox and Lead Developer, Bod, recommended me that I use a program called [Clumsy ](https://github.com/jagt/clumsy) to simulate network lag on local host. This program allowed me to replicate the issues we had on staging more reliably locally, and has led me to introduce #10082 to attempt hiding this issue.

My bandage partially hid the issue, sadly, but it ticked something in my head.

**The issue only happens in an environment where latency is high.**

I don't know how to confirm this, but if anyone is more knowledgeable in networking and/or mirror, please test out if this issue is indeed a networking related one.

I've ran out of things to test, and this is the final conclusion that I can come up with.
If its a networking issue, there is little I can do to fix this problem, and we'll continue having player rotations jank forever.

MaxIsJoe:
#10086

CL: [Fix] Finally catches (probably) the last issue that kept clashing with the LayDown component that kept messing with player rotations.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
+++ b/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
@@ -574,9 +574,6 @@ public class UniversalObjectPhysics : NetworkBehaviour, IRightClickable, IRegist
 			{
 				c.playerScript.RegisterPlayer.LayDownBehavior.ServerEnsureCorrectState();
 			}
-			else
-			{
-			}
 		}
 		else
 		{
@@ -1137,9 +1134,6 @@ public class UniversalObjectPhysics : NetworkBehaviour, IRightClickable, IRegist
 		if (CustomNetworkManager.IsServer)
 		{
 			c.playerScript.RegisterPlayer.LayDownBehavior.ServerEnsureCorrectState();
-		}
-		else
-		{
 		}
 	}
 
--- a/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
+++ b/UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
@@ -576,7 +576,6 @@ public class UniversalObjectPhysics : NetworkBehaviour, IRightClickable, IRegist
 			}
 			else
 			{
-				c.playerScript.RegisterPlayer.LayDownBehavior.ClientEnsureCorrectState();
 			}
 		}
 		else
@@ -1141,7 +1140,6 @@ public class UniversalObjectPhysics : NetworkBehaviour, IRightClickable, IRegist
 		}
 		else
 		{
-			c.playerScript.RegisterPlayer.LayDownBehavior.ClientEnsureCorrectState();
 		}
 	}
 
--- a/UnityProject/Assets/Prefabs/Player/Resources/Player_V4(uNet).prefab
+++ b/UnityProject/Assets/Prefabs/Player/Resources/Player_V4(uNet).prefab
@@ -1621,7 +1621,7 @@ MonoBehaviour:
   playerScript: {fileID: 114617133355526680}
   networkedLean: {fileID: -8106636780342977062}
   disabled: 0
-  clientAutoCorrectInterval: 0.75
+  uprightSprites: {fileID: 0}
 --- !u!114 &1812110424391118221
 MonoBehaviour:
   m_ObjectHideFlags: 0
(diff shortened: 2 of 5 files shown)


UnityProject/Assets/Scripts/Core/UniversalObjectPhysics.cs
```diff
@@ -1141,7 +1140,6 @@ public void ResetEverything()
 		}
 		else
 		{
-			c.playerScript.RegisterPlayer.LayDownBehavior.ClientEnsureCorrectState();
```
corp-0, line 1144: if the else is empty, remove the else

MaxIsJoe: done

Issue #7226, opened 2021-08-09
Hosts don't see other players lerp as they melee attack, for some reason.


PetMudstone:
### Environment
- Game Version: Staging 21080719
- OS Version: Windows 10

### Detailed description of bug
For some reason, hosts don't see other players' sprites move as they melee attack. This does not affect clients.

Host on left, client on right.
![PunchInconsi](https://user-images.githubusercontent.com/38266309/128659827-f583f8df-88db-4c48-874d-9f8588fdb04f.gif)

### Steps to reproduce
1. Launch a build of Unitystation from the Installations tab so you can be a host. Start a game.
2. Launch the same build of Unitystation again, join the host's game.
3. Move your host and client characters within view of eachother.
4. Have both instances of Unitystation side by side.
5. Melee attack using your client and host characters. Notice that from the host's point of view, the client character's sprite does not move while they are melee attacking.

PR #10122, opened 2023-10-10
Update bug_report.yml

Fr-Dae:
CL: [Fix] serveur options

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -23,7 +23,8 @@ body:
     description: Were you the client or the host when the issue occured?
     options:
     - Client
-    - Server
+    - Server LocalHost
+    - Server Headless 
   validations:
     required: true
 - type: dropdown


Bod9001 (approved):

corp-0 (changes requested):
this is not what the options are supposed to mean

Fr-Dae:
do you prefer, players, host headless, local host?

because that's exactly how people use this option. 

otherwise what's the use?

corp-0:
The use is exactly what is says in the label. Were you the client or the host when the issue occured?

Fr-Dae:
> The use is exactly what is says in the label. Were you the client or the host when the issue occured?

but there's a distinction to be made between 
- the client
- the headless which host + play, (error found via the server terminal or in joining game  in localhost)
- The LocalHost (game launched from the build in "host" mode and which closes as soon as the game closes)

these are the three different ways of connecting to the game, and some errors only occur in one of these modes. 

that's why I wanted to make this clear in the model

MaxIsJoe:
Closing this after a brief discussion with other maintainers.
This PR is largely abandoned, and the changes are unnecessary because we do not look for issues with hosts playing on the same exact client they're hosting from (the host is not meant to play the game from the same hosting process anyways).

Fr-Dae:
> the host is not meant to play the game from the same hosting process anyways

I'm sorry, what ? how does players who want to test in local solo mode be able to do so?

MaxIsJoe:
You host an instance, then start another one that connects to the server.

Fr-Dae:
> You host an instance, then start another one that connects to the server.

what you're saying isn't logical. 
a player wants to play solo or with a friend - he clicks on the [host] button.
Why should he have to do anything other than the button that is literally on the game menu?

MaxIsJoe:
Because unitystation and ss13 are not a co-op games, and are not designed with that in mind.

Fr-Dae:
they're not co-op? 
in other words? you mean that, although the game offers the possibility of playing by yourself, and hosting your own server locally, a user should not do this and ONLY 
connect to the EU-USA server to test the game? 

and if I only need to play with people on my local network, I'm not allowed to?

MaxIsJoe:
You're only allowed to play the game on the same instance as the host because:

1. Unity never had the ability to launch more than one instance at a time without rebuilding the entire game every time you wanted to test a change.
2. Our project structure from the start was so bad, that the server and the client both were crammed together. (This is bad for a million different reasons for games like SS13)

The game offers this functionality as a side effect of a series of uninformed decisions and engine limitations. We never use the local instance of the host as a reflection on what happens in the game, because the server intentionally has various different behaviors - or are completely missing ones - to lower the processing requirement of unitystation (check #7226 for example)

If you run into an issue as the local host playing on the same instance, we have no clue if this issue is intended behavior or an actual issue; and even if its an actual issue, we don't want to waste our time fixing something that is not intended to be seen or experienced by other people in the first place.

Unitystation is a demanding game, this is why we always tell users to use docker and headless instances to give the server all the resources it needs. So understandably, we expect all servers to be hosted in a headless mode where the host can't play the game at all unless he connects to the server as a client.

MaxIsJoe:
> and if I only need to play with people on my local network, I'm not allowed to?

No one is stopping you from playing the game the correct or wrong way. If you want to play on the same instance as the host, you are allowed to do so, even if it's not the intended way of playing.
Just don't clog our issue tracker with issues that are not worth our time because you peaked behind the curtains.

Fr-Dae:
in the real world, there are three scenarios.

- The player connects to the server on the hub (e.g. USA) and plays.
- the player clicks on the "host" button and plays. (why not with friends on their local ip)

- the rare person who wants to do a serious project and offer performance to a dedicated computer to run the server to become host, 
they will be interested in a headless management, with or without docker. but this last category is only a tiny %, nobody does that, unless he really want to set up a private server.


to test a specific mechanism in a build it's much more practical to run it locally (selfhosted) with admin rights, do the test and give the result on discord/github

MaxIsJoe:
In the real world:

A developer only tests the game from local host after making or testing a change, they see no issues, then merge it and then players discover later that headless and clients are broken and are unable to play the game properly, and we're forced to rollback to an older build immediately.

This is a scenario we ran into a million times, especially me and Bod. So I can confidently say that as someone who has worked on this project for more than 3 years at this point and has seen all sorts of issues this project has; your assumption of how things should operate is detached from how the project operates on every level.

What you *think* is practical does not align with how the game works. 

This discussion is getting locked 🔒.

PR #10150, opened 2023-10-25
Bug fixing

### LINKED ISSUE #4914: Window won't stay maximized
## Description

When running in windowed mode, the window won't stay maximized.
up to one second after being maximized, it will pop out of maximized state, and be slight offset from the top and left edge of the screen.
Something like this seems to have been reported before. #836

### Detailed Bug Report
I suspect only the display and its configuration is interesting here, so i'll include that for now.
I have two DELL U2515H 25" monitors installed. One horizontally oriented, and the other vertically oriented on the right side.
The first is set as 2560x1440 and the other as 1440x2560.
Maximizing the window on either of these produces the same result.

### Steps to Reproduce

1. Get a 25" monitor, i suspect. 💸
2. Configure it for 2560x1440 resolution.
3. Start game in windowed mode.
4. Maximize the game window.

### Extra
If you need me for some interactive debugging, i am available in GMT+2 usually between 8pm and 10pm.

### LINKED ISSUE #10043: [BUG] Movement Queue ignores inputs while animating from one tile to another.

### Game Version

23082521

### I was

Client

### OS

Windows 8, 10, 11

### What happened?

https://cdn.discordapp.com/attachments/273774715741667329/1145210820431728691/2023-08-27_00-17-53.mp4

Players have their inputs ignored while moving from one tile to another.

The game should queue the next position the player wants to go to so that they continue moving after they reach the next tile to match SS13's movement.

https://cdn.discordapp.com/attachments/273774715741667329/1145210820431728691/2023-08-27_00-17-53.mp4

1. Move to a tile
2. Spam a direction
3. Observe as you don't continue moving to the direction you're aiming for after reaching your destination

### Additional context

_No response_

Bod9001:
### Changelog:
CL: [Fix] Fixed inputs done while transitioning from a tile to another being ignored. Inputs will be queued until you initiate a new move then it will be reset or let go of all movement keys
CL: [Fix] Fixed odd resolutions not being able to stay maximized when the maximize button was pressed.

fixes #10043
fixes #4914

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Core/ResponsiveUI.cs
+++ b/UnityProject/Assets/Scripts/UI/Core/ResponsiveUI.cs
@@ -124,11 +124,6 @@ public class ResponsiveUI : MonoBehaviour, IInitialise
 
 		main.rect = new Rect(0, 0, width / (float) Screen.width, height / (float) Screen.height);
 
-		// if (requiresChange && !Screen.fullScreen)
-		// {
-		// 	Screen.SetResolution(width, height, false);
-		// }
-
 		screenWidthCache = Screen.width;
 		screenHeightCache = Screen.height;
 
--- a/UnityProject/Assets/Scripts/UI/Core/ResponsiveUI.cs
+++ b/UnityProject/Assets/Scripts/UI/Core/ResponsiveUI.cs
@@ -124,10 +124,10 @@ public class ResponsiveUI : MonoBehaviour, IInitialise
 
 		main.rect = new Rect(0, 0, width / (float) Screen.width, height / (float) Screen.height);
 
-		if (requiresChange && !Screen.fullScreen)
-		{
-			Screen.SetResolution(width, height, false);
-		}
+		// if (requiresChange && !Screen.fullScreen)
+		// {
+		// 	Screen.SetResolution(width, height, false);
+		// }
 
 		screenWidthCache = Screen.width;
 		screenHeightCache = Screen.height;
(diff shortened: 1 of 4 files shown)


eurtod:
hmm the input queue might be dangerous. if you are on the ground crawling super slow and decide to spam w or something you might get stuck being forced to move upwards for awhile idk.

MaxIsJoe:
> hmm the input queue might be dangerous. if you are on the ground crawling super slow and decide to spam w or something you might get stuck being forced to move upwards for awhile idk.

I think this can be solved by adding decay to the queued inputs

corp-0:
agree, can you add decaying time for the queued inputs? Personally, instead of having just two variables for caching them I would have an actual FIFO collection there, like a Queue<T>. You would then burn older inputs as the decay time is reached.

UnityProject/Assets/Scripts/UI/Core/ResponsiveUI.cs
```diff
@@ -124,10 +124,10 @@ private IEnumerator ForceGameWindowAspect()
 
 		main.rect = new Rect(0, 0, width / (float) Screen.width, height / (float) Screen.height);
 
-		if (requiresChange && !Screen.fullScreen)
-		{
-			Screen.SetResolution(width, height, false);
-		}
+		// if (requiresChange && !Screen.fullScreen)
```
corp-0: just remove the code

Bod9001:
well it resets if you let go of wasd

PR #10172, opened 2023-11-06
Move into project Tomlyn

Bod9001:
Why?
because the package manager often fails to grab Tomlyn leaving the project with compile errors and The only way to fix is to restart the editor that can be slow sometimes
and 
it also happens with the build process causing you to have to build twice


UnityProject/Assets/packages.config
```diff
@@ -1,4 +1,2 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<packages>
```
corp-0, line 2: this file is no longer needed if you uninstall NugetForUnity

Bod9001: Didn't uninstall NugetForUnity , just removed Tomlyn  from

corp-0: NugetForUnity isn't needed anymore if you move tomlyn to the project itself

PR #10181, opened 2023-11-12
 -trusted for headless Server

Bod9001:


Fr-Dae:
what does the option --trusted exacly ?

PR #10192, opened 2023-11-18
fixes for Valley

Bod9001:
Teleport inhibitor for Blue space bananas
no flashing indicator when no power for door buttons


FuntimeDudeski:
ILY <3
Is there a 'Button is unpowered' hud message? That would make it consistent with airlocks and consoles.

PR #10206, opened 2023-11-29
Map saving improvements

Bod9001:
Map saver will now automatically detect Cross object references and save them, 
maps saver now works with lists of cross object references and value types ( classes are a pain to compare )
Map saver now works in the editor when not playing
removes unnecessary logs from hub status

### DIFF
--- a/UnityProject/Assets/Scripts/Core/SecureStuff/AllowedReflection.cs
+++ b/UnityProject/Assets/Scripts/Core/SecureStuff/AllowedReflection.cs
@@ -279,12 +279,6 @@ namespace SecureStuff
 		public static Dictionary<Type, List<MethodsAndAttributee<T>>> GetFunctionsWithAttribute<T>()
 			where T : BaseAttribute
 		{
-			if (typeof(T) == typeof(ObsoleteAttribute))
-			{
-				Loggy.LogError("hey no, no obsolete stuff");
-				return null;
-			}
-
 			var result = new Dictionary<Type, List<MethodsAndAttributee<T>>>();
 
 			var allComponentTypes = AppDomain.CurrentDomain.GetAssemblies()
--- a/UnityProject/Assets/Scripts/Core/SecureStuff/AllowedReflection.cs
+++ b/UnityProject/Assets/Scripts/Core/SecureStuff/AllowedReflection.cs
@@ -6,10 +6,13 @@ using System.Reflection;
 using Logs;
 using NUnit.Framework;
 using UnityEngine;
+using UnityEngine.Events;
 
 namespace SecureStuff
 {
-	public class BaseAttribute : Attribute { }
+	public class BaseAttribute : Attribute
+	{
+	}
 
 	public struct MethodsAndAttributee<T> where T : BaseAttribute
 	{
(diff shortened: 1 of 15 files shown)


UnityProject/Assets/Scripts/Core/SecureStuff/AllowedReflection.cs
```diff
@@ -146,15 +276,16 @@ public static T CreateInstance<T>() where T : IAllowedReflection, new()
 		}
 
 
-		public static Dictionary<Type,List<MethodsAndAttributee<T>>> GetFunctionsWithAttribute<T>() where T : BaseAttribute
+		public static Dictionary<Type, List<MethodsAndAttributee<T>>> GetFunctionsWithAttribute<T>()
+			where T : BaseAttribute
 		{
 			if (typeof(T) == typeof(ObsoleteAttribute))
 			{
 				Loggy.LogError("hey no, no obsolete stuff");
```
MaxIsJoe: This error is unclear. You should report what exactly is having this issue.

UnityProject/Assets/Scripts/Core/SecureStuff/SecureMapsSaver.cs
```diff
@@ -138,123 +145,179 @@ public static class SecureMapsSaver
...
+									fieldData.Name = Prefix + Field.Name + "#" + i;
+									fieldData.Data = modified[i]?.ToString();
+									FieldDatas.Add(fieldData); //add data
+								}
+							}
+							else if (i < original.Count)
+							{
+								FieldData fieldData = new FieldData();
+								fieldData.Name = Prefix + Field.Name + "#" + i + "#" + "Removed";
+								fieldData.Data = "#removed#";
```
MaxIsJoe, line 297: What does removed entail? Is it just ghost data to tell mappers that a base component is no longer on this prefab?

UnityProject/Assets/Scripts/Core/SecureStuff/SecureMapsSaver.cs
```diff
@@ -269,5 +332,70 @@ public static class SecureMapsSaver
 				Loggy.LogError(e.ToString());
 			}
 		}
+
+		private static List<object> IEnumeratorToList(IEnumerator IEnumerable)
+		{
+			var List = new List<object>();
+			try
+			{
+				bool loop = true;
```
MaxIsJoe: You're not using this. Also, an exit strategy would be nice in case the loop gets stuck.

UnityProject/Assets/Scripts/Mapping/EventRouter.cs
```diff
@@ -0,0 +1,29 @@
...
+
+public class EventRouter : MonoBehaviour
+{
+	//TODO Specifying data sometime
+	//why Singular, gets complicated if you add multiple since you have to do a Equality of the elements of the list
+	public EventConnection Connection1 = new EventConnection();
+	public EventConnection Connection2 = new EventConnection();
+	public EventConnection Connection3 = new EventConnection();
+	public EventConnection Connection4 = new EventConnection();
+	public EventConnection Connection5 = new EventConnection();
```
MaxIsJoe, line 17: Why not use a list

MaxIsJoe (changes requested):

UnityProject/Assets/Scripts/Core/SecureStuff/SecureMapsSaver.cs
Bod9001: yeah, if a component has been removed from an item

UnityProject/Assets/Scripts/Core/SecureStuff/SecureMapsSaver.cs
Bod9001: theoretically it should never loop infinitely with the Enumerator , since it returns false when it Reaches the end

UnityProject/Assets/Scripts/Core/SecureStuff/SecureMapsSaver.cs
MaxIsJoe: ok 👍

UnityProject/Assets/Scripts/Core/SecureStuff/SecureMapsSaver.cs
MaxIsJoe: Alrighty

UnityProject/Assets/Scripts/Mapping/EventRouter.cs
Bod9001: the map saver, Would have to be expanded to support class comparison ( is the class exactly the same or is it different ) 
and then be able to serialise the differences if it's only slightly different, basically big pain in but

UnityProject/Assets/Scripts/Core/SecureStuff/AllowedReflection.cs
Bod9001: aaaa unrelated technically, shall I add it anyway

UnityProject/Assets/Scripts/Mapping/EventRouter.cs
MaxIsJoe: ... this still doesn't explain why you cant have a list of connections to loop through to make the code cleaner and have the ability to expand to having more than just 5 connections.

Are each connection statitc to something that will never change? Then its better to use proper names rather than just numbers

Bod9001: so, with the list, 
it needs to be able to compare with the prefab yes? , how do you compare a class?

Bod9001: because all it has is the object and type, I was thinking may be supporting structs but that seems like a dodgy implementation

Bod9001: you'd have to write a custom comparer for each class as well  since its something not standard with classes

PR #10213, opened 2023-12-04
Adds a particle effect for jetpacks releasing gas while moving.

### LINKED ISSUE #8199: Particle Effect For Jetpacks
#8191 Makes jetpacks work but they have no particle effect yet.

So an effect like the fire extinguisher needs to be made.

MaxIsJoe:
closes #8199

https://github.com/unitystation/unitystation/assets/34368774/d2f13f17-d776-43f0-b8ba-5dc7045e2319

CL: [New] Jetpacks now have particle effects when releasing gas and moving at the same time.


UnityProject/Assets/Scripts/Items/Others/Jetpack.cs
```diff
@@ -67,6 +67,10 @@ private void PushUpdate()
 			if (player.PlayerSync.IsPressedServer)//Is movement pressed?
 			{
 				PushPlayerInFacedDirection(player, gasContainer, gasReleaseOnUse);
+				if (releaseVisual != null)
+				{
+					Spawn.ServerPrefab(releaseVisual, gameObject.AssumedWorldPosServer());
```
corp-0, line 72: is this performant enough?

MaxIsJoe: It spawns a single splatter every 0.5 seconds if there is any movement on the server, and splatters handle their removal client side.

This shouldn't be a concern unless you're spawning hundreds of them, which would cause GC to pile up then.

corp-0 (approved):

PR #10215, opened 2023-12-05
Scriptable objects saving

Bod9001:
Add support for referencing different SOs and preffabs


MaxIsJoe:
this PR is uhh, a bit too large for review.
what is it trying to do?

Bod9001:
makes it so you can set custom SOs on objects on map saver as well as adding support for specifying different prefabs as well

PR #10218, opened 2023-12-06
Makes it so you can Toggle run While crawling

Bod9001:
### Changelog:
CL: [Improvement] Makes it so you can Toggle run While crawling


MaxIsJoe (approved):

corp-0:
will this make you crawl faster?

Bod9001:
nah, it caches the requested state then sets it to that when uncrawling

PR #10219, opened 2023-12-07
Fixes gas containers exploading before destruction

### LINKED ISSUE #10198: [BUG] Crates explode when destroyed via damage but don't despawn until destroyed a second time

### Game Version

WIP 3921 but issue has been around for a while

### I was

Server

### OS

Linux

### What happened?

When a crate is damaged to the point of destruction, it gives a 'crate explodes' hud message and plays a sound effect. The crate is then returned to an apparently undamaged state, despawning without special effects the second time it is destroyed (not sure whether the gas contents are actually released the first or second time it's destroyed).

Crate despawn on initial explosion/destruction

Spawn a crate
Shoot it, examining its damage state along the way

### Additional context

_No response_

MaxIsJoe:
fixes  #10198

CL: [New] Mappers and designers now have the ability to hook up functions to when an object/item get destroyed.
CL: [Fix] Fixed containers exploding before they are actually destroyed by explicitly asking if gas containers are meant to release their content due to excessive damage or when they are getting destroyed.

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/GasContainer.cs
+++ b/UnityProject/Assets/Scripts/Objects/GasContainer.cs
@@ -68,8 +68,7 @@ namespace Objects.Atmospherics
 		[Tooltip("If true : Cargo will accept gases found within this container and can be sold.")]
 		public bool CargoSealApproved = false;
 
-		[FormerlySerializedAs("exploadOnTooMuchDamage")] [SerializeField]
-		private bool explodeOnTooMuchDamage = true;
+		[SerializeField] private bool explodeOnTooMuchDamage = true;
 
 		#region Lifecycle
 
--- a/UnityProject/Assets/Scripts/Objects/GasContainer.cs
+++ b/UnityProject/Assets/Scripts/Objects/GasContainer.cs
@@ -68,7 +68,8 @@ namespace Objects.Atmospherics
 		[Tooltip("If true : Cargo will accept gases found within this container and can be sold.")]
 		public bool CargoSealApproved = false;
 
-		[SerializeField] private bool exploadOnTooMuchDamage = true;
+		[FormerlySerializedAs("exploadOnTooMuchDamage")] [SerializeField]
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/Objects/GasContainer.cs
```diff
@@ -66,6 +68,8 @@ public GasMix GasMix
 		[Tooltip("If true : Cargo will accept gases found within this container and can be sold.")]
 		public bool CargoSealApproved = false;
 
+		[SerializeField] private bool exploadOnTooMuchDamage = true;
```
corp-0:
```suggestion
		[SerializeField] private bool explodeOnTooMuchDamage = true;
```

corp-0: only change I see, looks good.

corp-0 (changes requested):

UnityProject/Assets/Scripts/Objects/GasContainer.cs
MaxIsJoe: done 👍

UnityProject/Assets/Scripts/Objects/GasContainer.cs
```diff
@@ -68,7 +68,8 @@ public GasMix GasMix
 		[Tooltip("If true : Cargo will accept gases found within this container and can be sold.")]
 		public bool CargoSealApproved = false;
 
-		[SerializeField] private bool exploadOnTooMuchDamage = true;
+		[FormerlySerializedAs("exploadOnTooMuchDamage")] [SerializeField]
```
corp-0: remove the formerlyserialized as. Re do your modification on the closet prefab instead.

MaxIsJoe: annnd done 👍

corp-0 (approved):

PR #10220, opened 2023-12-07
Adds specific map loading on game startup

### LINKED ISSUE #10112: [Feature] Allow servers to decide the first station to load when starting :earth_asia:

### What happened?

it would be appreciated if admin could choose
- The first loaded station at server startup.
- Change the stations available in rotation in-game.

MaxIsJoe:
closes #10112

CL: [New] servers can force a specific map on start now via the `-map` command line argument.

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/GameManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/GameManager.cs
@@ -218,6 +218,25 @@ public partial class GameManager : MonoBehaviour, IInitialise
 			QuickLoad = editorLoadPref;
 			Loggy.Log($"Currently using editor pref for quick-load checkup. Current value is {editorLoadPref}. To change this, please head to tools -> Enable QuickLoad.");
 #endif
+			var args = Environment.GetCommandLineArgs();
+			if (args.Length > 1)
+			{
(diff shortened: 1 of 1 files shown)


corp-0:
I have the impression that the linked issue would like a way to manage this in-game

MaxIsJoe:
1. The admin probably shouldn't have the ability to do stuff like this.
2. This is rather pointless because the admin has no clue when the server is ever going to restart, and servers on unitystation can sometimes last days without a single reboot.. unlike BYOND where every round end is a complete server reboot.
3. Dae's requested functionality is better covered in command line arguments, since most hosts will be using it anyways; so might as well leave it there.
4. The admin UI is too bloated and unorganized already, I don't want to add in very niche things that not a lot of people will use except testers.

corp-0:
I think the best compromise here is doing it in a config file, so eventually we can add a UI for it. It has all the benefits of being a command line argument + the ability to deserialize it and visualize it in game one day.

Fr-Dae:
> I think the best compromise here is doing it in a config file, so eventually we can add a UI for it. It has all the benefits of being a command line argument + the ability to deserialize it and visualize it in game one day.

is this a feasible solution?

PR #10226, opened 2023-12-11
Fixes for character creator

### LINKED ISSUE #9988: Character names should be allowed to have capital letters in the middle of their names
Related to #7878

image

MaxIsJoe:
fixes #9988

CL: [Fix] Fixed an issue where players could roll for the ability to play as a cow.
CL: [Fix] The strict filter for names have been loosened to allow for names that have capital letters in the middle of them.

### DIFF
--- a/UnityProject/Assets/Scripts/Clothing/RaceSOSingleton.cs
+++ b/UnityProject/Assets/Scripts/Clothing/RaceSOSingleton.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Linq;
 using UnityEngine;
 using ScriptableObjects;
 
@@ -25,11 +26,6 @@ public class RaceSOSingleton : SingletonScriptableObject<RaceSOSingleton>
 
 	public static List<PlayerHealthData> GetPlayerRaces()
 	{
-		var list = new List<PlayerHealthData>();
-		foreach (var race in Instance.Races)
-		{
-			if(race.Base.CanBePlayerChosen) list.Add(race);
-		}
-		return list;
+		return Instance.Races.Where(specie => specie.Base.CanBePlayerChosen).ToList();
 	}
 }
--- a/UnityProject/Assets/Scripts/Clothing/RaceSOSingleton.cs
+++ b/UnityProject/Assets/Scripts/Clothing/RaceSOSingleton.cs
@@ -24,7 +24,7 @@ public class RaceSOSingleton : SingletonScriptableObject<RaceSOSingleton>
 		return false;
 	}
 
-	public static List<PlayerHealthData> GetPlayerRaces()
+	public static List<PlayerHealthData> GetPlayerSpecies()
 	{
 		return Instance.Races.Where(specie => specie.Base.CanBePlayerChosen).ToList();
 	}
--- a/UnityProject/Assets/Scripts/Clothing/RaceSOSingleton.cs
+++ b/UnityProject/Assets/Scripts/Clothing/RaceSOSingleton.cs
@@ -22,4 +22,14 @@ public class RaceSOSingleton : SingletonScriptableObject<RaceSOSingleton>
 		race = null;
 		return false;
 	}
+
+	public static List<PlayerHealthData> GetPlayerRaces()
+	{
+		var list = new List<PlayerHealthData>();
+		foreach (var race in Instance.Races)
+		{
+			if(race.Base.CanBePlayerChosen) list.Add(race);
+		}
+		return list;
+	}
 }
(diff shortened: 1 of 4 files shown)


corp-0:
I would remove the restriction completely, honestly.

MaxIsJoe:
Lucky for you, It is removed!

UnityProject/Assets/Scripts/Clothing/RaceSOSingleton.cs
```diff
@@ -22,4 +22,14 @@ public static bool TryGetRaceByName(string raceName, out PlayerHealthData race)
 		race = null;
 		return false;
 	}
+
+	public static List<PlayerHealthData> GetPlayerRaces()
+	{
+		var list = new List<PlayerHealthData>();
```
corp-0: embrace LINQ

```suggestion
		return Instance.Races.Where(race => race.Base.CanBePlayerChosen).ToList();
```

as a second suggestion, I would replace any future mention of "race" with "specie" since it is more semantically correct.

UnityProject/Assets/Scripts/UI/Systems/Lobby/CharacterCustomization.cs
```diff
@@ -130,7 +131,7 @@ private void GrabAllSpecies()
 				Loggy.LogError("UNABLE TO GRAB ALL SPECIES!! CHARACTER CREATION SCREEN IS SURELY GOING TO BE BROKEN!!!");
 				return;
 			}
-			allSpecies = RaceSOSingleton.Instance.Races.Where(alien => alien.Base.CanBePlayerChosen).ToList();
+			allSpecies = RaceSOSingleton.GetPlayerRaces();
```
corp-0: wait, you did change this perfectly fine LINQ with that new method, why?

UnityProject/Assets/Scripts/Util/Independent/FluentRichText/StringExtensions.cs
```diff
@@ -38,5 +38,17 @@ public static class StringExtensions
 		public static string Superscript(this string text) => new SuperscriptStrategy().ApplyStyle(text);
 		public static string VerticalOffset(this string text, string amount) => new VerticalOffsetStrategy(amount).ApplyStyle(text);
 		public static string Width(this string text, string amount) => new WidthStrategy(amount).ApplyStyle(text);
+
+		public static string ToUpperFirstLetter(this string source)
```
corp-0: there is an extension method that does this already, if I'm not mistaken it is ``.Capitalize()``

UnityProject/Assets/Scripts/UI/Systems/Lobby/CharacterCustomization.cs
MaxIsJoe: We needed a general function to grab all player species without rewriting the same code over and over again.

corp-0: fair, but I still consider the LINQ version to be nicer.

UnityProject/Assets/Scripts/Util/Independent/FluentRichText/StringExtensions.cs
MaxIsJoe: fixed

UnityProject/Assets/Scripts/Clothing/RaceSOSingleton.cs
MaxIsJoe: done 👍

UnityProject/Assets/Scripts/Clothing/RaceSOSingleton.cs
```diff
@@ -22,4 +23,9 @@ public static bool TryGetRaceByName(string raceName, out PlayerHealthData race)
 		race = null;
 		return false;
 	}
+
+	public static List<PlayerHealthData> GetPlayerRaces()
```
corp-0:
```suggestion
	public static List<PlayerHealthData> GetPlayerSpecies()
```

MaxIsJoe:
Also, regarding the LINQ comment:

It's fine if we use LINQ with client-sided stuff because we don't really care about the extra GC that clients can easily deal with, but I generally prefer if we avoid LINQ as much as possible so we don't accidentally slip into a habit of overusing it, or giving newcomers the false sense that they can use LINQ freely.

https://unitystation.github.io/unitystation/contribution-guides/Development-Standards-Guide/#2-performance-gotchas

corp-0:
> Also, regarding the LINQ comment:
> 
> It's fine if we use LINQ with client-sided stuff because we don't really care about the extra GC that clients can easily deal with, but I generally prefer if we avoid LINQ as much as possible so we don't accidentally slip into a habit of overusing it, or giving newcomers the false sense that they can use LINQ freely.
> 
> https://unitystation.github.io/unitystation/contribution-guides/Development-Standards-Guide/#2-performance-gotchas

in your case, it was going to generate the same amount of GC because you also did a discardable list

MaxIsJoe:
better be safe than sorry ¯\_(ツ)_/¯

corp-0 (approved):

PR #10227, opened 2023-12-11
Adds bumping logic for tiles

### LINKED ISSUE #6416: Electrified grilles are only partially implemented
Electrified grilles do not trigger when bumped - electrocution can only occur if you click on the tile, which is not ideal.

One good solution would be to create a new interface, say, `IBumpable` that objects and tiles could implement, to handle getting bumped into. This might also be apt to simplify pushpull behaviour.

Also related: #6415, which makes it considerably harder to set up electrified grilles.

Maps also do not currently exploit electrifiable grilles, which is a significant map design element in /tg/ (though this probably shouldn't be done until powering grilles is simpler, via solving #6415).

MaxIsJoe:
closes #6416

This PR adds the ability for electrified grills to be bumpable, this does not add the functionality to any current maps as I have no clue how to set them up in the editor.
Enjoy creating traps with this

CL: [New] electrified grills will now actually work when you bump into them, instead of just relying on direct input.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Input System/InteractionV2/TileInteraction/ElectrifiedGrilleInteraction.cs
+++ b/UnityProject/Assets/Scripts/Core/Input System/InteractionV2/TileInteraction/ElectrifiedGrilleInteraction.cs
@@ -1,7 +1,9 @@
 using HealthV2;
+using Objects;
 using TileManagement;
 using UnityEngine;
 using Systems.Electricity;
+using Systems.Explosions;
 
 namespace Tiles
 {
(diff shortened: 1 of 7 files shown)


UnityProject/Assets/Scripts/Core/Input System/InteractionV2/TileInteraction/ElectrifiedGrilleInteraction.cs
```diff
@@ -11,7 +13,7 @@ namespace Tiles
 	/// </summary>
 	[CreateAssetMenu(fileName = "ElectrifiedGrilleInteraction",
 		menuName = "Interaction/TileInteraction/ElectrifiedGrilleInteraction")]
-	public class ElectrifiedGrilleInteraction : TileInteraction
+	public class ElectrifiedGrilleInteraction : TileInteraction, IBumpableObject
```
Bod9001, line 16: Issue if you make a offshoot cable, and don't have any resistance at the end of it it won't be electrified

MaxIsJoe: I have no idea how to fix this, so as of right now you just have to rely on wires that are already connected to other stuff

Bod9001 (approved):

PR #10236, opened 2023-12-17
Client side Gizmo system

Bod9001:
### Changelog:
CL: [New] stuff can be now rendered over FOV by setting it to UI layer
CL: [New] added new gizmo system


UnityProject/Assets/Scripts/UI/Systems/AdminTabButtons.cs
```diff
@@ -79,12 +79,12 @@ public void BtnOpenTileVV()
 
 		private void DisableAllGUI()
 		{
-			adminTools.gameObject.SetActive(false);
-			devSpawner.gameObject.SetActive(false);
-			devCloner.gameObject.SetActive(false);
-			devDestroyer.gameObject.SetActive(false);
-			devSelectTile.gameObject.SetActive(false);
-			tileChanger.Close();
+			//adminTools.gameObject.SetActive(false);
```
corp-0: why are these commented?

Bod9001: is, so you can have multiple mapping tools open at the same time

corp-0: then your intention is to change the behaviour and you should change the code to represent that. Just commenting out this leaves a functionality you don't want anymore lingering in the codebase without explanation

Bod9001: ok

PR #10239, opened 2023-12-21
Adds in the functionality for readable books

MaxIsJoe:
This adds the functionality for books to exist and be setup. However, it does not add the ability to craft/print books yet.

Theoretically, mappers can use this already to add books around stations and away sites since they can edit the page data directly, or they can add a list of page prefabs to its storage.


Bod9001:
Tests

Bod9001:
will merge after playtest

Bod9001 (approved):

PR #10240, opened 2023-12-22
Jetpack fixes and improvements

MaxIsJoe:
CL: [Improvement] The jetpack trails have been updated to be more visually appealing.
CL: [Improvement] Jetpacks can now be turned on and off by regular hand activation.
CL: [Improvement] Jetpack movement can now be steered by changing where the player is looking via mouse clicks.
CL: [Improvement] Jetpacks have their movement build gradually over time now when the player holds down a movement key. Similarly, that momentum is lost overtime when they're no longer holding down any movement keys.
CL: [Fix] Fixed an issue that caused jetpacks to be constantly applying their effects when they're no longer in a player's inventory.
CL: [Fix] Fixed an issue where players can still be influenced by the jetpack's logic even when they're no longer in their inventory.

### DIFF
--- a/UnityProject/Assets/Scripts/Player/PlayerParticle.cs
+++ b/UnityProject/Assets/Scripts/Player/PlayerParticle.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Linq;
 using Logs;
 using Mirror;
 using UnityEngine;
@@ -8,7 +9,7 @@ namespace Player
 	public class PlayerParticle : NetworkBehaviour
 	{
 		[SerializeField] private Transform particleHolder;
-		[SyncVar(hook = nameof(OnActiveIdListChange))] private SyncList<string> activeIDs = new SyncList<string>();
+		private SyncList<string> activeIDs = new SyncList<string>();
 		private List<PlayerParticleObject> particles = new List<PlayerParticleObject>();
 
 		private void Awake()
@@ -27,7 +28,11 @@ namespace Player
 		public override void OnStartClient()
 		{
 			base.OnStartClient();
-			OnActiveIdListChange(activeIDs, activeIDs);
+			activeIDs.Callback += OnActiveIdListChange;
+			for (int index = 0; index < activeIDs.Count; index++)
+			{
+				OnActiveIdListChange(SyncList<string>.Operation.OP_ADD, index, "", activeIDs[index]);
+			}
 		}
 
 		[Server]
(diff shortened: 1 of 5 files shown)


UnityProject/Assets/Scripts/Player/PlayerParticle.cs
```diff
@@ -0,0 +1,44 @@
...
+				PlayerParticleObject particleObject = new PlayerParticleObject()
+				{
+					Reference = obj,
+				};
+				particles.Add(particleObject);
+			}
+		}
+
+		[ClientRpc]
+		public void ToggleParticle(string id, bool state)
```
Bod9001: this should be a syncvar? because it's toggling so therefore it has a trackable state?

MaxIsJoe: done

UnityProject/Assets/Scripts/Player/PlayerParticle.cs
```diff
@@ -8,9 +8,10 @@ namespace Player
 	public class PlayerParticle : NetworkBehaviour
 	{
 		[SerializeField] private Transform particleHolder;
+		[SyncVar(hook = nameof(OnActiveIdListChange))] private List<string> activeIDs = new List<string>();
```
Bod9001: remember with synchronised lists you have to set the network identity dirty when adding and removing from them

Bod9001: I think it needs to be also synclist

MaxIsJoe: changed it

UnityProject/Assets/Scripts/Player/PlayerParticle.cs
```diff
@@ -47,16 +47,18 @@ public void ServerToggleParticle(string id, bool state)
...
 
-		public void OnActiveIdListChange(List<string> oldState, List<string> newState)
+		public void OnActiveIdListChange(SyncList<string> oldState, SyncList<string> newState)
 		{
 			if (particles == null || particles.Count == 0) return;
 			foreach (var particle in particles)
 			{
 				particle.Reference.SetActive(newState.Contains(particle.ID));
 			}
+			netIdentity.isDirty = true;
```
Bod9001, line 99: you shouldn't need to do this?

Bod9001: the update state for synclists is a bit more complicated than this

Bod9001: https://mirror-networking.gitbook.io/docs/manual/guides/synchronization/synclists

MaxIsJoe: > you shouldn't need to do this?

image

Bod9001: well Need to work out how to modify Preprocessor , since it does some magic stuff to link up the synclist

MaxIsJoe: Converted them to use synclist operations

Bod9001 (approved):

PR #10264, opened 2024-01-17
nerfs footprints making too much mess

MaxIsJoe:
CL: [Improvement] Footsteps will now make less mess when walking around.


corp-0:
Who complained about this?

MaxIsJoe:
I complained about this. It's not fun seeing the station get smeared in the first 2 minutes of the round.
This change makes footsteps last shorter, instead of across an entire hallway now.

Bod9001:
idk,  like if you have a janitor is cleaned up relatively quickly

AtnerNT:
I agree with max here. a tiny chemical spill will leave you trailing footprints for like 20 tiles. And it means that even with one spill, the entire department gets covered. And while its true a janitor can clean it up, how often are janitors going into the depths of departments to clean, not often. I like footsteps, but how long they trail for currently is a bit absurd.

MaxIsJoe:
closed after discussions on discord.
opting for another approach.

PR #8386, opened 2022-02-18
Adds in the NoFruit

MaxIsJoe:
Introducing : NoFruit!

For professional chefs that have a habit of gambling or a rocky assistant in the kitchen, NoFruit is the latest invention from Waffelz Co. that turns your kitchen into a casino where luck decides the ingredient for your next dish! Pull the tip, Grab a blunt object and HIT that NoFruit with harmful intent just like you do when you see your wife sleeping with the station clown.

NoFruit may taste like nothing, but with a good wack you can create the best mouth watering dish a space station could ever hope for from a wide selection of fruits that materializes from the recently discovered voids that is rich of delicious organic fruits. Get one now for only 19.99$ per packet! (Additional shipping fees apply, Product may be unavailable on Sol occupied systems).

CL: [New] Added the NoFruit

WARNING : Wafflez Co. is not responsible for  any facehugging entity that may or may not jump between the void and this world. Use that bat to good use when you see it in front of you before it couples with your head.

PR #10267, opened 2024-01-27
Faiths, Part 5 - Filth and Ants

MaxIsJoe:
This new part of Faiths introduces some radical changes to how faiths will be handled on unitystation from now on, and some lore tidbits here and there.
Faiths started off as a simple system that was designed to give a gameplay reason for the chaplain's existence, as their role was quite bare-bones due to the lack of cultist game modes and the lack of paranormal activities upon NT's stations. Now, the faith system is stepping into the shoes of a "chaos mediator" that encourages players to interact with more mundane aspects of the game that previously were ignored due to a lack of consequences and general interactivity.

Firstly, let's talk about the major changes.

# All FAITHS, ALL THE TIME.

image

Faiths are no longer bound to the existence of a chaplain or any faith members. They will continue to operate 24/7 even when there are no active faith members inside a faith, and no faith will no longer be disbanded even when there are 0 members and leaders within the faith.

# FEAR THE GERMS

image

What originally sparked my interest to come back to faiths and continue working on it, was a discussion discord related to #10264. Gilles and Bod were right, nerfing footprints was a terrible solution to the "players don't want to clean up their mess" problem. What are we, /TG/?

We don't want to trim or take an axe to any feature in unitystation simply because we don't have enough players engaging with it, even if that feature is incredibly flawed. As a solution, I've decided to add a new property to faiths that should make players take station hygiene more seriously starting from now.

The germaphobe faith property always keeps track of the station's filth score, and starts inviting the pesky wild life at specific score values.
image

At 150 filth objects, space ants will start visiting the station to feed on all the garbage and reagents left spilled on the ground.
Space ants are generally harmless at first, but will gradually become an active nuisance (and even a health hazard!) the more they feed on garbage and nearby food.

Ants also come in three variants, Space Ants, Fire Ants and Killer ants.

Fire ants evolve from space ants that spread over to nearby garbage, there is a low chance for them to evolve; but antagonists can spend their time farming/breeding ants by collecting the station's garbage and offering them as tribute.

Fire ants expose a lot of heat to nearby tiles, which is perfect if you want to trigger a Gibbtonite/plasma flood or burn unsuspecting victims.

Killer ants will never be evolved naturally, and are admin spawned only for the time being. Why? Because those small killers crawl inside your body and eat all your organs within seconds.

image

At 300 points, mold will start growing on bread; and will cause the bread to mutate into green monsters attempting to eat crewmembers near the kitchen.

While ants only existed to be annoying, bread monsters are there to hammer the point that the station's hygine is so bad; that germs are starting to mutate food inside the station now!

image

And lastly, at 550 points. Spores.

Station so stinky, it is causing a level 2 biohazard before there is even a blob around to spawn them.

#### TLDR for germaphobia:

- Ants will spawn the more the station has garbage and spilled reagents around.
- Ants will feed on garbage, food and reagents. They will evolve to fireants at some point if they continue spreading.
- If the station's filth score is too high, bread will mutate into bread monsters and spores will start growing out of nowhere.

With all of this, I hope cleaning up the station will become an engaging factor for all crew members; as they try and avoid having their feet bit by ants or mold grow on their food.
These changes also opens a couple new opportunities for antags to mess around with the station without murderboning everyone or relying on explosions.

# FEED ME, MISTER WAFFLE.

Unity_1rucHcvjwb

A new item has been introduced to help with low pop on unitystation, this item also taps into some lore implications to one of unitystation's background factions.
What, you thought #8386 description was a joke?

This new item will teleport food donations from all across the universe thanks to Waffle Co.'s R&D department and Sol's species rights program. No living creature should be denied access to food, so as a result; every 9 minutes a random piece of small food will be teleported via the Food Donations pad for consumption.

Do not attempt to break the glass.

# Notes

As you may have noticed, all Faith PRs reuse or expand upon already existing content in unitystation. I do not want to bloat unitystation with more features when we could instead use our already existing catalog of content, refine it and make it all connected to each other.

With the new changes introduced to Faiths and introduction to germphobia, I want this trend to continue. I want features to have more depth and meaning like the ones in this PR, as it makes the game feel more polished.

Some aspects of faiths will need to be rebalanced now that the member count requirement is removed, but that will need some testing first.

# Changelog:

CL: [New] Added the Quantum Food Donations Pad.
CL: [New] Added Ants.
CL: [New] Added a new property for faiths, Germaphobia.
CL: [Improvement] Some strings related to faiths have been improved.
CL: [Improvement] Faiths can no longer be disbanded when there are no members in them.
CL: [Improvement] Faiths are always active 24/7 now, even when there are no active members. (This still requires a lower than -75 or greater than 75 points for random events to trigger)
CL: [Fix] Fixed some NREs related to object scanning using a component tracker.


UnityProject/Assets/Scripts/Systems/Faith/FaithData.cs
```diff
@@ -38,6 +39,7 @@ public void RemoveAllMembers()
 
 		public void SetupFaith()
 		{
+			Loggy.Log("[FaithData/SetupFaith] Setting up faith data for " + Faith.FaithName);
```
Bod9001, line 42: log

MaxIsJoe: It's left there on purpose to keep track of what the faith system is doing

PR #10270, opened 2024-01-29
Makes bananium slippy

MaxIsJoe:
CL: [Improvement] bananium ore are now slippy.


Bod9001:
can you make butter slippery to

PR #10276, opened 2024-02-03
Mutes some spammy server logs that the server shouldn't keep track of

MaxIsJoe:


UnityProject/Assets/Scripts/Player/MovementV2/MatrixCash.cs
```diff
@@ -95,8 +95,10 @@ public void ResetNewPosition(Vector3 Centrepoint, RegisterTile Inon)
 			Positions[i] = null;
 		}
 
-		Positions[6] = Inon.Matrix.MatrixInfo;
-
+		if (Positions != null && Inon.OrNull()?.Matrix.OrNull()?.MatrixInfo is not null)
```
Bod9001, line 99: tthis should be catch since it's to do with movement?

MaxIsJoe: Updated to include a log

PR #10281, opened 2024-02-05
rebalances for filth and ant hills

MaxIsJoe:
CL: [Improvement] Ant bite inflictions now only occur when they're in their swarm state.
CL: [Improvement] Ant bites will only do 1 damage instead of a random set of damage between 2 and 4 if they're not in a swarm state.
CL: [Improvement] Ant hills will only spawn once around the station per faith event cycle (9 minutes) instead of in groups of two.
CL: [Improvement] Ant AI now runs every 60 seconds instead of 30 seconds.
CL: [Improvement] Ants filth lookup will only target one piece of filth to consume just like how they only target one piece of food, slowing down their evolution process.
CL: [Improvement] Ants will only spread in their swarm state now.
CL: [Improvement] Ants will now display hover tooltips, stating their current state and how to get rid of them.
CL: [Improvement] FilthGen has been rebalanced to generate more dirt than reagents.
CL: [Improvement] FilthGen now has an experimental formula for balancing filth.


Bod9001:
Can you make it so the damage gets split, since people are having issues if they had multiple limbs to damage a stacking

MaxIsJoe:
> Can you make it so the damage gets split, since people are having issues if they had multiple limbs to damage a stacking

Did you one better; I added a function to only target one body part.

UnityProject/Assets/Scripts/Systems/FilthGenerator/FilthGenerator.cs
```diff
@@ -72,14 +72,20 @@ public void RunFilthGenerator()
 				}
 			}
 
-			int numberOfTiles = (int) ((EmptyTiled.Count / 100f) * filthDensityPercentage);
+			int numberOfPlayers = PlayerList.Instance.AllPlayers.Count;
+			float densityPercentagePerPlayer = 1f / 2f; // 1% per 2 players
```
corp-0: why make a literal division here instead of just assigning the result of that known operation?

MaxIsJoe: thought it was clearer this way tbh

corp-0: well 1 / 2 is 0.5

I'd say this is clearer.
```cs
float densityPercentagePerPlayer = 0.5
```

corp-0: also the comment is a lie

MaxIsJoe: technically it isn't a lie, because after the calculation it *is* 1% per two players since (int) will round up

however, I'll remove it just incase someone gets confused about it

UnityProject/Assets/Scripts/Systems/FilthGenerator/FilthGenerator.cs
```diff
@@ -7,6 +7,7 @@
 using Tiles;
 using UnityEngine;
 using UnityEngine.Tilemaps;
+using Debug = UnityEngine.Debug;
```
corp-0: nani?

MaxIsJoe: whoops, cleaned those up

corp-0 (approved):

PR #10287, opened 2024-02-14
Species validaiton

MaxIsJoe:
CL: [Fix] Fixed an issue that would cause players to randomly create slimes and Arials in the main character creator page.
CL: [Improvement] Added an extra layer of species validation to avoid instances of players creating character sheets with species they're not supposed to play with.

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Lobby/CharacterSheet.cs
+++ b/UnityProject/Assets/Scripts/Systems/Lobby/CharacterSheet.cs
@@ -110,11 +110,16 @@ public class CharacterSheet : ICloneable
 
 	public void ValidateSpeciesCanBePlayerChosen()
 	{
+		var speciesToPick = RaceSOSingleton.GetPlayerSpecies();
 		if (GetRaceSo(true) == null)
 		{
-			Species = RaceSOSingleton.Instance.Races.Where(x => x.Base.CanBePlayerChosen).PickRandom().name;
+			Species = speciesToPick.PickRandom().name;
 		}
-
+		if (speciesToPick.Any(x => x.name == Species) == false)
+		{
+			Species = speciesToPick.PickRandom().name;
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/UI/Systems/Lobby/CharacterCustomization.cs
```diff
@@ -96,7 +96,7 @@ private BodyTypeName ThisBodyType
 		private List<ExternalCustomisation> ExternalCustomisationStorage = new();
 		private Dictionary<string, BodyPartCustomisationBase> OpenBodyCustomisation = new();
 
-		private List<PlayerHealthData> allSpecies;
+		public List<PlayerHealthData> AllSpecies;
```
Bod9001, line 99: need to add formally serialised as

Bod9001: scrap that

MaxIsJoe: Doesn't need to because it was never set via the inspector.

UnityProject/Assets/Scripts/Systems/Lobby/CharacterSheet.cs
```diff
@@ -110,11 +110,16 @@ private void ValidateAiName()
 
 	public void ValidateSpeciesCanBePlayerChosen()
```
Bod9001, line 111: this needs to stay the same because, otherwise then slimes break

Bod9001: and aliens

MaxIsJoe: How does it break aliens and slimes? as far as I'm aware, this only affects the character creator, and isn't called elsewhere.

PR #10300, opened 2024-03-16
refactor: minor clean-up of models, turning fields into properties.

corp-0:
### Purpose
The model's fields will now be properties and their names will respect C# code style.


MaxIsJoe (approved):
Looks good to me.

PR #10301, opened 2024-03-16
Discord Rich Presence

MaxIsJoe:
CL: [New] Unitystation now has discord rich presence. Show off which server and map you're playing on, and for how long as well.


corp-0:
@Bod9001 how can we make scan code happy here? It would be cool to merge this for this weeks playtest

Bod9001:
oh sorry for taking so long

Bod9001:
Sandbox violation: Method has illegal MethodAttributes: Discord.Discord.DiscordCreate(unsigned int32, Discord.Discord/FFICreateParams&, native int&) Returns Discord.Result  in Assembly Assets 
so, It will need to be wrapped in a secure stuff handler

Bod9001:
looks  like you'll have to move all the library stuff into secure stuff and then poke a few holes ffor access

MaxIsJoe:
Closing this because It's frustrating to get it working, will move this over to stationhub when we actually are going to push an update for it.

PR #10303, opened 2024-03-19
readded Fast reload

Bod9001:


corp-0:
:c

MaxIsJoe:
why isn't this added as a package from the package manager?

Bod9001:
added in different PR

PR #10314, opened 2024-03-29
April fools funnies

Bod9001:
### Changelog:
CL: [New] Oddly the assistantes seem more noisy then normal


corp-0:
is there no way to add the new package as a dependency instead of directly adding the code to our codebase?

UnityProject/Assets/Scripts/Core/Input System/KeyboardInputManager.cs
```diff
@@ -236,9 +236,10 @@ private bool CheckComboEvent(KeyCombo keyCombo, KeyEventType keyEventType = KeyE
 		{ KeyAction.ChatLocal,		() => { ChatUI.Instance.OpenChatWindow(ChatChannel.Local); }},
 		{ KeyAction.ChatRadio,		() => { ChatUI.Instance.OpenChatWindow(ChatChannel.Common); }},
 		{ KeyAction.ChatOOC,		() => { ChatUI.Instance.OpenChatWindow(ChatChannel.OOC); }},
-		{ KeyAction.ToggleHelp,    () => { ChatUI.Instance.OnHelpButton(); }},
+		{ KeyAction.ToggleHelp,     () => { ChatUI.Instance.OnHelpButton(); }},
 		{ KeyAction.ToggleAHelp,    () => { ChatUI.Instance.OnAdminHelpButton(); }},
 		{ KeyAction.ToggleMHelp,    () => { ChatUI.Instance.OnMentorHelpButton(); }},
+		{ KeyAction.PushToTalk,     () => { VoiceChatManager.Instance.ClientPushToTalkPressed = !VoiceChatManager.Instance.ClientPushToTalkPressed; }},
```
corp-0, line 242: in a real push to talk, you have to hold the key to talk. This sounds more like a toggle.

Bod9001: yeah,  sadly the current key bind manager isn't really set up for that, 
though there is UI to show when you microphone is active

corp-0:
so I imagine this chat isn't spatial, right? It is all global

Bod9001:
it's spatial

Bod9001:
only thing that looks fiddly for code scan
```
Sandbox violation: Access to type not allowed: [UnityEngine.AudioModule]UnityEngine.Microphone in Assembly Adrenak.UniMic.Runtime
```
how should we handle microphone access with secure stuff?

man This is stupid, who thought this was a good idea to add to c# xD
```
Sandbox violation: Access to type not allowed: [mscorlib]System.Runtime.Serialization.Formatters.Binary.BinaryFormatter in Assembly Assets
BinaryFormatter classes  implement the IRemotingFormatter interface to support remote procedure calls (RPCs), 

"The BinaryFormatter type is dangerous and is not recommended for data processing. Applications should stop using BinaryFormatter as soon as possible, even if they believe the data they're processing to be trustworthy. BinaryFormatter is insecure and can't be made secure."
```
Guess I'll have to swap it out for mirrors custom reader writers,

MaxIsJoe:
we already let our selves access things like the clipboard, I don't see why we should start worrying about microphone access.
though in the future, it would be nice if we can tell players on stationhub that a fork has the ability to access stuff like your microphone

If we really want to make this extra secure, simply do the same thing we do for safeurl but for microphones, and explicitly only allow audio listening when it is allowed by the user.

Bod9001:
yeah, I'm just wondering what people's thought on microphone access

Issue #9980, opened 2023-07-20
[BUG] The Photo Copier does not work.


eurtod:
### There is no other issue reporting the same problem.

- [X] I have searched the existing issues

### Game Version

23071519

### I was

Client

### OS

Windows 8, 10, 11

### What happened?

The photocopier in Libraries and where ever else do not work. You can't insert papers into it after opening the tray or fill it up with papers to be copied on to. The highlight on the item when holding the paper just goes away and you can't interact with it. onto

### What should have happened instead?

Should be able to put paper in scanner and tray.the 

### Steps to reproduce the issue

1. Go to a photo copier
2. Open the scanner
3. Try to put paper in it

eurtod:
i don't know how to reopen this but it's not fixed. it works on local now but not on servers still.

eurtod:
thank you

AtnerNT:
Can be closed now, fixed in #10320 and confirmed to be working during playtest

PR #10320, opened 2024-04-05
Anomaly Reports

AtnerNT:
This PR refactors the existing artifact data disk system to use pen and paper reports.

Currently, the old system works as follows:
Science contains an artifact data disk item. This item is inserted into the artifact console. Observations about the artifact are recorded on the console and then uploaded onto the disk. The disk is ejected when ready, and can be sold at cargo. The more accurate the data on the disk, the more valuable it is, up to 10,000 credits for a 100% accurate report.

How this PR changes this:
The data disk item and accompaning box of data disk have been removed from the game. Instead, after observations are recorded in the console, the console will print a paper report of the observations. This paper can be sold at cargo the same way as the data disk can be. Difference being, this report is _not_ a special item and is simply paper with text on it. This report can be read, so players can now view the observations on the report unlike before, also players can edit the contents of reports. Reports can be photocopied, filed, and all other features that accompany paper. Any piece of paper formatted correctly can be sold as a report. 

The purpose of this change is to:
A) Reduce the amount of unnecessary items in the game.
B) Allow players to review and edit research data after uploading.
C) Support roleplay and further bureaucracy gameplay.

See #progress-media for pictures 

### Changelog:

CL: [New] Artifact data disks have been removed. Anomaly research data is now printed on pen and paper reports that can be edited and reviewed at players pleasure. This reports retain their ability to be sold at cargo.
CL: [Balance] While lavaland remains easy, accessible and a better option than asteroids. Anomaly Reports now sell for 12,500 credits instead of 10,000
CL: [Balance] Unresearched Artifact Crates may be purchased from cargo for 10,000 credits for one. This is not a very monetarily profitable option for stations, but maintains a way for science to keep a steady stream of artifacts on rounds where cargo does not have the man power to send miners on asteroid expeditions. 
CL: [Improvement] The artifact research console has been renamed the Anomaly Reporting Console and has received a new look and features to accommodate the data disk changes.
CL: [Fix] Fixed the issue in which photocopier trays could not be loaded by clients.
CL: [Fix] Fixed the issue in which photocopier trays and scanners could both be open simultaneously, leading to confusing interactions.
CL: [Fix] Fixed radioactive artifacts emitting the wrong colour light.


AtnerNT:
Scene changes are removing the now missing datadisk prefabs, all stations with xenoarch mapped needed to be changed

AtnerNT:
This PR should also close #9980. Have tested on LocalHost with two clients and the bug seems to be resolved. Though probably best to also test on staging before closing.

AtnerNT:
Fixed important codacy issues. Tests will continue to fail because of code scan, scan whitelist is updated in this PR but obviosuly wont take effect until after it is merged

Bod9001 (approved):

PR #10328, opened 2024-04-12
The Printing Press Update

MaxIsJoe:
CL: [New] Books can finally be printed from photocopiers.
CL: [New] NT has upgraded all Photocopiers to provide more feedback to be more user-friendly.
CL: [New] Photocopiers can now scan and print several documents at once.
CL: [Improvement] Photocopiers can now hold up to 35 reserve papers for printing due to the new induced demand to print books.
CL: [Fix] Photocopiers will no longer behave as paper shredders in some circumstances where documents would disappear after being inserted inside scanners.


AtnerNT:
A: Documents not being shredded anymore is very nice.

but B: Is there any way to more effectively load paper than 1 by 1? (Having to load 30 pages one by one per print might be a hassle). Not sure what the best way to handle this might be, maybe allow players to interact with the printer using item containers, loading all paper in the containers. Could make paper stacks this way.

MaxIsJoe:
> but B: Is there any way to more effectively load paper than 1 by 1? (Having to load 30 pages one by one per print might be a hassle). Not sure what the best way to handle this might be, maybe allow players to interact with the printer using item containers, loading all paper in the containers. Could make paper stacks this way.

literally what you're asking for is in the changelogs

> Photocopiers can now scan and print several documents at once.

Bod9001 (approved):

PR #10329, opened 2024-04-12
Make switching on lights more visually pleasing

MaxIsJoe:
CL: [Improvement] There is a small delay on light switches affecting its connected light bulbs, making them more visually pleasing when turning on/off.

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Wallmounts/Switches/LightSwitchV2.cs
+++ b/UnityProject/Assets/Scripts/Objects/Wallmounts/Switches/LightSwitchV2.cs
@@ -75,8 +75,9 @@ namespace Objects.Lighting
 		{
 			foreach (var thingToInvoke in SwitchTriggerEvent.GetInvocationList())
 			{
-				yield return WaitFor.Seconds(Random.Range(0.09f, 0.6f));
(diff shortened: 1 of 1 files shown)


Bod9001:
What about loss of power?

MaxIsJoe:
> What about loss of power?

there isn't even anything remotely in this PR that changes how power stuff is handled

light source handles that, we don't change anything about it

Bod9001:
I think light switches get turned off when power goes off

Bod9001:
also did you check that the  state of the  lights aren't  synchronised  already, since you added a hook on the synchronisation of the light switch

MaxIsJoe:
I literally did not change anything, I just added a delay to the invocation

MaxIsJoe:
image

MaxIsJoe:
> since you added a hook on the synchronisation of the light switch

where in this PR did I add a syncvar, or any networking related changes.

Bod9001:
ok, still the issue with light switches though, powering off

MaxIsJoe:
> ok, still the issue with light switches though, powering off

If it was an issue, we would have noticed it long before I made this change.

Bod9001 (approved):

PR #10344, opened 2024-04-28
Retire the germaphobe faith property

MaxIsJoe:
Germaphobia, and by extension the ants, are being retired.

This does not mean that I'm pulling a /TG/. Faiths have always been a work-in-progress feature, and any feature is deemed not popular with the players is going to be removed at some point when people complain about it.

I still 100% believe that there's nothing wrong in how ants behave currently, but maybe it's just not our current player base's cup of tea due to how it exists to punish a specific behavior. So, to respect everyone's wishes, and to save all our future time from hearing yet another critique or complaint about how annoying ants are; Ants will no longer spawn naturally, as there will be no filth score to keep track of.

Admins and mappers can still manually spawn ants and their variants to fit whatever needs they want, but even then; they've received extra nerfs to reduce their destructiveness.

### Changelog:

CL: [Improvement] After several months of feedback, the germaphobe faith property is being retired. We thank all the crew members who got their organs eaten by space ants to test this new feature of faiths; we'll utilize it for better functions later.
CL: [Improvement] Ants will no longer spawn naturally due to the filth state of the station.
CL: [Improvement] Ants can now be removed using soaps and shovels, alongside fly swatters.
CL: [Balance] Ant search radius has been heavily nerfed from 2, 6, 12, 32 tiles to 2, 3, 4, 5 tiles maximum.
CL: [Balance] Ants are no longer able to inflict MobV2s.
CL: [Balance] Ant spread chance has been reduced from 5% to 3%
CL: [Fix] Fixed AntHill prefabs containing more than one attribute script.


AtnerNT:
Hope to see ants back at some point. I still think the main issue with the way they currently work is that there isn't enough user feedback. Food can disappear without warning from an ant nest in maints that someone never saw.
I made further comments about this in the forum post, but yeah, I don't think ants are a bad feature, just the way they function does need to be reworked before they should naturally spawn again.

corp-0 (approved):

PR #10351, opened 2024-05-04
Priority for IExaminable

corp-0:
### Purpose
Adds a new property to IExaminable with a default value of 1, representing the priority of this examine text to be considered when building the final examination text. Higher priority means the text appears first.

I also added tests to make sure this functionality works as intended.


MaxIsJoe (approved):
LGTM

PR #10360, opened 2024-05-10
More flammable fixes

MaxIsJoe:
this PR adds a way for the MapSaver to ignore specific objects that shouldn't be saved (like effects), also some fixes

Changelog:

CL: [Fix] Papers were flameproof for some odd reason.
CL: [Fix] Fixed an issue where flame sprites would take a couple of seconds to appear.
CL: [Fix] The game will only attempt to pre-load flame sprites on objects that actually need it
CL: [Fix] Fixed an issue where sprites would only appear on the server, but not on clients.

### DIFF
--- a/UnityProject/Assets/Scripts/MapSaver/MapSaver.cs
+++ b/UnityProject/Assets/Scripts/MapSaver/MapSaver.cs
@@ -11,6 +11,7 @@ using Objects;
 using SecureStuff;
 using Util;
 using Tiles;
+using UI.Core;
 
 namespace MapSaver
 {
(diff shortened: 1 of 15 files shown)


UnityProject/Assets/Scripts/MapSaver/MapSaver.cs
```diff
@@ -947,61 +948,67 @@ public static string Matrix4X4ToString(Matrix4x4 matrix4X4, StringBuilder SB)
 			CompactObjectMapData compactObjectMapData,
 			Vector3? CoordinateOverride = null, bool UseInstance = false)
 		{
-			var RuntimeSpawned = Object.GetComponent<RuntimeSpawned>();
-			if (RuntimeSpawned != null) return;
+			if (Object.TryGetComponent<MapSaverIgnoreObject>(out var _))
```
Bod9001, line 951: This would only be on the route of objects, So this wouldn't work, I think you should just pool it instead of adding a tag,  RecursiveSaveObject Is where it should go

MaxIsJoe: > This would only be on the route of objects, So this wouldn't work, I think you should just pool it instead of adding a tag, RecursiveSaveObject Is where it should go

I tried pooling it, it's a large headache.

PR #10381, opened 2024-06-07
shutup meg

MaxIsJoe:
no-cl

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -39,10 +39,10 @@ jobs:
         with:
           categories: Fix;New;Improvement;Balance
           write_to_file: false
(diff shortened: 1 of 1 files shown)


MaxIsJoe:
oh hey, it works

MaxIsJoe:
oh fuck off

corp-0:
Not bad. Probably would be the best to do all of that inside the changelog action itself, it would also allow you to have the no cl tag anywhere in the description instead of forcing you to start with it.

MaxIsJoe:
> Not bad. Probably would be the best to do all of that inside the changelog action itself, it would also allow you to have the no cl tag anywhere in the description instead of forcing you to start with it.

problem, the workflow test is always outdated; it doesn't use the latest one. So idk if this is even working or not

PR #10387, opened 2024-06-10
Make the vulpkanin species unique

AutumnNova:
### Notes:

This pr adjusts vulpkanin in the following ways:
Vulpkanin now have a species specific language (Vulpic)
A species specific emote for the vulpkanin that can be used to detect trace amounts of gas on the tile your on
Roughly 1.5 tiles worth of darkvision
Slightly increased movespeed
They can now store a single small item in their mouth (this is just a glorified pocket for now)
Their legs are now more fragile (75 limb hp compared to the regular 100)
Flashes/Welding blindness and other similar effects for vulpkin have their duration increased by roughly 25%
The safe temp range for vulpkin has been changed from -5c - 40c to -40c - 25c

A few of those are potentially a bit too wild, but I personally wanted to experiment a bit with a little more exotic differences for species

### Changelog:

CL: [New] Added Vulpic language, spoken by vulpkanin.
CL: [New] Added species specific emote to vulpkanin that can be used to detect trace amounts of gas.
CL: [Improvement] Given vulpkanin darkvision, increased movespeed, mouth storage, heat weakness, increased flash effect duration and modified limb health.


Bod9001:
I wonder if they should gain cold resistance?

Bod9001:
amusingly I thought there would be more code changes but surprising how much you can do via just modifying some fields xD

AutumnNova:
> I wonder if they should gain cold resistance?

It wouldnt be particularly useful? but I also personally dont think it'd hurt to give them some cold resist, possibly like -40 or so instead of -5

Bod9001 (approved):

PR #10388, opened 2024-06-11
Nuclear Reactor Fixes

AtnerNT:
### Changelog:
CL: [Improvement] Removed water pumps as they had identical functionality to gas pumps, but without the ease of use and the ability to be toggled or rotated.

CL: [Balance] Raised target temperature of reactor boilers to 100 degrees
CL: [Fix] Fixed turbines being unable to decrease their power output (lol)
CL: [Fix] Fixed turbines producing several MW of power even when not in use
CL: [Fix] Fixed reactors not receiving proper water levels
CL: [Fix] Fixed reactor GUI not showing proper K% levels on the sliders
CL: [Fix] Significantly reduced "Flip-Floppy" behaviour of reactor core temperature & pressure while at a stable neutron flux


UnityProject/Assets/Scripts/Objects/Engineering/Reactor/ReactorBoiler.cs
```diff
@@ -29,6 +30,9 @@ public class ReactorBoiler : MonoBehaviour, IMultitoolMasterable, ICheckedIntera
 		public void Awake()
 		{
 			ReactorPipe = GetComponent<ReactorPipe>();
+			ReactorPipe.pipeData.mixAndVolume.SetVolume(10);
```
Bod9001: should be able to set this in the editor, though when I allowed to be serialised apparently start zero so  it must get set somewhere and can't work out where so this is fine for now

Bod9001 (dismissed):
looks good, the only thing I have noticed may be pulling out the red bars from the dial so, You can set them in the editor, but that's be separate from this PR most properly looks good.

UnityProject/Assets/Scripts/Objects/Engineering/Reactor/ReactorBoiler.cs
AtnerNT: Done

Bod9001: well as I meant as in going into pipeData in the editor and setting there , Though it's defaulting to 0 so I imagine it gets set somewhere to 1 idk where that is

Bod9001 (approved):

PR #10395, opened 2024-06-13
Unarmed damage and Hearing range from bodyparts

AutumnNova:
### Notes:

adds a generic component for tracking stats
if something is meleeing and doesnt have human arms it should fallback to whatever value(s) are defined in weaponnetworkactions

as a result of getting localchat range from ears, if you have no ears you cant hear local chat (but can still hear the radio and ooc)

### Changelog:

CL: [Improvement] Players cannot hear local chat without ears.

### DIFF
--- a/UnityProject/Assets/Prefabs/Items/Implants/Organs/HumanRightArm.prefab
+++ b/UnityProject/Assets/Prefabs/Items/Implants/Organs/HumanRightArm.prefab
@@ -515,6 +515,12 @@ MonoBehaviour:
   RelatedPart: {fileID: 0}
   armEfficiency: 1
   crawlingSpeed: 0.3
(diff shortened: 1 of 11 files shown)


corp-0:
can you make sure nothing broke with non human players like aliens? Also what about borgs or cybernetic arms?

AutumnNova:
> can you make sure nothing broke with non human players like aliens? Also what about borgs or cybernetic arms?

Aliens work fine, Borgs I dont think can actually do unarmed melee attacks? but if they somehow do it should be fine as their arms have the relevant component and even if that component is missing it doesnt really break anything

MaxIsJoe (approved):
This looks good imo, If Gilles is okay with it; he can merge it.

PR #10399, opened 2024-06-14
More Reactor Balancing and Fixes

AtnerNT:
Turns out, without the infinite energy glitch, the reactor was fairly underpowered, producing less power than the P.A.C.M.A.N generator. 

### Changelog:

CL: [Improvement] Boiler console warnings now report high/low pressure instead of capacity.

CL: [Balance] Reactors and Boilers no longer spawn with 100u of water when created.
CL: [Balance] Boilers no longer absorb all the energy from the input gas at once. Instead, only absorbing 50% at a time. This reduces the average power output of the turbine but increases the stability of the output voltage.
CL: [Balance] Boilers have had their capacity increased from 10u to 65u. 
CL: [Balance] The maximum core pressure and temperature of reactors has been doubled, allowing them to safely operate at higher neutron-fluxes, in turn, increasing the potential power output of reactors but making them more dangerous to be around.

CL: [Fix] Fixed boiler-turbine console reporting incorrect pressure and capacity readings
CL: [Fix] Fixed Reactors having negative pressure


MaxIsJoe (approved):
LGTM

PR #10404, opened 2024-06-17
Make text bigger each time it stacks

MaxIsJoe:
CL: [New] Chat entries that get stacked will have their sizes increase gradually based on their stack count.

### DIFF
--- a/UnityProject/Assets/Scripts/Core/Chat/ChatEntry.cs
+++ b/UnityProject/Assets/Scripts/Core/Chat/ChatEntry.cs
@@ -186,12 +186,6 @@ namespace UI.Chat_UI
 		private void StackMessageSizeIncrease()
 		{
 			string pattern = @"<size(=|\+=)(\+?[0-9]+\+?)>";
-			var debugString = "";
-			foreach (var character in messageText.text)
-			{
-				debugString += " " + character;
-			}
-			Debug.Log(debugString);
 			Match match = Regex.Match(messageText.text, pattern);
 			if (match.Success)
 			{
--- a/UnityProject/Assets/Scripts/Core/Chat/ChatEntry.cs
+++ b/UnityProject/Assets/Scripts/Core/Chat/ChatEntry.cs
@@ -192,7 +192,6 @@ namespace UI.Chat_UI
 				messageText.text = Regex.Replace(messageText.text, pattern, match =>
 				{
 					float number = float.Parse(match.Groups[2].Value);
-					Debug.Log(number);
 					float newNumber = number + 2;
 					string newTag = $"<size{match.Groups[1].Value}+{newNumber}>";
 					return newTag;
--- a/UnityProject/Assets/Scripts/Core/Chat/ChatEntry.cs
+++ b/UnityProject/Assets/Scripts/Core/Chat/ChatEntry.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Text.RegularExpressions;
 using Managers.SettingsManager;
 using UnityEngine;
 using UnityEngine.UI;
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Core/Chat/ChatEntry.cs
```diff
@@ -172,20 +173,34 @@ public void SetText(string message, TMP_SpriteAsset languageSprite, TMP_FontAsse
...
 			stackObject.SetActive(true);
 			AnimateFade(1f, 0f);
 			StartCoroutine(AnimateStackObject());
+			StackMessageSizeIncrease();
 			this.RestartCoroutine(FadeCooldown(), ref fadeCooldownCoroutine);
 		}
 
+		private void StackMessageSizeIncrease()
+		{
+			string pattern = @"<size(=|\+=|\+)([0-9]+|[0-9]+\+|=[0-9]+\+)>";
```
corp-0: what's with the + at the end? Can the number be something like ``10+``?

MaxIsJoe: yes, it can be like this:

<size=[number]+>
<size=+[number]>
<size+=[number]>

corp-0: what are all the possible cases? Could it ever be ``<size+10`` for example?

MaxIsJoe: I just posted all the possible cases

corp-0: image

your second case doesn't match and it matches a case that shouldn't be valid

corp-0: what characters do you want to capture from the matches? I'm guessing the symbol and the value, right?

corp-0: image
this one should be enough

``<size(=|\+=|=+)(\+?[0-9]+\+?)>``

MaxIsJoe: I pushed a commit that should resolve this:
`<Size(=|\+?=)?([0-9]+|\+[0-9]+|[0-9]+\+)?>`

corp-0: ``<size(=|\+=)(\+?[0-9]+\+?)>``

I had one unused alternative inside the first group

MaxIsJoe: nevermind, it broke completely ingame

MaxIsJoe: > `<size(=|\+=)(\+?[0-9]+\+?)>`
> 
> I had one unused alternative inside the first case

image

this works, but now running into this funny issue

corp-0: that's funny. Could it be because of the multiline? Do you have that flag?

MaxIsJoe: removed it, but problem still presists

UnityProject/Assets/Scripts/Core/Chat/ChatEntry.cs
```diff
@@ -172,20 +173,35 @@ public void SetText(string message, TMP_SpriteAsset languageSprite, TMP_FontAsse
...
+		private void StackMessageSizeIncrease()
+		{
+			string pattern = @"<size(=|\+=)(\+?[0-9]+\+?)>";
+			Match match = Regex.Match(messageText.text, pattern);
+			if (match.Success)
+			{
+				messageText.text = Regex.Replace(messageText.text, pattern, match =>
+				{
+					float number = float.Parse(match.Groups[2].Value);
+					Debug.Log(number);
```
corp-0: remove the debug log

corp-0 (approved):

PR #10409, opened 2024-06-20
Matrix mapping

Bod9001:
### Changelog:
CL: [New] you can now construct new matrices/shuttles with 50 plasma steel 
CL: [New] Copy and paste tool now suports adding new matrixes
CL: [Improvement] you can now select the matrix you want to modify with the  copy and paste tool


corp-0:
> CL: [New] you can now construct new matrices/shuttles with 50 plasma steel

how? It spawns bellow my character? What happens if there is one already there?

Bod9001:
well, it's just there , it will explode if you move the matrix it's on or move it

UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
```diff
@@ -102,11 +104,18 @@ private void OnDisable()
...
 	}
 
-	public static Matrix MakeNewMatrix()
+	public static Matrix MakeNewMatrix(string Name = "Matrix")
 	{
-		//hummmm How to add matrix fluff?
+
+		if (string.IsNullOrEmpty(Name))
+		{
+			Name = "Matrix";
```
MaxIsJoe, line 112: what happens if there are multiple matricies that go by the name "Matrix"?

UnityProject/Assets/Scripts/Messages/Client/mapping/ClientRequestLoadMap.cs
```diff
@@ -19,9 +19,10 @@ public struct NetMessage : NetworkMessage
 		public Vector3 Offset; //Offset to apply 0,0 to get the position you want
 		public string Data;
 		public bool end;
-		public int MatrixID;
+		public int? MatrixID;
```
MaxIsJoe, line 22: why nullable? wouldn't we want all matricies to always have an ID? even if said ID is random?

UnityProject/Assets/Scripts/Shuttles/MatrixNamesSynchronise.cs
```diff
@@ -0,0 +1,32 @@
+using System.Collections;
+using System.Collections.Generic;
+using Mirror;
+using UnityEngine;
+
+public class MatrixNamesSynchronise : NetworkBehaviour
```
MaxIsJoe, line 6: namespace

UnityProject/Assets/Scripts/Shuttles/MatrixNamesSynchronise.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+	public string MatrixNamesSet = "";
+
+	[SyncVar(hook = nameof(SyncMatrixName))]
+	public string MatrixName;
+
+    // Start is called before the first frame update
+    void Start()
```
MaxIsJoe: `private void Start()`

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MatrixSpawner.cs
```diff
@@ -0,0 +1,23 @@
+using System.Collections;
+using System.Collections.Generic;
+using Tiles;
+using UnityEngine;
+
+public class MatrixSpawner : MonoBehaviour
```
MaxIsJoe, line 6: namespace

MaxIsJoe (changes requested):
Looks good, just have some questions and some minor stuff to be cleaned up

UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
Bod9001: it seems to be fine since everything uses the matrix ID (is int)

UnityProject/Assets/Scripts/Messages/Client/mapping/ClientRequestLoadMap.cs
Bod9001: well null new matrix, and it gets assigned to ID by the system

corp-0: we can do nullables?

Bod9001: well the C# feature that says "hey this could be potentially null"  I think it later on but we can do ? to highlight something could be null, useful value types

corp-0: yeah, I know what nullable means. I just thought we couldn't use them for some unity reason, why don't we have more nullable stuff? It would make catching bugs before they happen a lot easier

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MatrixSpawner.cs
Bod9001: Where is the namespace tool on rider? I can't find it

UnityProject/Assets/Scripts/Messages/Client/mapping/ClientRequestLoadMap.cs
Bod9001: don't think it works with Unity components properly CanBeNull it turns into that when it's a class reference on a class

Bod9001: [CanBeNull] GameObject obj yeash

PR #10411, opened 2024-06-24
Fix items spawning inside lights on asteroid scenes

AutumnNova:
Items mapped to the same tile as lights spawn the items inside the lights, this pr fixes the only instances of this occurring on asteroids

Moves light from ontop of the lower table to the wall above on ruins02

Moves container full of seeds left so its not under the lightbulb on ancientstation

Moves lightbulb from above the lower left rack to the bottom wall on ancientstation

### DIFF
--- a/UnityProject/Assets/Scenes/AdditionalScenes/AncientStation.unity
+++ b/UnityProject/Assets/Scenes/AdditionalScenes/AncientStation.unity
@@ -630,6 +630,18 @@ Transform:
     type: 3}
   m_PrefabInstance: {fileID: 9630316}
   m_PrefabAsset: {fileID: 0}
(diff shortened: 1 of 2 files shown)


Bod9001:
Technically it should be something in the code that stops The Storage from sucking it up @MaxIsJoe  could  you take a look?

MaxIsJoe:
This is a workaround, not a way to solve an issue.

There needs to be a better check here for shoving items into storage:
https://github.com/unitystation/unitystation/blob/develop/UnityProject/Assets/Scripts/Items/Others/RandomItemSpot.cs

MaxIsJoe:
why did this get merged? I said that this is a workaround, not a fix

PR #10416, opened 2024-07-01
Add Bar Signs

AutumnNova:
### Notes:

Adds functional bar signs
Bar signs can be locked or unlocked by using a keycard with access on the sign
Unlocked bar signs can have their sign changed by interacting with an empty hand
Bar signs can be emp'd, which will prevent interaction with the barsign and randomly change to a different sign
Bar signs require power to function and will not display anything if they dont have any

The barsign prefab can be configured to only have a specific set of signs when mapped, or alternatively the barsignprefilled prefab can be used for a preconfigured default selection of signs

https://github.com/unitystation/unitystation/assets/55225562/9b27f620-aedc-4d22-8c52-230abe0c4252

### Changelog:

CL: [New] Functional Bar signs.


MaxIsJoe (approved):

corp-0:
That is pretty cool

PR #10419, opened 2024-07-02
Maintrooms tweaks

AtnerNT:
This pr is just some tweaks I was doing for the maintrooms pr I am working on, but will be nessacary for if players will be submitting rooms using the in-game mapping tool.

This is just things like:
Maintrooms entrance and exit markers are now prefabs that can be mapped, aswell as blood splatters (which didnt have prefab trackers for some reason).

This PR also has a number of optimisations for the maint generation script that I thought I might aswell include as they are done. Maint generation should now use less memory overall and removes the old recursive implementation, that limited the size of maint generators due to stack overflows. Maint generators can now be far larger without causing issues. 
The arrays that the maze generator used are now 1D instead of 2D and exclusion zones now use a bulk Array.Fill which should improve both their respective performance aswell.


MaxIsJoe:
Are there any changelogs?
Also why are the decals getting updated?

AtnerNT:
Didn’t do a changelog as thought it was more backend?
And the blood spray decals were given a prefab tracker. They can’t be saved with the map saver without it. Think they predate us having the prefab tracker component.

late for me, will fix unit tests tmrw.

AtnerNT:
Nvm it was annoying, that should fix tests and codacy

MaxIsJoe (approved):

PR #10423, opened 2024-07-05
Mostt ageing fixing2

Bod9001:
### Changelog:

CL: [Improvement] dev tile changer no longer drops items when destroying pipes
CL: [Balance] added automatic scream (set control rod to max when too hot or too high pressure)
CL: [Improvement] Optimizes map loading and saving


MaxIsJoe:
the second log doesn't make sense, can you reword it?

MaxIsJoe (approved):

PR #10430, opened 2024-07-08
Fixes NREs for ACUs on startup

MaxIsJoe:
CL: [Fix] Fixed some NREs that prevented ACUs from being updated properly when they were first loaded into a round.

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Pipes/Devices/Scrubber.cs
+++ b/UnityProject/Assets/Scripts/Objects/Pipes/Devices/Scrubber.cs
@@ -393,7 +393,14 @@ namespace Objects.Atmospherics
 		};
 
 		private AcuSample atmosphericSample = new AcuSample();
-		AcuSample IAcuControllable.AtmosphericSample => atmosphericSample.FromGasMix(metaNode.GasMixLocal);
+		AcuSample IAcuControllable.AtmosphericSample
+		{
+			get
+			{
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/Objects/Pipes/Devices/Scrubber.cs
```diff
@@ -393,7 +393,14 @@ private void UpdatePowerUsage()
 		};
 
 		private AcuSample atmosphericSample = new AcuSample();
-		AcuSample IAcuControllable.AtmosphericSample => atmosphericSample.FromGasMix(metaNode.GasMixLocal);
+		AcuSample IAcuControllable.AtmosphericSample
```
corp-0, line 396:
```suggestion
		AcuSample? IAcuControllable.AtmosphericSample
```

I remember in a recent PR  that we support nullables, right? We should totally use them whenever possible.

MaxIsJoe: We don't entirely support them. If you add a `?` here it will just trigger Jetbrains to replace it with the `[CanBeNull]` annotation.

corp-0: that's what I thought initially, but I saw a PR recently that had ``?``. I think it was @Bod9001

corp-0 (approved):

PR #10434, opened 2024-07-11
Add Temperature Gates

AutumnNova:
Adds temperature gates, the last of these things
Temperature gates only pass gas when the input temp is lower then the threshold, this can be switched to when its higher by using a screwdriver on the gate, a minimum pressure difference of 10kpa is required between the pipenets however.

Also realigns a bunch of pipe sprites so they match with the rest of everything and additionally adds a name to the passive gate fitting prefab

Both temperature gates and pressure valves have animated sprites for when they are passing gas, however these are not implemented due to atmos being on a different thread making that rather difficult.

CL: [New] Add temperature gates.


UnityProject/Assets/Scripts/Objects/Pipes/Devices/PassiveGate.cs
```diff
@@ -12,7 +12,7 @@ public class PassivePump : MonoPipe
 
 		[NonSerialized] public float MaxPressure = 4500f;
 		[NonSerialized] public float ThresholdPressure = 10f;
-		[NonSerialized] public float TargetPressure = 101.325f;
+		[NonSerialized] public float TargetPressure = AtmosConstants.ONE_ATMOSPHERE;
```
corp-0, line 15: you could just use `AtmosConstants.ONE_ATMOSPHERE` wherever you are using `TargetPressure`

AutumnNova: TargetPressure can be changed via nettab, the default value is just AtmosConstants.ONE_ATMOSPHERE

corp-0: ah that makes sense

corp-0:
ignore codacy on this one, it will always trigger when you change one of the GUI_ classes

corp-0 (approved):

PR #10441, opened 2024-07-15
Mapping new prefabs!!

Bod9001:
destroys the non-parented prefabs!!* 
(Apart from old mob spawners, and floor decals)


MaxIsJoe (dismissed):
If there's any other issues, we can fix them later.
Merge if tests are good.

PR #10458, opened 2024-07-23
Get rid of Access Restriction and old door stuff

AutumnNova:
Deletes access control, the old door controller and the old door animators
Moves shutters, blastdoors, mineral doors and false walls to new door code
Moves turret control panel, vendors, ore redemption machine and emitters to use clearance restriction code


corp-0:
huge

Bod9001 (approved):

Bod9001:
A bit harder read on mobile but looks good.

PR #10465, opened 2024-07-28
Add Internal gas tank toggling

AutumnNova:
Adds the ability to toggle using a gas tank for internals via alt-click
All gas tanks have been changed to not be used for internals by default
A held gas tank that has been enabled for usage with your internals will disable itself if dropped or transfered to a gas tank dispenser.

### Changelog:
CL: [Improvement] Add the ability to toggle a gas tanks usage for internals, all gas tanks have been set to not be used for internals by default.


MaxIsJoe:
This is an odd change, why is it needed? and did classic /TG/ have this functionality?

MaxIsJoe (approved):

PR #10466, opened 2024-07-29
Fix healing item double action message and healing item behaviour

AutumnNova:
Fixes the double action messages when using bruise packs (or similar) on yourself
Also changes the healing behaviour of bruise packs (etc) to be more consistent

Prior behaviour would heal 40 damage per limb
e.g: if you had 2 left legs, one with 42 damage and one with 38 damage, one leg would be fully healed and the other would be left with 2 remaining damage

This new behaviour would heal 40 damage total between both legs, resulting in either one leg fully healed and the other with 40 damage left, or one leg with 2 damage left and the other with 38 damage left.
This also means that the amount of healing items used to healing any given amount of damage should be identical no matter how many limbs the damage is split between

### Changelog:
CL: [Fix] Fix double action message when healing with bruise packs.
CL: [Improvement] More consistent healing with bruise packs.


MaxIsJoe (approved):

MaxIsJoe:
Looks good, waitting for tests.

Bod9001 (approved):

PR #10469, opened 2024-07-30
Fix some antag objs

AutumnNova:
HasWorkingThrusters will always return false, so it has been removed as a check in all antag objs
Additionally marooned was checking the owner of the objective instead of the target, which has also been fixed

### Changelog:
CL: [Fix] Fix some antag objs.

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Antagonists/Objectives/EscapeObjectives/Escape.cs
+++ b/UnityProject/Assets/Scripts/Systems/Antagonists/Objectives/EscapeObjectives/Escape.cs
@@ -51,7 +51,7 @@ namespace Antagonists
 
 			return !Owner.Body.playerHealth.IsDead &&
 				ValidShuttles.Any( shuttle => shuttle.MatrixInfo != null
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/Systems/Antagonists/Objectives/EscapeObjectives/Escape.cs
```diff
@@ -51,7 +51,7 @@ protected override bool CheckCompletion()
 
 			return !Owner.Body.playerHealth.IsDead &&
 				ValidShuttles.Any( shuttle => shuttle.MatrixInfo != null
-					&& (CheckOnShip(Owner.Body.RegisterPlayer, shuttle.MatrixInfo.Matrix)) && shuttle.HasWorkingThrusters);
```
Bod9001, line 54: Might be worth checking if   the player is in x range of central command?, If this is just a quick fix then this is fine

Bod9001 (approved):

PR #10470, opened 2024-07-30
Weapon attachments

AutumnNova:
yep, thats it
Combat knife and survival knife are the only bayonets and Seclight is the only flashlight that are attachable right now
Rightclick menu used to detach em because any other approach would be a ux nightmare
IsSupressible has also been thrown out, an enum flag option for allowed attachments has been added instead
Similarly, the supressorPrefab var has been replaced with a generic attachments list, which will be added to the weapon at roundstart

Currently the pka every ballistic weapon except for the bow can have a flashlight and bayonet attachment
The makarov remains the only firearm compatible with the suppressor 
Firearm meleeing has been added and only functions on harm intent

Also fixes a statusdisplay on fallstation that had its channel set to doortimer for whatever reason
And adds the survival knife to the explorers bags, for slappin onto your pka, also so the bayonet interactions actually get tested.

### Changelog:
CL: [Improvement] Combat Knife, survival knife and Seclight weapon attachments

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
+++ b/UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
@@ -193,7 +193,7 @@ namespace Weapons
 		/// If true, displays a message whenever a gun is shot
 		/// </summary>
 		[SerializeField, SyncVar(hook = nameof(SyncIsSuppressed)), Tooltip("If the gun displays a shooter message")]
-		public bool isSuppressed;
+		private bool isSuppressed;
 
 		public bool IsSuppressed => isSuppressed;
 
--- a/UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
+++ b/UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
 using System.Text;
 using UnityEngine;
 using UnityEngine.Serialization;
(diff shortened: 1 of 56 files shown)


UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
```diff
@@ -347,24 +369,132 @@ public virtual void ServerPerformInteraction(InventoryApply interaction)
...
+					         !isSuppressed && isSuppressible)
+						{
+							if (Inventory.ServerTransfer(interaction.FromSlot, suppressorSlot))
+							{
+								SyncIsSuppressed(isSuppressed, true);
+							}
+							return;
+						}
+
+						if (Validations.HasItemTrait(interaction.UsedObject, CommonTraits.Instance.Knife) &&
```
MaxIsJoe: This implementation is going to get messy the more we add more attachments for guns.
It would be better if we had a more generic `GunAttachment` implementation that automatically sets those things up based on the attachment's behaviors.

MaxIsJoe (changes requested):

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
AutumnNova: I can add probably add a generic gunattachment implementation, however all of the itemstorage stuff would still need to be changed in the instance of adding a new attachment, also possibly the rightclick menu stuff(?)

MaxIsJoe: You can give every gun an attachment config, so when it starts; it reserves the item slots it needs for specific attachments to use.
And the right click can check each attachment if it has a specific interaction it wants to be implemented, and show it to the player.

MaxIsJoe (approved):

PR #10475, opened 2024-08-07
Better firearm melee behaviour

AutumnNova:
Fixes the firearm melee range to cover the entirety of the adjacent tiles to the player, instead of just most of them
Tweaks the firearm interactions to run the aimapply check first and then all other interaction checks afterwards which should be more intuitive behaviour

Still toying with the idea of making the melee check only work if there is something that can be melee'd under the clients mousecursor, so you would still be able to shoot if you clicked in the tiles adjacent to the player, but Ive not found a good way to do that yet

### Changelog:
CL: [Improvement] More consistent firearm melee behaviour

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
+++ b/UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
@@ -516,7 +516,7 @@ namespace Weapons
 			if (!DefaultWillInteract.Default(interaction, side)) return false;
 
 			//Melee behaviour for things like bayonets
-			if (interaction.Intent == Intent.Harm && (interaction.TargetPosition - interaction.Performer.RegisterTile().LocalPosition.To2()).magnitude <= 1.5f)
+			if (interaction.Intent == Intent.Harm && BayonetCheck(interaction.Performer.RegisterTile().LocalPosition.To2(), interaction.TargetPosition))
 			{
 				return false;
 			}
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
```diff
@@ -562,6 +562,21 @@ public virtual bool WillInteract(AimApply interaction, NetworkSide side)
 			return false;
 		}
 
+		//.magnitude leaves out small area in the outer corner of tiles diagonal to the player pos
+		//Doing it this way prevents that
+		private bool BayonetCheck(Vector2 playerPos, Vector2 targetPos)
+		{
+			const float checkRange = 1.5f;
+			if (playerPos.x >= targetPos.x - checkRange  && playerPos.x <= targetPos.x + checkRange)
```
MaxIsJoe, line 570: You can just use `Vector3.Distance()`

MaxIsJoe (changes requested):

UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
AutumnNova: Vector3.Distance() (and .magnitude) will make a circular radius around the player where they wont fire their gun and will instead melee (if in range), which is not what I feel would be intuitive here, given a circular radius cant cleanly cover the entirety of all tiles that are adjacent to the player
If I switch this back to use Vector3.Distance() (or .magnitude), then either the outer corners of the tiles directly diagonal to the player will be unmeleeable, or I extend the radius to cover those corners at the cost of having a weird deadzone where you cannot melee because it'd be outa range, but you also cant shoot covering half of the next tile in cardinal directions

MaxIsJoe (approved):

PR #10482, opened 2024-08-07
Add Dual wielding

AutumnNova:
Adds dual wielding, only works on harm intent
It should be noted that wielding a fullauto weapon in your main hand and a semiauto weapon in your offhand will not let you fire the semi weapon as if it was a fullauto weapon, additionally the weapon in your offhand will only ever fire as fast as the weapon in your main hand
The way this is implemented does mean you can tri-wield with 3 arms, quad wield if with 4 arms (etc), which leads to some interesting results

https://github.com/user-attachments/assets/f73f522c-3e92-4e8e-a3d5-f2fd96196560

### Changelog:
CL: [New] Add Dual wielding

### DIFF
--- a/UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
+++ b/UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
@@ -621,6 +621,20 @@ namespace Weapons
 			{
 				FiringPin.ServerBehaviour(interaction, isSuicide);
 			}
+
+			if (interaction.Intent == Intent.Harm && interaction.UsedObject == gameObject)
+			{
+				List<ItemSlot> hands = interaction.PerformerPlayerScript.DynamicItemStorage.GetHandSlots();
+				hands.Remove(interaction.PerformerPlayerScript.DynamicItemStorage.GetActiveHandSlot());
+				foreach (var hand in hands)
+				{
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Items/Weapons/Gun.cs
```diff
@@ -621,6 +621,20 @@ public virtual void ServerPerformInteraction(AimApply interaction)
 			{
 				FiringPin.ServerBehaviour(interaction, isSuicide);
 			}
+
+			if (interaction.Intent == Intent.Harm && interaction.UsedObject == gameObject)
+			{
+				List<ItemSlot> hands = interaction.PerformerPlayerScript.DynamicItemStorage.GetHandSlots();
+				hands.Remove(interaction.PerformerPlayerScript.DynamicItemStorage.GetActiveHandSlot());
```
Bod9001, line 628: how does it work if you then put something into that hand?

AutumnNova: this removes the active handslot from the newly created list of all handslots, so the foreach loop doesnt iterate over the gun thats triggering the dual wield interaction because that would trigger the dual wielding code twice
so it shoudnt really effect anything?

Bod9001: ok cool

PR #10487, opened 2024-08-11
Add Proto Kinetic Crusher

AutumnNova:
Adds the proto kinetic crusher, a melee version of the proto kinetic accelerator which can mark a target via a ranged attack that does no damage, to then deal bonus damage when that target is hit in melee, additionally theres extra bonus damage for backstabs

Few things to note, its impossible to mark mobs that arent using healthv2/mobv2 stuff, because old mobs dont support status effects
As such none of the trophies have been implemented either
Additionally theres currently no sprite indicator for if a player has been marked

Also implements the recharge icon indicator for all proto kinetic weaponry

Additionally ports some proto kinetic variants because whynot:tm:, sprites are from monkestation

hammer variant, knocks hit targets back, 70 bonus mark damage, cannot crit, can be worn on back, is a huge item
machete variant, 1s recharge time, can be worn on belt, is a large item, 35 bonus mark damage, 20 bonus backstab damage
spear variant, 1.5s recharge time, can be worn on back, is a hugeitem, 35 bonus mark damage, 20 bonus backstab damage, it should have a 2 tile reach but our melee system doesnt support that soooo
claw variant, 0.5s recharge time, is a normal size item, does 40 bonus mark damage, does 60 bonus backstab damage

All variants are available in the mining vendor for 1250 points
### Changelog:
CL: [New] Adds the proto kinetic crusher and similar weaponry


MaxIsJoe (approved):
Everything looks good

PR #10489, opened 2024-08-12
Add melee blocking

AutumnNova:
Adds the block chance stat to itemattributesv2, which dictates the odds of a melee attack being blocked if a player is holding that item in their active hand, 0% being never and 100% being always
Adds a block chance of 25 to the proto kinetic machete, 30 to the buckler, 50 to all other shields, the officers sabre and (activated) energy swords

Also prevents removal of firingpins from proto kinetic crushers (and variants)
Additionally fixes a bug where the right click removal options for weapon attachments wouldnt appear

### Changelog:
CL: [New] Adds the ability for shields to block melee attacks

### DIFF
--- a/UnityProject/Assets/Scripts/Items/ItemAttributesV2.cs
+++ b/UnityProject/Assets/Scripts/Items/ItemAttributesV2.cs
@@ -5,6 +5,7 @@ using UnityEngine;
 using Mirror;
 using AddressableReferences;
 using Core;
+using Core.Utils;
 using Systems.Clothing;
 using UI.Systems.Tooltips.HoverTooltips;
 using Util.Independent.FluentRichText;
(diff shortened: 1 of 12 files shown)


UnityProject/Assets/Scripts/Items/Weapons/WeaponNetworkActions.cs
```diff
@@ -247,6 +250,45 @@ public void ServerPerformMeleeAttack(GameObject victim, Vector2 attackDirection,
 		Cooldowns.TryStartServer(playerScript, CommonCooldowns.Instance.Melee);
 	}
 
+	private bool BlockCheck(GameObject victim)
```
corp-0, line 254: maybe decorate this with ``Server`` just to be safe.

UnityProject/Assets/Scripts/Items/ItemAttributesV2.cs
```diff
@@ -91,6 +91,17 @@ public float TraumaDamageChance
 		[EnumFlag]
 		public TraumaticDamageTypes TraumaticDamageType;
 
+		[SerializeField,
+			Range(0, 100),
+			Tooltip("How likely a player is to block an attack if they are holding this item in their active hand, 0% for never.")]
+		private float blockChance = 0;
+
+		public float BlockChance
```
MaxIsJoe: It would be better to choose `MultiFloat` so blocking has interest sources.
`blackChance` is the base, while `BlockChance` manages all sources. That way you don't have to manage when the stats change on things like energy swords when they're on/off.

MaxIsJoe (changes requested):

MaxIsJoe (approved):

PR #10498, opened 2024-08-18
Add an ingame option to clear sound cache

MaxIsJoe:
CL: [New] Added a new option to clear sound cache in-game.

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Core/OptionsMenu/SoundOptions.cs
+++ b/UnityProject/Assets/Scripts/UI/Core/OptionsMenu/SoundOptions.cs
@@ -1,6 +1,9 @@
-﻿using Audio.Managers;
+﻿using System.Linq;
+using Audio.Managers;
 using Audio.Containers;
+using TMPro;
 using UnityEngine;
+using UnityEngine.AddressableAssets;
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/UI/Core/OptionsMenu/SoundOptions.cs
```diff
@@ -142,7 +159,7 @@ public void ResetDefaults()
 					AudioManager.MusicVolume(0.2f);
 					AudioManager.TtsVolume(0.2f);
 					AudioListener.volume = 1;
-					AudioManager.CommonRadioChatter(false);
+					AudioManager.CommonRadioChatter(true);
```
Bod9001: wasn't agreed upon for it to be false?

MaxIsJoe: I didn't agree to anything like that?
Why would I agree to disabling a feature that I added

Bod9001: I think it should be off by default

PR #10513, opened 2024-09-02
Add hand cranked weapons

AutumnNova:
Adds a component that allows for hand crank recharge weapons, which require activating in your hand to recharge, which takes 2s

Adds a single shot hand cranked musket that fires a modified disabler bolt that also does 15 burn
Adds a single shot hand cranked smoothbore disabler pistol which has a very wide spread 

Haven't added a way to obtain either of these ingame yet as I dont really know if they should be obtainable naturally, and also I dont really know how they should be obtainable, so eh

Also contains a fix to energy weapons so they dont play the dryfire sound ten billion times if you hold the mouse down after running out of shots

### Changelog:
CL: [New] Adds a hand cranked energy musket and a hand cranked smoothbore disabler pistol


Bod9001 (approved):
looks good

Issue #10163, opened 2023-11-04
[Feature/Dev] Improving the Mobv2 development experience


MaxIsJoe:
As of right now, we have a pretty solid and good structure for creating advanced mobs on HealthV2 that can also be playable by other players thanks to the efforts of I and Bod; However, the sheer amount of power we gave ourselves came with a slight pinch of complexity and made the approach to V2 mobs not an easy thing to get into from the perspective of outsiders who have not touched V2 mobs before.

This issue aims to list down a few tasks and discussions to help make working with MobV2s a lot easier and quicker with less chances of creating mistakes by newcomers or other current developers.

# Prerequisite info: How do MobV2s function? 

* MobV2s work exactly like current players, any change you make for players also applies to mobs; as `MobV2`s are prefab varients of `Player_V4(Unet)` 
* MobV2s have a special `Mob` component that defines things such as their name and character sheet, because MobV2s are spawned as NPCs, not players, so they require a special component to initialize important info about them for the `PlayerScript` component.
* When MobV2s spawn, they spawn with souls and minds, just like players; The only difference is that Mobs spawn with a `NotImportantMind` flag which disregards them as NPCs. 
* When a MobV2's gets initialized, it will use RaceSOs to define what body it will have and what Health systems it will spawn with. (See Arial for example)
* MobV2s can have "blueprint" prefabs, which allows us to make one prefab of an NPC then change its properties via lightweight prefabs that can also act as items in-game. (See slimes for example)

## Prerequisite info: How does the AI for MobV2s function?

* V2s use a state-machine like system for processing different state of a mob's AI, called `MobAI` and `BrainMobAI`.
* `MobAI` allows mobs to have more than one state active at the same time, allowing complex behaviors; such as running and gunning at the same time, without creating monolithic components that handle more than one behavior at the same time.
* Each state has its own blacklist of states it cannot co-exist with, giving programmers and designers an easy way to avoid state conflicts without writing a lot of checks in-code (Example: A happy state has a blacklist for a sad state, thus preventing the AI from processing the happy and sad state at the same time)
* States can be pre-configured to run periodically over time or over every `Update` and `FixedUpdate` call on the `UpdateManager`
* All states have a `EnterState`, `ExitState` and `HasGoal` and `Tick` functions, names are self-explanatory.
[...]

PR #10535, opened 2024-09-25
Mob creator/editor tool

MaxIsJoe:
part of  #10163

Allows contributors to easily create new mobs without having to worry about going through the manual process of creating them.
This has no changes related to gameplay or anything, this is just some tooling for developers.

### DIFF
--- a/UnityProject/Assets/ScriptableObjects/Health/BodyPartsBaseSO.asset
+++ b/UnityProject/Assets/ScriptableObjects/Health/BodyPartsBaseSO.asset
@@ -1,0 +1,26 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!114 &11400000
+MonoBehaviour:
+  m_ObjectHideFlags: 0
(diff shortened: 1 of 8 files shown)


Bod9001:
does it initialise the Body health systems automatically?

MaxIsJoe:
> does it initialise the Body health systems automatically?

It doesn't. You are given the option to edit them manually, however.

Bod9001:
just saying this will result in stuff like the new species Spawning in without any blood

MaxIsJoe:
> just saying this will result in stuff like the new species Spawning in without any blood

do you want me to add in something like presets for people to use automatically?

Bod9001:
ideally it could be like prefab variance but I can't think of a good way of doing it

Bod9001:
if you think it's outside the scope of the PR is good

Bod9001 (approved):

MaxIsJoe:
I can add that functionality later when I expand upon this tool.
But right now it does its intended use-case fully, anything else can be considered QoL additions to make the mob creation process faster.

PR #10538, opened 2024-09-27
Printable Cloning Records

MaxIsJoe:
image
image

final
image

CL: [New] Cloning machines now print cloning records on paper for medical staff to use.
CL: [Improvement] Updated cloning UI to follow the diegetic design of unitystation.
CL: [Improvement] Improved the flow and usability for the cloning console UI.


Bod9001:
tessts are failing

UnityProject/Assets/Scripts/Objects/Medical/CloningConsole.cs
```diff
@@ -158,12 +142,22 @@ public void ServerTryClone(CloningRecord record)
 		{
 			if (cloningPod && cloningPod.CanClone())
 			{
-				var status = record.mind.GetCloneableStatus(record.mobID);
+				Mind mind = NetworkUtils.FindObjectOrNull(record.mindID)?.GetComponent<Mind>();
```
Bod9001, line 146: there's an extension to uint to get the game object/networkk identity
NetIdToGameObject();
NetworkIdentity();

UnityProject/Assets/Scripts/Objects/Medical/CloningConsole.cs
```diff
@@ -223,26 +248,83 @@ public CloningRecord()
...
+			sb.Append("|");
+
+			// Serialize sicknessList
+			sb.Append(string.Join(",", sicknessList));
+
+			return sb.ToString();
+		}
+
+		// Method to deserialize the object from a compact string
+		public static CloningRecord FromCompactString(string compactString)
```
Bod9001: why not use Jsonconvert?

UnityProject/Assets/Scripts/UI/Objects/Medical/Cloning/GUI_CloningItemList.cs
```diff
@@ -1,53 +0,0 @@
...
-	{
-		public void AddItems(int count)
-		{
-			for (int i = 0; i < count; i++)
-			{
-				AddItem();
-			}
-		}
-
-		public bool AddItem()
```
Bod9001, line 16: oh so, it no longer saves it in machine?

UnityProject/Assets/Scripts/Objects/Medical/CloningConsole.cs
MaxIsJoe: Json is slow

UnityProject/Assets/Scripts/UI/Objects/Medical/Cloning/GUI_CloningItemList.cs
MaxIsJoe: It no longer does.

UnityProject/Assets/Scripts/Objects/Medical/CloningConsole.cs
MaxIsJoe: this is fine

UnityProject/Assets/Scripts/Objects/Medical/CloningConsole.cs
Bod9001: This is not performance critical code xD
map loading uses Jsonconvert and it's not the slow bit

UnityProject/Assets/Scripts/UI/Objects/Medical/Cloning/GUI_CloningItemList.cs
Bod9001: so you just have to keep the Records next the machine? and Scan get paper put it into machine then clone?

UnityProject/Assets/Scripts/Objects/Medical/CloningConsole.cs
MaxIsJoe: either way this works just fine and looks good

UnityProject/Assets/Scripts/UI/Objects/Medical/Cloning/GUI_CloningItemList.cs
MaxIsJoe: yeup

AtnerNT: Think it might be good for the cloner to store like the most recent scan in like a buffer.
If you want to store any more than that you need to file your papers.

AtnerNT: Unless it already does that then yeah

MaxIsJoe: > Think it might be good for the cloner to store like the most recent scan in like a buffer. If you want to store any more than that you need to file your papers.

The current implementation is better for usability, as It's pretty much straight forward and more interactive.

UnityProject/Assets/Scripts/Objects/Medical/CloningConsole.cs
Bod9001: What about if someone edits the paper and has invalid data? also we want to add stuff in the future that is safe such as genetic defects

MaxIsJoe: It will be rejected if its invalid.
Either way, they can't directly mess with character sheet stuff because it's stored on the mind.
Also, that can be added by extending the string that gets printed on the paper with the info that we need.

Bod9001: But my point is it's easier just to add to a class versus having  add to the class then add theserialise and then do  deserialise steps and if it's conditional God help you xD

MaxIsJoe: json doesn't look good imo

MaxIsJoe: we could probably turn it into binary, but that would probably take more than the page can hold

AtnerNT: I think Binary would be good, unless you do some weird stuff, the binary should always be equal to or shorter than your string method

Bod9001: the problem is you have to do special stuff for example holding mutations since you can have a variable amount so therefore you have to say how many mutations it has so you can de serialise number,  yyou might as well just use json  works

MaxIsJoe: Can't do binary, the result takes more than the page can take.

Bod9001: what about json

MaxIsJoe: I've said this before two times already, JSON looks ugly and should not be put infront of players.

corp-0: What do you guys mean when you say binary?

Regarding showing this to users, are supposed to be able to read it? Otherwise you could just base64 it

MaxIsJoe: Users should be able to read it.

PR #10546, opened 2024-10-01
Implement breakable shields

AutumnNova:
Creates the breakable shield component, which will make any item with it applied take damage from an attack when an attack is blocked

Additionally the component has a configuration option to allow repairing it with an item

Riot/Tele/Strobe shields have been set to be repairable using a singular titanium sheet (takes 2s), whilst the Roman shields and the wooden buckler cannot be repaired.

### Changelog:
CL: [New] The riot, strobe, telescopic, roman and wooden shields now take durability damage when blocking attacks.
CL: [New] The riot, strobe and telescopic shields can be repaired using a titanium sheet


MaxIsJoe (dismissed):
Looks good 👍

corp-0 (approved):

PR #10550, opened 2024-10-04
Adds DIMGUI finally as a UI solution

MaxIsJoe:
A way to do UI using declarative code inside standard Mono/Network behaviors. Nothing actually uses this yet, but damn was this a hellish experience to get it working.


UnityProject/Assets/Scripts/IMGUIDemo.cs
```diff
@@ -0,0 +1,33 @@
...
+		if (Input.GetKey(KeyCode.LeftShift) && Input.GetKeyDown(KeyCode.Insert))
+		{
+			showDemo = !showDemo;
+			if (showDemo)
+			{
+				ImGuiUn.Layout += OnLayout;
+			}
+			else
+			{
+				ImGuiUn.Layout -= OnLayout;
```
Bod9001: Do you need to allow this on the code scan list?

MaxIsJoe: Oh yeah thanks for reminding me.
this needs to be allowed on the code scan list

PR #10556, opened 2024-10-06
Chill bug fixing

Bod9001:
### Changelog:

CL: [New] Bedsheets now cover you when snuggled in a bed
CL: [Balance] Update Hot ice reaction to be accurate to temperature ranges and consume plasma
CL: [Fix] Fixes constructing sinks Giving you power sinks
CL: [Fix] Fixes infinite restart after restarting round once


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/Reactions/FreonFireReaction.cs
```diff
@@ -50,7 +50,7 @@ public void React(GasMix gasMix, MetaDataNode node)
 
 					gasMix.RemoveGas(Gas.Freon, freonBurnRate);
 					gasMix.RemoveGas(Gas.Oxygen, freonBurnRate * oxygenBurnRate);
-
+					gasMix.RemoveGas(Gas.Plasma, freonBurnRate);
```
AutumnNova, line 53: I very much object to this change honestly
While I understand the intent to make the freon event not have the station burn down, this isnt really going to do much to solve that and the change to the temperature range (~~which as a side note, if we want to mirror tg should be 120k to 160k~~ EDIT: I was wrong about that, it should be changed to 273.15k) should fix this

Bod9001: Though what does it change in terms of making though?
I don't feel like it has that much of a big impact, if you're going out of your way to make it

AutumnNova: well, freonfire handles the freon cooling effect, and it does so in the proper way, requiring oxygen to cool and producing co2 as a byproduct whilst also slowly cooling it over time, so you are effectively gating the proper cooling method behind the presence of plasma

Id personally say removing this change (and the change to the energy release part) in addition to removing the improper freon cooling implementation would be nice, but yknow

AtnerNT:
The power sink fix is already in #10555 which is merged. 
As for the Hot Ice changes, ive already discussed my feeling on that change on the discord, and and have eleaborated a bit further on Autumns PR

UnityProject/Assets/Scripts/Managers/GameManager.Escape.cs
```diff
@@ -64,12 +64,6 @@ private void InitEscapeStuff()
 
 		Vector3 newPos;
 
-		if (LandingZoneManager.Instance.centcomDocking == null)
```
AtnerNT, line 67: Why are you removing the error log?

AtnerNT: And how does this fix the infinite restart?

Bod9001: well this is a hardcoded thing for the old system of docking, so it's no longer used

Bod9001: also it would never reset the shuttle status so it was always thinking of heading back to central command after It was being called

AtnerNT: So new mapping meant this was always null and returning when that didnt matter anymore?

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/Reactions/FreonFireReaction.cs
Bod9001: the problem is it gets too cold for gameplay, I picked a temperature that winter coats can protect you from, and is not to devastating to be in without protection, making it simulated  base will make it really hard to balance

UnityProject/Assets/Scripts/Managers/GameManager.Escape.cs
Bod9001: The object that handled this was a custom object so it didn't get saved in the new json map, so it only cropped up now

MaxIsJoe:
There's a gameplay change here that I don't quite fully understand, so I can't weigh in and say which is more correct.
I'll instead invite everyone to discuss the Freon change on discord, and once a conclusion + agreement is made on what to do; I'll review this PR again.

MaxIsJoe (approved):

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Meta/Atmospherics/Data/Reactions/FreonFireReaction.cs
Bod9001: will work on your PR to add the changes required to balance numbers around but would be good+

PR #10558, opened 2024-10-07
Fix freon Cooling behaviour

AutumnNova:
Deletes freon freeze entirely, the behaviour responsible  for instantly setting the roomtemp to 220.15k (-50c) as this implementation is bad. very bad.

As the reaction is set to only trigger between -45c and 45 and instantly sets the temperature to -50c, there is no way to heat the station up to a survivable temperature without first scrubbing the entire station of freon. Which is insanely overpowered for traitors and those looking to grief the station.

Along with that it pretty much breaks any semblance of balance for the supermatter, not that any of the maps that are played on at low-pop have the sm, but I dont believe that particularly justifies breaking the balance this badly.

Additionally for extra salt in the wound, as Freon Burn has the proper cooling implementation for freon, if there is oxygen present the freon will use that behaviour to cool, and it will do so quite rapidly as the proper freon implementation cools faster the colder the freon is. such that it takes less then a minute to go from -50c to -120c

After it gets to about -120c it will start spitting out hot ice and it will continue to cool until it reaches the temperature limit set in atmosdefs (60k or -220c ish) 

Along with deleting freon freeze this pr also sets the max burn temp to freon to 0c, so freon will not do its exothermic reaction unless the tile its on is 0c or below

Tiles below -40c or so remain slippery, that behaviour is entirely untouched

Video testing the behaviour from this pr because git doesnt like videos greater then 10mb: https://www.youtube.com/watch?v=Bin9SiYOokQ

CL: [Fix] Fix freon cooling behaviour to not instantly throw the station into extremely cold temps


corp-0:
I understand the sentiment but maybe we should think in ways to keep the functionality. Tweak numbers and what not.

I don't think Bod will approve plainly deleting that interaction

AutumnNova:
Freon Burn Still cools like freon freeze does, it just doesnt instantly set the temperature to -50c. Its also pretty much the implementation most ss13 servers use for freon (granted modern tg uses slightly different calculations then what we have currently, but its close enough:tm:)

theres also just not really any way to adjust freon freeze in a way that wont cause issues, as it sets the temperature instantly to a predefined value

AtnerNT:
Hmmm, mixed, the freon burn reaction does cool down, but it also consumes the freon and can make hot ice form simply from opening a canister of freon.
I agree that the freon cool effect shouldn't instantly set the temp to -50 but it should still exist. I discussed a possible implementation with bod here: https://discord.com/channels/273774715741667329/312454684021620736/1292534293599490241

Also the freon burn reactions shouldnt occur until like -60. Otherwise freon will quickly cool the entire station to -160 and place hot ice everywhere, which is what happened during the playtest, where a simple freon leak flooded the station with hot Ice

AutumnNova:
> Hmmm, mixed, the freon burn reaction does cool down, but it also consumes the freon and can make hot ice form simply from opening a canister of freon. I agree that the freon cool effect shouldn't instantly set the temp to -50 but it should still exist.

Im very much not sure what you mean by "it should still exist"? the only behaviour change this pr really makes is not having freon set itself to -50c when within the 45c to -45c degree temperature range, because that instantly deletes almost 100 degrees worth of temperature for free and thats bad:tm: (and also making freons cooling effect only happen at 0c instead of 20c, but like eh)

For clarification, The implementation behaviour of freon fire reaction.cs is and has always been the following:

when below the temperature defined by FREON_MAXIMUM_BURN_TEMPERATURE and when there is oxygen present, a portion of freon and oxygen is consumed and turned into co2, which slowly cools down the enviroment

it takes with these atmosdefs about 5 minutes to go from 0 degrees to -220 with a pure 1:1 ratio of oxy/freon at ~1000kpa
at the end of that you are left with roughly 200kpa of oxy freon and carbon (about 20% carbon, equal of the other)

freon still causes the frosted tile effect, that remains unchanged from this pr

additionally if the burn rate of freon is higher then a specific threshold and between 160k and 120k there is a 1/3rd chance of spawning some hot ice

changing the starting threshold for the cooling effect and the minimum temperature freon reaches can be easily done and should be done via the relevant atmosdefs

if someone wishes to modify behaviour for hot ice spawning, this should be done with the relevant if statement within the freon fire reaction code, instead of modifying the heat behaviour for freons cooling effect entirely

which is not what bod is doing in their pr and will not result in the desired behaviour mentioned in discord

I also feel the behaviour seen on playtest is largly in part due to freon freeze instantly putting the entire station at -50c, which speeds up freons cooling effect quite rapidly

AtnerNT:
I haven't played a whole bunch with freon, but if your numbers are correct with the example you provided, then I approve pf these changes

Bod9001:
The only issue as it uses up oxygen so makes the atmosphere unBreathble

AtnerNT:
> The only issue as it uses up oxygen so makes the atmosphere unBreathble

Well, I mean, the solution there is dont flood the station with freon if you dont have an air supply in the vents.
If you are flooding the station with freon you should flood it with O2 aswell

Bod9001:
another issue is it makes it hard to balance in terms of freeze level  and pressure
because if you make it too cold the pressure drops too low and you start taking pressure damage and if it's too cold to start taking overwhelming cold damage where even a winter coat will protect you, 
I balanced it so don't take pressure damage and a winter coat can protect you, and it's not too devastating being caught out with out a winter coat

AutumnNova:
You balanced it such that freon can delete slightly under 95 degrees worth of temperature per atmos update  and  will still rapidly cool itself to unlivable temperatures afterwards. Plus it is almost impossible to reheat the station if theres greater then 1 mole of freon on a tile and it additionally destroys sm balance, as mentioned in the pr description.

Which really isnt balanced in the slightest? and also I feel kinda takes the complex and intricate atmos system and kinda just throws it in the garbage bin honestly.

Bod9001:
I don't mind making a more complicated mechanism for how it becomes cold, I just want it to be
freon flood = Snow mode, where the dangers are the cold and the slipperiness of the tiles
and the the traitors can put in a bit of effort to cause hot ice to form making it a death trap (e.g adding BZ as to the air)

Bod9001:
image
Smashing a canister of freon results in zero cooling in Square station on the central Wing, so it does not in fact cool anything

AutumnNova:
as mentioned in both the pr description and one of the comments explaining the exact behaviours of the freon cooling implementation, only is triggered below 0c

if you want to change that, you can do so easily by changing the atmosdefs
if you also wish the change the minimum temperature freon can cool to, you can also change that in the atmos defs as ive mentioned several times at this point bod

Bod9001:
I'll be working on this PR to tweak the values to try out TG style and see if that's satisfactory

PR #10569, opened 2024-10-13
Add Strobe shield's targeted flash

AutumnNova:
Implements an OnMelee Action to ItemAttributesV2, uses this to implement strobe shields melee targeted flash, this shares a cooldown with the non-targeted flash

Also adds the ICustomMeleeBehaviour Interface, which will allow implementing custom melee behaviour like backstabs or random damage without having to copy-paste a bunch of melee code every time.
 Changes MarkedMelee and ExtradimBlade to use this, eventually Stun batons are also going to be converted to use this

Additionally cleans up WeaponNetworkActions a bit.

CL: [New] Meleeing a player with the Strobe shield will stun them.


AtnerNT:
Does it just stun them? or does it flash them aswell?

AutumnNova:
> Does it just stun them? or does it flash them aswell?

Stun + Flash, as its currently impossible to have one without the other given FlasherBase just does nothing if passed 0 for stun duration

Bod9001 (approved):

MaxIsJoe (approved):
Looks good

PR #10570, opened 2024-10-13
Playtest Fixes

### LINKED ISSUE #4127: Additions to the Chief Engineer's tools.
## Description
The CE's advanced tools (Experimental welding tool, Jaws of Life, Hand Drill) could use a few more features.

#### Jaws of Life and Hand Drill should have toggleable modes.
Right now the Jaws of Life has both crowbar and wirecutter traits and the hand drill has both screwdriver and wrench traits. They both have these dual traits active at the same time. While this works alright most of the time, it does cause odd behavior, such as when trying to use the hand drill on girders. 

You may intend to disassemble a girder after unsecuring it using the screwdriver trait but the wrench trait overrides it and makes you resecure the girder. 

Toggling modes on the jaws of life and hand drill should update the sprite, show a message, and perhaps even play a sound.

#### Jaws of Life could force open doors.
In /tg/, the Jaws of Life are able to force open doors, even powered ones, if they are used on them after a few seconds. If this behavior is added, I think it would be best if it was an individual component as other items (such as the Fire Axe and Crowbar) are able to force open doors (though not all can force open powered doors). 

#### Experimental Welding Tool needs to generate fuel.
The Experimental Welding Tool should generate it's own fuel over time as in /tg/.

- [x] Jaws of Life and Hand Drill toggleable modes.
- [x] Jaws of Life force open doors.
- [ ] Experimental Welding Tool regenerates fuel.

### LINKED ISSUE #10553: Synthetic and Cybernetic limbs should be unified.
Currently in game we possess two sets of items, the synthetic limb set, and the cybernetic limb set.
Synthetic limbs look identical to cybernetic limbs, but their purpose is to be implanted on living creates to improve mobility and function.
Cybernetic limbs on the other hand, require power and sole purpose is to be implanted as part of cyborg construction.

It doesn't make sense to have these two items separate, it is unintuitive and can lead to mistakes when attempting to perform surgery on players. 
Synthetic limbs should be dissolved, and their functionality added to cybernetic limbs. Battery packs should be able to be implanted in organics safely and organics should be able to charge them by entering charging bays just like cyborgs.

Not only is this an important balancing factor to synthetic limbs (You can be 10x the normal move speed, but you will need to recharge your limbs) it also is a lot more natural gameplay wise.

**Note to developers:** If you are addressing this issue, synthetic limbs and cyborg limbs both exist on the techweb, design files and their own exosuit fabricator categories. Ensure this is all corrected as part of the unification.

AtnerNT:
The bugs with MMI I wanted to fix, but i do not understand our brain code and how to adjust it, so have excluded those fixes from this PR.

Closes #10553
Closes #4127 

BTS Changes: Alot of the cyborg body parts have be renamed to fit with the generic 'Capitals, No spaces' naming convention of prefabs

### Changelog:
CL: [New] Added the ability for the CE's welding tool to regenerate fuel

CL: [Improvement] Cyborg surgeries have been giving more text and are no longer instant
CL: [Improvement] Chemistry reagent mixes now have their amounts rounded to 2 decimal places. This should make it easier to get precise values in mixes
CL: [Improvement] Synthetic limbs have been removed and their functionality merged with the cyborg limbs. Cyborg limbs may now be implanted on organics for the same benefits synthetics used to supply. Battery packs now may also be implanted on organics to power these limbs.
CL: [Improvement] Empty eye sockets no longer! Cyborg lights now have bodypart sprites finally giving cyborgs eyes!

CL: [Balance] Cyborg limbs are now researchable with no prior nodes on the tech web and at a reduced price (Due to synthetics no longer being available from round start). Synthetic Organs now require cyborgs to be researched first

CL: [Fix] Fixed implant explosives exploding at 0,0
CL: [Fix] Fixed knives attempting to cut the pneumatic cannon instead of being placed inside
CL: [Fix] Fixed cargo crates spawning in the walls of the shuttle
CL: [Fix] Fixed smoke only consuming 1 unit of reagent
CL: [Fix] Fixed cyborgs starting off with engineering sprites instead of default
CL: [Fix] Fixed cyborgs sprites not updating after adding a tool carousel
CL: [Fix] Fixed certain cyborg surgeries providing bleed stacks


Bod9001:
Tests

AtnerNT:
Ready for review

UnityProject/Assets/Scripts/Systems/Chemistry/ReagentMix.cs
```diff
@@ -480,7 +483,7 @@ public void Multiply(float multiplier)
 				foreach (var key in reagents.m_dict.Keys)
 				{
 					var nuber = reagents.m_dict[key];
-					nuber = nuber * multiplier;
+					nuber = (float)Math.Round(nuber * multiplier, 2);
```
Bod9001: The only issue I see is with stuff like health where, blood has like a really small amount x and y

Bod9001: this might result in chemical duplication

Bod9001: /destruction

UnityProject/Assets/Scripts/Systems/SmokeAndFoam/SmokeAndFoamManager.cs
```diff
@@ -40,6 +40,8 @@ public static void StartSmokeAt(Vector3 PositionWorld, ReagentMix Container, int
 		SourceReservoir.PerTile.Multiply(SourceReservoir.ReagentPurity);
 		SourceReservoir.StacksLeft = AmountOfSmoke;
 
+		Container.Take(AmountOfSmoke - 1);
```
Bod9001, line 43: ok, 
theoretically if you want to make something smoke , should it use up all the reagents in a container, like is the entire contents of the container getting Carried in the smoke?

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
```diff
@@ -259,7 +264,8 @@ public void BodyPartAddHealthMaster(LivingHealthMasterBase livingHealth) //Only
 
 			SetHealthMaster(livingHealth);
 			BodyType bodyType = BodyType.NonBinary;
-			if (playerSprites.ThisCharacter != null)
+			if (isCybernetic == true && playerSprites.ThisCharacter.Species != "Cyborg") bodyType = BodyType.Other2; //Synthetic
```
Bod9001, line 267: this is technically some of the auto init code for players getting used ,  probably a better way is when you make a cyborg arm it sets the Gender automatically so it doesn't think it's unset

UnityProject/Assets/Scripts/Systems/Chemistry/ReagentMix.cs
AtnerNT: Does health use reagent amounts smaller than 0.01u? didnt think it did, but even so I could just change it to 3dps.

UnityProject/Assets/Scripts/Systems/SmokeAndFoam/SmokeAndFoamManager.cs
AtnerNT: SO think about it this way:
I have a beaker, 500 tiles of smoke, 50u of acid. Right now, that smoke will place 0.1u of smoke on everytile, but only conusme 1u of acid from the beaker.

This is a bug fix so that 500 tiles of smoke, will place 0.1u on each tile, consuming 50u of acid in the process,

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
AtnerNT: But confused what you mean on this one.
So explanation why this is here, cyborg limbs, as im sur eyou know, determine their variant based on the body type. Its how engi and janitor cyborgs exist. Problem is, if you implanted a cyborg limb on like a male human, it would have engi borg sprites because the body type of the organic. This here checks if the limb is cybernetic, but the host organic, look it to have the organic cyborg limb sprites.

UnityProject/Assets/Scripts/Systems/Chemistry/ReagentMix.cs
Bod9001: it might do since, each body part consumes a small amount of nutriment and oxygen and stuff,  so that might not play nicely

AtnerNT: Would 3dps be enough? I highyl doubt we dealing with numbers smaller than 0.001u

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
Bod9001: well what I'm saying is that, you set the initial sprites  when it's made through machinery, 
so it doesn't get auto set when implanted

UnityProject/Assets/Scripts/Systems/SmokeAndFoam/SmokeAndFoamManager.cs
Bod9001: When working on it I don't mind the reagent duplication in the air, 
since with stuff like healing smoke if you get a lot of it to get an actual decent amount of healing you would have to have like 300u and I don't think any containers have that

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
AtnerNT: Body parts always have their sprites set to match the body once implanted, its not cyborg specific. Even if you set the body part type before implanting it'll get overridden once implanted. This line exists to disable that autoset in the case of cyber => organic implants

UnityProject/Assets/Scripts/Systems/Chemistry/ReagentMix.cs
Bod9001: I feel like it would upset the maths for the automatic balance of the organs, since it does the maths to work out how much each organ should consume for nutriment and stuff,  maybe have it so beakers and transferring rounds it if you're using reagent containers, but not reagent mix directly

UnityProject/Assets/Scripts/Systems/SmokeAndFoam/SmokeAndFoamManager.cs
AtnerNT: I havent changed the logic of how much reagent is in the cloud at all. I just made it actually use up the reagent it puts in the cloud

AtnerNT: Also while it wasnt the point you were making, we have bluespace beakers which have 300u i believe

UnityProject/Assets/Scripts/Systems/Chemistry/ReagentMix.cs
AtnerNT: The issue is, we have stuff happening right now like:
20u - 10u = 9.99998u, which makes precise chem amounts impossible to work with. Also does any organ use less than 0.001u per second? That seems obsurdly low

Bod9001: image
this is a tiny amount, but it's every second so + all the other body pparts it adds up to the right aamount

AtnerNT: Hmmm, If I make it 4dp, might still prevent some floating point erros and nutrient consumption will be still 0.0004. which is only 5% lower than current consumption. I dont think a 5% decrease in consumption will change much at all, an in return chemistry will be easier to work with

Bod9001: ok, fair enough but I feel like it might cause issues with the small consumption, I would rather go for whenever you do reagent transfers as a player it Rounds it, up to you though

UnityProject/Assets/Scripts/Systems/SmokeAndFoam/SmokeAndFoamManager.cs
Bod9001: yeha, it's fine to consume the entire beaker

Bod9001: probably best to make it do that, what about a scenario where it's 1 smoke and 299u acid?
is it just very acidic smoke xD in one tile

Bod9001: nah, the current weight works would be it would just be 1u acid on that tile, 
so it just consumes the entire container?

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
Bod9001: that shouldn't be happening, Once it set with sprites it should be keeping those

UnityProject/Assets/Scripts/Systems/SmokeAndFoam/SmokeAndFoamManager.cs
AtnerNT: If its just one tile... yeah that smoke is just gonna have some incredible concentration lol

Bod9001: or should it only consume the equivalent amount of acid one smoke equals one acid consumed?

UnityProject/Assets/Scripts/Systems/Chemistry/ReagentMix.cs
AtnerNT: Its not just reagent transfers though, its whenever reagents are added or removed, floating point precision makes you get stuff like 1u + 2u = 3.00001u.
Also with small consumption, 4dps is incredibly small, if something is consuming less than 0.0001u in a given timeframe, just extend the timeframe by that point

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
AtnerNT: The unchanged code is:

BodyType bodyType = BodyType.NonBinary;
if (playerSprites.ThisCharacter != null)
{
	bodyType = playerSprites.ThisCharacter.BodyType;
}

Whenever it was added to a new body, its Sprites were set to match that, no matter what.

Also even if wasnt the case, the solution you proposed wouldn't work as they NEED to change when implanted to cyborgs. So in either scenario code here is needed.

UnityProject/Assets/Scripts/Systems/SmokeAndFoam/SmokeAndFoamManager.cs
AtnerNT: You saying to add like a 'max saturation' to the clouds? I mean could do, but is there any downside to letting one tile of smoke be super dense? It doesnt break anything as at that point its arguably less effective than just throwing the beaker and letting it spill

AtnerNT: Wait it already does that nvm. `Container.Take(AmountOfSmoke - 1);`
It only consumes the amount of acid for the number of smoke. If theres only 1 smoke tile itll only consume 1 acid

UnityProject/Assets/Scripts/Systems/Chemistry/ReagentMix.cs
Bod9001: yeah, Fair enough

UnityProject/Assets/Scripts/Systems/SmokeAndFoam/SmokeAndFoamManager.cs
Bod9001: I think it's good

UnityProject/Assets/Scripts/HealthV2/Living/BodyParts/BodyPart.cs
Bod9001: they shouldn't need to change when implanted to cyborgs if they default to non-binary,  
 the only  time they should need to uupdate as if the cyborg chose a different body, 
 but then it's like getting a cleaning robot leg and putting it on engineer Of course it would look odd

AtnerNT: The thing is, the cyborg NonBinary sprites aren't the ones that are needed for humans. Basically, if you print a cyborg leg, it should be able to implanted on a cyborg or a human without being strange.
You should have to take the carousel out, and put it back in each time you want to add more legs to an Engiborg for instance. They should change to match the borg they are implanted on. And if implanted on an organic, they should be the humanoid borg sprite.
There shouldnt be a specific object for humans, one for borgs, one for engi borgs etc... that goes against the whole point of unifying synth and cyborg limbs, as it was just confusing

Bod9001: well the problem with the prosthetic arm versus Robotic arm 
prosthetic arm has a hand, cyborgs don't have hands

Bod9001: so you need something to differentiate between the two

AtnerNT: Cyborgs have hands as part of this, it's more of an unintended consequence of the unification, but it never made sense that for some arbritary reason you couldn't implant hands onto cyborgs in the first place.

AtnerNT: Its mostly because we dont have any seperation between hands and arms as body parts. If we ever get that distinction, you can seperate cyborgs arms into the two parts. But it never made sense tht you can implant a new hand onto a human but not a cyborg

Bod9001: well the entire gameplay poiint of cyborgs as they don't have hands they have the tool Carousel instead, meant to differentiate between playing a normal character

AtnerNT: They still have a tool carousel, but arms give hands in unitystation. If the intent is that cybirgs that spawn in dont have hand slots, then they shouldn't spawn in with arms. Otherwise they arent arms they are just slabs on metal attached to their side.
Again this is because we have no distinction between hands and arms as body aprts (cant surgically remove hands without removing arms). But as long as that's the case, cyborg arms should give hand slots

AtnerNT: Also, cyborgs still dont have back slots, chest slots, belts, pockets etc. Giving them hands does not ruin cyborgs, and allows then to do stuff like push buttons. The tool carousel is still very valuable for them

Bod9001: hummm, it's just departing quite a big way from ss13 cyborgs

Bod9001 (approved):

PR #10572, opened 2024-10-13
main stations converted to json

Bod9001:
### Purpose
 <br>


MaxIsJoe:
converted to draft until everything is done

UnityProject/Assets/Scripts/Core/RootSillys/SingletonScriptableObject.cs
```diff
@@ -0,0 +1,60 @@
...
+namespace ScriptableObjects
+{
+	/// <summary>
+	/// Abstract class for making reload-proof singletons out of ScriptableObjects
+	/// Returns the asset created on the editor, or null if there is none
+	/// Based on https://www.youtube.com/watch?v=VBA1QCoEAX4
+	/// </summary>
+	/// <typeparam name="T">Singleton type</typeparam>
+
+	public abstract class BadSingletonScriptableObject<T> : ScriptableObject where T : ScriptableObject
```
corp-0, line 14: why is it called bad?

Bod9001: this is due to lighting system Being in its own assembly and being annoying, so have to copy this to make the common materials SO singleton worrk ,it's bad because it doesn't reference the  statiic manager with all the SO preloaded

PR #10638, opened 2024-11-28
chore: modifies the log format string

corp-0:
It is now closer to what Max was already doing, for example:

``[TimedEventsManager.cs/Awake:26] Setting up event hooks.``

I also removed now redundant text from some logs in TimedEventManager.


MaxIsJoe (approved):
looks good

PR #10661, opened 2024-12-19
Fixes an instance where chat would block character naming

MaxIsJoe:
CL: [Fix] Fixes a long-standing issue where attempting to edit a character's name in the character UI would cause the chat UI to popup while typing.

### DIFF
--- a/UnityProject/Assets/Scripts/UI/Systems/Character/CharacterSettings.cs
+++ b/UnityProject/Assets/Scripts/UI/Systems/Character/CharacterSettings.cs
@@ -63,6 +63,7 @@ namespace UI.Character
 			ShowCharacterPreviewOnCharacterSelector();
 			characterEditor.SetActive(false);
 			characterSelector.SetActive(true);
+			UIManager.Instance.isInputFocus = false;
 		}
 
 		public void EditCharacter(int key)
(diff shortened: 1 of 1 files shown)


Bod9001 (dismissed):

UnityProject/Assets/Scripts/UI/Systems/Character/CharacterSettings.cs
```diff
@@ -79,6 +81,7 @@ public void ShowCharacterEditor(CharacterSheet character)
 			characterSelector.SetActive(false);
 			characterEditor.SetActive(true);
 			characterEditorScript.LoadCharacter(character);
+			UIManager.Instance.isInputFocus = true;
```
Bod9001, line 89: Do you remember to unfocused it?

Bod9001: it might be worth putting on  disable just to make sure

corp-0:
Since you are there, can you remove the name validation? Just let people call the characters whatever they want, for now. 

We can find strategies to validate names  accordingly to the character status later.

UnityProject/Assets/Scripts/UI/Systems/Character/CharacterSettings.cs
MaxIsJoe: pushed a change where it disables the focus when the entire window disappears

MaxIsJoe:
> Since you are there, can you remove the name validation? Just let people call the characters whatever they want, for now.
> 
> We can find strategies to validate names accordingly to the character status later.

I'll do it in another PR because Unity is being a bitch right now and I can't do anything.

Bod9001 (approved):

PR #10674, opened 2025-01-19
Adds TAGs to admin actions

Bod9001:


corp-0:
let's make a single file with all of these strings as contants, so we never mistype them in the code

UnityProject/Assets/Scripts/Messages/Client/VariableViewer/OpenBookIDNetMessage.cs
```diff
@@ -16,7 +16,7 @@ public override void Process(NetMessage msg)
 
 		private void ValidateAdmin(NetMessage msg)
 		{
-			if (IsFromAdmin() == false) return;
+			if (IsFromAdmin("VV") == false) return;
```
corp-0: w?

UnityProject/Assets/Scripts/Messages/Client/ClientMessage.cs
```diff
@@ -32,11 +34,9 @@ public static void SendUnreliable(T msg)
 			NetworkClient.Send(msg, 1);
 		}
 
-		internal bool IsFromAdmin()
+		internal bool IsFromAdmin(string PermissionCode)
```
corp-0:
```suggestion
		internal bool HasPermission(string PermissionCode)
```

UnityProject/Assets/Scripts/Messages/Client/VariableViewer/OpenBookIDNetMessage.cs
Bod9001: variable viewer VV

UnityProject/Assets/Scripts/Messages/Client/ClientMessage.cs
corp-0: actually, the permissions code should already have a ``HasPermission`` method

UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
```diff
@@ -379,12 +380,15 @@ private bool ValidatePlayerAdminStatus(PlayerInfo player)
...
 		//PlayerLimit Checking:
 		//Deny player joining if limit reached and this player wasn't already in the round (in case of disconnect)
 		var playerLimit = GameManager.Instance.PlayerLimit;
-		if (ConnectionCount > GameManager.Instance.PlayerLimit && roundPlayers.Contains(player) == false)
+
+
+
+		if (ConnectionCount > GameManager.Instance.PlayerLimit
+		    && roundPlayers.Contains(player) == false
+		    && AdminCommandsManager.IsAdmin(player,"ADMIN_BYPASS_PLAYER_LIMIT") == false)
```
MaxIsJoe: This is going to be problematic in the future if we want to check for more than one perms at the same time.
Reflections can probably be used to automatically generate enums that exist in the tags/perms files, but if you don't want to use enum flags; at least make the argument an array.

MaxIsJoe (changes requested):

UnityProject/Assets/Scripts/Messages/Client/VariableViewer/OpenBookIDNetMessage.cs
corp-0: `VARIABLE_VIEWER` then, no reason to be a cheapskate with letters

UnityProject/Assets/Scripts/Managers/PlayerList.Admin.cs
Bod9001: why not overload that takes array?

MaxIsJoe: that works as well

MaxIsJoe: just make sure to have the strings as consts so we don't mistype them

Bod9001: cool

UnityProject/Assets/Scripts/Messages/Client/VariableViewer/OpenBookIDNetMessage.cs
Bod9001: cool

UnityProject/Assets/Scripts/Messages/Client/ClientMessage.cs
Bod9001: this has extra logging on

PR #10678, opened 2025-01-29
Radio effect, 3D sounds, and some fixes for TTS

MaxIsJoe:
This PR was meant to include more QoL improvements for TTS to make the station feel more lively by making chatter be heard through walls, however, this ended up being problematic due to how TTS and chat "security" is being handled currently.
I'll revisit this later after discussing a solution with Bod.

CL: [New] Radio chatter that goes through TTS will now have a radio filter applied to it.
CL: [Improvement] TTS messages are now 3D, which allows them to be effected by spatial properties (such as reverb, distance fall off, etc).
CL: [Fix] Fixed an issue with TTS audio that made them unaffected by volume settings.
CL: [Fix] Fixed an issue where TTS audio would get repeated when speaking in multiple radio channels.

### DIFF
--- a/UnityProject/Assets/Prefabs/Chat/ChatSystem.prefab
+++ b/UnityProject/Assets/Prefabs/Chat/ChatSystem.prefab
@@ -11,6 +11,8 @@ GameObject:
   - component: {fileID: 929989539608245524}
   - component: {fileID: 114008945277933904}
   - component: {fileID: 82862869719533520}
+  - component: {fileID: 1399361671882644652}
+  - component: {fileID: 8097790567109026546}
   m_Layer: 5
   m_Name: Mary_TTS
   m_TagString: Untagged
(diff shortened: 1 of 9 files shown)


corp-0 (approved):

corp-0:
> chatter be heard through walls, however, this ended up being problematic due to how TTS and chat "security" is being handled currently.
I'll revisit this later after discussing a solution with Bod.

I think it is fair that if you can't read it, you can't hear it. If you really wanted to do it, you could maybe play a generic muffled chatting sound whenever someone speaks behind walls, so that you can't really make out what they are saying (and also skip the need to make tts work with it)

MaxIsJoe:
> I think it is fair that if you can't read it, you can't hear it.

True, but older SS13 servers that used TTS (before the current AI methods were a thing) made speech echo throughout the station; which made the station feel more alive and inhabited, since you could always hear chatter of some kind. We can't really replicate that with a generic sound effect; but we could try implementing blips (like how most forks and ss14 does currently).

PR #10680, opened 2025-02-01
Admin acccess levels

Bod9001:
### Changelog:
CL: [New] new system for handling admins ranks and permissions


Bod9001:
is not perfeect with showing on the client e.g the next round drop-down is still visible but I can't really think of a good way of handling it without  making a bespoke Thing

and another good point with next round drop-down  
should it be a separate Permission to see what a different admin has set next round drop-down  , since that is currently lumped in with TAG.ADMIN_INFO that has loads of other stuff too

Bod9001:
~~NOTE 2 
it doesn't have live updating, 
since I don't know where to put it, 
since it would need to access the logged in admins and refresh those and send out messages but also the path for permissions~~
fixed

MaxIsJoe (approved):

UnityProject/Assets/Scripts/UI/Systems/AdminTools/Interfaces/UITAGShowAndHiding.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using UnityEngine;
+
+public class UITAGShowAndHiding : MonoBehaviour
+{
+
+	public TAGType TAGType;
+
+	public List<TAGType> NeedsAny;
```
corp-0, line 11: instead of two lists, you could do what we do in clearance code. It is basically the same logic

PR #10681, opened 2025-02-02
Adds a visible warning when `CharacterManager` fails to fetch characters

MaxIsJoe:
CL: [Improvement] Added a clear warning/error message for players who cannot fetch characters from the database. This should reduce the confusion players and developers have when some issues occur.

### DIFF
--- a/UnityProject/Assets/Scripts/Systems/Lobby/CharacterManager.cs
+++ b/UnityProject/Assets/Scripts/Systems/Lobby/CharacterManager.cs
@@ -272,8 +272,18 @@ namespace Systems.Character
 					{
 						Loggy.Error(
 							$"Failed to load characters online. because: {accountResponse.Exception!.Message}");
+						UIManager.InfoWindow.Show("Failed to load characters online" +
+						                          $"{accountResponse.Exception!.Message}",
+							false, "Error");
 					});
-					throw accountResponse.Exception;
+					if (accountResponse.Exception != null)
+					{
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/Systems/Lobby/CharacterManager.cs
```diff
@@ -272,8 +272,18 @@ public async Task LoadOnlineCharacters()
 					{
 						Loggy.Error(
 							$"Failed to load characters online. because: {accountResponse.Exception!.Message}");
+						UIManager.InfoWindow.Show("Failed to load characters online" +
+						                          $"{accountResponse.Exception!.Message}",
+							false, "Error");
 					});
-					throw accountResponse.Exception;
+					if (accountResponse.Exception != null)
```
corp-0, line 279: there already is an error logged above. No need to throw an exception that won't be handled

MaxIsJoe: Rider shows a warning  if its left without a null check

Bod9001: why not handle log a big error but generate a new character?

MaxIsJoe: because it's not always about generating a new character, it could be a network error

Bod9001 (approved):

PR #10686, opened 2025-02-06
Increase audio source limit

MaxIsJoe:
CL: [Improvement] Increases audio source limit to x4 the previous limit. This has a slight increase in CPU usage on clients, but will be beneficial in the long run.
CL: [Improvement] Added a buffer for audio streaming, which should improve audio latency and edge-cases that would cause short one-shot sounds (like gun shots) to spike system resource usage.


corp-0:
> but will be beneficial in the wrong long.

MaxIsJoe:
fixed

Bod9001 (approved):

PR #10688, opened 2025-02-07
Update CodeScanList.json to add new transports

MaxIsJoe:


CodeScanList.json
```diff
@@ -3131,6 +3131,17 @@
                 "All": true
             }
         },
+        "Mirror.Transports": {
+            "Mirror.MultiplexTransport": {
+                "All": true
```
Bod9001: due to the security risk only poke holes what you need

Bod9001: same for the others

MaxIsJoe: This is exactly what Codescan is complaining about
you posted the errors

Bod9001: well as in, you don't need to access everything, you only need to add access to the stuff you're using, yyacht using every single variable inside of Mirror.MultiplexTransport

Bod9001: can remove the         "All": true from them and it should give a bit more detailed error

PR #10705, opened 2025-02-19
Fixes some issues with round starts and gives more details to clients when loading

MaxIsJoe:
CL: [Improvement] Clients now receive more info about their loading states to keep them more in the loop.
CL: [Fix] Fixed an issue where everyone who joins would attempt to start a new round, even if they did not have permissions or reason to do so.
CL: [Fix] Character sheets are now pre-fetched during the loading process for players to prevent some unwanted bugs with the character creator and join process.
CL: [Fix] Fixed an issue where the character creator would not pop up automatically when no character sheets are detected once the game finishes loading.

### DIFF
--- a/UnityProject/Assets/Scripts/Player/JoinedViewer.cs
+++ b/UnityProject/Assets/Scripts/Player/JoinedViewer.cs
@@ -13,6 +13,8 @@ using Messages.Client;
 using Messages.Client.NewPlayer;
 using Messages.Client.SpriteMessages;
 using UI;
+using UnityEngine;
+using Util.Independent.FluentRichText;
 
 namespace Player
 {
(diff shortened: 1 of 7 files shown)


UnityProject/Assets/Scripts/Player/JoinedViewer.cs
```diff
@@ -316,6 +330,7 @@ private IEnumerator WaitForLoggedOffObserver(Mind loggedOffPlayer)
 			var netIdentity = loggedOffPlayer.GetComponent<NetworkIdentity>();
 			if (netIdentity == null)
 			{
+				GUI_PreRoundWindow.Instance?.OnClientLoadUpdateStatus?.Invoke("<color=red>An error occurred. Press F5 to check for what error had occured.</color>");
```
corp-0: use the styling extension, I think it should be

```cs
"An error occurred. Press F5 to check for what error had occured.".Red();
```

Bod9001 (approved):

PR #10708, opened 2025-02-22
Mob Traversal + AStar pathfinding.

MaxIsJoe:
CL: [New] New Mobs (including players) can now utilize the new pathfinding solutions to traverse around the map.
CL: [New] Added A* Pathfinding as an alternative method for pathfinding. This method results in generally lower quality paths, but should be faster when attempting to cover larger distances.
CL: [New] Pathfinding solutions now have the option to generate paths through doors. (It is highly recommended to not over-use them as they can heavily spike memory usage and CPU on larger matrices)
CL: [New] Added a new mutation that allows the skin to become invisible. The invisibility mutation is only available for feline and lizard based species.
CL: [Improvement] Arials now decloak from their invisibly after receiving damage.
CL: [Improvement] Arials can now chase down their prank victims.
CL: [Improvement] Arials can now open doors when walking to different destinations.
CL: [Improvement] Arials can now attempt to climb tables when attempting to cross a path that's blocked by a table.
CL: [Improvement] Arial mobs have had their walking speeds greatly increased, and they now use the same movement that players use.
CL: [Fix] Our resident pranksters have been finally fixed, and Arials will once again be able to exchange their culture with Nanotrasen employees.


UnityProject/Assets/Scripts/Tilemaps/Behaviours/Pathfinding/PressurePathfinder.cs
```diff
@@ -19,71 +19,190 @@ private bool PositionIsOutOfBounds(ChunkedTileMap<MetaDataNode> terrain, Vector3
...
+
+			foreach (var direction in directions)
+			{
+				var neighbor = position + direction;
+				if (PositionIsOutOfBounds(terrain, neighbor)) continue;
+				var tile = terrain.GetTile(neighbor);
+				if (checkForDoors)
+				{
+					if (tile.PositionMatrix.GetFirst<DoorMasterController>(tile.LocalPosition, CustomNetworkManager.IsServer) != null)
+					{
```
AtnerNT: Would it be worth adding a clearance check here so they don't generate paths unless they have access to them? Or is the intention here that hostile mobs will navigate through doors attacking them? 
(Thinking of stuff like simple bots that might interact with some doors but not all)

UnityProject/Assets/Scripts/Mobs/BrainAI/States/Arial/LookForTroubleArialAi.cs
```diff
@@ -14,91 +18,76 @@ namespace Mobs.BrainAI.States.Arial
...
 			}
 
-			if (pathfinder == null) return;
-			path = pathfinder.FindNewPath(
-				MatrixManager.WorldToLocal(LivingHealthMaster.playerScript.playerMove.OfficialPosition,
-					LivingHealthMaster.RegisterTile.Matrix).RoundTo2Int(),
-				target.gameObject.AssumedWorldPosServer().RoundTo2Int());
-			Loggy.Info($"{target} - {target.OrNull()?.AssumedWorldPosServer()}");
-			if (path != null)
+			if (pathfinder.QueueMovementGoal(target.gameObject.TileLocalPosition().To3Int(),
```
AtnerNT, line 79: If the new target position here is the same as the old one, dont generate new path. 
Should only need to generate a new path if the target moves and/or they encounter an un expected obstacle in their path.

(Not sure if these conditions are implemented elsewhere, hard to tell only seeing the diffs)

UnityProject/Assets/Scripts/Mobs/Traversal/MobTraversal.cs
```diff
@@ -0,0 +1,190 @@
...
+using UnityEngine;
+
+namespace Mobs.Traversal
+{
+	public class MobTraversal : MonoBehaviour
+	{
+		public PlayerScript Mob;
+		public MovementSynchronisation Movement;
+		public int MaxQueuedTargets = 12;
+		public int MaxRetries = 6;
```
AtnerNT, line 20: What is the reasoning for 6? I feel if a pathfinding attempt has already failed twice its already reason enough to move to the next target in the queue, unless I am mis interpreting what these retries are for

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Pathfinding/PressurePathfinder.cs
AtnerNT: I wonder if theres a way to optimise this;

A) Is a tile with a door consider occupied? if so you can skip this check if the tile is unoccupied,
B) Maybe using room number if that doesnt work, is a door considered part of a room? if not, if the tile has a room number other than -1 you can also skip this check

AtnerNT (commented):
I only did a quick read through, so some of the comments might be mis informed but let me know

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Pathfinding/PressurePathfinder.cs
MaxIsJoe: All mobs need to know paths that go through doors so they can handle them on their own when they reach them.

MaxIsJoe: room numbers are broken for a very long while and I gave up on fixing them

UnityProject/Assets/Scripts/Mobs/BrainAI/States/Arial/LookForTroubleArialAi.cs
MaxIsJoe: I thought about that when I first added this function, but then I remembered that we might want mobs to go back to their original position after they reached their goal; so I intentionally removed that from the queue code.

UnityProject/Assets/Scripts/Mobs/Traversal/MobTraversal.cs
MaxIsJoe: There could be a problem during their traversal that doesn't let them reach their next tile, so they need to retry traversal again to `pos` in case they encounter something like a table, door, a player swapping with them, a player pulling them, atmos throwing them around, etc.

If they stopped immediately after the second or third attempt, then they would look like they give up easily, or they lost track of what they're doing mere seconds after they are given a goal. 6 is a sweet spot between giving them time to perform actions (like attacking a tile or waiting for something to open) while also not making them look like they are stuck in place.

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Pathfinding/PressurePathfinder.cs
AtnerNT: On this I checked, isOccupied is true when a door is closed at that location, I still think you should skip this door check if isOccupied is false as the door check is significantly more expensive and doing it on every tile is wasteful

MaxIsJoe: This isn't new information

image

MaxIsJoe: realised that what you were talking about was missing for some reason so I inverted it again

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Pathfinding/PressurePathfinder.cs
```diff
@@ -19,71 +19,196 @@ private bool PositionIsOutOfBounds(ChunkedTileMap<MetaDataNode> terrain, Vector3
...
+				var tile = terrain.GetTile(neighbor);
+				if (tile == null) continue; // No tile at this position. Probably space.
+				if (tile is { IsOccupied: false })
+				{
+					yield return neighbor;
+				}
+				else
+				{
+					//(Max): This is a performance killer.
+					//We need to find a way to update tiles to register if their occupied type is made by something like a door/window/etc.
```
AtnerNT: There is a RegisterDoor class that doors specifically use, if you can access the register tile you could possibly do something like:

if(tile is RegisterDoor)
Though not sure if accessing a registertile on a given layer is easier or harder than checking for a door

AtnerNT: And for windows, you should just be able to check the highest layer on a tile. Which on a tile with windows should be the window layer (Or Object for the Grille, cant remember).
Floor bots used this to find base tiles when they worked

MaxIsJoe: using this idea now since it doesn't do `GetComponent` anymore
though I'm unsure how faster this is, sooo, warning still there.

UnityProject/Assets/Scripts/Core/Input System/InteractionV2/TileInteraction/TableInteractionClimb.cs
```diff
@@ -56,54 +58,84 @@ public override bool WillInteract(TileApply interaction, NetworkSide side)
...
 	{
+		if (interaction.UsedObject == null || interaction.UsedObject.TryGetComponent(out UniversalObjectPhysics objectPhysics) == false)
+		{
+			return;
+		}
 		if (!interaction.UsedObject.RegisterTile().Matrix.IsPassableAtOneMatrixOneTile(interaction.TargetCellPos, true, true, null, excludeTiles))
 		{
 			return;
 		}
+		StartClimbing(true, interaction.PerformerPlayerScript,
```
Bod9001, line 69: what did you change about this function since the git compare has just brutalised it

MaxIsJoe: I moved it into its own function so I can call it from the CilmbTable strat

Bod9001: cool

Bod9001 (approved):

PR #10738, opened 2025-03-05
shutters/blastdoors indestructible open

FuntimeDudeski:
### Purpose
adds prefabs for these. can't test them because linux but i scrupulously copied the prefab overrides from the non-indestructible versions.

### Notes:
n/a

### Changelog:
n/a


FuntimeDudeski:
just realized i PR'd 2 different prefab updates that both need to update networkmanager. fortunately i forgot to include it in both branches so i fixed that one and will just wait till the other is merged then update this one.

PR #10740, opened 2025-03-06
indestructible shutters/blastdoors but open

FuntimeDudeski:
### Purpose
adds prefabs for these. can't test them because linux but i scrupulously copied the prefab overrides from the non-indestructible versions.

### Notes:
n/a

### Changelog:
n/a


FuntimeDudeski:
will proceed to test these shortly!

PR #10741, opened 2025-03-06
falsewall sprites

FuntimeDudeski:
### Purpose
adds all needed spritesheets and SOs to make falsewall variants of almost every artificial wall type

### Notes:
once merged, will test and PR new falsewall prefabs
also moved a couple old style shuttle spritesheets out of the way

### Changelog:
n/a for now


FuntimeDudeski:
hm, i think the conflicts are because i moved some old shuttle wall files and put the new ones in their place with the same names. easiest way to fix?

PR #10744, opened 2025-03-06
file metadata

FuntimeDudeski:
### Purpose
unity keeps wanting to update these files with metadata. i don't understand why they weren't and it's scary and i want it to go away. i don't know what this PR does but it's almost all just putting something in a foreverID field.

### Notes:
n/a

### Changelog:
n/a


Bod9001:
Forever IDs are used to save them in the map so they need to be saved

Bod9001 (approved):

PR #10754, opened 2025-03-12
Adds auto connect to logins when connecting from the hub

MaxIsJoe:
CL: [Improvement] Players logging into their account for the first time when connecting to a server will automatically connect to the server once they successfully log in now.


MaxIsJoe:
this was annoying to test because login code use Task instead of UniTask, so had to test for multiple edge cases where the main thread might get accidentally blocked. It works now, but make a note for later that we should update all our Task code to UniTask, and if we *really* need to use different threads, UniTask offers a thread pool that we can use that also automatically goes back to the main thread when its done its job.

Bod9001 (approved):

corp-0:
Big

PR #10759, opened 2025-03-16
teleport inhib matrix system item

FuntimeDudeski:
### Purpose
- adds teleport inhibitor to matrix system items
- one falsewall spritesheet i forgot (the one i'm trying to actually use of course)

image

### Notes:
n/a

### Changelog:
n/a


MaxIsJoe:
what does this item do?

FuntimeDudeski:
> what does this item do?

teleport inhibitor is a component that prevents teleportaion via hand teleporter or bluespace banana peels in a radius (i don't know for sure whether it prevents reactive armor and teleport-on-hit effects--it probably should). currently using a wide-area one in valley cause the map could be cheesed/made incoherent with random teleportation. this PR just makes a prefab with it.

MaxIsJoe:
That's a bit curious, considering I don't recall anything in the teleportation code that checks for anything like this.

MaxIsJoe (approved):
ah well, nothing sus is here. Approved.

AtnerNT:
> That's a bit curious, considering I don't recall anything in the teleportation code that checks for anything like this.

```cs
if (ignoreInhibitors == false)
{
	foreach(TeleportInhibitor inhib in TeleportInhibitor.Inhibitors)
	{
			var inhibPosition = inhib.GetComponent<UniversalObjectPhysics>().OfficialPosition.RoundToInt();
			if(Vector3.Distance(inhibPosition, worldPosEntrance) <= inhib.Range)
			{
				SparkUtil.TrySpark(worldPosEntrance, expose: false);
				Chat.AddExamineMsgFromServer(interaction.Performer, $"The teleporter sparks as it attempts to open a portal. An unknown force appears to be interfering with the device!");
				return;
			}
}
```

It doesnt stop teleport actions (as that would break other systems, potentially messing with aGhosts. It instead just inhibits the ability for players to create portals. Hand teles and the like will spark instead of creating portals. Though this function should also be extended to quantum pads, ill add that to the list when updating maintrooms

FuntimeDudeski:
> > That's a bit curious, considering I don't recall anything in the teleportation code that checks for anything like this.
> 
> ```cs
> if (ignoreInhibitors == false)
> {
> 	foreach(TeleportInhibitor inhib in TeleportInhibitor.Inhibitors)
> 	{
> 			var inhibPosition = inhib.GetComponent<UniversalObjectPhysics>().OfficialPosition.RoundToInt();
> 			if(Vector3.Distance(inhibPosition, worldPosEntrance) <= inhib.Range)
> 			{
> 				SparkUtil.TrySpark(worldPosEntrance, expose: false);
> 				Chat.AddExamineMsgFromServer(interaction.Performer, $"The teleporter sparks as it attempts to open a portal. An unknown force appears to be interfering with the device!");
> 				return;
> 			}
> }
> ```
> 
> It doesnt stop teleport actions (as that would break other systems, potentially messing with aGhosts. It instead just inhibits the ability for players to create portals. Hand teles and the like will spark instead of creating portals. Though this function should also be extended to quantum pads, ill add that to the list when updating maintrooms

if it's modified to inhibit quantum pads w/o an option to toggle, it would break valley bigly and mess with caldera a tiny bit (both use qpads for fixed map routes and teleport inhibitors to prevent players from jumping over adminwalls). i also have no problem with players building and positioning their own quantum pads in my awaysites. it's really just any kind of random teleportation effect that could be used to sequence break that i use this for currently.

PR #10762, opened 2025-03-19
Fix ore machine UI not updating correctly

MaxIsJoe:
CL: [Fix] Fixed an issue with ore machines not updating their UI correctly.


MaxIsJoe:
This doesn't solve the full underlying issue with dynamic lists in NetUI being broken on random UIs.
A deeper bug needs to be found and squashed.

PR #10781, opened 2025-03-31
Remap Fallstation sci

AutumnNova:
Remaps fallstation sci to make it a bit modern and also add ordnance 
Hopefully I havent forgot anything

### Changelog:
CL: [Improvement] Remap fallstation's science department and add ordnance

### DIFF
--- a/UnityProject/Assets/StreamingAssets/Maps/MainStations/FallStation.json
+++ b/UnityProject/Assets/StreamingAssets/Maps/MainStations/FallStation.json
@@ -6005,7 +6005,6 @@
               {
                 "Tel": "FluidStraightPipe",
                 "Z": 1,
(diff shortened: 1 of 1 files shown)


MaxIsJoe:
image

I'd fill this area with something practical like how I filled medbay on that same map so it's not just a giant empty hallway. Even a single table with some items on it can make it less bland.

MaxIsJoe:
Reference to what I mean: #10127

AutumnNova:
> I'd fill this area with something practical like how I filled medbay on that same map so it's not just a giant empty hallway. Even a single table with some items on it can make it less bland.

MaxIsJoe:
yup, this works

MaxIsJoe (approved):

PR #10806, opened 2025-04-12
Checks if any parts of the player is null before attempting to gasp

MaxIsJoe:
CL: [Fix] Fix a weird bug where some aspects of gasp and respiratory checks would pass in null values to the emote manager.
CL: [Fix] Fixed an issue where some emotes should have not publicly displayed their failure text.


Bod9001:
did you fix the infinite loop?

Bod9001 (approved):

PR #10807, opened 2025-04-13
Show serverinfo on escape key

MaxIsJoe:
CL: [Fix] Brought back the ability to view server info when pressing the escape key


Bod9001:
Does itclose  it if you press escape again?

Bod9001 (approved):

PR #10825, opened 2025-04-27
Reworked Lobby Screen - Part 1

MaxIsJoe:
image

This needs a bit of an explanation as to **why** this specific bit of UI needed a rework, and why this was tackled first before more problematic pieces of UI like the character creator.

For the past few months, we've had several issues related to players getting stuck on the lobby screen for various different reasons. Of course, when the problem is first apparent on the lobby screen, everyone rushes to the code that controls the lobby to diagnose what the issue might be; but the code for the UI is so bad, that we often spend more time than necessary debugging issues that aren't even relevant to the main issue that's causing the players to get stuck in the first place.

The old design was also very limiting, as it lacked the flexibility to be expanded upon by both us and future fork maintainers. I've been toying with the idea of creating a small "spin off" game-mode for a while, and the lobby screen has been always an annoying challenge to deal with.

The current iteration of this PR intends to do the following:
- Divide the lobby screen into different sections and components that are all responsible for their own behaviors.
- Eliminate the page system entirely, as it was inflexible and made code harder to maintain.
- Pave the way for forks and unique game modes to have finer control over the lobby screen via code, without worries about hardcoded systems or future merge conflicts.
- Allow for an artwork section to be added on the lobby screen.

This rework also fixes a lot of weird quirks that the lobby screen has had for the longest time now, such as players getting stuck on the 00:00 screen until they hit the ready button.

Overall, I think this screen is in a better shape right now, and still offers a 1 to 1 functionality that is to be expected from the old screen. However, I'd like to further refine this screen in a later date; but this is enough Unity UI for now, I don't want to go bald just yet.

### Changelog:

CL: [Improvement] The lobby screen has been reworked both visually and in terms of functionality to make it more robust and flexible.


MaxIsJoe:
tested and works, and since there seems to be no comments on this; merged.

PR #10833, opened 2025-05-01
offset evil asteroids

FuntimeDudeski:
### Purpose
adds some offset to asteroids at 0,0 to fix the object spawn issue

### Notes:
n/a

### Changelog:
n/a


Bod9001:
well this is more of an issue with how they spawn initially needs to be redone

FuntimeDudeski:
fair enough. but if starshipstation is merged before that's fixed, this should be merged as well to prevent the bug in the meantime.

Bod9001 (approved):

PR #10846, opened 2025-05-07
Staging Fixes & Changes

AtnerNT:
### Changelog:
CL: [Balance] With more powerful explosives more readily available, the yield requirements for several ordnance bounties have been increased to reflect the range of modern explosives.
CL: [Balance] Azide has had its explosive yield nerfed from 6 to 4.
CL: [Fix] Fixed the ability to remove all organs from a simplebot, leaving an empty husk.
CL: [Fix] Fixed simplebots still being able to perform their tasks without a body.
CL: [Fix] Fixed wrenching being blocked by mess decals, vents and other passable objects.


AtnerNT:
Should mention as construction and wrenching shares the same script, this also fixes related construction issues. Most notably with open blast doors and shutters, which previously would prevent you from being able to secure things on the same tile (such as windoors, doors etc)

Bod9001 (approved):

PR #10862, opened 2025-05-13
chore: more debugging

corp-0:
fart shit piss


MaxIsJoe:
did the tests pass on this?

corp-0:
Nah.

Issue is unity wants all credentials to log in (apparently), so we have email and password on secrets but pull requests from forks can't access secrets because that'd be dangerous

PR #10863, opened 2025-05-14
Attempt Number 2 for updating

MaxIsJoe:
### Purpose
 <br>


corp-0:
that's what they told you on gameci discord? Doesn't seem like it should fix anything

MaxIsJoe:
> that's what they told you on gameci discord? Doesn't seem like it should fix anything

they didn't say anything, I just randomly tested something and it worked

tests run now so yay! but now it has some weird perms issue

PR #10907, opened 2025-06-04
signs, matrix item updates

FuntimeDudeski:
### Purpose
updated the directional dept. sign graphics for a bit more consistency, added arrivals

MappingLightSource - <s>set a default light sprite and brightness level, added a gameobject for a mapping helper (left blank so there isn't a big sprite getting in the way of clicking things by default)</s>, removed lavaglowv2 from catalogue

set default volume on periodicgassetter from 0 to 2.5

RuinAutonomousPlatform - changed gun on turrets from energy carbine to energy pistol
[ed: and changed energy pistol's kill mode bullet type to Bullet_LaserWeak which was apparently unused. decreased energy usage per shot from 1000 to 800 which gives the energy pistol 6 lethal shots instead of 5 (its small power cells hold 5000 watts)]

### Notes:
n/a

### Media Preview:
image

### Changelog:
n/a


FuntimeDudeski:
the mapping light source change may be causing problems with mapped lights. investigating...

FuntimeDudeski:
reverting that change for now. good to go!

MaxIsJoe (approved):
merge once tests pass

PR #10527, opened 2024-09-15
Refactors player actions and Adds Spell Cards

wraith-54321:
### Purpose
Currently game actions are hard linked to player clients, this is not great for a variety of reasons but the largest of which being it means AIs cant use them, the goal of this is to move action control/access to Mind as well as add action source control for easy ownership changing.

Also old yeller's xeno code because they dont work at all with the new system and refactoring them(which they needed anyway) is out of scope.

### Notes:
DNM, STILL HAS DEBUGGING CODE

### Changelog:

CL: [New] Wizards can now buy the Spell Cards spell

PR #10931, opened 2025-06-28
ActionsV2

MaxIsJoe:
Supersedes a part of #10527

I wanted to finish my work on arachnid invasion, and didn't want to wait for others to finish reworking the actions system.

This doesn't replace the old system fully (yet), because there's still a lot of spaghetti that needs to be untangled so we can port the old actions to the new system. However, to ensure that people are **100% AWARE** that the V1 system is getting **PHASED OUT**, every v1 action now has a giant **OBSOLETE** marker on their buttons and their tooltips.

What does V2 even offer that's new compared to the old system? Absolutely nothing. In fact, it's exactly the same system as before, but cleaned up for a better dev experience. There are probably a lot of bugs fixed now with this as well, since actions are now stored and checked in a more elegant way. The only real new "feature" we have currently with this new approach is having the ability to give ghosts actions; which can be great for adding some gameplay elements to round observers.

I still have some gripes with some things that are being handled currently, but to avoid over-engineering this system and being a perfectionist, I think everything is *good* as it is right now.

There is a doc that briefly explains how everything works, so read up on it when you want.

# Changelog

CL: [New] We're moving to a new system for HUD actions that's more robust than the old one. The old system will be slowly phased out (just like V1 mobs) as we slowly port everything to it.
CL: [Improvement] Updated Action tooltips to use TMP_Text, which will provide a more clear and crispier reading experience.


UnityProject/Assets/Scripts/Actions/V2/ActionManager.cs
```diff
@@ -0,0 +1,321 @@
...
+				case ActionTriggerType.Both:
+				default:
+					ServerRemoveAction(data.ID);
+					ClientRemoveAction(data.ID);
+					break;
+			}
+		}
+
+		[Command]
+		public void CmdTriggerAction(string actionId, Vector2 mouseLocation)
```
Bod9001, line 173: Should be probably a  net message or try catch

Bod9001: Wait a minute who owns the manager since only owners can trigger commands?

UnityProject/Assets/Scripts/Core/SecureStuff/SerializedAction.cs
```diff
@@ -0,0 +1,174 @@
...
+			if (CachedVector3IntAction == null) Cache<Vector3Int>();
+			CachedVector3IntAction?.Invoke(param);
+		}
+
+		private void Cache()
+		{
+			if (target == null || string.IsNullOrEmpty(methodName))
+				return;
+
+			var method = target.GetType().GetMethod(methodName, BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic);
```
Bod9001: Maybe remove non-public?

UnityProject/Assets/Scripts/Items/SuitSensor.cs
```diff
@@ -184,24 +184,21 @@ private IDCard GetIdentification()
 			return null;
 		}
 
-
-		private void SwitchMode(PlayerScript player)
+		public void SwitchMode(Vector2 mousePosition)
```
Bod9001, line 187: this is not used ServerPerformInteraction(HandActivate interaction)

UnityProject/Assets/Scripts/Actions/V2/ActionManager.cs
MaxIsJoe: Ghost and Player_V4 owns them

also if you want net messages, use the `ClientOnly` trigger and handle them yourself.

UnityProject/Assets/Scripts/Core/SecureStuff/SerializedAction.cs
MaxIsJoe: why so?

UnityProject/Assets/Scripts/Items/SuitSensor.cs
MaxIsJoe: It's used by the action

UnityProject/Assets/Scripts/Actions/V2/ActionManager.cs
Bod9001: cool cool

UnityProject/Assets/Scripts/Core/SecureStuff/SerializedAction.cs
Bod9001: Might be insecure function that can be accessed by private? low access the code I've already done for the event rooter

UnityProject/Assets/Scripts/Items/SuitSensor.cs
Bod9001: ok cool

UnityProject/Assets/Scripts/Core/SecureStuff/SerializedAction.cs
MaxIsJoe: hmm, UnityEvent does allow private methods; but sure.

Bod9001:
```
	if (methodInfo.IsStatic) return false;
					if (methodInfo.IsPrivate) return false;
					```

Bod9001: use ValidateMethodInfo to check the method

PR #10932, opened 2025-06-28
[Tool] - Code Scan editor

MaxIsJoe:
image


Bod9001:
how to run?

MaxIsJoe:
> how to run?

Get Godot mono: https://godotengine.org/download/windows/ (4.4)

Then run it from the editor.

Bod9001:
so, is not the best you have to download something but Looks ookay if you're willing to maintain it

MaxIsJoe:
> so, is not the best you have to download something but Looks ookay if you're willing to maintain it

you have to download python to run some of our other tools

MaxIsJoe:
This is ready

corp-0:
this should go on their own repo as well. It would be best if we had the source code on its own repo and create binaries on releases so people don't need godot editor and they simply run the program

PR #10941, opened 2025-07-04
Hotfixes for 25070416

AtnerNT:
Draft while I do a local build an actually test it all

### Changelog:
CL: [Fix] Fixed being unable to remove beakers/dishes from viro machines
CL: [Fix] Fixed Jail button being a second kick button


AtnerNT:
Okay tested with two clients on local build, this should fix

PR #10961, opened 2025-07-28
Adds campfires

MaxIsJoe:
image

### Changelog

CL: [New] Added craftable campfires. You can craft one using 4 wooden logs.
CL: [Improvement] Objects that are on fire will produce 2 smoke moles per fire rate tick instead of 6.
CL: [Improvement] Increased fire tick rate to 2.25 seconds for flammable objects.
CL: [Fix] Fixed an NRE that would block crafting recipes from being accurately processed.


Bod9001:
Tests are failing

PR #10968, opened 2025-08-14
Some SpriteSO sterilization fixes

MaxIsJoe:
CL: [Fix] Fixed an issue where textures could not networked when an object has just been spawned.

### DIFF
--- a/UnityProject/Assets/Scripts/Actions/ActionData.cs
+++ b/UnityProject/Assets/Scripts/Actions/ActionData.cs
@@ -100,7 +100,7 @@ public class ActionData : ScriptableObject
 			Type = ActionType == UIActionType.Toggle ? Actions.V2.ActionType.Activated : Actions.V2.ActionType.Trigger,
 			AnimatedIconCatalogue = Sprites,
 			HasCustomCursor = HasCustomCursor,
-			CursorTexture = CursorTexture,
+			CursorTexture = null,
 			HasCustomCursorOffset = HasCustomCursorOffset,
 			OffsetType = OffsetType,
 			CursorOffset = CursorOffset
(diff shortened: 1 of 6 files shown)


UnityProject/Assets/Scripts/Actions/V2/ActionButtonData.cs
```diff
@@ -34,7 +33,7 @@ public class ActionButtonData
 		public bool HasCustomCursor;
 		public CursorOffsetType OffsetType;
 		public Vector2 CursorOffset;
-		public Texture2D CursorTexture;
+		public SpriteDataSO CursorTexture;
```
Bod9001, line 36: You need to update all the SOs that use a custom texture, I think stuff like Fire built all that other stuff does, If you reverted back to the texture rider will tell you what overrides it, So you can make a note what to update

MaxIsJoe: this isn't an SO

MaxIsJoe: also all components that use this have been updated

UnityProject/Assets/Scripts/Actions/ActionData.cs
```diff
@@ -100,7 +100,7 @@ public ActionButtonData ToActionButtonData()
 			Type = ActionType == UIActionType.Toggle ? Actions.V2.ActionType.Activated : Actions.V2.ActionType.Trigger,
 			AnimatedIconCatalogue = Sprites,
 			HasCustomCursor = HasCustomCursor,
-			CursorTexture = CursorTexture,
```
Bod9001, line 103: ok, Technically this is a lazy way of doing it, you're supposed update CursorTexture in the SOs too, that's what I meant 
image
There's a bunch of overrides like this

MaxIsJoe: Bod
This doesn't touch SOs at all
The `ToActionButtonData()` function is a function to convert old v1 stuff to v2 stuff, but its not being used anywhere at the moment.

Bod9001: ok cool

PR #10984, opened 2025-10-05
Drop shadow Shader

Bod9001:
### Changelog:
CL: [New] Add drop shadow to items and objects can be toggled in the settings
CL: [Improvement] Machines like hydroponics, lath, XO Suit fabricator and Department lathes now get improvements with part Upgrades
CL: [Fix] Fixes Stomacchs apppearing on cyborgs surgery stepss


MaxIsJoe:
This will need to be a toggle in the settings once its done btw

PR #10989, opened 2025-10-15
Staging fixxing2

Bod9001:
### Changelog:
CL: [New] added portable heaters
CL: [New] added Rapid Part Exchange Device used for upgrading parts in machines can be made in lathe
CL: [Improvement] Department batteries and SMES Now use their internal battteries for charge
CL: [Improvement] Makes laser line from laser puzzle colour more visible
CL: [Balance] slightly reduces pest and weed growth on Hydro tray
CL: [Fix] Removes eye shadow ( the shadow effect on eyes )
CL: [Fix] Fixed not be able to empty bags out on to tables
CL: [Fix] Fixed machine crafting
CL: [Fix] Fixed SMES and department batteries having their charge reset by disassembling and reassembling


AtnerNT:
How does the laser line colours look with your changes? I would prefer the visuals try be maintained just with a stronger light colour.

Do the new capacities of SMES's and DP's match their old ones? otherwise power could be massively thrown off

AtnerNT:
> How does the laser line colours look with your changes? I would prefer the visuals try be maintained just with a stronger light colour.

I see you are chaging the top sprite colour too.
Please DO NOT do this. This is just a regression in the visuals, the lasers are supposed to be a white core and emit coloured light. If you want to make them more saturated, increase the strength ogf the light sprite, NOT the top sprite

MaxIsJoe:
fix the changelog text

UnityProject/Assets/Scripts/Objects/Machines/FoodProcessor.cs
```diff
@@ -73,7 +74,6 @@ private void Awake()
 		{
 			registerTile = GetComponent<RegisterTile>();
 			spriteHandler = GetComponentInChildren<SpriteHandler>();
-			storage = GetComponent<ItemStorage>();
```
MaxIsJoe, line 76: We need to figure out some sort of tagging system for cases where there might be multiple components of the same type on a single object.
Because this feels like a workaround, and will cause problems in the future when Unity randomly decides to destroy references.

UnityProject/Assets/Scripts/Systems/Botany/GrownFood.cs
```diff
@@ -113,7 +113,7 @@ public void Start()
 		{
 			if (reagentContainer.ReagentMixTotal == 0)
 			{
-				SetUpFood(plantData, PlantTrayModification.None, null);
+				SetUpFood( SeedPacket.GetComponent<SeedPacket>().plantData, PlantTrayModification.None, null);
```
MaxIsJoe, line 120: Shouldn't this be null checked just incase?

UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
```diff
@@ -998,30 +961,6 @@ public void CmdFilledDynamicInput(GameObject forGameObject, string input)
 		}
 	}
 
-	[Command]
-	public void CmdTriggerStorageTrap(GameObject storage)
```
MaxIsJoe, line 1002: Where is this implemented now? I don't see any trap changes except ones where you remove stuff related to it?

MaxIsJoe (changes requested):

UnityProject/Assets/Scripts/Objects/Machines/FoodProcessor.cs
AtnerNT: In general though, isn't better to serialize component references in this case? I dont see why GetComponent should be used here

UnityProject/Assets/Scripts/Player/PlayerNetworkActions.cs
AtnerNT: InteractableStorage.cs?

UnityProject/Assets/Scripts/Objects/Machines/FoodProcessor.cs
MaxIsJoe: > Unity randomly decides to destroy references.

That's why.

Bod9001:
"Do the new capacities of SMES's and DP's match their old ones? otherwise power could be massively thrown off", The capacity is roughly the same little bit less though
18,000,000 -> 15,000,000
to Line up with having 15 batteries inside

My only concern with the battery power changes is charging eguns using loads of power

UnityProject/Assets/Scripts/Objects/Machines/FoodProcessor.cs
Bod9001: Haven't checked In terms of performance but one benefit is it's better for non-coders to interact with

Bod9001:
"I see you are chaging the top sprite colour too.
Please DO NOT do this. This is just a regression in the visuals, the lasers are supposed to be a white core and emit coloured light. If you want to make them more saturated, increase the strength ogf the light sprite, NOT the top sprite"

You can't see any colour though in a lit room they might as well just be white, 
Commonly coloured Lasers don't have a white core

MaxIsJoe (approved):
approved.
but gonna leave it up incase atner has something to say

PR #10999, opened 2025-10-24
Staging fixing

Bod9001:
### Changelog:

CL: [Fix] Fixes hugs being visible through walls
CL: [Fix] Fixes Broken construction with plant DNA manipulator
CL: [Fix] Fixes drinks dispenser not working properly

### DIFF
--- a/UnityProject/Assets/Scripts/Clothing/BackPack/InteractableStorage.cs
+++ b/UnityProject/Assets/Scripts/Clothing/BackPack/InteractableStorage.cs
@@ -231,6 +231,9 @@ public class InteractableStorage : NetworkBehaviour,
 		switch (pickupMode)
 		{
 			case PickupMode.DropClick:
+			case PickupMode.All:
+			case PickupMode.Same:
+			case PickupMode.Single:
 				if (canQuickEmpty)
 				{
 					var slots = itemStorage.GetItemSlots();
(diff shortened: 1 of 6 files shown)


UnityProject/Assets/Scripts/Clothing/BackPack/InteractableStorage.cs
```diff
@@ -231,6 +231,9 @@ public void ServerPerformInteraction(HandActivate interaction)
 		switch (pickupMode)
 		{
 			case PickupMode.DropClick:
+			case PickupMode.All:
+			case PickupMode.Same:
+			case PickupMode.Single:
```
AtnerNT, line 236: Are these all the PickUpModes? (Or like all of them)?
Because if so this switch statement is basically serves no purpose. Perhaps this would be better as an if(pickupMode != PickUpMode.None)

UnityProject/Assets/Scripts/Clothing/BackPack/InteractableStorage.cs
```diff
@@ -305,8 +307,6 @@ public bool WillInteract(InventoryApply interaction, NetworkSide side)
 
 		if (Cooldowns.IsOn(interaction, cooldown, side)) return false;
 
-
-
 		if (IsFull(interaction.UsedObject, interaction.Performer))
 		{
 			if (Cooldowns.TryStart(interaction, cooldown, side) == false) return false;
```
AtnerNT, line 312: if(condition) return false;
return false;

what is the point of this?

Bod9001: it doesn't really matter too much about the duplication

UnityProject/Assets/Scripts/Clothing/BackPack/InteractableStorage.cs
Bod9001: well people might want to extend the behaviour in the future

PR #11000, opened 2025-10-26
Doorfix

Bobtheapple:
### Purpose
Fixes a couple of bugs from my recent door changes related to doors that spawn on the server open.  Adjusts how girder works to wait for the girder to despawn before spawning the object that is being built.

### Notes:
The door code is full of a lot of fiddly older code, a lot of redundant checks and some orphaned booleans that aren't actually doing anything. I am working on a refactor to make it a bit more intuitive, should be done in a few days if life doesn't get in the way.  This included some of that because it was related to the issue, I removed the non-essential checks from Open() and Close(), those functions will now close/open the door no matter what, the logic for firelocks and whether the door is already opened and what not is now solely on the TryOpen and TryClosed functions, so use those if you want to check for those kind of things, use close/open if you just want the server to close the darned thing.

Continue to @ me on discord if anything spots anything related to doors.

### DIFF
--- a/UnityProject/Assets/Scripts/Objects/Construction/Girder.cs
+++ b/UnityProject/Assets/Scripts/Objects/Construction/Girder.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System.Collections;
+using System.Collections.Generic;
 using Core;
 using UnityEngine;
 using Mirror;
@@ -206,31 +207,42 @@ namespace Objects.Construction
 		[Server]
 		private void ReinforceGirder(HandApply interaction)
 		{
-			interaction.HandObject.GetComponent<Stackable>().ServerConsume(1);
-			Spawn.ServerPrefab(reinforcedGirder, SpawnDestination.At(gameObject));
-			_ = Despawn.ServerSingle(gameObject);
+			StartCoroutine(SyncSpawn(interaction, reinforcedGirder, 1));
 		}
 
 		[Server]
 		private void ConstructFalseWall(HandApply interaction)
 		{
-			GameObject theWall = Spawn.ServerPrefab(FalseWall, SpawnDestination.At(gameObject)).GameObject;
-			DoorMasterController doorController = theWall.GetComponent<DoorMasterController>();
-			tileChangeManager.MetaTileMap.SetTile(registerObject.LocalPositionServer, falseTile);
-			interaction.HandObject.GetComponent<Stackable>().ServerConsume(2);
-			_ = Despawn.ServerSingle(gameObject);
-			doorController.TryClose();
+			StartCoroutine(SyncSpawn(interaction, FalseWall, 2));
 		}
 
 		[Server]
 		private void ConstructReinforcedFalseWall(HandApply interaction)
 		{
-			GameObject theWall = Spawn.ServerPrefab(FalseReinforcedWall, SpawnDestination.At(gameObject)).GameObject;
-			DoorMasterController doorController = theWall.GetComponent<DoorMasterController>();
-			tileChangeManager.MetaTileMap.SetTile(registerObject.LocalPositionServer, falseTile);
-			interaction.HandObject.GetComponent<Stackable>().ServerConsume(2);
-			_ = Despawn.ServerSingle(gameObject);
-			doorController.TryClose();
+			StartCoroutine(SyncSpawn(interaction, FalseReinforcedWall, 2));
 		}
+
+		/// <summary>
+		/// Waits for the girder to despawn before spawning the desired object
+		/// </summary>
+		/// <param name="interaction">The HandApply action that doing the constructing</param>
+		/// <param name="toSpawn">What we are constructing</param>
+		/// <param name="toConsume">How many resources to consume from the active hand slot</param>
+		IEnumerator SyncSpawn(HandApply interaction, GameObject toSpawn, int toConsume)
+		{
+			interaction.HandObject.GetComponent<Stackable>().ServerConsume(toConsume);
+			SpawnDestination destination = SpawnDestination.At(gameObject);
+			yield return _ = Despawn.ServerSingle(gameObject);
+			GameObject spawned = Spawn.ServerPrefab(toSpawn, destination).GameObject;
+			
+			//Some extra logic to get that Open -> Close animation for the FalseWalls spawning
+			//We want this rather than spawning the door closed so the builder can see they did it right
+			if(toSpawn == FalseReinforcedWall || toSpawn == FalseWall)
+			{
+				DoorMasterController doorController = spawned.GetComponent<DoorMasterController>();
+                tileChangeManager.MetaTileMap.SetTile(registerObject.LocalPositionServer, falseTile);
+				doorController.Close();
+            }
+        }
 	}
 }
--- a/UnityProject/Assets/Scripts/Objects/Construction/Girder.cs
+++ b/UnityProject/Assets/Scripts/Objects/Construction/Girder.cs
@@ -207,42 +207,31 @@ namespace Objects.Construction
 		[Server]
 		private void ReinforceGirder(HandApply interaction)
 		{
-			StartCoroutine(SyncSpawn(interaction, reinforcedGirder, 1));
+			interaction.HandObject.GetComponent<Stackable>().ServerConsume(1);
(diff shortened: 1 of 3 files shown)


UnityProject/Assets/Scripts/Objects/Doors/DoorMasterController.cs
```diff
@@ -522,15 +513,7 @@ public void Close(bool byForce = false)
 
 		public void Open(bool byForce = false)
 		{
-			if (IsClosed == false) return;
-
-			if (isFireLock == false)
```
MaxIsJoe, line 527: Can you test and make sure that firelocks aren't broken from this change?

You can also test if it's working on multiplayer by using this: https://unitystation.github.io/unitystation/development/Multiplayer-PlayMode/

UnityProject/Assets/Scripts/Objects/Construction/Girder.cs
```diff
@@ -206,31 +207,42 @@ private void ConstructWall(HandApply interaction)
...
+			GameObject spawned = Spawn.ServerPrefab(toSpawn, destination).GameObject;
+			
+			//Some extra logic to get that Open -> Close animation for the FalseWalls spawning
+			//We want this rather than spawning the door closed so the builder can see they did it right
+			if(toSpawn == FalseReinforcedWall || toSpawn == FalseWall)
+			{
+				DoorMasterController doorController = spawned.GetComponent<DoorMasterController>();
+                tileChangeManager.MetaTileMap.SetTile(registerObject.LocalPositionServer, falseTile);
+				doorController.Close();
+            }
```
AtnerNT: I have a few concerns with all this.
This script is on the girder, it will get disabled when the girder gets despawned, yet you are still trying to use other methods/properties of the gameObject.

Trying the get the Girder's registerObject AFTER you've destroyed it is going to cause issues.

AtnerNT: Also coroutines only work on active gameobjects so it will break from that aswell

AtnerNT: If your intention is to just remove the visual overlap for a frame.
Instead of destroying the girder and then still trying to use it to spawn something.
Try ForceClearing the griders spriteHandler.
There should be a method like SpriteHandler.ForceClear(); (or similliar) that will allow you to hide it visually before destroying it.

UnityProject/Assets/Scripts/Objects/Doors/DoorMasterController.cs
```diff
@@ -385,6 +379,7 @@ public bool CheckStatusAllow(HashSet<DoorProcessingStates> states)
 
 		public void TryOpen(GameObject originator, bool blockClosing = false)
 		{
+			if (!IsClosed) return; //Can't open if we are open. Figures.
```
AtnerNT, line 379: Just for Unitystation style guidelines, we don't use ! to invert bools in conditions as they can be easy to miss.
Please replace this with if(IsClosed == false) return;

UnityProject/Assets/Scripts/Objects/Doors/DoorMasterController.cs
```diff
@@ -522,15 +513,7 @@ public void Close(bool byForce = false)
...
-			if (IsClosed == false) return;
-
-			if (isFireLock == false)
-			{
-				var fireLock = matrix.GetFirst<FireLock>(registerTile.LocalPositionServer, true);
-				if (fireLock != null && fireLock.fireAlarm.activated && fireLock.DoorMasterController.IsClosed) return;
-			}
-
-			if (!this || !gameObject) return; // probably destroyed by a shuttle crash
+			if (!gameObject) return;  // probably destroyed by a shuttle crash
```
AtnerNT, line 513: same thing here for inverting bools

UnityProject/Assets/Scripts/Objects/Doors/DoorMasterController.cs
```diff
@@ -174,7 +174,14 @@ private void Awake()
 				openSortingLayer = SortingLayer.NameToID("Machines");
 			}
 
-			doorAnimator.SyncDoorStatus(doorAnimator.SyncDoorUpdateType, DoorAnimatorV2.DoorUpdateType.Close);
+			if (IsClosed)
```
AtnerNT: This if else could be compressed to one line, or at the least 2 lines. Not a big deal, up to you which you prefer

UnityProject/Assets/Scripts/Objects/Doors/DoorMasterController.cs
```diff
@@ -183,18 +190,6 @@ public void OnSpawnServer(SpawnInfo info)
 			HackingProcessBase.RegisterPort(TryBump, this.GetType());
 			HackingProcessBase.RegisterPort(TryClose, this.GetType());
 			HackingProcessBase.RegisterPort(ConfirmAIConnection, this.GetType());
-			if (CustomNetworkManager.IsServer)
```
AtnerNT, line 186: You removed the server side test... why? have you tested to make sure this is okay?

UnityProject/Assets/Scripts/Objects/Construction/Girder.cs
```diff
@@ -206,31 +207,42 @@ private void ConstructWall(HandApply interaction)
 		[Server]
 		private void ReinforceGirder(HandApply interaction)
 		{
-			interaction.HandObject.GetComponent<Stackable>().ServerConsume(1);
-			Spawn.ServerPrefab(reinforcedGirder, SpawnDestination.At(gameObject));
-			_ = Despawn.ServerSingle(gameObject);
+			StartCoroutine(SyncSpawn(interaction, reinforcedGirder, 1));
```
AtnerNT: Imma beat Max to this and say, we have the UniTask package in our project, its adivsed to use it over Coroutines

AtnerNT (changes requested):

UnityProject/Assets/Scripts/Objects/Doors/DoorMasterController.cs
Bobtheapple: No other code references the Close() function directly, and the firelock check was already in the TryClose() function, so the code was checking the firelocks status twice.  I did go and test it to be sure, and they appear to be working fine.  

I will note that currently the firelocks don't try to close doors when the come down and prevent doors from closing, so if a door is open and the fire alarm is pulled the door is stuck open under the firelock.  I tested a version of the game before I made any changes and it was like that before as well.  I could adjust that.

UnityProject/Assets/Scripts/Objects/Construction/Girder.cs
Bobtheapple: You are absolutely right.  The code was working for my testing, but if I tested putting a longer wait in and it did break, so there was definitely some potential for issues there.  I've reverted the code for now, with the adjustment of calling the door to close before despawning the gameobject to be extra sure.  I've tested it and the fix still works, so this wasn't necessary to fix the bug I introduced.  Its something to consider fixing later.

UnityProject/Assets/Scripts/Objects/Doors/DoorMasterController.cs
Bobtheapple: I just moved this check from Open() to TryOpen(), the door code was already full of this style.  I've gone through and change all these checks to the preferred style now.

UnityProject/Assets/Scripts/Objects/Doors/DoorMasterController.cs
Bobtheapple: I'm fairly certain it was okay, but it also doesn't hurt anything either, so I've put it back

UnityProject/Assets/Scripts/Objects/Construction/Girder.cs
Bobtheapple: I've reverted the code to not use a coroutine now, I'll look into UniTask for the door code refactor and see if I can remove all the coroutines that were already there.

UnityProject/Assets/Scripts/Objects/Doors/DoorMasterController.cs
Bobtheapple: Adjusted!

Bod9001 (approved):

PR #11007, opened 2025-11-03
Door refactor

Bobtheapple:
### Purpose
This is a major refactor of the DoorMasterController and related code in an effort to make it more legible and to get rid of some of the remaining bugs

### Notes:
This covers a lot of little stuff all over the place, most of it is pretty behind-the-scenes:

1. The biggest change is that the wide range of open/close methods in DoorMasterController have been cut down to six: Open and Close, which handles the opening/closing and shouldn't be called directly except for serverside functionality, TryForceOpen and TryForceClosed, for crowbaring/prying, and PulseTryOpen and PulseTryClose, for opening and closing the door normally

2. All the coroutines on the doors have been replaced with UniTaskVoids.  Since opening an airlock would start a minimum of 3 coroutines (the access animation, the door animation, and the autoclose timer) this may have some positive impact on server load.

3. Some of the Access Remote checks were moved onto DoorMasterController as per a //FIXME comment.

4. The hacking panel stuff that was on DoorMasterController was consolidated on ConstructibleDoor

5. The masking layer and sprite layer code that was on DoorMasterController was moved to the DoorAnimatorV2

6. The AI connection cable in the hacking interface was fixed per a //FIXME comment.

7. The AI player chat messages were being handled in two places, on DoorMasterController and in GUI_Airlock, with slightly different feedback messages.  I moved this all to DoorMasterController, and moved the logic for updating the GUI to GUI_Airlock.  I also fixed the color change on the electrify button per a //FIXME note.  I also added a lot of chat notifications for the AI player to help them understand why the doors aren't working correctly.

8. FireAlarm was tweaked, its coroutine is now a UniTaskVoid and it handles toggling firelocks slightly differently.  Firelocks were also changed, this fixed a subtle bug where the firelock stopped the door from closing while down, it also makes it so instead of checking the tile every time a player interacts with a door to see if there is a firelock that's down, the firelock just tells the door when its down, and while its down the door checks to see if the firelock is still there or it got destroyed.

9. I changed how CanCloseDoorAt() worked because calling Matrix was causing some issues.  I removed the function from Matrix and now its part of DoorMasterController, this code was only being used by DoorMasterController anyway.

10. I fixed a bug I introduced earlier related to GenericTriggerOutputs, I had changed Open()/Closed() to accept a boolean which made them no longer be actions, the byForce boolean is now handled in a way that doesn't interfere with GenericTriggerOutput

### Changelog:
CL: [New] The "AI Connection" cable for doors has been implemented, you can now cut off AI player's door access by hacking.
CL: [New] The "Door Access" cable for doors has been implemented, if cut the door will deny all access.
CL: [Improvement] The hacking interface is more colorblind friendly.
CL: [Improvement] Electrifying doors will now change the color of the electrify icon in the AI player's GUI.
CL: [Improvement] AI will now receive more notifications about why doors aren't working right.
CL: [Improvement] Opening airlocks in unusual ways (AI players, switches, NPCs) will now correctly use the airlock's internal wiring.
CL: [Fix] Fixed a bug where doors couldn't close when firelocks were engaged, and now they will close properly when you pull the alarm.
CL: [Fix] Fixed a bug where cutting the "Close" wire on a door would sometimes block the door from autoclosing when the wire was reconnected.
CL: [Fix] Fixed a bug where bolting a door open would disable the door's autoclose after it was unbolted.
CL: [Fix] Fixed a bug where some doors/shutters would have the wrong sprite when players first join a server.

### Still to do:
I only got through to the AI functions on this pass, I figured I had found enough fixes that I should go ahead and release what I've done, especially because Bod9001 pointed out the shutters/pressure plate bug which was an issue.  The multitool section and the admin function section on DoorMasterController are still not reworked.  Door pushing is still not in, I'll add that in the next go around.  After that I plan to go through the modules and take a look at hacking.  And then I should be about done with doors.

### Known Issues:
If you open a door, then bolt it, then unbolt it with the Bolt Toggle Card, the door will close immediately rather than on the autoclose timer.  This is NOT the case for toggling the bolts in any other way.  Its not a problem per se, but it does indicate there is some jank going on somewhere, I'm looking in to it.

If the AI player clicks a door to open its GUI, then a player comes along and opens the hacking interface and cuts the prevent electrocution wire and walks away, the AI's door GUI doesn't register that the door is electrified until they or another player interacts with the door in some way.  A really minor edge case but I'm working on it.

The hacking process is spawning multiple ports with the same color.  This has always been the case, I thought some of the colors were just really similar which is why I swapped them to color blind friendly colors, but no its clear that they are actually just the same.  I'll have this fixed in my next PR.


UnityProject/Assets/Scripts/Objects/Doors/DoorAnimatorV2.cs
```diff
@@ -311,12 +381,12 @@ public void RemovePanelOverlay()
 		}
 
 		/// <summary>
-		/// Used to call coroutines from outside monobehaviors
+		/// Used to call UniTasks from outside monobehaviors
 		/// </summary>
 		/// <param name="anim"></param>
-		public void RequestAnimation(IEnumerator anim)
+		public void RequestAnimation(UniTaskVoid anim)
```
corp-0: technically, since this isn't a coroutine anymore, this pattern shouldn't be needed

Bobtheapple: I was thinking it was saving me from including Cysharp.Threading.Tasks, but apparently its not needed to use .Forget().  Adjusted!

UnityProject/Assets/Scripts/Objects/Doors/DoorAnimatorV2.cs
```diff
@@ -177,9 +240,12 @@ public IEnumerator PlayOpeningAnimation(bool skipAnimation = false, bool panelEx
 			}
 			else
 			{
-				overlayHackingHandler.SetCatalogueIndexSprite((int) Panel.NoPanel, false);
+				overlayHackingHandler.SetCatalogueIndexSprite((int)Panel.NoPanel, false);
 			}
 
+			//Toggle layers after animation to hide firelocks behind door
```
AtnerNT, line 228: Is the intention that you can't see firelocks when they are open? How would one pry them shut?

UnityProject/Assets/Scripts/Objects/Doors/FireLock.cs
```diff
@@ -19,13 +19,38 @@ private void Awake()
...
+			foreach (var door in MatrixManager.GetAt<DoorMasterController>(registerTile.WorldPositionServer, true))
+			{
+				if (door.IsFireLock)
+					continue;
+
+				door.IsFireLockEngaged = false;
+			}
 
-			//registerTile.SetNewSortingOrder(SortingLayer.NameToID("Door Closed"));
+			doorMasterController.PulseTryOpen(bypassSoftware: true);
```
AtnerNT, line 59: Doors shouldn't open when firelocks disengage?

UnityProject/Assets/Scripts/Objects/Doors/Modules/AccessModule.cs
```diff
@@ -53,16 +53,19 @@ public override void ClosedInteraction(HandApply interaction, HashSet<DoorProces
 
 		public override void BumpingInteraction(GameObject byPlayer, HashSet<DoorProcessingStates> States)
 		{
-			if (!master.HasPower || !CheckAccess(byPlayer))
+			if (byPlayer != null)
 			{
-				States.Add(DoorProcessingStates.SoftwarePrevented);
+				if (!master.HasPower || !CheckAccess(byPlayer))
```
AtnerNT: bool thing again here (Know it was already like this, but still good to change).

UnityProject/Assets/Scripts/Objects/Doors/Modules/AccessModule.cs
```diff
@@ -53,16 +53,19 @@ public override void ClosedInteraction(HandApply interaction, HashSet<DoorProces
...
+				{
+					States.Add(DoorProcessingStates.SoftwarePrevented);
+				}
```
AtnerNT: Also all this could be one if instead of being nested. I dont see any behaviour that happens when byPlayer isnt null but the other conditions are false

UnityProject/Assets/Scripts/Objects/Doors/Modules/BoltsModule.cs
```diff
@@ -59,7 +59,7 @@ private void SetBoltsState(bool state)
 				master.DoorAnimator.TurnOffAllLights();
 			}
 
-			master.UpdateGui();
+			master.UpdateGUI();
```
AtnerNT: Codacy WILL complain about this, but codacy is stupid and doesnt know what an acronym is

AtnerNT (changes requested):
Great PR! Just a few minor things

UnityProject/Assets/Scripts/Objects/Doors/DoorAnimatorV2.cs
Bobtheapple: The original code went out of its way to specifically do this, it even had multiple methods, both the line openSortingLayer = SortingLayer.NameToID("Machines"); and a boolean "UseMachinesForOpenLayeer", so I thought this was the intended functionality.  If its not I will happily fix this as doing this layer swapping is messy.

UnityProject/Assets/Scripts/Objects/Doors/FireLock.cs
Bobtheapple: This is the firelock itself that's opening, the code for the door under it is in the foreach statement above it, which just tells the door that the firelock has been disengaged.

UnityProject/Assets/Scripts/Objects/Doors/Modules/AccessModule.cs
Bobtheapple: That's seperated out because DenyAccess will play the "AccessDenied" sound via CheckAccess->ProcessCheckAccess->DenyAccess if the player is null. Yeah, this code desperately needs some cleaning, its not intuiutive at all.

UnityProject/Assets/Scripts/Objects/Doors/Modules/BoltsModule.cs
Bobtheapple: Annoying, I'll change it back

UnityProject/Assets/Scripts/Objects/Doors/Modules/AccessModule.cs
Bobtheapple: Also looking at it this section is bad in general, going to rework this, give me a sec

UnityProject/Assets/Scripts/Objects/Doors/Modules/AccessModule.cs
Bobtheapple: Blarg, adjusted

UnityProject/Assets/Scripts/Objects/Doors/Modules/AccessModule.cs
Bobtheapple: Okay, that should be a lot better.  I haven't gotten to fully refactoring the modules, there is a lot weird in them.

UnityProject/Assets/Scripts/Objects/Doors/FireLock.cs
Bobtheapple: I've added comments to disambiguate what's going on, also added a check for a corner case where the Firelock loses power after it has come down.

UnityProject/Assets/Scripts/Items/Devices/AccessRemote.cs
```diff
@@ -88,57 +88,51 @@ public bool WillInteract(HandApply interaction, NetworkSide side)
...
+				{
+					case AccessRemoteState.Open:
+						TryOpenDoor(doorController, interaction.Performer);
+						break;
+					case AccessRemoteState.Emergency:
+						AccessModule accessModule = interaction.TargetObject.GetComponentInChildren<AccessModule>();
+						accessModule.ToggleAuthorizationBypassState();
+						break;
+					case AccessRemoteState.Bolts:
+						BoltsModule boltsModule = interaction.TargetObject.GetComponentInChildren<BoltsModule>();
```
MaxIsJoe: This needs to be cached. GetComponent is bad for performance.

UnityProject/Assets/Scripts/Items/Devices/AccessRemote.cs
```diff
@@ -88,57 +88,51 @@ public bool WillInteract(HandApply interaction, NetworkSide side)
...
-					TryOpenDoor(doorController, interaction.Performer);
-					break;
-			}
+                switch (currentState)
+				{
+					case AccessRemoteState.Open:
+						TryOpenDoor(doorController, interaction.Performer);
+						break;
+					case AccessRemoteState.Emergency:
+						AccessModule accessModule = interaction.TargetObject.GetComponentInChildren<AccessModule>();
```
MaxIsJoe: This needs to be cached. GetComponent is bad for performance.

MaxIsJoe: The whole point of modules is that not all doors will have it, because each door will want to implement their own behaviors in the future. (Example, Wooden Doors)

MaxIsJoe (changes requested):

UnityProject/Assets/Scripts/Items/Devices/AccessRemote.cs
Bobtheapple: Got it, should be better now

UnityProject/Assets/Scripts/Items/Devices/AccessRemote.cs
Bobtheapple: The door not having an access module is being handling a little bit above this in the code with the CheckRemoteConnectivity() call.  I've added some comments to make that a little clearer.  This PR's focus wasn't the remotes so I wasn't changing too much code, I just saw the //FIXME while I was changing it so it didn't bypass the door's wiring and figured I could move the checks over.  It should be cleaner now, I'll add refactoring the Access Remote to my todo list.

Bobtheapple: Hmm, though this brings up an interesting issue for the future, as the code stands (and it was like this when I got here) all doors, even wooden doors, have some oddities: They all have a HackingProcessBase, which means they have internal wiring even if that wiring can't be accessed because they don't have a ConstructibleDoor module.  That wiring handles 4 things by default: opening, closing, bumping, and probably the most glaring one is AI Connection.  Opening, closing, and bumping are whatever, they are called wires in the code but since nobody can ever see on Wooden Doors them we can think of them as the door's hinges.  But AI connection could potentially be an issue to always have on a door by default.  The AI code could be spun off into its own module to resolve this.

MaxIsJoe: >  The AI code could be spun off into its own module to resolve this.

That would be preferable, yeah.

Bobtheapple:
Added the requested fix for shutters on OutpostStation, toggling the "IsClosed" boolean in mapping rather than having to add the specific open or closed version of the door should work fine now.

UnityProject/Assets/Scripts/Items/Devices/AccessRemote.cs
Bobtheapple: Sound good, I will do this when I do the Door Module refactor PR.

Bod9001 (approved):

PR #11010, opened 2025-11-07
Stun on stun baton thrown

### LINKED ISSUE #5722: Stun batons should have a chance to stun people when thrown.
### Description
In /tg/, thrown stun batons have a 35% chance of stunning people. Obviously, this should only happen if the stun baton is turned on and if successful it should deduct charge from the power cell (see #5720). This would be a nice option for officers and other baton wielders, albeit not an always successful one.

Marongon:
### Purpose
When throwing a stun baton at a player (it must be turned on and have enough charge in it's power cell) there is a 35% chance to stun the player on hit, as was wanted in issue #5722 . The actual percentage was added to MeleeEffect script as a variable, making it possible to change the percentage per weapon thrown.

### Notes:
I used the fact, that there already is a system where a player can be hit in the chest/other bodypart when somebody throws an item that can deal damage and just added a check, if the thrown item contains a MeleeEffect component, if so, then it checks what the chance of being hit is and then tries to apply the effect if the RNG works out. The actual MeleeEffect had to be remade a little bit, because all of the inner workings were in one method that was called by server. I didn't change anything about how it works, just moved the actual doing the effect part into a separate public method, that can be called from other parts of the code. I tested it on build + editor and it worked for me, I hope I didn't ruin anything, but there wasn't a lot of changed code, so I presume not.

### Changelog:
CL: [New] When thrown a stun baton has a 35% chance to stun you, if it is turned on and has sufficient power in it's cell.


UnityProject/Assets/Scripts/Core/Physics/UniversalObjectPhysics.cs
```diff
@@ -1375,6 +1376,16 @@ private void ProcessThingsToHit()
 					livingHealthMasterBase.ApplyDamageToBodyPart(ddamagedBy, damage, AttackType.Melee, DamageType.Brute, currentAim);
 					if (currentAim == BodyPartType.Mouth && TryGetComponent<Edible>(out var edible)) edible.TryConsume(null, hit.gameObject, true);
 
+					if (TryGetComponent<MeleeEffect>(out var meleeEffect))
```
MaxIsJoe: `GetComponent` is not ideal for performance. Try using `CommonComponents`.

MaxIsJoe: An even better thing:
Use events.
`MeleeEffect` should listen for an event triggered by `ProcessThingsToHit()` to do stuff. That way the code is cleaner, and UOP does not need to depend on the existence of that component to function.

UnityProject/Assets/Scripts/Items/Weapons/Melee/MeleeEffect.cs
```diff
@@ -143,6 +146,52 @@ public void ServerPerformInteraction(HandApply interaction)
...
+			RegisterPlayer registerPlayerVictim = target.GetComponent<RegisterPlayer>();
+
+			TryApplyEffect(interaction.Intent, performer, target, dir, registerPlayerVictim, wna, toggleableEffect);
+		}
+
+		#endregion
+
+		private void OnEnable()
+		{
+			UpdateManager.Add(Timer, 1f);
```
MaxIsJoe, line 163: Avoid using the update manager for this.
Use `UniTask` instead. 

As a general reminder, avoid using the `UpdateManager` whenever possible; because we don't want to tax the process loops for every little check we want to make.
https://unitystation.github.io/unitystation/contribution-guides/Development-Standards-Guide/

If you want to know how to use `UniTask`, check:
https://github.com/Cysharp/UniTask?tab=readme-ov-file#getting-started

MaxIsJoe (changes requested):

MaxIsJoe:
Good PR overall, but you can improve it.

UnityProject/Assets/Scripts/Items/Weapons/Melee/MeleeEffect.cs
corp-0: didn't we have a cooldown functionality somewhere?

Marongon: this was marked as my contribution, but i actually didn't change anything in the OnEnable, Timer, OnDisable or Examine .. i just made the CheckBattery and TryApplyEffects

corp-0: ah, I see. Your IDE auto formatter moved it. I think it makes sense to leave it as is, but you can improve that if you want

UnityProject/Assets/Scripts/Core/Physics/UniversalObjectPhysics.cs
MaxIsJoe: We have a class called `SerializedAction` that we use to assign events in our project.
You declare it like this: `public SerializedAction OnHit;`
You can then assign a function to trigger when `OnHit?.Invoke()` is called like this: `physics.OnHit.CachedAction += StunCheck;`

Make sure to unsubscribe it when the object is destroyed, so you don't leave a leak, like this:
`physics.OnHit.CachedAction -= StunCheck;`

Marongon:
### Changelog:
CL: [Fix] Instead of TryGetComponent, a SerializedAction is used to signal the throw to the MeleeEffect class

MaxIsJoe (approved):
LGTM.
It can still be improved further, but this is good as is for now.
Waiting on tests to pass now.

PR #11013, opened 2025-11-07
feat: move secrest from config.json to their own file we can mount on…

corp-0:
… docker

AAAAAAAAAAAAAAAA


MaxIsJoe:
image

Yeah..

PR #11014, opened 2025-11-08
Endless Dry Heaving and PWR Game Soda fix

Bobtheapple:
### Purpose
Fixes a couple of bugs mentioned in Discord regarding how vomiting is working and the PWR Game Soda having mold in it by default

### Notes
Oops, forgot to do this on a different branch than my earlier one, let me know if that will screw with the commits history, I don't know Github that well

### Changelog:
CL: [Fix] Fixed a bug where once a player starts dry heaving, they will never stop.
CL: [Balance] PWR Game Soda is no longer a toxic gamer soda, feel free to drink it without vomiting


corp-0:
the commits will be squashed into a single  one, so it is not a big issue

PR #11016, opened 2025-11-12
Error cleaning

Bod9001:


AutumnNova:
Ill point out some of these "random z offsets" are not actually random and are fully intentional
specifically they are done to get multiple items on a table or rack to not clip horribly and layer well

Bod9001:
There should be an automatic system for the layering of items since It does that in normal gameplay @AutumnNova

PR #11028, opened 2025-11-22
Supporter Data

MaxIsJoe:
This system by itself is fairly bare bones and only exists to load the data, and show custom flares for supporters at the moment.
We can extend the fetch system to use online APIs or something else if we need, so we are not just bound to fetching from a json file in streaming assets.

Supporters are stored in a `supporters.json` file in the data section of the streaming assets folder.

In a later PR, this system will be used for displaying supporters on the pre-round screen.

### Changelog:

CL: [New] Added a new supporter system that keeps track of people who donate to Unitystation through Patreon. This system needs to be manually updated for now, but it will be automated in the future.
CL: [New] Added a supporter rank that displays an exclusive icon in OOC chat next to their name.


UnityProject/Assets/Scripts/Managers/Supporters/FetchSupporterMethods/FetchSupportersFromJsonInStreamingAssets.cs
```diff
@@ -0,0 +1,42 @@
...
+                    return JsonConvert.DeserializeObject<List<Supporter>>(jsonContent) ?? new List<Supporter>();
+                }
+
+                // If the file does not exist, create a default one
+                var defaultSupporters = new List<Supporter>();
+                defaultSupporters.Add( new Supporter
+				{
+					Identifier = "12345678901234567",
+					Flare = "[S] - "
+				});
```
corp-0, line 27: Is the flare supposed to be a prefix to appear in OOC? Because if that's the case, you just need to create the supporter role in permissions.toml and that's it.

MaxIsJoe: uses the perms/tag system now

i was planning on doing that later anyways for specific supporter features that should be accessible by them only.

PR #11067, opened 2025-12-25
Blueprint Resources for AdventureStation & Future Asteroid Field Use

FuntimeDudeski:
### Purpose
[update] test runner doesn't like having maps with the same name. will wait to blueprintize the existing asteroidsystem ruins/derelicts in the next general asteroid update.

Waht the title says. Includes blueprintized versions of both existing and new whiteships/ruins.

### Media Preview:
All over #mapping

### Changelog:
nothing player-facing yet


MaxIsJoe:
tests failing

PR #11083, opened 2026-01-11
CaveStation (& a few text fixes)

FuntimeDudeski:
### Purpose
Adds CaveStation, a lowpop mainstation map focusing on excavation and construction (and hide-and-seek style deathmatch if it's a nuke ops round)

also fixes some text--mostly AI module text and ProTip messages (spelling/capitalization/added a little more specificity in a couple places)

### Notes:
the big ship is technically the mainstation matrix, so i might have to make adjustments or the nukedisk teleport could put the disk in one of the unreachable locations inside it. blob gameplay might be technically possible if the player immediately goes to the ship before spawning as blob.

currently no gateway, no lavaland, and no flyable shuttles included (besides the nuke ops'). i might throw in a little more content incl. a dock with a tiny shuttle or two, but keeping all the action inside one huge mostly contained area was a lowpop design goal.

because the asteroid field, lavaland, and awaysite are all currently unused, having the option not to load them would be nice for the server, but it's not terribly important

### Media Preview:
nuke ops rounds:

other:

### Changelog:
CL: [New] New Mainstation: CaveStation - Your crew has been mistakenly assigned to an early-stage construction project within the hollow of a giant asteroid, much of which was mined out long ago. Try and make the best of it. An NT freighter housing the AI is stationed just outside the site to provide supplies and receive excavated resources. It's the most valuable asset in the area, so any potential attack by an enemy corporation will likely target it. Your crew has two ships on site: a command shuttle from which Command and Security operate, and a transport ship that carries the crew and has a cargo bay full of construction supplies.


corp-0:
dungeon-keeper

PR #11091, opened 2026-01-19
Fix reagents spawning in space

MaxIsJoe:
CL: [Fix] Fixed an issue where reagents would spawn in space when they're not supposed to.

### DIFF
--- a/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
+++ b/UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
@@ -765,16 +765,16 @@ public partial class MatrixManager : SingletonManager<MatrixManager>
 		}
 
 		matrixInfo ??= AtPoint(worldPos, true);
+		var usableTilePos = GetNearestNonSpaceTilePosition(worldPos, matrixInfo, true);
 
-		if (IsSpaceWithNoTilesNearbyAt(worldPos, matrixInfo))
+		if (usableTilePos == Vector3.zero)
 		{
 			Loggy.Warning($"Attempted to do a reagent reaction in the middle of space at {worldPos}", Category.Chemistry);
 			return;
 		}
 
-
-		Vector3Int localPos = WorldToLocalInt(worldPos, matrixInfo);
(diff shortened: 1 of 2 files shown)


UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
```diff
@@ -765,16 +765,16 @@ public static void ReagentReact(ReagentMix reagents,
 		}
 
 		matrixInfo ??= AtPoint(worldPos, true);
+		var usableTilePos = GetNearestNonSpaceTilePosition(worldPos, matrixInfo, true);
```
Bod9001, line 768: why  Nearest, What scenarios would need to jump a tile? 
I suppose if something blew up and had a bunch of reagents and was instantly over space from the explosion , but had floors around it?

UnityProject/Assets/Scripts/Tilemaps/Behaviours/Layers/MetaDataLayer.cs
```diff
@@ -336,6 +336,8 @@ private void HandleSplats(ref ReagentMix reagents, ref bool didSplat,
 			}
 			else
 			{
+				//(Max): This sucks ass. Stop hardcoding this shit in places that it doesn't belong to, and make it part of the reagent logic itself.
+				//TODO: Refactor this part so that Reagents handle their own logic for splats instead of MetaDataLayer.
```
Bod9001, line 340: Slight difference is needs to have non Evaporation For space lube , And how does Mop work with wet floors if you make the floors wet with water?

Blood Is needed for the splatter effects imo looks good

UnityProject/Assets/Scripts/Managers/MatrixManager/MatrixManager.cs
MaxIsJoe: You're not jumping a tile, you're finding the closest non-space tile.
Which can also be under yourself.
The idea is that instead of just making reagents disappear into nothingness, they at least spawn below or next to the person who's trying to use it.

Bod9001: humm, fair enough,  do you think that for say had a water tank but then a bomb below it, the bomb destroyed the water tank but then also destroyed the floor beneath should all the water be deleted because it went into space or should it still spill out on the adjacent tile?

PR #11096, opened 2026-01-22
feat: add option to set an auto rank on permissions config

corp-0:
### Purpose
If defined, any player that doesn't have a rank already will get this rank automatically.

### Changelog:
CL: [Improvement] Added option to set an auto rank in permissions config. Any player without a rank will get this rank when they join.


UnityProject/Assets/StreamingAssets/Config/Admin/permissions.toml
```diff
@@ -24,106 +30,356 @@
...
 color = "#6400ff"
 permissions = ["MENTOR_MESSAGE", "PLAYER_INFO"]
+
+# Ranks can also have sprites to show in chat.
 [ranks.supporter]
 show_in_chat = true
 abbreviation = "<sprite=\"icon_id_gold\" index=1>"
 color = "#8fbc8f"
 permissions = ["SUPPORTER"]
-[[players]]
```
Bod9001, line 112: You probably need to keep around the host/admin one since it's used for Giving the local player admin when they start

corp-0: they are right there. Also this is just the example file

Bod9001: ok cool

Bod9001 (approved):

PR #11108, opened 2026-02-08
Proper namespaces for all of the project

MaxIsJoe:
This PR is chonky, and has a lot of automated changes. There is a 100% chance that there will be a lot of unintended bugs and things breaking.

This PR will make any current unmerged branch completely incompatible with upstream once merged.

# What does this PR do?

For a long while now, Unitystation has never had a proper namespace for the upstream side of the project; which would have made the life of community servers very difficult when they attempt to code their own content in their downstreams.
Everyone else also just simply never really cared about properly using namespaces, which resulted in a lot of namespace pollution, type name conflicts, poor code discovery/readability, and more.

This PR uses Rider's refactoring tools to automatically "correct" all our namespaces by fixing existing ones, and introducing namespaces to classes that are missing them. 
It also creates a unified root namespace called "US13", where all our upstream code resides. 

This is important for downstream servers, because it creates an important distinction between the "core" experience we provide, and their own content that will exist outside the `US13` namespace.

# Assembly definitions

Assembly definitions (.asmdef) have one major advantage that helps a giant project like ours, and it's called "only compile the code under a few specific namespaces if they change".
While unity did improve compilation times over the past few years (remember when we used to have 1+ minute compile times? lol), we still have to wait 15-20+ seconds every time we want to recompile the project. This also assumes that everyone has a beefy enough PC for that time to not be longer.

ADs are now possible because everything is in a proper namespace. 
We can define specific chunks in our codebase to be only re-compiled when our changes are linked to those chunks, instead of recompiling the entire project every time.

There is also a very neat security aspect to this for codescans. Because ADs isolate code, we can ensure that no one can tamper with our own safe code, and even roll out hot security patches without forcing all servers to be in sync with upstream.

For the time being, this PR does not introduce any new assembly definitions. The sheer scale of everything here is already too much, and I know we'll need to update codescans to be more context aware of ADs.

# Interface selection lister/autofixer tool

image

This PR introduces a new tool to list all interface selectors that we currently have in the project, and also auto-fix them if we ever break namespaces again.
If we ever run into a similar situation with namespaces again, this tool can automatically fix them.

# Current problems

All noticable game-breaking problems have been resolved.
Further testing will be required on staging to see what else we could have missed.

### Changelog:

CL: [Improvement] We've introduced a major change to how we organize unitystation's code to give clear separation for server forks and their content. This is a major "under-the-hood" update, so expect some bugs and unexpected changes here and there.


corp-0:
it is failing to compile for the codescan test

MaxIsJoe:
> it is failing to compile for the codescan test

Yeah, I kinda expected that.
Let me just fix all the current game-breaking bugs created by all the namespace changes, then I'll see what can be done about codescans.

PR #11109, opened 2026-02-08
rockets explode at the end of flight + new .38 revolver ammos and existing ammo fixes.

PerfectTangent:
### Purpose

This PR adds a small invoke when a projectile with ProjectileRangeLimited reaches its maximum range and uses that to allow explosive projectiles to explode with it. What happened before was rockets and projectile grenades would simply vanish once they reached their maximum range (which they don't typically do irl). This change adds an option to make them explode at the end of their flight.

This PR also fixes some prefab mistakes on a few types of ammo, and adds the functionality to .38 special rubbershot and HotShot rounds. Scatterlaser shells fire lasers as they should, now, too.

### Notes:

scatterlaser shells in /tg/ fire a weaker laser projectile than the ones here do. This might mean that scatterlaser shells might overperform, but they are rare, so it should be fine for now. Our buckshot also is 33% stronger than /tg/, so it's not like scatterlasers are comparatively unmatched. I think an audit of all shotgun shell damage should happen at some point. 

Everything was tested in editor. I noticed that incendiary bullets were not causing any visual effects (though they did start fires and deliver firestack to shot players), so something may have broken there at some point.

### Changelog:

CL: [New] missiles and grenade launcher nades explode when they reach their maximum range without hitting anything.
CL: [Improvement] shotgun scatterlaser shells fire lasers now! 5 bolts dealing 15 damage each.
CL: [Improvement] .38 special hotshot bullets now have their incendiary component. 12.5 damage per bullet and applying 2 firestacks.
CL: [Balance] buckshot and rubbershot have had their maximum projectile ranges reduced from 50 to 20.
CL: [Balance] 40mm grenades have had their range reduced to 20.
CL: [Balance] 84mm missiles have had their range reduced to 30.
CL: [Fix] .38 special rubbershots are now correctly nonlethal.
CL: [Fix] rubber shotgun shells fired the wrong type of nonlethal slug, leading to excessive brute damage. They now fire rubber pellets.
CL: [Fix] slug shotgun shells were firing nonlethal beanbags. They now fire the correct highly lethal projectile.


UnityProject/Assets/Scripts/Projectiles/Behaviours/ProjectileExplode.cs
```diff
@@ -36,6 +67,8 @@ public bool OnHit(MatrixManager.CustomPhysicsHit hit)
 		private void Explode(Vector3 worldPosition)
 		{
 			if (CustomNetworkManager.IsServer == false) return;
+			if (hasExploded) return;
```
corp-0, line 73: did you find cases where it exploded twice or something?

PerfectTangent: I read the unitystation development guide and it was advised to think about future proofing things like that. checks like that exist all over gun code, so I did it to be safe. If you don't think it's necessary, it can be removed.

corp-0: I myself, am not sure if it is needed, but if it is it should probably be resetted back to false in `OnEnable`. I'm assuming that when the projectile explodes, it isn't really destroyed and goes back into a pool with everything disabled. If that is the case, and you happen to reuse the same prefab later on, it will come with `hasExploded` set to true and it won't explode

PerfectTangent: ahh, right. that makes sense.

PerfectTangent:
First PR in 3 years. Good to be back.

MaxIsJoe:
tests are failing.
Also, welcome back!

corp-0:
> Bullet_.38 or Bullet_.38 HotShot NEEDS to be committed with it's new Forever ID. Expected: False But was: True 

looks like you just have to generate a forever id for the new prefab

MaxIsJoe:
Go to the prefab -> check the `Prefab Tracker` component -> hit the "generate unique ID" button or create your own unique ID.
Then
Go to the network manager prefab -> Hit the button that fills all references to prefabs
Then commit your changes.

corp-0 (approved):

PR #11110, opened 2026-02-12
More Room Generation Options

AtnerNT:
Per request of Funtime to help with their mapping projects

### Changelog:
CL: [New] Added the ability for Maintrooms to have a maximum number of copies, to limit duplicate rooms where needed
CL: [New] Added the MaintRoomGuarantor, a mapping object to ensure a given room always generates on a map, even if not in the same place.


MaxIsJoe (approved):
LGTM

PR #11112, opened 2026-02-18
gibb Reason and the admin chat window fix

Bod9001:
### Changelog:

CL: [Improvement] logs the reason for gibb

CL: [Fix] fixes admin chat window not updating properly

### DIFF
--- a/UnityProject/Assets/Scripts/US13/Core/Camera/CameraEffectControlScript.cs
+++ b/UnityProject/Assets/Scripts/US13/Core/Camera/CameraEffectControlScript.cs
@@ -97,6 +97,15 @@ namespace US13.Core.Camera
 		public List<IBumpableObject> Bumps = new List<IBumpableObject>();
 
 
+		public void OnDestroy()
+		{
+			base.OnDestroy();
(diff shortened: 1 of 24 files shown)


UnityProject/Assets/Scripts/US13/Core/Camera/CameraEffectControlScript.cs
```diff
@@ -97,6 +97,15 @@ public void Awake()
 		public List<IBumpableObject> Bumps = new List<IBumpableObject>();
 
 
+		public void OnDestroy()
+		{
+			base.OnDestroy();
+			if (CustomNetworkManager.IsHeadless == false)
```
corp-0, line 103: shouldn't this happen before the base? What if the base destroys the object?

Bod9001: base is just passed on logic is fine, since it inherits from the singleton manager and that does some stuff in there

corp-0 (approved):

MaxIsJoe (approved):
LGTM

PR #11122, opened 2026-03-06
[Draft] Reserve Core

havietisov:
(offer a different name for this PR)

### So, the idea

Idea is to have SOMETHING that can play as station AI. A bot that will open you a door, for instance, or raise an alert that there's a fire, or even order something in the cargo, WHATEVER. Whatever the basic duties of the station AI player are, these can be automated when noone's arsed to actually play as one (it's also hella cool to have anyway). 

## Resources

After a bunch of research and experimenting it became apparent this should be possible on a "bag of peanuts and a high five" kind of computational budget. @corp-0, being aware of my rambling in discord, told to create this PR so we could see discuss it here.
 
## Implementation 

High-level overview:

1. Tiny 22m embedding model already trained by other people, 
2. set of cluster centroids obtained by [k-means](https://en.wikipedia.org/wiki/K-means_clustering) or [DBSCAN](https://en.wikipedia.org/wiki/DBSCAN), from our training+calibration examples
3. a matrix running principal components extraction (trained by us in minutes),
4. An echo state network  - [a random set of weights with only one matrix actually trained on anything](http://www.scholarpedia.org/article/Echo_state_network?source=post_page-----f910809d23d4----------------------). In this case, we don't even need to train the linear part because GBDT replaces it, and we have to train it anyway. 
5. [GBDT](https://en.wikipedia.org/wiki/Gradient_boosting) (trained in minutes), 
6. a set of training+calibration pairs (coming from our testing and play sessions + some stuff predefined by us) 
7. [autoencoder network](https://en.wikipedia.org/wiki/Autoencoder) (trained by us in minutes-to-hours), so it can fail gracefully
8. and, ideally, a [PDDL domain + solver + PDDL problem statement](https://en.wikipedia.org/wiki/Planning_Domain_Definition_Language) generator for fancier stuff than "open the door" and "tell me a joke". 
9. in the game? to either have it as a separate NPC that does things, and let people enable/disable it, or replace an actual AI player when nobody is arsed to play as station AI. 

Looks like a lot, but the most expensive part is actually running the embedding model. Rest's running on an actual throwaway-tier potato hardware, especially if we let players run the 22m embedding model + PCA extraction matrix (nothing's gonna happen if someone does something funny with these values, literal self-sabotage by the player), with rest running on the server, taking miliseconds at worst. 

Everything mentioned has been validated on smaller scale in lab conditions, and it ... worked. What's left is actual scaling and test deployment in actual environment it's supposed to run in, it's the only proper way to test if it has any legs (i'm confident it has).

[markdown file with inference pipeline for clarity sake](https://github.com/havietisov/unitystation/blob/auto_ai/inference_pipeline.md)

## Hopes and dreams

To cover decent amount of possible requests by playeres and fail gracefully if players as for something unsupported. Basically - it needs to be useful and NOT be ruining anything. To also use this for other robots doing different things, created by robotics dept. 

## Other points

Tolerant to slang, typos, and will never say anything that we didn't program it to say. 

## Questions

Free to leave any questions. 

## Timeline

Will take a while to see something remotely useful - weeks to months

### Changelog: 
CL: [New] AI for the AI when no player is playing the AI


corp-0:
> Tiny 22m embedding model already trained by other people,

Is this based on those models that interpret language into commands for smart homes and stuff?

havietisov:
> > Tiny 22m embedding model already trained by other people,
> 
> Is this based on those models that interpret language into commands for smart homes and stuff?

not just them, smart homes use embedding models among other stuff. so, in a sense, we should get functionality similar to what of a smart home. just bigger. and probably funnier

corp-0:
you will keep getting comments of the bot about not having changelog if you update the PR description. You can add a placeholder changelog line for now

DooblyNoobly:
I like this idea actually. Definitely deserves some exploring/prototyping

PR #11123, opened 2026-03-07
make speedloaders reusable/reloadable.

PerfectTangent:
### Purpose

This PR fixes a small issue with clip type magazines such as speedloaders. Before, when they reach 0 ammo, they would all delete themselves. this was problematic for speedloaders, which are supposed to persist and be reusable. This adds a check to clips to ensure only stripper clips get deleted while speedloaders don't.

### Notes:

Tested in editor.

### Changelog:

CL: [Fix] revolver speedloaders will not automatically despawn once out of bullets, and can be reloaded and reused.


MaxIsJoe (approved):
LGTM

PR #11136, opened 2026-03-20
feat: deep-fryer construction/deconstruction and cleaning

corp-0:
### Purpose
When merged this PR will allow the Deep-fryer to be constructed/deconstructed, moved around after unsecuring with the wrench and also cleaned.

It also includes a couple more bug fixes.

### Notes:
- renamed `GetComponentCustom` to `GetCachedComponent` because that name makes more sense.
- Had to update all call-sites to the new name, which is why there are so many files changed. Use commits to review changes.

### Changelog:
CL: [Improvement] The deep-fryer can now be constructed and deconstructed.
CL: [Fix] Fixed any non-edible item being eaten when thrown aiming to the mouth.
CL: [Fix] Fixed being able to force feed you any non-edible item by interacting while aiming to the mouth.


MaxIsJoe (approved):
das allota changes
LGTM

corp-0:
> das allota changes LGTM

because of the rename, use the commits to review. The actual meaningful changes aren't that many

corp-0:
also, the SO browser. I forgot about that:

imagen

I didn't want to have to find the SO in that stupid right click menu

PR #11141, opened 2026-03-22
fixes reagent containers being absolutely broken

corp-0:
### Purpose
This is the path of least resistance. I considered the options of having multiple containers in a single `ReagentContainer` component, and doing surface/internal, but I decided against it because it is a bigger change and I don't want to break more things.

So issue was that every time we used `GetComponent` it was a coin-flip on which of the two reagent containers it would get, in theory at least because I always saw it getting the base disabled one.

To fix this, I ensured that every time we want to use a `ReagentContainer`, we use the enabled one no matter what. So that if the base was enabled by deep-frying, we use that one. If the item already had a reagent container, then deep-frying wouldn't enable the base and we always get the right one.

For this reason, the components cache now supports multiple components per key instead of a single one. This makes the `includeDisabled` argument when querying them actually work because we can iterate the components until we find the first enabled one.

In terms of performance, this should be alright. 

UPDATE: 

Well, I did the asked changes. Now there is only a single instance of both ReagentContainer and Edible on every single item in the game. I did a bulk migration tool that inspected every item and in the case of duplicates, it copied over the data from the added component to the base one.

800+ items were modified. I could only inspect manually so many of them and they looked correct. I did a second tool to find any broken reference to the removed components and the result came back clean.

### Changelog:

CL: [Fix] Fixed being unable to transfer reagents between containers.
CL: [Fix] The deep-fryer can be refilled now.
CL: [Fix] Fixed the spray bottle not spraying anything.
CL: [Fix] Fixed every single reagent container in the game declaring themselves as empty.
CL: [Fix] Fixed food not satiating.
CL: [Fix] Fixed the deep-fryer reporting not enough oil to clients.


Bod9001:
Why not just check for only one reagent container, so therefore the base one becomes the added one, then you don't have to do This?

MaxIsJoe:
> Why not just check for only one reagent container, so therefore the base one becomes the added one, then you don't have to do This?

This is a fine fix for now.

A better way to handle this in the future is to disallow multiple reagent container components entirely, and instead make the singular reagent component handle multiple.

MaxIsJoe (dismissed):

Bod9001:
yeha, that's what I'm saying I don't mind going through the PR and removing the accidental duplicates

MaxIsJoe:
> yeha, that's what I'm saying I don't mind going through the PR and removing the accidental duplicates

image

If you see this checkmark, jump ahead

MaxIsJoe:
though of course, seek Gilles' perm first

corp-0:
it is not so simple. If you want to go that route you have to:
1. Copy over the configuration of ReagentContainer from everything that already had it into the base.
2. Only now you can remove the extra (added by legit reagent containers)
3. You have to do the same thing for edible
4. Now you can remove the multi-component cache and make it single component again
5. **BUT you still have to ask on call-sites if the ReagentContainer/Edible is enabled** or not, otherwise you will keep having issues with unintended behaviour, examination and tooltip strings

corp-0:
did the requested changes

corp-0:
well shit, all the prefabs are broken. I suggest we merge the previous fix and work on cleaning up the component later, otherwise the game will be broken for next play test

MaxIsJoe:
previous fix it is then

PR #11150, opened 2026-04-05
feat: effect on consume component

corp-0:
### Purpose
adds a new component for custom effects (other than the reagent transfer) when a consumable item is consumed.

As a proof of concept, I have replaced the `HealableJobEdible` component that extended `Edible` to make donuts heal characters with sec job.

It is made with composition in mind and you can add many effects on a single prefab and it will just work. You can also use composable conditions to make some effects only work when certain conditions are met. As an example, here is the implementation of the replacement of `HealableJobEdible`

imagen

Some ideas to extend this system further (with very simple implementations:
- Make glass shards do internal damage when eaten.
- Make a speed potion that adds a haste status effect.
- Hot food that applies a Mouth Burning status effect or does some minimal burn damage.
- Grandma's healing herbs that can heal some illnesses.

To replace `HealableJobEdible` I created the `JobIsEqualToCondition`, but we can expand this as well, for example:
- `SpeciesIsEqualTo` so that the effect is only applied on certain species.
- `HasOrgan`so we only do something if the eater has an organ that in-universe can process the effect.
- `RandomChance`.

More ideas combination of effects and conditions:
- Eating raw meat makes you ill, unless you are a vulp, cat or lizard.
- Eggs giving you salmonella unless you are a lizard.
- Fortune cookie that gives you a random buff but only once per round.

More things to replace with this system (that currently extend Edible or Consumable in an ad-hoc script):
  - XenomorphFood: larvae implant becomes an effect with a species condition
  - WishSoup: random effect selection via a random-chance condition
  - PillToxHeale: toxin healing becomes a simple heal effect
  - PillStackRemover: inject reagents using the effect.

### Notes:
- Have also removed the duplicated ReagentContainer and Edible from everything, so we are back to single components.
- There are many files changed, so review using the commits.

### Changelog:
nothing facing the user yet, but this will make new cool stuff possible.


MaxIsJoe:
why are tests fai- WOAH WHAT DO YOU MEAN 92 MINUTES?

MaxIsJoe:
image

UnityProject/Assets/Scripts/US13/Core/Editor/Tools/DuplicateComponentCleaner.cs
```diff
@@ -0,0 +1,283 @@
...
+using US13.Items.Food;
+
+/// <summary>
+/// Editor tool to find and remove duplicate Edible or ReagentContainer components on prefabs.
+/// Processes prefabs in hierarchy order (parents before children) so that fixing a parent
+/// automatically resolves inherited duplicates in its variants.
+/// When duplicates are found, it removes the one with the oldest ancestry (deepest in the
+/// variant chain) and keeps the one added closest to the current prefab level.
+/// </summary>
+public class DuplicateComponentCleaner : EditorWindow
```
MaxIsJoe: namespace

MaxIsJoe (changes requested):

UnityProject/Assets/Scripts/US13/Core/Editor/Tools/DuplicateComponentCleaner.cs
corp-0: Ah my bad, I wasn't supposed to commit the tool. It is a one time use

corp-0: removed. Any comment on the new code at all?

MaxIsJoe (dismissed):

MaxIsJoe:
Approved.
Need to figure out what's going on with the scenes, however.

corp-0:
why scenes failing tho

Bod9001 (dismissed):

PR #11151, opened 2026-04-06
Simpler Hunger calculation strategy

MaxIsJoe:
`NutrimentInBlood` is a simpler hunger calculation strategy that uses a single source of truth for how hunger is generally handled and balanced, the blood.

Different properties can also be set when the system first starts on this strategy, such as how much nutrition exists in the blood or stomach.

We can calculate how much nutrition is needed over a period of time by checking the `NutrimentConsumedPerTick` variable.
All humanoid mobs have 1 tick per 1 second at the time being.
`NutrimentConsumedPerTick` is 0.01 consumption per tick by default.

`nutrition consumed=0.01×(seconds)`
1 minute = 0.6 nutrition consumed.
5 minutes = 3 nutrition consumed.
10 minutes = 6 nutrition consumed.
25 minutes = 15 nutrition consumed.
50 minutes = 30 nutrition consumed.

image

[THESE VALUES ARE NOT FINAL. NEED TESTING FOR PROPER TWEAKS]
You're considered well-fed if you have more than 20 nutriment reagents in your body. Going above 40 makes you fat due to stomach/body fat shenanigans.

Going below 20 reagents is when you start getting hungry, and going below 10 is when you start starving.

### Notes:

- There is still work to be done to simplify hunger further and make it much more bearable to work with. (Moving slowness to the status effects, simplifying body fat, etc)
- There are still some external factors that can affect these numbers. Mainly on the body fat side.
- A single sliced Arnold pizza can sustain an army of greytides. Balancing may be required. From what side? we'll have to see later.

https://github.com/user-attachments/assets/73365c05-cf53-47bf-8cd3-ad3c86e3012a

### Changelog

CL: [New] All humanoid species now use a simpler calculation method to see how hungry they are. Feedback is required to balance things around this in later PRs.


corp-0:
Scene test failing, so it wasn't just me

PR #11152, opened 2026-04-06
Removes unneeded tests and removes rooms from Tests

Bod9001:


MaxIsJoe (changes requested):
No.

Removing tests is not a solution to fixing tests.
I'm okay with removing blueprints for maintrooms out tests because they kinda just bloat everything atm, but I'm not okay with the idea of removing important tests entirely.

Bod9001:
what Tests  was removed?
that we needed, they are old ones we don't need any more

MaxIsJoe:
MatrixHasMatrixSyncAndIsCorrectParent, MatrixHasAllLayers, LayersHaveCorrectParentAndAreNotDuplicated, PrefabsAreNotAt00

all seem to have valid use cases, such as PrefabsAreNotAt00 checking for missing prefabs.

Bod9001:
well is for old scene tests these are all Json mapps now and don't have the same issues

MaxIsJoe (approved):
eh

PR #11187, opened 2026-04-26
chat - stack only repeats of most recent message

FuntimeDudeski:
### Purpose
fixes repeated text message stacking so it doesn't look back several lines and cause the message log to misrepresent the chronological order of events (very annoying when repeatedly performing a sequence of 2-3 actions). adds an inspector field to ChatUI for configuration from no lookback up to the old limit of 6.

### Notes:
need to figure out how to add this to the in-game options menu

i would also like to figure out how to keep the font from growing to absurd sizes when a message is repeated many times but have not been successful yet (the current max size is unnecessarily large)

### Changelog:
CL: [Fix] Repeated message stacking now only triggers when repeating the most recent line in the chatlog, maintaining an accurate order of events--Currently configurable up to 6 recent messages or disableable via Unity editor inspector.

### DIFF
--- a/UnityProject/Assets/Scripts/US13/Core/Chat/ChatUI.cs
+++ b/UnityProject/Assets/Scripts/US13/Core/Chat/ChatUI.cs
@@ -38,6 +38,11 @@ namespace US13.Core.Chat
 		public GameObject chatEntryPrefab = default;
 		public int maxLogLength = 90;
 
+        [SerializeField]
+        [Range(0, 6)]
+        [Tooltip("Number of entries back from most recent to scan for duplicate messages to stack. The chatlog won't always preserve an accurate order of events at values >1. 0 to disable.")]
+        private int duplicateLookbackDepth = 1;
+
 		[SerializeField]
 		private TMP_Text chatInputLabel = null;
 		[SerializeField]
@@ -363,7 +368,7 @@ namespace US13.Core.Chat
 		private bool WillUpdateStack(ref string message)
 		{
 			if (allEntries.Count <= 5) return false;
-			for (int i = 0; i < 6; i++)
+			for (int i = 0; i < duplicateLookbackDepth; i++)
 			{
 				var entryToCheck = allEntries[allEntries.Count - i - 1];
 				string cleanedEntryMessage = Regex.Replace(entryToCheck.Message, @"<size=[^>]+>|</size>", string.Empty);
--- a/UnityProject/Assets/Prefabs/UI/ChatUI.prefab
+++ b/UnityProject/Assets/Prefabs/UI/ChatUI.prefab
@@ -4374,6 +4374,7 @@ MonoBehaviour:
   chatEntryPrefab: {fileID: 1096268754478474, guid: 6db9b4c8e7f61427b8eab85171ff4950,
     type: 3}
   maxLogLength: 100
+  duplicateLookbackDepth: 1
   chatInputLabel: {fileID: 2823315393706369458}
   channelToggleTemplate: {fileID: 968003211179716248}
   background: {fileID: 3884359157366066247}


UnityProject/Assets/Scripts/US13/Core/Chat/ChatUI.cs
```diff
@@ -363,7 +363,7 @@ public void AddChatEntry(string message, TMP_SpriteAsset languageSprite = null)
 		private bool WillUpdateStack(ref string message)
 		{
 			if (allEntries.Count <= 5) return false;
-			for (int i = 0; i < 6; i++)
+			for (int i = 0; i < 1; i++)
```
MaxIsJoe: Probably best if this is moved to a variable, instead of it being hardcoded.
Also, maybe raise it to 2?

MaxIsJoe (changes requested):

UnityProject/Assets/Scripts/US13/Core/Chat/ChatUI.cs
FuntimeDudeski: at a value of 2, wouldn't a repeated A, B, A, B sequence of actions still make it ambiguous which step occurred most recently? i don't think violating temporal sequence is ever worth the space savings--i'd look at verbosity/necessity of chatlog messages and do everything i could to reduce that (condition/integrity text on most objects, narrower distance checks on display/nondisplay of observational messages), as well as reduce the max size of large text (alert and gateway connection messages, stacked repeat scaling limit), before resorting to letting the chatlog violate chronology to save lines.

re: implementation, can you tell why this didn't work?
image

MaxIsJoe: Why did you remove the for loop? I said just replace the magic number with a variable that can be set from the inspector.

Also, you're only checking the very last entry; you're not checking anything else.

FuntimeDudeski: the above was something i had already tried previously, not a response to your suggestion--it's a copy of the screenshot i posted to the discord a few weeks ago and linked to in the notes for this PR.

'implementing it as a variable that can be set from the inspector'
off the top of my head, i don't know precisely how to do this--where to declare the variable, what component/gameobject to make an inspector field for, etc.--or why exactly i would be doing that when my intent is to exclusively check the most recent line. i'm still trying to memorize the most basic syntax and terminology. not at the point of 'just do X' yet.

if by 'last entry' you mean 'most recent', yes. the point is it's important that the message log not make it look like things happened in an order other than that in which they did. in an extreme case, it can make the message log misrepresent, for instance, who was the aggressor in a fight or other 'dispositive facts' that someone can, in the heat of the moment, easily miss seeing in realtime to catch the discrepancy. but far more common, it just causes confusion whenever one engages in a repeated sequence of different actions and is trying to keep track of what they're doing. there are many un- or under-exploited ways of reducing chatlog clutter/scrolling that don't make as critical a sacrifice as preserving a clear, accurate order of events.

MaxIsJoe: > off the top of my head, i don't know precisely how to do this

You did in that other PR you pushed.
You just place all variables at the top of the class with each other.

> at component/gameobject to make an inspector field for

It's on the same script that has the for loop.

----

In regard to limiting stacking only to the most recent message, I doubt we really want that as the default behavior.
SS13 is a very noisy game in terms of the text output. US13 is a bit quieter due to us forgetting that the game is mainly a MUD at its origin, but it still outputs a lot of repetitive text while playing. (Just look at any instance of a fight going on, or a bomb going off.)
Making it so that it only reacts to the most recent message only will result in a lot of repetitive noise being spammed.

It's much better if message stacking is configurable so that players can decide their own preferences.
For now, make it configurable via the inspector; then look into turning it into a setting for players to change.

FuntimeDudeski: i get that--it's exactly that fidelity to the MUD style text narration that made me say this:

> i don't think violating temporal sequence is ever worth the space savings--i'd look at verbosity/necessity of chatlog messages and do everything i could to reduce that (condition/integrity text on most objects, narrower distance checks on display/nondisplay of observational messages), as well as reduce the max size of large text (alert and gateway connection messages, stacked repeat scaling limit), before resorting to letting the chatlog violate chronology to save lines.

i'm down with making it configurable though. the text box is pretty well implemented at base, but there's a lot of verbosity/flow improvement possible beyond this idea, so i'll plan on looking into that as well.

FuntimeDudeski: added a nice inspector config var with a TOOLTIP (!!!) and SLIDER BAR <3<3<3 -- range from 0 to 6, with 0 disabling the stack functionality entirely. tested and seems to work perfectly! no idea how to integrate into the options menu yet.

MaxIsJoe (approved):

Issue #11184, opened 2026-04-26
Tile palette missing animated connected texture sprite


FuntimeDudeski:
update 4/30:

#11201 fixed everything except animated+connected lava.

<img width="703" height="556" alt="Image" src="https://github.com/user-attachments/assets/83087af8-e621-456e-a589-5cce3f96714b" />

update 4/27: the common pattern is it's the connected tiles in base floors. that's the only palette i'm aware of having the issue.

this issue appeared with #11079 which updated tiles to all support animation/connection

before:

<img width="825" height="615" alt="Image" src="https://github.com/user-attachments/assets/6fafdeef-2183-4f5c-9e62-c2d7bc0ec497" />

after (ignore the grid formatting shift):

<img width="744" height="543" alt="Image" src="https://github.com/user-attachments/assets/6ea6b3cf-0b5c-4017-918f-afdaa2061a04" />

it affects catwalk, lattice, and the smooth grass tiles in the Base Floors palette.

in some cases, the same tile (lattice, catwalk) appears in the floors palette and has a sprite shown. that same tile is in base but doesn't show a sprite (or collider type) in inspector.

you can see in the inspector that the tile is there in Base Floors, but its sprite is not present.

if i create a new tile palette and try to add the grass tiles, they are invisible, but they are visible when adding additional copies to Floors.

adding something to the first field called 'sprite' and removing the texture from 'sprite sheet' makes grass appear in the palette again (but this isn't a fix since it would break connected tiles)

<img width="595" height="421" alt="Image" src="https://github.com/user-attachments/assets/8e04d119-ee2a-4820-a0b4-187092f4d3ff" />

PR #11198, opened 2026-04-28
Tile and tile interaction work

FuntimeDudeski:
### Purpose
- adds indestructible plating (progresses visually through damage stages but can not ultimately be destroyed or damaged to the point that holes appear in it)
- removes ‘is sealed’ from damaged plating with exposed space
- base plating now progresses to lattice when its final/most damaged state is destroyed
- repairing damaged plating now takes slightly longer the more damaged it is
- prying up damaged plating now takes slightly less time the more damaged it is; plating which is only lightly damaged has a chance of prying into an intact floor tile.  
- adds loose plating, a palette-excluded plating variant which looks slightly lighter around the edges and is pryable--sets this special variant as the kind used when installed manually on lattice by a player. unsecured plating has the same stats as regular base plating aside from losing half of its bomb resistance, and instead of progressing to a damaged state if destroyed, it is knocked loose as a floor tile (50%), scrap metal (25%), or nothing (25%). it takes 1 second of welding to secure it to the lattice, making it into standard, unpryable base plating and its sprite/name/desc distinguish it as loose plating.
- grass base tiles now become dirt base when destroyed, grass health reduced 300 => 100 (the total damage to destroy the tile thus rises from 300 to 400 since dirt has 300 health of its own)
- indestructible grass can now be visually damaged, becoming indestructible dirt
- top snow floor tiles no longer have plating as a required tile underneath
- adds animated + connected lava (currently invisible in the palette and shows up as catwalk in the editor out of game, presumably due to bug in issue #11184)

- marked the missing base tiles in the base floors palette with icons while bug is in effect and added the indestructible plating variants with a marker
image

- window deconstruction step with welder was still set to harm intent--now disarm like other deconstruction steps
- window repair steps restored to help intent (was disarm, likely an error)
- pod wall connections changed to connect-to-self to improve their visuals
image
image

- updated a couple spawnable lists for grilles that had deprecated versions and a reference to a deprecated version on one of the grille objects. unchecked ‘is not pushable’ on the new grill object prefab to match the prefab it deprecates.
- changed the hardcoded time value in RepairWithWelder.cs to an inspector-settable one, making the formerly hardcoded value (4 sec) the default.
- adds a couple random item pools to vary the output of destruction to structures
- fixes a shuttle wall deconstruction step
- couple spelling/punctuation/capitalization fixes

### Notes:
~~PRing as a draft so bod can take a look at why the animated connected lava doesn't seem to be usable~~

### Changelog:

CL: [New] Map editor: Indestructible base plating added for awaysites, etc. It can be dented/cracked/repaired but not destroyed.

CL: [New] Map editor: Animated + connected lava base with basalt edges added!

CL: [Improvement] Terrain destruction is more dynamic, with grass able to be destroyed & show the dirt underneath and plating producing debris & revealing lattice when destroyed.

CL: [Improvement] Plating is no longer permanently installed when first placed on lattice--It can be pried up again until welded to the lattice beneath. It will, however, stick to foam base automatically.

CL: [Fix] Jaws of life now correctly expedites shuttle wall deconstruction when used on the proper step.

CL: [Fix] Reinforced window repair/deconstruction intent requirements are now consistent with the rest (help to repair, disarm to disassemble)


UnityProject/Assets/Scripts/US13/Core/Input System/InteractionV2/TileInteraction/RepairWithWelder.cs
```diff
@@ -13,7 +13,8 @@ namespace US13.Core.Input_System.InteractionV2.TileInteraction
 	public class RepairWithWelder : TileInteraction
 	{
 
-
+		[Tooltip("Seconds taken to perform this action. Set to 0 for instant.")] [SerializeField]
```
MaxIsJoe, line 16: for future reference, you can compile all attributers in the same bracket; like this:
`[Tooltip("Seconds taken to perform this action. Set to 0 for instant."), SerializeField]`

otherwise this is good

MaxIsJoe (approved):

PR #11205, opened 2026-05-01
Reworks blueprints

AtnerNT:
No changelog as no real player facing changes;

This PR seperates the blueprint loading/choosing logic out of maint room generators into a new BlueprintSpawner static class that can be used to load blueprints from any location.
This also changes how room choices are made, a list of rooms, their weights and max occurances are registered under a key to BlueprintSpawner. Then, any script can pass in this key to get an appropriate random room.
This PR also adds a GenericBlueprintSpawner prefab to load blueprints on roundspawn or in response to a generic trigger seperated from the maint generation system.

The BlueprintSpawner class should serve as a much more flexible class to implement random generation in non-maintroom systems such as random engines for stations and lavaland. 

Side benefits: Subdirectories for blueprints are now supported, No more SO spam


AtnerNT:
PR for security tests: #11206

MaxIsJoe (approved):

Issue #11200, opened 2026-04-29
[BUG] Falsewalls block turret vision when open and reveal themselves through UI


FuntimeDudeski:
### There is no other issue reporting the same problem.

- [x] I have searched the existing issues

### Game Version

26040419

### I was

Client

### OS

Linux

### What happened?

false_open blocks turret targeting LoS
https://discord.com/channels/273774715741667329/312454684021620736/1490929386642214935

closed falsewalls give away their presence via the condition text, the deconstruction procedure text, and the shift+hover examine tooltip
https://discord.com/channels/273774715741667329/312454684021620736/1490928568010674337

### What should have happened instead?

false_open not block turret vision/target acquisition and falsewalls not being meta-detectable when closed

### Steps to reproduce the issue

open a falsewall between self and hostile turret
examine a closed falsewall and compare to regular wall

FuntimeDudeski:
re: design around intentional detectability https://github.com/unitystation/unitystation/pull/11210#issuecomment-4403239607

PR #11210, opened 2026-05-03
More fixes

AtnerNT:
### Changelog:
CL: [Improvement] Small Optimisations in turret code

CL: [Fix] Fixed Turrets LOS being interrupted by open false walls
CL: [Fix] Fixed Turrets having their clearances miss-set
CL: [Fix] Fixed brains being able to be destroyed via laser science.

### DIFF
--- a/UnityProject/Assets/Scripts/US13/Systems/Research/LaserPuzzle/ItemPlinth.cs
+++ b/UnityProject/Assets/Scripts/US13/Systems/Research/LaserPuzzle/ItemPlinth.cs
@@ -11,6 +11,7 @@ using US13.Core.Input_System.InteractionV2.Interfaces;
 using US13.Core.Lifecycle;
 using US13.Core.Sprite_Handler;
 using US13.HealthV2;
+using US13.Items.Traits;
 using US13.Managers.NetworkManagement;
 using US13.Projectiles;
 using US13.Projectiles.Behaviours;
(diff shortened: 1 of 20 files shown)


AtnerNT:
I think this is suitable to close #11200
The other part of the issue - that false walls are inspectable I feel is intended? as a way to find them.

AtnerNT:
I broke some stuff somehow, one mo

AtnerNT:
Should be ready to merge once tests pass

UnityProject/Assets/Scripts/US13/Systems/Research/LaserPuzzle/ItemPlinth.cs
```diff
@@ -81,8 +84,7 @@ public void ServerPerformInteraction(PositionalHandApply interaction)
 			else
 			{
 				if (interaction.HandObject == null) return;
-
-				if (blackListedItemNames.Contains(interaction.HandObject.Item().InitialName))
+				if (interaction.HandObject.Item().HasAnyTraitZeroAlloc(blackListedTraits) || blackListedItemNames.Contains(interaction.HandObject.Item().InitialName))
```
Bod9001, line 87: it may be better to have a Component that we attach to Mark as an indestructible, and so When it gets to server.deSpawn You have to do some special Override indestructible = true , to make it actually be destroyed

MaxIsJoe: Traits can be adjusted during runtime.
Adding components is difficult and a headache because of mirror.
This is fine.

MaxIsJoe (dismissed):

MaxIsJoe:
Need to fix security tests

MaxIsJoe (approved):

FuntimeDudeski:
> I think this is suitable to close #11200 The other part of the issue - that false walls are inspectable I feel is intended? as a way to find them.

i think some way of distinguishing them on deliberate examination is good, but probably not from arbitrarily far away and preferably via intentional examination rather than just mouse hover. and the condition text being bright green doesn't help.

arguably, walking up and clicking the falsewall and seeing if it opens is close to an ideal way to check/inspect, but it could also be cool to append a string to the description of a falsewall when shift+click examining it from 3 or fewer tiles away, e.g. 'This wall looks a bit odd.'

PR #11211, opened 2026-05-05
feat: Firebots and Securitron

AtnerNT:
### Changelog:

CL: [New] Adds firebot
CL: [New] Adds securitron
CL: [Improvement] Firefighting foam can now extinguish fires and will do so with a higher potency than water
CL: [Improvement] Simplebots now stop one tile earlier instead of pathing atop their target
CL: [Improvement] Sulphuric acid is now a liquid and deals acid damage when on the player
CL: [Fix] Fixed bots being unable to be emagged


UnityProject/Assets/Scripts/US13/Core/Chat/ChatRelay.cs
```diff
@@ -79,7 +79,9 @@ public void Start()
 
 		private void WhisperCheck(ChatEvent chatEvent)
 		{
-			var willWhisper = whisperPrefix.Any(prefix => chatEvent.message.Contains(prefix));
+			var willWhisper = whisperPrefix.Any(prefix =>
```
Bod9001, line 82: While you're touching this could you quickly make it so, 
words that have like w at the end w! Doesn't cause it to trigger whisper

Bod9001 (approved):

Issue #11226, opened 2026-05-16
[Refactor] iron chef should be made into an admin triggered gamemode/location


PerfectTangent:
### Describe the problem with the current implementation.

Iron chef is a good little location to have some fun admeme managed fun, but isn't appropriate as an Awaysite, where adventure is supposed to be had and materials relevant to the station are to be found. Iron chef a gimmick idea and Awaysites should have general every round appeal.

### Why is a refactor required?

discussion was held in the mapping channel and there was general consensus on this idea. it isn't appropriate as an Awaysite and other than providing immense amounts of food, it doesn't have any value for station crewmembers to go there *except* to participate in the iron chef challenge.

### Explain the goals of your suggested refactor, and what will be changed from the current implementation.

It would be more suitable as a scene available to admins to spawn in some corner of the map, OR, perhaps a mainstation concept with its own supporting gamemode if someone is interested in developing that.

### To-do checklist

- [ ] add iron chef as an admin spawned scene or admin triggered gamemode

### Describe current workarounds or bandage solutions until a refactor is in motion.

- [x] remove iron chef from the map pool.

PR #11227, opened 2026-05-16
Removes Ancient Temple, Palace, Dark Jungle, Shuttle Race, and Iron Chef from Awaysite map list.

PerfectTangent:
### Purpose

Awaysites have been undermaintained for a long time. several of them were implemented a long time ago when the game had much less content and the maps haven't been updated. several of the awaysites also have conceptual issues, have design issues, or have broken features (like Ancient Temple skeletons). We held a discussion on the topic ([here](https://discord.com/channels/273774715741667329/285240893160685570/1500482717480124577)) and determined that some should be removed for now until updated or spun into separate things.

Sites removed:
- Ancient Temple
- Palace
- Dark Jungle
- Shuttle Race
- Iron Chef

this leaves the current list of awaysites as:
- caverna
- relicta
- backrooms
- mystic valley

which are all much stronger, more actively maintained, and better awaysites at this time. this means that players will more reliably get awaysites that have things worth doing at them.

### Notes:

the framework I suggested and got agreement or no disagreement for was that
- palace should probably be cut for good. it just isn't a good awaysite conceptually or in execution
- dark jungle should be cut for being a trap, and being buggy, with no good rewards for the players. if it were to be brought back, it should probably be done on a new scene because the one we have has seen some map rot.
- iron chef should be its own thing separate from awaysites (#11226)
- shuttle race is slightly broken and gets stale very fast. it should be reworked to use some kind of random generation, and perhaps would be best served as a gamemode of its own, like iron chef.

### Changelog:

CL: [Balance] Ancient Temple, Palace, Dark Jungle, Shuttle Race, and Iron Chef have been removed from the Awaysite list.

### DIFF
--- a/UnityProject/Assets/ScriptableObjects/SubScenes/AwayWorldList.asset
+++ b/UnityProject/Assets/ScriptableObjects/SubScenes/AwayWorldList.asset
@@ -13,12 +13,7 @@ MonoBehaviour:
   m_Name: AwayWorldList
   m_EditorClassIdentifier: 
   AwayWorlds:
-  - AwaySites/AncientTemple.json
   - AwaySites/Caverna.json
-  - AwaySites/Palace.json
   - AwaySites/Relicta.json
-  - AwaySites/DarkJungle.json
   - AwaySites/Backrooms.json
   - AwaySites/MysticValley.json
-  - AwaySites/ShuttleRace.json
-  - AwaySites/IronChef.json


FuntimeDudeski:
i can remake shuttle race if people are interested, though i think shuttle flying needs work and changes to the mechanics might affect the design (just as the original course was obsoleted by new shuttle mechanics). good to cut for now.

jungle deserves modernization. i always loved its lighting colors and the xeno-infested base in the cave. i've mildly felt like reworking it for a long time, just didn't quite have sufficient inspiration. will start brainstorming more on it and maybe try making that impassable water tile.

should i make iron chef a blueprint? it's in a bit of a limbo if it's no longer a default awaysite but there isn't yet a system for admins to trigger a midround gamemode/event w/ blueprint spawn (and a means to get there).

Bod9001 (approved):

PR #11235, opened 2026-05-20
Faster explosions

MaxIsJoe:
A while back, I did a lot of optimizations to explosions that allowed us to very easily have very big explosions that would otherwise grind the server to a halt. These optimizations were not fully noticeable to the average player due to how most explosions are often just small ones.

Today, I removed the hardcoded step time to make explosions feel much snappier, quicker, and much more dangerous.
The step time for explosions are now configurable inside the server's game config, and the default step time is now 0.12 seconds. This value can be much lower (or higher) depending on your server's performance, experiment freely with the value.

#### Before

https://github.com/user-attachments/assets/5dd8d135-b944-460a-bd8e-65bc3865a04d

#### After
https://github.com/user-attachments/assets/e4d790a0-433f-4ab1-81e8-e719789dc014

# Changelog

CL: [New] Explosions by default are now much quicker in traversal between tiles.
CL: [New] Servers can now configure how fast or slow explosion expand over time. (Default is 0.12 seconds)

### DIFF
--- a/UnityProject/Assets/Scripts/US13/Managers/GameConfigManager.cs
+++ b/UnityProject/Assets/Scripts/US13/Managers/GameConfigManager.cs
@@ -59,7 +59,7 @@ namespace US13.Managers
 		public int LowPopCheckTimeAfterRoundStart = 300;
 		public int RebootOnAverageFPSOrLower = 35;
 		public string AccountAPIHost;
-		public float ExplosionStepTimeInSeconds = 0.12f;
+		public float ExplosionStepTimeInSeconds = 0.05f;
 
 		//how many rounds of logs Should be stored before they get deleted,  null = 100, -1 Do not delete (will lag admin log UI After a while so manage yourself)
 		//= n The number you want to keep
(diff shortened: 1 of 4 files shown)


Pockets-byte:
default could stand to be significantly faster

MaxIsJoe:
> default could stand to be significantly faster

True, but I'd like to remain cautious about it. Explosions still do a lot of expensive stuff behind the scenes, and I don't think most server hosts will go out of their way to change this value.

However, if everyone is fine with me setting it at a lower value; I can make it like this: (0.05)

https://github.com/user-attachments/assets/c1fb0d06-8a86-416f-b1c6-20edb6f19773

MaxIsJoe:
People seem to really like 0.05, so I'll make it the default for now.

During next playtest on staging, if we find out that 0.05 is too fast for networking changes and are causing problems with GC, we can bump it back up to 0.12 via the game config file

Bod9001:
I know make it the same as run speed!

Bod9001:
0.145 is move speed, so can you set it to that?

PR #11240, opened 2026-05-23
Add AI disclosure check in PR template

MaxIsJoe:
Adds AI disclosure check to PR template.

Mostly copied from the Godot PR template, but changed to fit our needs.


corp-0:
I would probably add it to CONTRIBUTING.md instead of as part of the PR template. I don't like adding more frictions to PRs, more things to answer. They are doing us a favour by contributing, or at least I see it like that.

Now, like I said in discord, I would only allow the usage of AI as assistance for a human programmer who can think by themselves and understand the code they are contributing. If there is no human intervention in the workflow of the contribution, like we had today, we can safely reject the PR

PR #11247, opened 2026-05-30
Suspend world rendering in the pre-game lobby to improve performance

DooblyNoobly:
### Purpose
Provides a way to suspend world camera and lighting cameras to improve the performance of full screen menus. 
Can also be used if you need to disable game camera for any other reason. 
Applied this to the pre game lobby and tested. Results below.

Tested on 4k screen.

Before culling the game cameras:
build_before_camera_cull

After culling the game cameras:
build_after_camera_cull

### Changelog:
CL: [Improvement] Improved performance on the pre-round lobby screen.

### DIFF
--- a/UnityProject/Assets/Prefabs/SceneConstruction/NestedManagers/UIManager/CharacterCreator/SubUIElement/BodyPartCustomisation/Windows/PreRoundScreen.prefab
+++ b/UnityProject/Assets/Prefabs/SceneConstruction/NestedManagers/UIManager/CharacterCreator/SubUIElement/BodyPartCustomisation/Windows/PreRoundScreen.prefab
@@ -103,6 +103,7 @@ MonoBehaviour:
   m_VerticalAlignment: 512
   m_textAlignment: 65535
   m_characterSpacing: 0
+  m_characterHorizontalScale: 1
   m_wordSpacing: 0
   m_lineSpacing: 0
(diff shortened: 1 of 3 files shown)


MaxIsJoe:
Doesn't this break the space background on the lobby screen?

DooblyNoobly:
> Doesn't this break the space background on the lobby screen?

No, looks like the BG is a fullscreen UI texture.

MaxIsJoe:
image

I was correct. It does break it.
This isn't how the pre-round screen should be rendered.

MaxIsJoe:
image

For reference, this is how space is rendered. It's always like this, even on the pre-round screen.

Bod9001:
"I was correct. It does break it.
This isn't how the pre-round screen should be rendered."
does look like that in the editor I think? before the changes
it's just showing the radar preview

DooblyNoobly:
@MaxIsJoe I've simplified the reference counting code so its easier to read and created a static bg with the new space tile: 
Screenshot 2026-05-31 153718

I would like to come back to PreRoundScreen and improve it somewhat in another PR. What I want to do is make the parent fill screen instead of centered and then adjust the children so that it scales better on larger screens. But I think that is beyond the scope of this PR.

MaxIsJoe (approved):
Alright, I think this is fine for the current scope at the moment.

PR #11251, opened 2026-05-31
Better error logging for ItemLightControl NRE's

DooblyNoobly:
### Purpose

Re orders the null checks so a null objectLightEmission is caught before de referencing it which in turn prints the improved error logs so mappers can fix connection problems.

### Media Preview:
Before fix:
itemLightControlNRE

After fix:

ImprovedErrors

### Changelog:
CL: [Fix] Fixes null reference checks on ItemLightControl and allows the more detailed error logs to show.


MaxIsJoe (approved):

FuntimeDudeski:
will put in fixes for glowshroom and anything else that turns up

PR #11252, opened 2026-05-31
Improve the null checks for optional directional / integrity references

DooblyNoobly:
### Purpose
There were missing null checks for optional refs that caused NRE spam from LightSource OnEnable and OnDisable

### Media Preview:
Before fix:

LightSourceNRE

### Changelog:
CL: [Fix] Prevents NRE for optional Directional / Integrity components on LightSource


MaxIsJoe (approved):
Oi very noice

FuntimeDudeski:
thanks! i implemented floor lights best i could but the script is set up around machine lights being wallprotrusions. the other thing i need to figure out is making the item version of them placeable, which light tube and bulb fixtures use WallMountHandApplySpawn for.

PR #11254, opened 2026-06-04
Playtest Fixes & Awaysite Consoles

FuntimeDudeski:
### Purpose

- updates computer.cs to support an optional, selectable enableComponentWhenPowered field that lets the console gameobject have an arbitrary component that is only enabled when the console is powered. [PLEASE REVIEW!--the ‘console is unpowered’ message doesn't display when these consoles aren't powered but they otherwise work as expected.]
- updates messageoninteract.cs to have a multi-line input in the inspector (mainly so the text can include line breaks) [LMK if there is a preferable way to do this]
- adds a couple non-station-specific console types:
- - Info Console - has message on interact component which is set to be disabled when power is off
- - ButtonConsole - has generic switch component which i trust will work as soon as map saver gets fixed to save entries in its generic trigger objects list, also linked to power state
- adds spritesheets for an advanced/heavy emitter variant
- fixes a couple misrotated things on fallstation
- few text edits to message prints in scripts/prefab descs
- updates generic wall switch sprite to buttondoor instead of a logic gate for now
- fix missing reference on glowshroom/glowcap
- clockwork_floor1 was missing from objectmanager
- adds desc and disables constructable from fake solarpanel floor
- corrects sprite import settings on a couple vendor lightmasks

### Media Preview:
emitter_advanced

emitter_advanced_open

emitter_advanced_lights_on

emitter_advanced_lights_lowpower

### Changelog:

CL: [New] New console types for mappers: Info Console which displays a message when powered and interacted with, and ButtonConsole which behaves like a Generic Switch (also when powered)


MaxIsJoe (approved):

FuntimeDudeski:
any simple, conflict-free way to modify the block of code that gives the 'console is unpowered' message based on the CanOpenNetTab bool to include all attempts to interact with an unpowered console, even ones with Network Tab set to none?

> public bool CanOpenNetTab(GameObject playerObject, NetTabType netTabType)
		{
			if (!hasPower)
			{
				Chat.AddExamineMsgFromServer(playerObject, $"{gameObject.ExpensiveName()} is unpowered");
				return false;
			}
			return true;
		}

PR #11256, opened 2026-06-04
Fix DirectionalPassable NRE on map load when the object has no Matrix yet

DooblyNoobly:
### Purpose
Fixes NRE on map load when the Matrix isn't loaded yet. When the matrix finishes loading, the register fires the event again and the object updates its subsystems properly.

### Media Preview:
Fixes these NRE's on server: 

Screenshot 2026-06-04 193952

### Changelog:
CL: [Fix] Fixed a NullReferenceException in DirectionalPassable during map load on the server

### DIFF
--- a/UnityProject/Assets/Scripts/US13/Objects/Directionals/DirectionalPassable.cs
+++ b/UnityProject/Assets/Scripts/US13/Objects/Directionals/DirectionalPassable.cs
@@ -134,12 +134,14 @@ namespace US13.Objects.Directionals
 		private void UpdateSubsystems()
 		{
 			if(ObjectPhysics.HasComponent == false) return;
+			if(Matrix == null) return; // not registered to a matrix yet (e.g. during map load spawn)
 			Matrix.TileChangeManager.SubsystemManager.UpdateAt(ObjectPhysics.Component.OfficialPosition.ToLocalInt(Matrix));
 		}
 
 		private void UpdateSubsystemsAt(Vector3Int localPos)
 		{
 			if(ObjectPhysics.HasComponent == false) return;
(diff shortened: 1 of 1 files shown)


UnityProject/Assets/Scripts/US13/Objects/Directionals/DirectionalPassable.cs
```diff
@@ -134,12 +134,14 @@ private void OnLocalTileChange(Vector3Int oldLocalPos, Vector3Int newLocalPos)
 		private void UpdateSubsystems()
 		{
 			if(ObjectPhysics.HasComponent == false) return;
+			if(Matrix == null) return; // not registered to a matrix yet (e.g. during map load spawn)
```
MaxIsJoe, line 137: This might need to be queued up until Matrix stops being null, because I think it only gets called once; which might result in this function never doing what it's supposed to do.

MaxIsJoe (changes requested):

UnityProject/Assets/Scripts/US13/Objects/Directionals/DirectionalPassable.cs
DooblyNoobly: No need, the Matrix re-fires the event when it is initialized: https://github.com/unitystation/unitystation/blob/7ea49819ec5ee961df2a2b031ddc0be7b3492dba/UnityProject/Assets/Scripts/US13/Managers/MatrixManager/MatrixManager.cs#L219

That then flows on to eventually called OnLocalTileChange that calls UpdateSubsystemsAt. So its safe to return when the matrix is null in that function.

MaxIsJoe (approved):

PR #11257, opened 2026-06-04
Turret Fire Delay Fixes, Human NPC Mobv2, & More Prefab Organization

FuntimeDudeski:
PRing as draft since this is my biggest scripting attempt so far

### Purpose
Scripts:

- turret.cs - Turret fire delay code was a jumble of clamping manipulations. It might still be a bit goofy in its design*, but I fixed the bugs w/ rate of fire and made it much more transparent to prefab creators/mappers what the fields actually do and how to control desired results.
- Fire delay modifier affects only installed gun now--Default and stun bullets have their fire delays set directly and it can now be set independently for each.

Prefabs:

- adds a generic human NPC mobv2 template. i have no idea what i'm doing so please check my work, but the intention is to have a simple, customizable mob for players to interact with in awaysite contexts for now.
- removes some unused overrides from the turret inheritance tree
- does some basic setup on syndicate and centcom turrets like configuring clearances
- tweaks ballistic turrets slightly--final fire delay is almost identical to previous but multiplied just enough to slightly differentiate the different installed ballistic guns. ai ballistic turret (unused anywhere AFAIK) gun set from buzzsaw LMG to r550c since it's presumably linked to the station. 
- another liberty: centcom turret (unused anywhere AFAIK) projectile changed to laserheavy (shot speed is slightly lower than station turrets, but overall higher DPS. range 8 => 12 (made this change after recording the vid below)
- bit of gun folder organization. moves guns with unfinished implementations (default laser instead of its proper projectile, the weird magic hand ‘guns’ parented from a 7.62 rifle with 1 shot, etc.) to subfolders marking them as unimplemented. will seek to redo these when i replace the improperly parented melee weapons.
- removes constructable from loose plating (still evaluating how this tile is working out--may take some time due to lowpop)
- updates IonBoltExplosion to EMP type instead of a standard explosion. it doesn't seem to work on lightbulbs i test fired it at (centcom self-powered ones) but at least it looks correct and keeps the ion rifle from being a semiauto rocket launcher (ion rifle/carbine still moved to unfinished folder and not included in any maps/pools--by me at least, or anywhere i've noticed--pending ion explosion confirmed fully working).
- dead body prefabs now hold some blood for FX purposes
- random text fixes per usual
- renames LaserTurret TurretBase since it's what all turrets inherit from.

### Notes:
*Question: There's a check for whether the turret has a gun with no magazine, in which case it fires the default/backup projectile. Is this desirable? (Only time this should occur would be if a player installed a magazineless gun into a turret, in which case I'd assume you wouldn't want it to fire.)

### Media Preview:

https://github.com/user-attachments/assets/8a90cd06-4f36-4599-80d0-e3b479f9e677

https://github.com/user-attachments/assets/c5e1fca0-88c3-41b2-8f82-0c7ed1c9adef

### Changelog:
CL: [Fix] Turret shot cooldown with a gun installed now scales correctly. Streamlined turret shot cooldown code and improved component field names & tooltips. Configuring turrets is now much more transparent and predictable.


UnityProject/Assets/Scripts/US13/Objects/Other/Turret.cs
```diff
@@ -463,55 +462,51 @@ public void ChangeBulletState(TurretState newState)
...
+                    bulletName = defaultProjectile != null ? defaultProjectile.name : "MissingDefaultProjectile";
+                    bulletSound = gun.FiringSoundA;
+                }
+            }
+            else
+            {
+	            bulletName = defaultProjectile != null ? defaultProjectile.name : "MissingDefaultProjectile";
+                if (spawnGun != null)
+                {
+                    bulletSound = spawnGun.GetComponent<Gun>().FiringSoundA;
```
MaxIsJoe: This probably should be cached, especially if the gun is never going to change.

UnityProject/Assets/Scripts/US13/Objects/Other/Turret.cs
```diff
@@ -463,55 +462,51 @@ public void ChangeBulletState(TurretState newState)
...
+                {
+                    // Default to assigned projectile otherwise
+                    // Do we actually want to let the turret fire an arbitrary projectile if the installed gun exists but has no magazine? -FD
+                    bulletName = defaultProjectile != null ? defaultProjectile.name : "MissingDefaultProjectile";
+                    bulletSound = gun.FiringSoundA;
+                }
+            }
+            else
+            {
+	            bulletName = defaultProjectile != null ? defaultProjectile.name : "MissingDefaultProjectile";
```
MaxIsJoe: turn that string into a const.
whenever you see two strings repeat, put them in a constant variable.

UnityProject/Assets/Scripts/US13/Objects/Other/Turret.cs
```diff
@@ -463,55 +462,51 @@ public void ChangeBulletState(TurretState newState)
...
-			shootSpeed *= shootSpeedMultiplier;
-			shootSpeed = Mathf.Min(0.5f, shootSpeed);
+        private void SetUpBullet()
+        {
+            if (turretState == TurretState.Stun)
+            {
+                bulletName = stunBullet.name;
+                bulletSound = taserSound;
+                // Limit the final shot cooldown to 0.5sec minimum
+                shootSpeed = Mathf.Max(defaultFireDelay * fireDelayMultiplier, 0.5f);
```
MaxIsJoe: since this number is repeated with the same porpuse; but it into a variable.
Magic numbers are not allowed.

MaxIsJoe (changes requested):

UnityProject/Assets/Scripts/US13/Objects/Other/Turret.cs
FuntimeDudeski: i was going to ask about that--a lot of the fire delay code seems based around players building turrets and putting various guns in them (otherwise why not just have authors specify fire delay directly in inspector and be done?)

but if that kind of thing isn't on the roadmap (i know turret frames can be built but have no idea if completed turrets can be finished) or it isn't relevant to caching the strings, i can attempt that.

UnityProject/Assets/Scripts/US13/Objects/Other/Turret.cs
FuntimeDudeski: this one's taken care of. will work on others soon.

FuntimeDudeski:
it also occurs to me that there's no reason stun bullet and default/backup lethal bullet need to share a fire delay var with each other. much more versatile with them configurable independently (and of course, if player-buildable turrets aren't gonna be a big focus, all the fire delay manip code could be simplified)

hm, and there's also no really good reason i can think of to force the fire delay modifier on the directly-configurable projectiles' cooldowns in any case...

UnityProject/Assets/Scripts/US13/Objects/Other/Turret.cs
MaxIsJoe: A lot of this code seems to rely on the existence of a gun in the first place, so it should be designed around that.
Basically, what you need to do is check if gun is null first, then cache a reference to it to be used later on.

UnityProject/Assets/Scripts/US13/Objects/Other/Turret.cs
```diff
@@ -156,6 +156,8 @@ private enum TurretType
 		private const float DetectTime = 1.5f;
 		private const float MinimumShotCooldown = 0.5f;
 
+		private const string MissingProjectile = "MissingDefaultProjectile";
```
MaxIsJoe: `private const string MISSING_PROJECTILE"
follow naming standards for consts

UnityProject/Assets/Scripts/US13/Objects/Other/Turret.cs
FuntimeDudeski: am i doing this check and caching it inside the SetUpBullet() method or closer to the base of the script? getting pretty confused by this part

MaxIsJoe:
tests are failing

UnityProject/Assets/Scripts/US13/Objects/Other/Turret.cs
FuntimeDudeski: so, where i'm at currently is: i got the other 2 things done, tried caching the gun (which i now realize might not be the 'this' you were referring to--gun? bullet? sound?)... and that ended in tears. as of current commit everything is working when tested though.

PR #11267, opened 2026-06-12
Tile Traits + Tile Trait updates for base floor tiles

### LINKED ISSUE #5185: Allow traits to be attached to tiles.
## Description
Title, it should be possible to attach the same traits that item .prefabs use to tiles as well. This would be helpful for making .prefabs be able to reference multiple kinds of tiles and also for making tiles reference each other.
- In #4842, the PlaceableTileReplaceBase component on the floor tile has to reference every single kind of asteroid floor tile individually. If all those asteroid floors shared one trait that the PlaceableTileReplaceBase component could reference instead, then the component on the floor tile would appear cleaner in the inspector.
- It could be possible to set tiles to connect/smooth with other tiles with particular traits. This would be good for making windows smooth into walls (important for 3/4 sprites) and compatible windows and for making particular walls, windows, and floors _not_ smooth into eachother when they're not supposed to.

DooblyNoobly:
### Purpose
Closes #5185 

Tiles can now carry traits using the TileTrait asset. Removes the need for long compatible tile list for waysToPlace check. 

Before:

Screenshot 2026-06-12 175214

After:

Screenshot 2026-06-12 181549

All natural base floor tiles have been updated to use the NaturalBase asset and tested:

Screenshot 2026-06-12 183137

### Changelog:
CL: [New] Added Tile Traits


DooblyNoobly:
P.S I know there is double up with PlacableTile.cs and PlaceableTileReplaceBase.cs but I thought it was outside the scope of this PR to properly refactor it. The fix would be to merge the two classes together as they basically do the same thing and add a flag for the replace feature. That would touch a lot of prefabs that would need to be fixed though so its not a simple refactor.

UnityProject/Assets/Scripts/US13/Tilemaps/Tiles/TileTrait.cs
```diff
@@ -0,0 +1,18 @@
+using ScriptableObjects;
+using UnityEngine;
+
+namespace US13.Tilemaps.Tiles
+{
+	/// <summary>
+	/// Defines a particular trait that a tile can have (assigned on the tile asset).
+	/// </summary>
+	[CreateAssetMenu(fileName = "TileTrait", menuName = "Tiles/TileTrait")]
+	public class TileTrait : SOTracker
```
Bod9001, line 10: why not just reuse the item tray?
since we could just turn that into a generic tray system

DooblyNoobly: It was a delibrate separate type so it keeps the inspector picker only showing the tile traits for that system. We can pull a shared base out later if we expand traits.

Bod9001: ok cool

Bod9001 (approved):

PR #11277, opened 2026-06-23
Crafting Menu Rework: Part 1

MaxIsJoe:
image

The crafting menu is getting an overhaul so that it matches the UI design of the rest of the project.
This has required a lot of wrangling to be able to get it to this state, and I've got bored from how long it took just to move a few things around and add single QoL feature.

In later PRs, the aim is to make the crafting UI more inline with /TG/'s own crafting menu:
image

### Known Issues

- Search input seems to randomly decide not to capture inputs at all? Look into Unity fuckery with negative scales again.

### AI Disclosure

- Claude was used to make the Legacy Text -> TMPro replacement tool.

### Changelog:

CL: [New] The crafting menu supports an "Everything" category that displays a creature's all-learnt recipes.
CL: [Improvement] Updated the Crafting Menu to display category names next to their icons.
CL: [Improvement] Updated the Crafting Menu flow, so everything can be easily navigated from left to right on controllers.
CL: [Fix] Fixed an instance where the crafting menu would default to an empty page when opened for the first time.


MaxIsJoe:
Closing this because there's some unity black magic happening that I don't understand, and Doob is doing stuff with the UIManager; which will basically make this harder to fix.

PR #11292, opened 2026-07-04
Adds mob knockdown when standing in a moving shuttle

MaxIsJoe:
Parity feature with /TG/.
When a mob is standing in a moving shuttle, and the shuttle suddenly makes a turn or moves faster, that mob will be knocked off its beat and stunned momentarily.

This is kinda hard to translate here in Unitystation, so I made sure to keep the requirement check a config value; so servers can disable them if they want.

# Changelog

CL: [New] Mobs standing in a moving shuttle will get knocked down when the shuttle makes a sudden turn, or when they increase in speed.
CL: [Improvement] Whenever a player gets knocked down from a stun, a sound will always play now to notify others nearby that a body has fallen.

### DIFF
--- a/UnityProject/Assets/Scripts/US13/Tilemaps/Behaviours/Objects/RegisterPlayer.cs
+++ b/UnityProject/Assets/Scripts/US13/Tilemaps/Behaviours/Objects/RegisterPlayer.cs
@@ -388,6 +388,7 @@ namespace US13.Tilemaps.Behaviours.Objects
 
 			this.RestartCoroutine(StunTimer(stunDuration), ref unstunHandle);
 			ServerUpdateStunStatus(true);
+			Sound.At(CommonSounds.Instance.Bodyfall, playerScript.gameObject);
 		}
 		private IEnumerator StunTimer(float stunTime)
 		{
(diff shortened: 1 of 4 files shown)


UnityProject/Assets/Scripts/US13/Tilemaps/Behaviours/Objects/RegisterPlayer.cs
```diff
@@ -388,6 +388,7 @@ bool CheckArmorStunImmunity()
 
 			this.RestartCoroutine(StunTimer(stunDuration), ref unstunHandle);
 			ServerUpdateStunStatus(true);
+			Sound.At(CommonSounds.Instance.Bodyfall, playerScript.gameObject);
```
corp-0: <3

corp-0: I think this is missing the trigger, 
```suggestion
			Sound.At(CommonSounds.Instance.Bodyfall, playerScript.gameObject).PlayNetworked();
```

MaxIsJoe: ah, that explains why the sound was missing. thanks.

PR #11295, opened 2026-07-09
Adds PrimeTween

MaxIsJoe:
Adds initial support for `PrimeTween` as a replacement to `LeanTween`.
PrimeTween is much more performant than LeanTween, and has zero allocations. It also feels much more familiar to Godot's Tween library, which gives us a lot of freedom to animate whatever however we like.

You can read about it here:
https://github.com/KyryloKuzyk/PrimeTween

### Changelog

CL: [Fix] Fixed a long-standing issue with the backflip animation on poor internet connections, as the animation sequences are client sided now.


MaxIsJoe:
non-zero chance that code scans will accept this library on the first try, so I'll return to this PR later to whitelist whatever it needs to funciton.

corp-0:
> non-zero chance that code scans will accept this library on the first try

https://www.youtube.com/watch?v=HQwxUAF_YCQ

MaxIsJoe:
image


https://github.com/user-attachments/assets/35f1d8dd-6087-45f9-aa94-94911d5cb71f

PR #11303, opened 2026-08-03
[CodeScan] Improve error formatting, and add hints for type errors

MaxIsJoe:
Turns out taking a break does clear your mind enough to be able to take a stab at seemingly complicated things that are not actually complicated at all.

Tis PR improves the formatting of errors in codescan reports so they're much more readable.
Errors related to non-whitelisted types will also display a hint on how to whitelist them.


MaxIsJoe:
More precise hints will be added later.
for now I want to keep this simple

PR #11324, opened 2026-09-20
feat: update pr2changelog action yaml

corp-0:
### Purpose
I updated pr2changelog, so now we can declare a body to skip the changelog parse with `[NOCL]`

I also updated a lot of shit on the action, since it was untouched ever since I first wrote it.


corp-0:
^that won't happen again after this is merged

